PR #1, opened 2015-10-10
Fix two missing link references

carsonmcdonald:


bhadrip:
Hello @carsonmcdonald,

Thank you for your input on the SDK!  Submitting changes via a pull request is helpful to us because it makes your suggested changes very explicit.  
Our process for handling these requests is to pull the changes into our internal system and push them to GitHub after following our testing and release process(Even though in this case it is documentation). We will update the documentation.

PR #3, opened 2015-10-15
Calculate the payload length on every update.

chadcatlett:
The publish sample doesn't calculate the payload's length during each loop, after the counter reaches 10 the original length is going to result in minor data truncation.


bhadrip:
Hello @chadcatlett 

Thank you for your input on the SDK! Submitting changes via a pull request is helpful to us because it makes your suggested changes very explicit.

Our process for handling these requests is to pull the changes into our internal system and push them to GitHub after following our testing and release process.

We will soon release a new version with the updated sample.

PR #9, opened 2015-12-18
Add disconnectHandler to ShadowParameters_t

gthole:
Issue: `aws_iot_shadow_connect` defaults the `MQTTClientParameters.disconnectHandler` to `NULL` with no way to override through the `ShadowParameters_t pParams` argument.  Cases where `aws_iot_shadow_connect` is used to connect the MQTT Client should be able to specify a disconnect handler.

This PR adds `disconnectHandler` to `ShadowParameters_t` with a `NULL` default value for `ShadowParametersDefault` so no change is required to sample apps or existing implementations.  This same approach could be applied to other parameters in `MQTTConnectParams` as well, if needed.

Alternately, we could change `ShadowParameters_t` to take `MQTTConnectParams` directly to avoid this overlap, but that would be a breaking change.

### DIFF
--- a/aws_iot_src/shadow/aws_iot_shadow.c
+++ b/aws_iot_src/shadow/aws_iot_shadow.c
@@ -27,7 +27,8 @@ const ShadowParameters_t ShadowParametersDefault = {
 		.port = AWS_IOT_MQTT_PORT,
 		.pRootCA = NULL,
 		.pClientCRT = NULL,
-		.pClientKey = NULL
+		.pClientKey = NULL,
+		.disconnectHandler = NULL
 };
 
 void aws_iot_shadow_reset_last_received_version(void) {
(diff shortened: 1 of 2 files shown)


johnrotach:
Hello Greg,

Thanks much for submitting this!  From the initial look the change makes a lot of sense.  Our normal procedure is to pull PR code into our internal repo and run it through our development process and then release it.  When we target a release for this we'll comment back and close this PR.

John
AWS IoT

gthole:
Sounds great!  Thanks John.

gthole:
Any progress on this?  There has been a release since I originally posted this, and it's still an issue in the current SDK.  Thanks!

chaurah:
Hi @gthole,
      The disconnect handler has been added to ShadowInitParams in v2.0.0 of the SDK. Please accept my apologies for the delayed response to this issue. I should have closed this when the new version of the SDK was released.

Rahul

PR #26, opened 2016-06-12
Docs, make invocation

revmischa:
* `aws_iot_shadow_init()` docs really should mention that it now initializes the MQTT client as well for you.
* invoke `make` recursively the right way. use `$(MAKE) -C` not `cd dir && make` - https://www.gnu.org/software/make/manual/html_node/Recursion.html


chaurah:
Hi @revmischa,
        Thank you for the suggestions. Our policy for pull requests is to make the changes internally and include them for the next release of the SDK. I have included these changes for the upcoming v2.1 release of the SDK. If you have any further suggestions, please feel free to let me know.

I will update and close this request once the 2.1 release is out.

Rahul

revmischa:
Personally I think it's weird to use camelCase in C. Not a big thing though. Documentation for shadow connect and MQTT connect params would be nice. Ability to set the connect params when doing shadow connect (right now they're hardcoded in the client).
Also the ability to send large files via MQTT would be really really useful for me. Hugely useful. The max message size limit is a big issue. Amazon has an internal ticket to see if that can be increased right now.
Also I'm unclear on how device name and certification provisioning should work in an automated fashion. If I want to make a lot of devices, how do I assign device IDs and certificates _en masse_?
And when I send a message via MQTT and have it invoke a lambda, how do I get a reference to the device that sent it? I didn't see any info passed in about the client ID on the server side.

chaurah:
Addressing the suggestions below:
1) Camel case - Too late to do anything about this now I think. Return would not be worth the effort.
2) Documentation updates - I see the lack of documentation for Shadow params. Adding that in.
3) Set connect params when doing shadow connect - Looking into this now. It will be a breaking change if I suddenly remove all the hardcoding.
4) Large MQTT messages - That is a service side question unfortunately. You are correct that there are internal tickets on this right now. No new information at the moment.
5) For more information on device name and certification provisioning, please take a look at or BYOC feature here - https://aws.amazon.com/about-aws/whats-new/2016/04/aws-iot-now-supports-using-your-own-certificates/
This will allow you to mass generate and upload certificates instead of having to go through the console.
6) You can get the client ID while creating the rule using the clientId() function. Details here - http://docs.aws.amazon.com/iot/latest/developerguide/iot-sql-functions.html

Hope that helps. Please let me know if you have further questions.

Rahul

chaurah:
This change has been merged in v2.1 of the SDK. Closing the pull request.

Rahul

PR #34, opened 2016-07-28
Constify string parameters

tim-nordell-nimbelink:
String literals, such as "Hello world" need to be const char \* pointers;
as such, using AWS_IOT_MQTT_HOST produce compiler warnings when the
pHost variable is assigned directly to this.

Since the library does not need to modify these string constants, modify
the interfaces to be const char pointers instead of char pointers.

Signed-off-by: Tim Nordell tim.nordell@nimbelink.com


tim-nordell-nimbelink:
I guess I'll add that technically in C99 they're char *, but in C++ they're const char *; since the library has no need to modify these, it should default to const.  (In an embedded microcontroller environment, string literals are often placed within flash which cannot be modified.)

chaurah:
Hi Tim,
       Thank you for this pull request. Our process to merge pull requests is by making the changes internally and including them as a part of the next release. I have added this to the list of fixes for the next release.

Rahul

gordonwang0:
Hello @tim-nordell-nimbelink,

PR #51, opened 2016-11-15
Added Win32 support to aws-iot-device-sdk-embedded-c

joneal:
Added ability to use aws-iot-device-sdk-embedded-c library on Win32 platforms by compiling library into a Win32 DLL.  

win32 folder created under both platform and samples folder.

platform/win32 also has aws-iot-sdk-dll folder to host Visual Studio 2015 solution and project files for aws-iot-sdk.dll.  aws-iot-sdk.dll uses existing mbedTLS library for TLS communications and NOT Win32 libraries.

samples/win32 contains two sample projects - subscribe_publish_sample and shadow_sample.  Both samples are Win32 console applications.  These applications dynamically load the Win32 aws-iot-sdk.dll and do not compile against a static library. 

Solution file in platform/win32/aws-iot-sdk-dll contains all three projects - aws-iot-sdk DLL, subscribe_publish_sample console application, shadow_sample console application.  Build dependencies are configured to build in proper order, and project include paths are configured to find both aws-iot-device-sdk-embedded-c source and aws-iot-sdk DLL.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+platform/win32/aws-iot-sdk-dll/.vs
+platform/win32/aws-iot-sdk-dll/Debug
+platform/win32/aws-iot-sdk-dll/Release
+platform/win32/aws-iot-sdk-dll/*.VC.db
+platform/win32/aws-iot-sdk-dll/*.user
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,18 @@
+certs
+external_libs/mbedTLS
+
 platform/win32/aws-iot-sdk-dll/.vs
 platform/win32/aws-iot-sdk-dll/Debug
 platform/win32/aws-iot-sdk-dll/Release
 platform/win32/aws-iot-sdk-dll/*.VC.db
+platform/win32/aws-iot-sdk-dll/*.VC.opendb
 platform/win32/aws-iot-sdk-dll/*.user
 
 samples/win32/shadow_sample/shadow_sample_console/Debug
 samples/win32/shadow_sample/shadow_sample_console/Release
+
+samples/win32/subscribe_publish_sample/subscribe_publish_sample_console/Debug
+samples/win32/subscribe_publish_sample/subscribe_publish_sample_console/Release
+
+
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@ platform/win32/aws-iot-sdk-dll/Debug
 platform/win32/aws-iot-sdk-dll/Release
 platform/win32/aws-iot-sdk-dll/*.VC.db
 platform/win32/aws-iot-sdk-dll/*.user
+
+samples/win32/shadow_sample/shadow_sample_console/Debug
+samples/win32/shadow_sample/shadow_sample_console/Release
+
--- a/include/aws_iot_config.h
+++ b/include/aws_iot_config.h
@@ -23,13 +23,21 @@
 
 // Get from console
 // =================================================
-#define AWS_IOT_MQTT_HOST              "" ///< Customer specific MQTT HOST. The same will be used for Thing Shadow
-#define AWS_IOT_MQTT_PORT              8883 ///< default port for MQTT/S
-#define AWS_IOT_MQTT_CLIENT_ID         "c-sdk-client-id" ///< MQTT client ID should be unique for every device
-#define AWS_IOT_MY_THING_NAME 		   "AWS-IoT-C-SDK" ///< Thing Name of the Shadow this device is associated with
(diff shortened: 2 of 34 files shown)


chaurah:
Hi @joneal,
  This is awesome. We already had a plan to add support for Windows in the next release. This will be tremendous help. Our current process to merge pull requests is to make the changes internally and include them, with credit to the original author, in the next release. I will be sure to include this. Please do let us know if you have further suggestions. Thank you for using AWS IoT.

Rahul

joneal:
Glad to help.  Given time, I would like to also include a .NET (C#) module that has the P/Invoke definitions for the new aws-iot-sdk.dll.

chaurah:
Hi @joneal,
     We now have a C++ SDK that also works on Windows. You can find it [here](https://github.com/aws/aws-iot-device-sdk-cpp). There's still a lot to be done on the Windows side of things include some of the takeaways I have after looking at your pull request. Please let me know if you have any suggestions on how we can improve that SDK or any thoughts on C vs C++ for Windows development. It would be extremely useful information for us.

Rahul

joneal:
Hello Rahul,

I have quickly reviewed the C++ SDK.  I have not developed in C++ in several years, and have not kept up with C++ 11, so my questions and suggestions may not be valid.

I suppose my main question is what benefit does having a platform neutral SDK written in C++ provide over the existing SDK for C?   It seems that now there are two SDKs for potentially low resource to server level platforms that essentially provide the same benefits, from a consumer perspective.  Yet now there is a wildly divergent code base between the two (C/C++).  

Having to download CMake, OpenSLL, git bash, etc just to build a VS2015 solution is going to turn off a lot of Windows developers.  This will be a source of configuration issues in order just to get the development environment up and running by all the extra gears in the machine, when providing a working VS2015 solution file would greatly simplify matters.

Not sure of the benefit of support Websockets...?

Why the OpenSSL option?  I considered it and other libraries for TSL when I built the Win32 port for the embedded C SDK, but came back to mbedTLS in order not to disrupt the "spirit" of the SDK, in that all platforms would use mbedTLS even if other platform specific options were available.  I would think this would simplify long-term maintenance of the SDK.

I think I still prefer the simple Win32 DLL support added to the embedded C SDK.  This allows many various development tools/languages to utilize the SDK on a Win32 box; again the consumer of the SDK not caring if the internal implementation is C or C++ in the DLL.....or caring if integrating at the source code level in which I feel the C SDK is much easier to integrate than the C++ SDK.

I may be way off base, but that is my perspective.  Would like to hear any rebuttals especially if I am in error.

Jerry

chaurah:
Hi @joneal,
     Apologies for the extreme delay in responding. I missed this comment completely in my notifications. Answers to your questions below:
1) Why OpenSSL/WebSocket/C++ specific/Windows support? Very simple reason, we have customers who explicitly asked for it :) There are pros and cons to each of those of course, but we do have to provide SDKs to as large an audience as necessary. In this case there have been sufficient requests for each of those.

2) Your callout about VS solutions definitely makes sense. I will try to push through a new minor version that includes them as soon as possible.

3) I disagree with you about the spirit of the SDK. I think the current architecture of the C SDK doesn't make this clear enough. But we Want developers to port the SDK to their platforms. As much as we try, there's no way we can write anything below the MQTT layer in a generic enough manner that it extracts the best performance out of the large variety of target devices we are hearing about. There's simply too many. There are also devices where MbedTLS/OpenSSL simply wouldn't work at all and it is required to port the SDK to whatever TLS implementation is available on the device.

MbedTLS and OpenSSL are simply the most common and that is why we have reference implementations for both of them. If we get enough requests, we will consider adding more in the future as well.

4) C++ SDK is intended to be targeted at higher end devices where developers want to and will use features like multi-threading and message queuing. It also lets developers get away from the static only allocations that are required for the Embedded C SDK.

The idea here is that there are definitely devices that are both very capable and have modern compilers (eg. Raspberry Pi which, relative to a low power embedded device, is in a different league altogether). There's no reason those capabilities should not be used. So we want to have SDKs that are capable of addressing those use cases. We will be refactoring the C SDK so it better supports low power devices as well where the C++ SDK simply can't be used. There will be overlap. And we hope to differentiate the SDKs based on the feature sets they provide.

This is all of course being worked on based on the feedback we are receiving from all of the developers who are using the SDK. So your perspectives are VERY important for us to hear. So please do keep sending us your thoughts about these changes. It is extremely valuable information and drives our team priorities.

Rahul

joneal:
Hi Rahul,

Thanks for the feedback; makes sense.  I look forward to seeing the next release of the embedded C SDK.

Jerry

truonggiangbk:
Hi jonel,
Can you share tutorial for build aws-iot-device-sdk-embedded-C into Win32 dll? Thank

gordonwang0:
Hi @joneal,

PR #59, opened 2016-12-08
Fixed MQTT header constructing and parsing

lkaino:
Bit fields must not be used in communication protocol frame parsing. See [http://stackoverflow.com/questions/1490092/c-c-force-bit-field-order-and-alignment](http://stackoverflow.com/questions/1490092/c-c-force-bit-field-order-and-alignment).

The bit-fields used in AWS C library caused issues in PowerPC target I'm using.

This might not be the most beautiful way to fix it, but it works. Tested on x86 and PowerPC architectures.

### DIFF
--- a/include/aws_iot_mqtt_client_common_internal.h
+++ b/include/aws_iot_mqtt_client_common_internal.h
@@ -69,26 +69,17 @@ typedef enum msgTypes {
 	DISCONNECT = 14
 } MessageTypes;
 
+/* Macros for parsing header fields from incoming MQTT frame. */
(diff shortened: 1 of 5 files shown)


chaurah:
Hi @lkaino,
   Thank you for submitting this pull request. We are already working on removing these bit-fields for the next release of the SDK. This change will be very useful towards that goal. Our usual process for including pull requests is that we include the change in the next release of the SDK after merging it internally. I will update and close this issue when that release is out.

Please do let us know if you have further suggestions. Thank you for using AWS IoT.

Rahul

PR #78, opened 2017-02-22
Fixed compilation warnings

nikitanerkar:
Fixed gcc compilation warnings


chaurah:
Hi Nikita,
    Thank you for this pull request. I have added it to our internal tracker as well to make sure it gets merged as applicable to the next release of the SDK. Please do let me know if you have any further suggestions.

Rahul

nikitanerkar:
Please view following change as well

chaurah:
Thank you for this update Nikita. I think an additional option, where the timer increases to the maximum value and then keeps retrying with the max value would also be useful. I will add this to our internal tracker.
As always, please let me know if you have any further suggestions.

Rahul

PR #88, opened 2017-06-16
Fix username and password are confused

tipfei:
Use the username as password


liuszeng:
Hi @tipfei ,

Thank you very much for pointing this out. I have added it to our internal tracker to make sure it gets merged as applicable to the next SDK release. Really appreciate your effort in making the SDK better.

Thanks,
Liusu

PR #110, opened 2017-11-28
Allow keepAliveIntervalInSec to be configured when calling aws_iot_shadow_connect

pctj101:
I have issues with the keep alive interval being "too long" which causes my SSL tunnel to close down.

It would be nicer to have keepAliveIntervalInSec be configurable even when using shadows.

### DIFF
--- a/src/aws_iot_shadow.c
+++ b/src/aws_iot_shadow.c
@@ -33,10 +33,10 @@ extern "C" {
 #include "aws_iot_shadow_records.h"
 
 const ShadowInitParameters_t ShadowInitParametersDefault = {(char *) AWS_IOT_MQTT_HOST, AWS_IOT_MQTT_PORT, NULL, NULL,
-															NULL, false, NULL};
+															NULL, false};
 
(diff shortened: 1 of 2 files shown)


vareddy:
Hi @pctj101,
Thank you for your pull request! We will merge it into mainline after the next release. 
Please let us know if you have any more questions.
Thanks!
Varun

src/aws_iot_shadow.c
```diff
@@ -33,10 +33,10 @@ extern "C" {
 #include "aws_iot_shadow_records.h"
 
 const ShadowInitParameters_t ShadowInitParametersDefault = {(char *) AWS_IOT_MQTT_HOST, AWS_IOT_MQTT_PORT, NULL, NULL,
-															NULL, false, NULL};
+															NULL, false};
```
gordonwang0: I think there should be a NULL after the false here to initialize `ShadowInitParametersDefault.disconnectHandler`.  Can you take a look at this?

escherstair: I cannot speak for @pctj101 but I think the `NULL` is necessary.

gordonwang0:
Hello @pctj101,

PR #124, opened 2018-01-08
Thing Shadow: Fix potential shadow buffer overflow

projectgus:
This isn't a problem for the default SHADOW_MAX_SIZE_OF_RX_BUFFER value, as the payload will be at most AWS_IOT_MQTT_RX_BUF_LEN bytes so there is room for the NULL byte. However if this value is changed to any lower value, it can overflow when writing the terminal NULL byte.

Thanks!


vareddy:
Hi @projectgus ,
Thank you for your pull request! You are correct in that the buffer will overflow when it is changed to a lower value. We will test this fix and merge into master after our testing is complete.
Please let us know if you have any more questions or suggestions about the SDK.
Thanks!
Varun

PR #131, opened 2018-01-22
Changes for JSON Object in delta callback. Note - Only tested on STM32

chrismrutherford:
also has variable keepalive and delta call only on value change.


vareddy:
Hi @chrismrutherford,
We will merge and test the pull request and push to master when we have completed our testing.
Thanks a lot!
Varun

PR #135, opened 2018-02-05
mbedtls_wrap: Fix unintialized variable usage

vpetrigo:
Hello,

Thank you for your library as it really helps a lot.
I would like to propose a small fix that allows applications that utilize AWS library to be built correctly with `-Wall -Werror` flags.

The root of issue is uninitialized variable `ret` that lead to the following error:

```c
../../../platform/linux/mbedtls/network_mbedtls_wrapper.c:280:69: error: ‘ret’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   written_so_far < len && !has_timer_expired(timer); written_so_far += ret, frags++) {
                                                                     ^~
cc1: all warnings being treated as errors
```

I have not digged through the context of usage of that method, but it seems that it might be a problem if `has_timer_expired` evaluates to `true`.


huguesBouvier:
Thank you for this pull request! You are correct, there is a condition where “ret” would be accessed while not initialized.
We are evaluating your fix for inclusion in a subsequent release.

PR #152, opened 2018-04-05
Fix potential buffer overflow in parseStringValue

gordonwang0:
Requires the inclusion of a buffer size in jsonStruct_t.

In response to #151.

### DIFF
--- a/src/aws_iot_json_utils.c
+++ b/src/aws_iot_json_utils.c
@@ -170,12 +170,9 @@ IoT_Error_t parseBooleanValue(bool *b, const char *jsonString, jsmntok_t *token)
 		IOT_WARN("Token was not a primitive.");
 		return JSON_PARSE_ERROR;
 	}
-	if(jsonString[token->start] == 't' && jsonString[token->start + 1] == 'r' && jsonString[token->start + 2] == 'u'
-	   && jsonString[token->start + 3] == 'e') {
+	if(strncmp(jsonString + token->start, "true", 4) == 0) {
 		*b = true;
-	} else if(jsonString[token->start] == 'f' && jsonString[token->start + 1] == 'a'
-			  && jsonString[token->start + 2] == 'l' && jsonString[token->start + 3] == 's'
-			  && jsonString[token->start + 4] == 'e') {
+	} else if(strncmp(jsonString + token->start, "false", 5) == 0) {
 		*b = false;
 	} else {
 		IOT_WARN("Token was not a bool.");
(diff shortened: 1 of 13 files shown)


tests/unit/src/aws_iot_tests_unit_json_utils_helper.c
```diff
@@ -58,26 +60,48 @@ TEST_C(JsonUtils, ParseStringBasic) {
...
+
+	jsmnReturn = jsmn_parse(&test_parser, pJsonString, strlen(pJsonString), t, sizeof(t)/sizeof(t[0]));
+	CHECK_EQUAL_C_INT(3, jsmnReturn);
+
+	rc = parseStringValue(parsedString, STRING_BUFFER_LENGTH, pJsonString, t + 2);
+
+	CHECK_EQUAL_C_INT(SHADOW_JSON_ERROR, rc);
+
+	/* Ensure there was no attempt to write to a buffer that's too small. */
+	for(i = 0; i < STRING_BUFFER_LENGTH; i++)
```
huguesBouvier: NP: You could do a mem compare

tests/unit/src/aws_iot_tests_unit_json_utils_helper.c
```diff
@@ -58,26 +60,48 @@ TEST_C(JsonUtils, ParseStringBasic) {
...
-	rc = parseStringValue(parsedString, json, t + 2);
+	rc = parseStringValue(parsedString, STRING_BUFFER_LENGTH, json, t + 2);
 
 	CHECK_EQUAL_C_INT(3, r);
 	CHECK_EQUAL_C_INT(SUCCESS, rc);
 	CHECK_EQUAL_C_STRING("this is a longer string for test 2", parsedString);
 }
 
+/* Test that parsing a string doesn't overflow the given buffer. */
+TEST_C(JsonUtils, ParseStringWithBufferTooSmall) {
```
huguesBouvier: You should test the happy path too :)

src/aws_iot_shadow_json.c
```diff
@@ -354,28 +354,28 @@ bool isJsonValidAndParse(const char *pJsonDocument, void *pJsonHandler, int32_t
 }
 
 static IoT_Error_t UpdateValueIfNoObject(const char *pJsonString, jsonStruct_t *pDataStruct, jsmntok_t token) {
```
huguesBouvier: A look up table could look better. Though it is a risky change.

src/aws_iot_json_utils.c
```diff
@@ -170,12 +170,9 @@ IoT_Error_t parseBooleanValue(bool *b, const char *jsonString, jsmntok_t *token)
 		IOT_WARN("Token was not a primitive.");
 		return JSON_PARSE_ERROR;
 	}
-	if(jsonString[token->start] == 't' && jsonString[token->start + 1] == 'r' && jsonString[token->start + 2] == 'u'
-	   && jsonString[token->start + 3] == 'e') {
+	if(strncmp(jsonString + token->start, "true", 4) == 0) {
```
huguesBouvier: Thanks!

tests/unit/src/aws_iot_tests_unit_json_utils_helper.c
gordonwang0: Like when the buffer is big enough?  I think that's adequately tested by the integration tests (jobs integration test).

tests/unit/src/aws_iot_tests_unit_json_utils_helper.c
gordonwang0: Wouldn't that require keeping an additional block of all-zero memory around?

PR #155, opened 2018-04-07
Fixing other potential memory corruption

huguesBouvier:
Fixing other potential memory corruption bugs.
Also added a function to free up mutexes created by mqtt init and shadow init

Passed unit tests and integration tests


src/aws_iot_shadow_json.c
```diff
@@ -424,13 +462,13 @@ bool isReceivedJsonValid(const char *pJsonDocument) {
 	return true;
 }
 
-bool extractClientToken(const char *pJsonDocument, char *pExtractedClientToken) {
+bool extractClientToken(const char *pJsonDocument, size_t jsonSize, char *pExtractedClientToken, size_t clientTokenSize) {
 	int32_t tokenCount, i;
 	uint8_t length;
```
gordonwang0: Are you sure `uint8_t` (up to 255) is always sufficient?  How about `size_t` instead?

tests/unit/src/aws_iot_tests_unit_shadow_action_helper.c
```diff
@@ -838,9 +838,75 @@ TEST_C(ShadowActionTests, NoClientTokenForShadowAction) {
...
+	
+	//Test by cutting the JSON document
+	ret_val = isReceivedJsonValid(getRequestJson, 3);
+	CHECK_EQUAL_C_INT(false, ret_val);
+		
+	//Happy path
+	ret_val = isReceivedJsonValid(getRequestJson, TEST_JSON_SIZE);
+	CHECK_EQUAL_C_INT(true, ret_val);
+	
+	IOT_DEBUG("-->Success - sReceivedJsonValid");
```
gordonwang0: Wrong test name.

tests/unit/src/aws_iot_tests_unit_shadow_action_helper.c
```diff
@@ -838,9 +838,75 @@ TEST_C(ShadowActionTests, NoClientTokenForShadowAction) {
 	IOT_DEBUG("-->Success - No client token for shadow action \n");
 }
 
+TEST_C(ShadowActionTests, IsJsonValid)
+{
+	bool ret_val;
+	char getRequestJson[TEST_JSON_SIZE];
+	
+	IOT_DEBUG("-->Running Shadow Action Tests - sReceivedJsonValid \n");
```
gordonwang0: Wrong test name.

tests/unit/src/aws_iot_tests_unit_shadow_action_helper.c
```diff
@@ -838,9 +838,75 @@ TEST_C(ShadowActionTests, NoClientTokenForShadowAction) {
...
+	//Try JSON with token but not enough memory
+	snprintf(getRequestJson, TEST_JSON_SIZE, TEST_JSON_RESPONSE_FULL_DOCUMENT);	
+	ret_val = extractClientToken(getRequestJson, TEST_JSON_SIZE, extractedClientToken, 1 );
+	CHECK_EQUAL_C_INT(false, ret_val);
+	
+	//Happy path
+	ret_val = extractClientToken(getRequestJson, TEST_JSON_SIZE, extractedClientToken, MAX_SIZE_CLIENT_ID_WITH_SEQUENCE );
+	CHECK_EQUAL_C_INT(true, ret_val);
+	
+	IOT_DEBUG("-->Success - getAndDeleteRequest");
```
gordonwang0: Wrong test name.

gordonwang0:
Can you also change `aws_iot_mqtt_client.c`?

https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/ba30f681c5ffb529b8f7230b1220f31c4a19c176/src/aws_iot_mqtt_client.c#L207-L221

In `aws_iot_mqtt_init`, the mutexes `state_change_mutex` and `tls_read_mutex` and `tls_write_mutex` need to be destroyed before returning if there's an error after their creation to avoid a memory leak.

src/aws_iot_mqtt_client.c
```diff
@@ -169,6 +169,34 @@ IoT_Error_t aws_iot_mqtt_set_connect_params(AWS_IoT_Client *pClient, IoT_Client_
...
+    IoT_Error_t rc = SUCCESS;
+
+    if (NULL == pClient) {
+        rc = NULL_VALUE_ERROR;
+    }
+
+#ifdef _ENABLE_THREAD_SUPPORT_
+    if ((&(pClient->clientData.state_change_mutex) != NULL)&&(rc == SUCCESS))
+    {
+        rc = aws_iot_thread_mutex_destroy(&(pClient->clientData.state_change_mutex));
```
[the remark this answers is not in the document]
escherstair: I'm not 100% sure about handling in this way.
`(&(pClient->clientData.state_change_mutex) != NULL)` checks if the **address** of variable `pClient->clientData.state_change_mutex` is not `NULL`.
Then this address is passed to `aws_iot_thread_mutex_destroy()` by value; the function can free the content, but cannot modify the parameter passed by value (that is the address).
Even setting `pClient->clientData.state_change_mutex` to `NULL` (which I think it's a good idea), its address is not NULL.
And so calling `aws_iot_mqtt_free()` one more time the condition `(&(pClient->clientData.state_change_mutex) != NULL)` is `true` and so `aws_iot_thread_mutex_destroy()` would be called ahain.

Is this the expected beahvior?

samples/linux/shadow_sample/aws_iot_config.h
```diff
@@ -23,7 +23,7 @@
 
 // Get from console
 // =================================================
-#define AWS_IOT_MQTT_HOST              "" ///< Customer specific MQTT HOST. The same will be used for Thing Shadow
+#define AWS_IOT_MQTT_HOST              "a3qggb6vsdlf5s.iot.us-west-2.amazonaws.com" ///< Customer specific MQTT HOST. The same will be used for Thing Shadow
```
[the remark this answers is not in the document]
huguesBouvier: Correct, fixed

src/aws_iot_mqtt_client.c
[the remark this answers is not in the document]
huguesBouvier: Actually it is a bug, checking NULL is pointless, it is a structure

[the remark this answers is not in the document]
huguesBouvier: It seems there is not need to check, destroy will set it to an invalid value so subsequent call will return EINVALID

src/aws_iot_shadow_json.c
huguesBouvier: Fixed

tests/unit/src/aws_iot_tests_unit_shadow_action_helper.c
huguesBouvier: Fixed

tests/unit/src/aws_iot_tests_unit_shadow_action_helper.c
huguesBouvier: Fixed

tests/unit/src/aws_iot_tests_unit_shadow_action_helper.c
huguesBouvier: Fixed

huguesBouvier:
Ok, I will change that.

src/aws_iot_mqtt_client.c
```diff
@@ -169,6 +169,34 @@ IoT_Error_t aws_iot_mqtt_set_connect_params(AWS_IoT_Client *pClient, IoT_Client_
...
+        rc = NULL_VALUE_ERROR;
+    }
+
+#ifdef _ENABLE_THREAD_SUPPORT_
+    if (rc == SUCCESS)
+    {
+        rc = aws_iot_thread_mutex_destroy(&(pClient->clientData.state_change_mutex));
+    }
+
+    if (rc == SUCCESS)
```
gordonwang0: I think we should still attempt to destroy mutexes, even if the previous call to `aws_iot_thread_mutex_destroy` failed.  We could also print a debug message if a call to `aws_iot_thread_mutex_destroy` fails.

src/aws_iot_mqtt_client.c
```diff
@@ -212,10 +240,13 @@ IoT_Error_t aws_iot_mqtt_init(AWS_IoT_Client *pClient, IoT_Client_Init_Params *p
...
+		(void)aws_iot_thread_mutex_destroy(&(pClient->clientData.state_change_mutex));
 		FUNC_EXIT_RC(rc);
 	}
 	rc = aws_iot_thread_mutex_init(&(pClient->clientData.tls_write_mutex));
 	if(SUCCESS != rc) {
+		(void)aws_iot_thread_mutex_destroy(&(pClient->clientData.tls_read_mutex));
+		(void)aws_iot_thread_mutex_destroy(&(pClient->clientData.state_change_mutex));
 		FUNC_EXIT_RC(rc);
 	}
 #endif
```
gordonwang0: Line 263 also contains an exit on failure, should also destroy mutexes there.

huguesBouvier: I am not sure I agree totally with that process. The mqtt free should be called after init anyway. I think we are bringing complexity when not needed.

src/aws_iot_mqtt_client.c
```diff
@@ -169,6 +169,39 @@ IoT_Error_t aws_iot_mqtt_set_connect_params(AWS_IoT_Client *pClient, IoT_Client_
...
+			rc = aws_iot_thread_mutex_destroy(&(pClient->clientData.tls_read_mutex));
+		}else{
+			(void)aws_iot_thread_mutex_destroy(&(pClient->clientData.tls_read_mutex));
+		}
+
+		if (rc == SUCCESS)
+		{
+			rc = aws_iot_thread_mutex_destroy(&(pClient->clientData.tls_write_mutex));
+		}else{
+			(void)aws_iot_thread_mutex_destroy(&(pClient->clientData.tls_read_mutex));
```
gordonwang0: This line should destroy `tls_write_mutex`.

Also, is it necessary to re-assign `rc`?  Seems we could just do the following to clean everything up.
```c
if(rc == SUCCESS) /* Check not NULL_VALUE_ERROR. */
{
    (void)aws_iot_thread_mutex_destroy(&(pClient->clientData.state_change_mutex));
    (void)aws_iot_thread_mutex_destroy(&(pClient->clientData.tls_read_mutex));
    (void)aws_iot_thread_mutex_destroy(&(pClient->clientData.tls_write_mutex));
}
```

src/aws_iot_mqtt_client.c
gordonwang0: But if `mqtt_init` returns something other than `SUCCESS`, users may not know they need to call `mqtt_free`.  I think it makes sense for `init` to clean everything up if it fails.

src/aws_iot_shadow_json.c
gordonwang0: It's still `uint8_t` in the latest version?

tests/unit/src/aws_iot_tests_unit_shadow_action_helper.c
gordonwang0: This is still `sReceivedJsonValid` in the latest version?

src/aws_iot_mqtt_client.c
huguesBouvier: I though of that but it will not work unless you refactor the function because it has exit points everywhere

PR #196, opened 2018-06-22
Keep a local copy of topicName on subscription

justeph:
On subscription, we should keep a local copy of the topicName, because we need it when receiving a notification to forward it to the correct user handler. If user free memory of the topicName, he will never get notification. 

So use of calloc on subscribe and free on unsubscribe


gordonwang0:
Hi @justeph,

Thanks for taking the time to put together this pull request.  Unfortunately, we can't accept this because one of the [design goals of this SDK](https://github.com/aws/aws-iot-device-sdk-embedded-C#design-goals-of-this-sdk) forbids the use of dynamic memory allocation.

PR #198, opened 2018-06-27
Keep a local copy of topicName on subscription

justeph:
following pull request #196 

On subscription, we should keep a local copy of the topicName, because we need it when receiving a notification to forward it to the correct user handler. If user free memory of the topicName, he will never get notification.

This patch does not use dynamic memory allocation. 

The TOPIC_NAME_MAX_LEN is defined to be 256 as described here: https://aws.amazon.com/sns/faqs/#Features_and_Functionality

If this patch is not correct please don't close the case, I would be happy to discuss and modify it. 

### DIFF
--- a/src/aws_iot_mqtt_client_subscribe.c
+++ b/src/aws_iot_mqtt_client_subscribe.c
@@ -176,7 +176,7 @@ static uint32_t _aws_iot_mqtt_get_free_message_handler_index(AWS_IoT_Client *pCl
 	FUNC_ENTRY;
 
 	for(itr = 0; itr < AWS_IOT_MQTT_NUM_SUBSCRIBE_HANDLERS; itr++) {
-		if(pClient->clientData.messageHandlers[itr].topicName == NULL) {
+		if(pClient->clientData.messageHandlers[itr].topicNameLen == 0) {
 			break;
 		}
 	}
@@ -260,8 +260,8 @@ static IoT_Error_t _aws_iot_mqtt_internal_subscribe(AWS_IoT_Client *pClient, con
 	//	return RX_MESSAGE_INVALID_ERROR;
 	//}
 
-	pClient->clientData.messageHandlers[indexOfFreeMessageHandler].topicName =
-			pTopicName;
+	memcpy(pClient->clientData.messageHandlers[indexOfFreeMessageHandler].topicName,
+			pTopicName, topicNameLen);
 	pClient->clientData.messageHandlers[indexOfFreeMessageHandler].topicNameLen =
 			topicNameLen;
 	pClient->clientData.messageHandlers[indexOfFreeMessageHandler].pApplicationHandler =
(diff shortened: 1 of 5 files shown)


src/aws_iot_mqtt_client_subscribe.c
```diff
@@ -260,8 +260,8 @@ static IoT_Error_t _aws_iot_mqtt_internal_subscribe(AWS_IoT_Client *pClient, con
 	//	return RX_MESSAGE_INVALID_ERROR;
 	//}
 
-	pClient->clientData.messageHandlers[indexOfFreeMessageHandler].topicName =
-			pTopicName;
+	memcpy(pClient->clientData.messageHandlers[indexOfFreeMessageHandler].topicName,
```
gordonwang0: Need to check that `topicNameLen < TOPIC_NAME_MAX_LENGTH`.  Otherwise, this introduces a memory corruption bug.

lt72:
Hi @justeph, this is a difficult case to solve, and here is why:
1) we *likely* cannot have a topicName members sized at TOPIC_NAME_MAX_LEN, because that would cost 1 or more Kb of static memory to go potentially unused. MITIGATION: user could trim AWS_IOT_MQTT_NUM_SUBSCRIBE_HANDLERS to limit the unused memory amount.
2) we cannot allocate that memory dynamically because one of the goals of the C-SDK is to support running with no dynamic memory allocation.
3) We cannot use a variable size struct (char topicName[ ]) because we would need to allocate that dynamically anyway.

I believe that your solution is the best possible solution, other than asking the user not to de-allocate the topicName storage for the duration of the program. 

I have no definite opinion about what solution is more desirable, but we are definitely open to hear the case for one or the other approach, or suggest alternatives.

Thank you for working with us, we truly appreciate it!

gordonwang0:
Hi @justeph,

Thanks for taking the time to put this together.  However, I'm worried about this introducing breaking changes for our current users.

Limiting the MQTT topic length to `TOPIC_NAME_MAX_LEN` (256) will cause problems for current customers using MQTT topics longer than this.  While they could increase this value, it's unreasonable to force customers to keep large blocks of likely unused memory in use for storing topic names, especially on an SDK that considers small, memory-constrained devices.

I think the SDK's current implementation is best considering these constraints.  Customers may choose to allocate the topic names on the heap or global memory to preserve them if needed.  Please let me know your thoughts on this.

justeph:
Hi @lt72 and @gordonwang0,

I do understand all the reasons you mention and the need to save as much memory as possible. 

@gordonwang0 I don't understand why it would introduce breaking changes for current users as topicName longer than 256 is not allowed anyway (see the link from my first message). By the way, I tried on my aws iot account online with the "Test" feature: trying to subscribe to a topic with more than 256 char fails (disconnection and re-connection). 

I understand that dynamic memory allocation is avoided. By the way, I completely missed the fact that it was mentioned on the doxygen documentation that memory should be static.. 

From my user point of view, I want to subscribe to notifications from a given topic and that's all. After that, I do not care about the topicName any more, so I can reuse memory location for something else..  My little experience in C programming taught me to not rely on user pointer inside a library.. That is why it took me some time to understand what was happening there.

Anyway, I really appreciate you take the time to discuss about it :)

gordonwang0:
Hi @justeph,

Version 4 keeps local copies of topic name by default.  Please give it a try if you have time.

PR #207, opened 2018-07-20
Delete unnecessary variables

JunhwanPark:
This patch is a simple fix. I removed unnecessary variable settings.


gordonwang0:
Thanks for doing this!

PR #217, opened 2018-08-07
samples/linux: remove initialise globals to 0

JunhwanPark:
This PR is also simple fix. 
I think, global variables do not need initialise 0 or false.


gordonwang0:
Hi @JunhwanPark,

I prefer those to be initialized to 0 or false because it makes it makes those values easier to debug.  I think seeing an uninitialized "garbage" value in a debugger is confusing.

I'm willing to change it though if you have a good reason to.

JunhwanPark:
Hello @gordonwang0 
I agree with you. I think I do not need the initial value of a global variable, but I think it may be needed for readability and debugging. I'll delete this PR.
Thank you.

PR #222, opened 2018-08-22
support 'certificates in sources' option

JunhwanPark:
Some embedded devices must have the certificate files in their source
when they are built. If you disable the 'MBEDTLS_FS_IO' option of
mbedTLS, you can not use the function to import files. This patch
includes a script that inserts the certificate into the source code.
Here's how to use the script:

$ cd ./certs
$ sh mkcertData.sh [root CA file] [certificate file] [private key file]

The certData.c file is generated, and this patch contains mbedtls
function substitution to use it.

Signed-off-by: Junhwan Park <junhwan.park@samsung.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/certs/.gitignore
+++ b/certs/.gitignore
@@ -1,0 +1,1 @@
+certData.c
--- a/certs/certData.h
+++ b/certs/certData.h
@@ -1,0 +1,13 @@
+#ifndef __CERTDATA_H__
+#define __CERTDATA_H__
+
+extern unsigned const char root_ca_pem[];
+extern const int rootCaLen;
+
+extern unsigned const char client_cert_pem[];
+extern const int clientCertLen;
+
+extern unsigned const char client_private_key_pem[];
+extern const int clientPrivateKeyLen;
+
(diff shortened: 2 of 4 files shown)


JunhwanPark:
The RTOS I use is not easy to insert additional certificate files. So I need to insert a certificate into the source to create a one-binary.

gordonwang0:
Hi @JunhwanPark,

Thank you for your continued contributions to this SDK!

This PR is changing the Linux portability layer.  On Linux, I strongly prefer the use of the filesystem, as there are security concerns with placing the private key plaintext in a header file.  So I don't think it's a good idea to provide this functionality on Linux.

If you require the ability to place credentials in header files, I suggest making a new portability layer for your RTOS and implementing this functionality only for that platform.

Please let me know your thoughts.

JunhwanPark:
Hello @gordonwang0 
I agree with you. I will only use this feature on my RTOS. I am already testing this by applying it, and I will let you know if there is a better way. 
Thank you.

PR #257, opened 2019-01-20
Replace queue/list implementation

gordonwang0:


lib/include/iot_linear_containers.h
```diff
@@ -0,0 +1,888 @@
...
+ * element. Optional; pass `NULL` to ignore.
+ * @param[in] linkOffset Offset in bytes of a link member in its container, used
+ * to calculate the pointer to pass to `freeElement`. This value should be calculated
+ * with the C `offsetof` macro. This parameter is ignored if `freeElement` is `NULL`
+ * or its value is `0`.
+ */
+/* @[declare_linear_containers_list_double_removeall] */
+static inline void IotListDouble_RemoveAll( IotListDouble_t * const pList,
+                                            void ( *freeElement )( void * ),
+                                            size_t linkOffset )
```
lt72: No need to pass offset as a parameter, it can be more efficiently computed with 
IotLink_Container

lib/include/iot_linear_containers.h
```diff
@@ -0,0 +1,888 @@
...
+ * to calculate the pointer to pass to `freeElement`. This value should be calculated
+ * with the C `offsetof` macro. This parameter is ignored if `freeElement` is `NULL`
+ * or its value is `0`.
+ */
+/* @[declare_linear_containers_list_double_removeallmatches] */
+static inline void IotListDouble_RemoveAllMatches( IotListDouble_t * const pList,
+                                                   bool ( *isMatch )( const IotLink_t *, void * ),
+                                                   void * pMatch,
+                                                   void ( *freeElement )( void * ),
+                                                   size_t linkOffset )
```
lt72: See other comment RE: offset. Too many parameters increase the required stack size, and that is also a perf overhead since stack access may be one order of magnitude slower than register at each access.

PR #266, opened 2019-02-04
Allow calling wait after disconnect and reentrancy.

gordonwang0:


lib/source/mqtt/aws_iot_mqtt_api.c
```diff
@@ -1696,110 +1738,150 @@ AwsIotMqttError_t AwsIotMqtt_Wait( AwsIotMqttReference_t reference,
...
-        {
-            IotMutex_Lock( &( pOperation->pMqttConnection->timerMutex ) );
-            publishRetryActive = ( pOperation->pPublishRetry != NULL );
-            IotMutex_Unlock( &( pOperation->pMqttConnection->timerMutex ) );
-        }
+        /* Check any status set by the send thread. Block the receive callback
+         * during this check by locking the mutex for operations pending responses. */
+        IotMutex_Lock( &( _IotMqttPendingResponseMutex ) );
+        status = pOperation->status;
+        IotMutex_Unlock( &( _IotMqttPendingResponseMutex ) );
```
lt72: After you release the mutex, your status variable can be changed, and there s really no reason therefore to read it under a lock.
Different would be if you read it AND update it under lock, so that any thread that comes after will take a different path. But that is not the case here...

lib/source/mqtt/aws_iot_mqtt_api.c
```diff
@@ -1696,110 +1738,150 @@ AwsIotMqttError_t AwsIotMqtt_Wait( AwsIotMqttReference_t reference,
...
 
     /* Check reference. */
     if( AwsIotMqttInternal_ValidateReference( reference ) == false )
     {
         return AWS_IOT_MQTT_BAD_PARAMETER;
     }
 
-    IotLogInfo( "Waiting for operation %s to complete.",
-                AwsIotMqtt_OperationType( pOperation->operation ) );
+    /* Lock the connection mutex to prevent the connection from being closed. */
```
lt72: You are making a blocking call holding a mutex. That is a no-no (that is why POSIX has condition variables and waiting on a conditon variable releases the mutex and waits atomically). You may have a situation where the thread that will resolve the wait may need the same connection mutex. So this code will be difficult to maintain. The way to do this is to change the state of the connection and/or the outstanding operations for this connection.

You can implement without holding alock in this way:
1) if DISCONNECT comes before WAIT: you can set a flag in the operation that says it is CANCeLLED. Inside disconnect, you can clean up all operations that are NOT waitable (and therefore are NOT being waited on already, and never will be) and all operations that are not being waited on actively. You can mark the waitable operations as CANCELLED. You just need the memory for the operation and the flag that says CANCELLED. You can aleady de-allocate the semaphore, if it practical to do so. You can place the survivor operations in a zombie list or keep them in the same list. The invariant is that, AFTER disconnet comes it, all operatins in the operation list are waitable and marked as CANCeLLED. Or all operations in the zombie list are waitable and CANCeLLED.
The zombie list can be global. 
When WAIT is called, the connection is gone and  the operation is marked as CANCeLLED. WAIT will check that flag, so WAIT just goes through the list, and cleanup up the remaining memory for the operation being waited on, and return error OPERATION_CANCELED to teh user, or somethgn like that.

2) If WAIT come before DISCONECT: Then you can set a flag in the operation the user wants to wait on that says that the operation is being WAITED. You will clear that flag only when wait completes. When DISCONNECT comes in, you will clean up all operations that are NOT waitable (and therefore are NOT being waited on already, and never will be) and all operations that are not being waited on actively. You can mark the waitable (but not waited on) operations as CANCELLED, adn clean up what you can, or not, just as above. For all operations that are already marked as WAITED, you set the flag that says CANCELLED in the opeation, and signal the semaphore. wait will wake up, see the flag, let theuser know the bad news, and clean up all memory. The user should not touch the operation after WAIT returns with OPERATION_CANCELLED.


the only things you need a mutex for is setting the flags and removing operations from the lists.


Makes sense?

lib/source/mqtt/aws_iot_mqtt_operation.c
```diff
@@ -580,25 +595,65 @@ void AwsIotMqttInternal_DestroyOperation( void * pData )
...
         IotSemaphore_Destroy( &( pOperation->notify.waitSemaphore ) );
+
+        /* Decrement the MQTT connection's count of waitable operations. Ignore
+         * operations which are tied to the connection and do not modify
+         * the reference count: CONNECT, DISCONNECT, and PINGREQ. */
+        if( ( pOperation->operation != AWS_IOT_MQTT_CONNECT ) &&
+            ( pOperation->operation != AWS_IOT_MQTT_DISCONNECT ) &&
+            ( pOperation->operation != AWS_IOT_MQTT_PINGREQ ) )
+        {
+            if( lockMutexes == true )
```
lt72: This code is repeated elsewhere as well, so an helper method such as _decreaseConnectionRefCount would be good.

PR #267, opened 2019-02-05
Add priority and stack size to thread creation.

gordonwang0:

### DIFF
--- a/platform/source/posix/iot_threads_posix.c
+++ b/platform/source/posix/iot_threads_posix.c
@@ -137,23 +137,75 @@ static void * _threadRoutineWrapper( void * pArgument )
 /*-----------------------------------------------------------*/
 
 bool Iot_CreateDetachedThread( IotThreadRoutine_t threadRoutine,
-                               void * pArgument )
+                               void * pArgument,
+                               int32_t priority,
+                               size_t stackSize )
 {
     bool status = true;
     int posixErrno = 0;
+    _threadInfo_t * pThreadInfo = NULL;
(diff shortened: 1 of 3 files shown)


platform/source/posix/iot_threads_posix.c
```diff
@@ -179,10 +231,69 @@ bool Iot_CreateDetachedThread( IotThreadRoutine_t threadRoutine,
...
+
+                            status = false;
+                        }
+                    }
+                }
+            }
+
+            if( status == false )
+            {
+                ( void ) pthread_attr_destroy( &threadAttributes );
```
cobusve: I do not know this well enough to understand what this destroy call is doing, the variable is on the stack and will be cleaned up and I did not see any calls which seemed to obviously allocate anything. This may be fine, just consider if this is required.

platform/source/posix/iot_threads_posix.c
```diff
@@ -137,23 +137,75 @@ static void * _threadRoutineWrapper( void * pArgument )
...
-                               void * pArgument )
+                               void * pArgument,
+                               int32_t priority,
+                               size_t stackSize )
 {
     bool status = true;
     int posixErrno = 0;
+    _threadInfo_t * pThreadInfo = NULL;
     pthread_t newThread;
     pthread_attr_t threadAttributes;
```
cobusve: This definately needs to be a pointer and not an instance on the stack if you look at init and destroy. This will cause WEIRD behavior

PR #304, opened 2019-03-06
update defender doc and add AFR specific code

qiutongs:
*Issue #, if available:*

*Description of changes:*
- update defender doxygen document
- add AFR specific code so that both AFR and C-SDK share exactly same source code for defender


qiutongs:
> Although this
> #ifdef _DEFENDER_ON_AMAZON_FREERTOS
> may be needed now, we should strive to eliminate it.
> What controls this #define?

I totally agree. I just don't have better solution for now.

I want to hide this macro _DEFENDER_ON_AMAZON_FREERTOS in defender internal header. What do you think?

lt72:
That could be a good place for now, ultimately this should be controlled by cmake, if it is necessary.

qiutongs:
> That could be a good place for now, ultimately this should be controlled by cmake, if it is necessary.

That makes sense. Thanks for suggestion.

PR #313, opened 2019-03-13
atomic & its abstraction layer. 

yuhui-zheng:
*Issue #, if available:*
N/A

*Description of changes:*
- atomic operation default implementation for GCC, and user extension (macro ```COMPILER_OPTION_GCC_ATOMIC_BUILTIN```, header ```atomic_user_override.h```).
- API test cases for atomic.
- an implementation update on ```_nextPacketIdentifier()``` in ```iot_mqtt_serialize.c```.
  ```IOT_ATOMIC_OPERATION``` macro can be used to flip between original implemnetation and atomic approach.

With this PR, we are consistent with what we have in FreeRTOS PR. Things might be updated in next PR, and thus no need to comment:
- Test cases for atomic arithmetic APIs (add, sub, inc, dec) operating on signed integers. Detailed examples need to be given to guide how casting can be done in user code.
- Documentation for (1) how to flip between atomic implementation and original implementation; (2) ```compiler.h``` is a temporary solution, macros inside which need to go into ```CMakeLists.txt```. 


lib/include/platform/atomic_user_override_template.h
```diff
@@ -0,0 +1,129 @@
+/*
```
yuhui-zheng: This file provides a template for user. It will not compile, which is intentional.

lib/include/platform/atomic.h
```diff
@@ -0,0 +1,294 @@
...
+ * (e.g. ARM, Xtensa, ... )
+ * there's no reason to not use GCC built-in atomics.
+ * Thus, no user extension is provided in this branch.
+ *
+ * If you believe atomic built-in is not compiled correctly by GCC for your target,
+ * please use the other branch and provide your own implementation for atomic.
+ *
+ * attribute flatten is not needed in this branch. 
+ */
+    #define FORCE_INLINE_FLATTEN    inline __attribute__( ( always_inline ) )
```
yuhui-zheng: A note for Lorenzo -- flatten is removed here, since with gcc built-in (default implementation), flatten does not affect the compilation result. The macro name still has "FLATTEN" in it, and user needs to make sure nested function calls are flattened and inlined.

lt72: I perused other code bases, and it seems nobody uses flatten. I cannot find any evidence as to why though. Should we remove it for now? Seems confusing if it does not had value... we could instead add a comment on/

lib/include/platform/atomic.h
```diff
@@ -0,0 +1,294 @@
...
+ * Atomic add
+ *
+ * @brief Adds count to the value of the specified 32-bit variable as an atomic operation.
+ *
+ * @param[in,out] pAddend  Pointer to memory location from where value is to be loaded and written back to.
+ * @param[in] lCount       Value to be added to *pAddend.
+ *
+ * @return previous *pAddend value.
+ */
+static FORCE_INLINE_FLATTEN int32_t Atomic_Add_i32( int32_t volatile * pAddend,
```
lt72: Here we need to complete our evaluation of signed- vs. unsigned- integers...

lib/include/platform/atomic.h
yuhui-zheng: Agreed. I'll updated. Comments will be left in ```atomic_user_override_template.h```, saying user needs to check inline behavior in disassembly since there's a function call. 

Limited example I can find -- intel graphic driver (SandyBridge's New Acceleration component) https://github.com/intel/iotg-lin-gfx-ddx/blob/master/src/sna/sna_glyphs.c#L583. Their [README](https://github.com/intel/iotg-lin-gfx-ddx/blob/master/src/sna/README) explained a little bit about the overall design. (I won't pretend to have a full understanding of stuffs, but it seems the goal is to have specific hw acceleration behavior.)

lib/include/platform/atomic.h
yuhui-zheng: updated per offline discussion.

CMakeLists.txt
```diff
@@ -25,7 +25,16 @@ endif()
 set( CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin )
 set( CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib )
 
+# Define macro to use atomic implemnetation with GCC built-in.
+# Up until GCC 4.6.3, __sync_* shall be called. From 4.7.0, 
+# __atomic_* are available. We only implemented atomic
+# with __atomic_*.
+#
+# todo: check compiler is GCC AND GCC version is above 4.7.0.
```
yuhui-zheng: one thing at a time. this is a solid todo.

We require CMake min version 3.5. A couple of macros I tried, haven't found a good solution. 
```CMAKE_COMPILER_IS_GNUCC``` --> works
```CMAKE_CXX_COMPILER_VERSION``` --> doesn't work
```CMAKE_CXX_COMPILER_ID``` --> doesn't work.

Comment (gordonwang0) [reply CMakeLists.txt]: Try `CMAKE_C_COMPILER_VERSION` and `CMAKE_C_COMPILER_ID`

(`_CXX_` is for C++ compilers)

File: CMakeLists.txt [hunk 5]
```diff
@@ -25,6 +25,18 @@ endif()
...
+add_definitions( -DCOMPILER_OPTION_GCC_ATOMIC_BUILTIN=1 )
+
+# Whether to use atomic or not.
+# Default to "no".
+add_definitions( -DIOT_ATOMIC_OPERATION=0 )
```
gordonwang0: I think it should default to enabled (with compiler support).

lib/include/platform/atomic.h
```diff
@@ -0,0 +1,299 @@
+/*
```
gordonwang0: Consider moving this file to platform/, as it seems specific to gcc.

tests/common/unit/iot_tests_atomic.c
```diff
@@ -0,0 +1,364 @@
...
+    TEST_ASSERT_MESSAGE( ulOp1 == 0x55555555, "Atomic_XOR_u32 -- did not XORed correctly." );
+    TEST_ASSERT_MESSAGE( ulReturnValue == MAGIC_NUMBER_32BIT_1, "Atomic_XOR_u32 -- expected return value 0xA5A5A5A5." );
+}
+
+TEST( Common_Unit_Atomic, AtomicCasFailToSwap )
+{
+    uint32_t ulCasDestination_32;
+    uint32_t ulCasComparator_32;
+    uint32_t ulCasNewValue_32;
+    bool bExecutinoStatus;
```
gordonwang0: typo: execution (also naming)

CMakeLists.txt
gordonwang0: Also consider moving this out of the root CMakeLists.txt.  Since atomic is in platform, it should go in the platform CMakeLists.txt.

lt72: +1

CMakeLists.txt
lt72: +1

lib/include/platform/atomic.h
yuhui-zheng: ```platform/include/ ``` could be a potential candidate. This folder only has ```posix``` dir, where atomic doesn't fit. Thoughts on how to organize them? Or ```platform/include/atomic.h``` is good enough for now?

CMakeLists.txt
yuhui-zheng: Is the plan to soft launch atomic with this PR? If yes, I'll flip it to 1. 

This macro controls whether libraries depend on atomic. For now, the only place we use atomic in library is MQTT packet identifier, which is more like an example application scenario.

CMakeLists.txt
yuhui-zheng: Thanks. Will update to below once we finalize where to put atomic.h.
```
if ( CMAKE_COMPILER_IS_GNUCC AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 4.7 )
    add_definitions( -DCOMPILER_OPTION_GCC_ATOMIC_BUILTIN=1 )
endif()
```

gordonwang0:
Can you fix the compiler warnings in the CI?  (Only the ones for the atomic files, you can ignore warnings in other files).

PR #320, opened 2019-03-17
CBMC proofs fixed in qiutongs/metrics_fix branch

markrtuttle:
Fixing the CBMC proofs to use the new API for deserialization.


gordonwang0:
This appears to be identical to what already in the integration branch?

`qiutongs/metrics_fix` is a temporary branch.  I'm going to delete it now.

markrtuttle:
That is correct:  The new branch was generating proof failures, and the fix for the integration branch was enough to fix this branch.

PR #388, opened 2019-04-22
Fix possible overflow in PUBLISH packet size calculation

gordonwang0:


markrtuttle:
This patch does avoid the buffer overflow.  But what about defining the helper function
```
static bool _safe_accumulation(size_t *sum, size_t val) {
  if (*sum > _MQTT_MAX_REMAINING_LENGTH)
    return false;
  if ( val > _MQTT_MAX_REMAINING_LENGTH - *sum)
    return false;
  *sum += val;
  return true;
}
```
and then replacing in the original code every publishPacketSize increment of form
```
publishPacketSize += pPublishInfo->topicNameLength + sizeof( uint16_t );
```
with
```
if (!_safe_accumulation(&publishPacketSize, pPublishInfo->topicNameLength + sizeof( uint16_t ))
  status = false
```
Then the code would be absolutely bullet proof and possibly easier to read.

PR #405, opened 2019-05-06
Move Defender metrics into platform

gordonwang0:


lib/include/types/iot_platform_types.h
```diff
@@ -30,6 +30,9 @@
 /* The config header is always included first. */
 #include "iot_config.h"
 
+/* Linear containers (lists and queues) include for metrics types. */
+#include "iot_linear_containers.h"
```
lt72: Is this the right spot for this header?  (same question for metrics types down below...)

gordonwang0: The type `IotLink_t` is needed for the metrics structure.

PR #425, opened 2019-05-14
Add aborts to functions that shouldn't fail

gordonwang0:


lt72:
Mmmmmhhhh, this change seems unnecessary, we should focus on the release, hence board testing.
Also, abort() should be hidden behind a macro, and in debug builds should be a log+assert. In retail should return error and continue.
Calling abort() in the code just injects yet another dependency on c- runtime. We need less of those, not more. Also, we need to be consistent across the e tire code base, and this change does not look like it does that. Please change it.
Let's close the fit&fi ish of the c-sdk here and focus on the release this sprint. Fix a couple of boards, maybe 3 or 4, then resume...  let's ship this thing.



On May 13, 2019 19:19, Gordon Wang <notifications@github.com> wrote:
________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/425

Commit Summary

  *   Add abort calls to POSIX clock
  *   Update iot_threads_posix.c

File Changes

  *   M platform/source/posix/iot_clock_posix.c<https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/425/files#diff-0> (20)
  *   M platform/source/posix/iot_threads_posix.c<https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/425/files#diff-1> (54)

Patch Links:

  *   https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/425.patch
  *   https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/425.diff

gordonwang0:
This change was made to the POSIX platform layer only, and systems implementing POSIX will have the standard library.  I placed the aborts in code that should never be hit -- if it is, something is badly wrong.

PR #521, opened 2019-07-16
Flush incoming data on MQTT receive malloc failure

gordonwang0:


tests/mqtt/unit/iot_tests_mqtt_receive.c
```diff
@@ -775,6 +776,47 @@ TEST( MQTT_Unit_Receive, InvalidPacket )
 
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Tests the behavior of @ref mqtt_function_receivecallback when memory
+ * allocation fails.
+ */
+TEST( MQTT_Unit_Receive, ReceiveMallocFail )
```
qiutongs: How does this test prove the "_flushPacket" is invoked?

lib/source/mqtt/iot_mqtt_network.c
```diff
@@ -727,6 +751,23 @@ static void _sendPuback( _mqttConnection_t * pMqttConnection,
...
+static void _flushPacket( void * pNetworkConnection,
+                          const _mqttConnection_t * pMqttConnection,
+                          size_t length )
+{
+    size_t bytesFlushed = 0;
+    uint8_t receivedByte = 0;
+
+    for( bytesFlushed = 0; bytesFlushed < length; bytesFlushed++ )
+    {
+        ( void ) _IotMqtt_GetNextByte( pNetworkConnection,
```
qiutongs: Is the returning status important here?

tests/mqtt/unit/iot_tests_mqtt_receive.c
qiutongs: My question is why this test would fail with the code before this change.

gordonwang0: If _flushPacket were not invoked, then the second call to the ReceiveCallback would fail because a partial, invalid packet would be left in the data stream.

lib/source/mqtt/iot_mqtt_network.c
gordonwang0: No, can't do anything because there's no memory to process the packet.  It's discarded so the next packet may be processed.

tests/mqtt/unit/iot_tests_mqtt_receive.c
qiutongs: Make sense.

PR #551, opened 2019-09-09
fix:aws defender unit and integration tests

abhidixi11:
Fixed compilation issues for windows port for defender test project

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/platform/network/iot_network_mbedtls.c
+++ b/platform/network/iot_network_mbedtls.c
@@ -32,6 +32,12 @@
 #include <stdio.h>
 #include <string.h>
 
+#if !defined( _WIN32 ) && !defined( _WIN64 )
+    #include <arpa/inet.h>
+    #include <netdb.h>
+    #include <sys/socket.h>
+#endif
+
 /* mbed TLS network include. */
 #include "iot_network_mbedtls.h"
 
@@ -140,6 +146,7 @@
 #define FLAG_SECURED                 ( 0x00000001UL ) /**< @brief Secured connection. */
 #define FLAG_HAS_RECEIVE_CALLBACK    ( 0x00000002UL ) /**< @brief Connection has receive callback. */
 #define FLAG_CLOSED                  ( 0x00000004UL ) /**< @brief Connection is closed. */
+
 
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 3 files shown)


platform/network/iot_network_mbedtls.c
```diff
@@ -1192,6 +1201,90 @@ IotNetworkError_t IotNetworkMbedtls_Destroy( void * pConnection )
...
+    struct sockaddr_storage server = { 0 };
+    size_t length = sizeof(struct sockaddr_storage);
+    const void * pServerAddress = NULL;
+    char * pAddressStart = NULL;
+    const char * pPortFormat = NULL;
+    uint16_t remotePort = 0;
+    size_t addressLength = 0;
+
+    /* Cast function parameter to correct type. */
+    _networkConnection_t * const pNetworkConnection = pConnection;
```
leegeth: better to have a Null check for pConnection before proceeding

platform/network/iot_network_mbedtls.c
```diff
@@ -1192,6 +1201,90 @@ IotNetworkError_t IotNetworkMbedtls_Destroy( void * pConnection )
...
+        if (inet_ntop(server.ss_family,
+            pServerAddress,
+            pAddressStart,
+            addressLength) != NULL)
+        {
+            /* Add the port to the end of the address. */
+            addressLength = strlen(pServerInfo->pRemoteAddress);
+
+            portLength = snprintf(&(pServerInfo->pRemoteAddress[addressLength]),
+                7,
```
leegeth: a constant would be better here

platform/network/iot_network_mbedtls.c
```diff
@@ -32,6 +32,15 @@
 #include <stdio.h>
 #include <string.h>
 
+#if defined( __GNUC__ )
```
gordonwang0: Would probably be better to use OS-specific (instead of compiler defines).

```suggestion
#ifndef( _WIN32 )
```

(Windows being the only OS we support that needs different headers)

platform/network/iot_network_mbedtls.c
```diff
@@ -32,6 +32,15 @@
...
+#include <netdb.h>
+#include <arpa/inet.h>
+#else
+/* Windows sockets */
+#include <winsock2.h>
```
gordonwang0: These don't need to be included by this header.  They're already handled by the porting files
```suggestion
```

platform/network/iot_network_mbedtls.c
```diff
@@ -32,6 +32,15 @@
...
+#include <ws2tcpip.h>
```
gordonwang0:
```suggestion
```

tests/defender/system/aws_iot_tests_defender_system.c
```diff
@@ -28,8 +28,13 @@
 #include <stdio.h>
 
 /* Network includes. */
-#include <netdb.h>
-#include <arpa/inet.h>
+#if defined( __GNUC__ )
```
gordonwang0: See comment in iot_network_mbedtls.c

Also: This file should be cross-platform, so must not use any platform-specific includes.

tests/defender/system/aws_iot_tests_defender_system.c
```diff
@@ -911,17 +916,34 @@ static void _verifyTcpConections( int total,
 static char * _getIotAddress( void )
```
gordonwang0: This function should be re-implemented using the Defender metrics function to keep it independent of the network stack.

gordonwang0:
Please also run the code formatting tool

platform/network/iot_network_mbedtls.c
abhidixi11: This is not the header but C file, I have not added any WSL headers in other headers.

platform/network/iot_network_mbedtls.c
abhidixi11: Pleas see comment above.

tests/defender/system/aws_iot_tests_defender_system.c
abhidixi11: OK will change to #ifndef( _WIN32 ).

tests/defender/system/aws_iot_tests_defender_system.c
abhidixi11: Note: Currently in freertos repository, it uses secure sockets API, which will also need change when we want to go sub-module route.

platform/network/iot_network_mbedtls.c
abhidixi11: OK will change that, was not sure which one to use.

tests/defender/system/aws_iot_tests_defender_system.c
abhidixi11: Removed the test of verifying server address as discussed.

platform/ports/posix/types/iot_platform_types_posix.h
```diff
@@ -32,6 +32,9 @@
     #include POSIX_TYPES_HEADER
 #else
     #include <sys/types.h>
+    #include <arpa/inet.h>
```
gordonwang0: This won't work for the mac OS port, which doesn't include this file.

Can you move these to iot_network_mbedtls.c?
```c
#if !defined( _WIN32 ) && !defined( _WIN64 )
    #include <arpa/inet.h>
    #include <netdb.h>
    #include <sys/socket.h>
#endif
```

tests/defender/system/aws_iot_tests_defender_system.c
```diff
@@ -911,17 +884,34 @@ static void _verifyTcpConections( int total,
...
```
gordonwang0: This function can be deleted now, correct?

platform/network/iot_network_mbedtls.c
```diff
@@ -141,6 +147,7 @@
 #define FLAG_HAS_RECEIVE_CALLBACK    ( 0x00000002UL ) /**< @brief Connection has receive callback. */
 #define FLAG_CLOSED                  ( 0x00000004UL ) /**< @brief Connection is closed. */
 
+
```
gordonwang0: Extra newline?

PR #563, opened 2019-09-16
Fix Defender Compilation warnings

abhidixi11:
*Issue #, if available:*

*Description of changes:*


gordonwang0:
After looking at what that function does (generate topics), you could replace it with this implementation:

https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/1f89a4872c5ea99a031795cdeec1b4831ae8e442/lib/include/private/aws_iot.h#L261

It's a common function used to generate topics for the libraries.

PR #583, opened 2019-09-30
fix: Re-architect Defender.

abhidixi11:
Use shared MQTT connection and async MQTT calls for Defender library.

Before this change, Defender library created and maintained its own
MQTT connection. Use of shared MQTT connection will save memory and
processing time.
Defender library used taskpool task which made blocking calls to
MQTT library. Because MQTT library also uses taskpool, blocking calls
may cause deadlock if limited number of taskpool tasks are available..

*Issue #, if available:*

*Description of changes:*


lib/include/aws_iot_defender.h
```diff
@@ -95,8 +95,10 @@
  * @brief Intializers of data handles.
  */
 /**@{ */
-#define AWS_IOT_DEFENDER_START_INFO_INITIALIZER \
-    { .mqttNetworkInfo = { 0 }                  \
+#define AWS_IOT_DEFENDER_START_INFO_INITIALIZER         \
+    { .pClientIdentifier = NULL,                        \
+      .clientIdentifierLength = 0,                      \
+      .mqttConnection = IOT_MQTT_CONNECTION_INITIALIZER \
```
jainamt: ```AwsIotDefenderStartInfo_t``` has ```AwsIotDefenderCallback_t callback;``` as a member as well. Shouldn't ```AWS_IOT_DEFENDER_START_INFO_INITIALIZER``` initialize it too ?

demos/source/aws_iot_demo_defender.c
```diff
@@ -212,35 +250,78 @@ static AwsIotDefenderError_t _defenderDemo( const char * pIdentifier,
...
+                                      &_mqttConnection );
+
+    if( mqttError != IOT_MQTT_SUCCESS )
+    {
+        IotLogError( "Failed to Create MQTT Connection" );
+        /* TODO map to defender error */
+    }
+    else
+    {
+        startInfo.mqttConnection = _mqttConnection;
```
jainamt: Parts of ```startInfo``` are initialized in different places. Can this be consolidated together so that its easier to read ?

lib/include/private/aws_iot_defender_internal.h
```diff
@@ -272,6 +272,13 @@
     ( millisecondValue ) / 1000                \
 
 
+/**
+ * MQTT connection mock for unit testing defender API.
+ * This mock will be only used inside defender MQTT wrapper.
+ */
+
+#define DEFENDER_MQTT_CONNECTION_MOCK    ( ( IotMqttConnection_t ) 0xDEF30CED )
+
```
jainamt: Shouldn't this be wrapped with some sort of TEST flag?

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -126,17 +168,18 @@ AwsIotDefenderError_t AwsIotDefender_SetMetrics( AwsIotDefenderMetricsGroup_t me
 
 AwsIotDefenderError_t AwsIotDefender_Start( AwsIotDefenderStartInfo_t * pStartInfo )
 {
-    if( pStartInfo == NULL )
-    {
-        IotLogError( "Input start info is invalid." );
+    IOT_FUNCTION_ENTRY( AwsIotDefenderError_t, AWS_IOT_DEFENDER_SUCCESS );
 
-        return AWS_IOT_DEFENDER_INVALID_INPUT;
+    if( ( pStartInfo == NULL ) || ( pStartInfo->mqttConnection == IOT_MQTT_CONNECTION_INITIALIZER ) )
```
jainamt: how about if ```callback``` is not set?

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -367,120 +414,134 @@ static void _metricsPublishRoutine( IotTaskPool_t pTaskPool,
...
+    reportCreated = AwsIotDefenderInternal_CreateReport();
+
+    /* If Report is created successfully. */
+    if( reportCreated )
     {
-        AwsIotDefenderEventType_t eventType;
+        /* Publish report to defender topic. */
+        mqttError = AwsIotDefenderInternal_MqttPublish( AwsIotDefenderInternal_GetReportBuffer(),
+                                                        AwsIotDefenderInternal_GetReportBufferSize(),
+                                                        &_mqttCallbackInfo );
```
jainamt: where was ```_mqttCallbackInfo ``` set? Should it be cleared off in ```AwsIotDefender_Stop```?

lib/include/private/aws_iot_defender_internal.h
abhidixi11: Yes, I didn't get time to see if there is an existing test flag.   I was also waiting to see if this concept is OK. It avoids unnecessary connects to brokers during testing.

lib/source/defender/aws_iot_defender_api.c
abhidixi11: callback was optional in earlier API therefore I did not add that check:
https://docs.aws.amazon.com/freertos/latest/lib-ref/c-sdk/defender/structAwsIotDefenderStartInfo__t.html
Since we are changing that, we should make it mandatory.

lib/source/defender/aws_iot_defender_api.c
abhidixi11: _mqttCallbackInfo is global and initialized on line 98. It should not be cleared off as it is internal to library and contains callback processing for MQTT library async calls.

demos/source/aws_iot_demo_defender.c
```diff
@@ -62,6 +62,11 @@ static AwsIotDefenderError_t _startDefender( const char * pIdentifier,
                                              void * pNetworkCredentialInfo,
                                              const IotNetworkInterface_t * pNetworkInterface );
 
+/**
```
gordonwang0: Please take a look at the flow of the other demos, this demo should be similar.

demos/source/aws_iot_demo_defender.c
```diff
@@ -145,9 +158,34 @@ int RunDefenderDemo( bool awsIotMqttMode,
 void _defenderCallback( void * param1,
                         AwsIotDefenderCallbackInfo_t * const pCallbackInfo )
 {
+    const char * pEvent = NULL;
+
     ( void ) param1;
 
-    IotLogInfo( "User's callback is invoked on event %d.", pCallbackInfo->eventType );
+    /* Convert defent event to string - should this be made into API function ? */
+    switch( pCallbackInfo->eventType )
```
gordonwang0: Yes, this should be an API function.

lib/include/private/aws_iot_defender_internal.h
gordonwang0: There is already a mocked MQTT implementation.  See if you can use that.

lib/include/private/aws_iot_defender_internal.h
```diff
@@ -329,13 +336,27 @@ IotMqttError_t AwsIotDefenderInternal_MqttConnect( void );
  * Subscribe accept/reject defender topics.
  */
 IotMqttError_t AwsIotDefenderInternal_MqttSubscribe( IotMqttCallbackInfo_t acceptCallback,
-                                                     IotMqttCallbackInfo_t rejectCallback );
+                                                     IotMqttCallbackInfo_t rejectCallback,
```
gordonwang0: These function do not follow the naming guidelines for internal functions.

Also, these structs (IotMqttCallbackInfo_t) should not be passed by value.

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -236,48 +279,52 @@ void AwsIotDefender_Stop( void )
...
-    if( taskPoolError != IOT_TASKPOOL_SUCCESS )
-    {
-        IotLogWarn( "Failed to cancel metrics publish job with return code %d and status %d.", taskPoolError, status );
-        IotClock_SleepMs( WAIT_METRICS_JOB_MAX_SECONDS * 1000 );
-    }
+        /* If cancel failed, let it sleep for a while and hope everything finishes. */
+        if( taskPoolError != IOT_TASKPOOL_SUCCESS )
+        {
+            IotLogWarn( "Failed to cancel metrics publish job with return code %d and status %d.", taskPoolError, status );
+            IotClock_SleepMs( WAIT_METRICS_JOB_MAX_SECONDS * 1000 );
```
gordonwang0: This is not guaranteed to work... we should fix this at some point.

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -551,18 +589,105 @@ void _rejectCallback( void * pArgument,
...
+        _handleApplicationCallback( AWS_IOT_DEFENDER_FAILURE_MQTT, NULL );
+    }
+    else
+    {
+        switch( mqttOp )
+        {
+            case IOT_MQTT_SUBSCRIBE:
+                _subscribed = true;
+                /* Create metrics subscribe job. */
+                taskPoolError = IotTaskPool_CreateJob( _metricsPublishRoutine, NULL, &_metricsPublishJobStorage, &_metricsPublishJob );
```
gordonwang0: This is not safe.  MQTT is a network protocol, you could receive a duplicated response.  That would cause this callback to be invoked more than once, with both invocations modifying this global data.

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -551,18 +589,105 @@ void _rejectCallback( void * pArgument,
...
+                /* Silence warnigns when asserts are disabled. */
+                ( void ) taskPoolError;
+                AwsIotDefender_Assert( taskPoolError == IOT_TASKPOOL_SUCCESS );
+                /* Schedule Publish Job */
+                taskPoolError = IotTaskPool_Schedule( IOT_SYSTEM_TASKPOOL, _metricsPublishJob, 0 );
```
gordonwang0: Why is this scheduling a job (for immediate execution) to handle a nonblocking publish?  Can just do the nonblocking publish in this function.

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -551,18 +589,105 @@ void _rejectCallback( void * pArgument,
...
+                taskPoolError = IotTaskPool_CreateJob( _unsubscribeRoutine, NULL, &_unsubscribeJobStorage, &_unsubscribeJob );
+                /* Silence warnigns when asserts are disabled. */
+                ( void ) taskPoolError;
+                AwsIotDefender_Assert( taskPoolError == IOT_TASKPOOL_SUCCESS );
+                IotTaskPool_ScheduleDeferred( IOT_SYSTEM_TASKPOOL,
+                                              _unsubscribeJob,
+                                              _defenderToMilliseconds( AWS_IOT_DEFENDER_WAIT_SERVER_MAX_SECONDS ) );
+                break;
+
+            case IOT_MQTT_UNSUBSCRIBE:
```
gordonwang0: Constantly modifying subscriptions is significant overhead.  Currently, the sequence of jobs is:

Subscribe -> Publish -> Unsubscribe -> Subscribe -> Publish -> Unsubscribe, etc...

You don't have to unsubscribe if the next thing to do is to resubscribe to the same topic.

In the Defender_Start function, you should make blocking calls to add the subscriptions.  In Defender_Stop, you should make blocking calls to remove the subscriptions.  In between, your Jobs should only do Publishes.

lib/source/defender/aws_iot_defender_mqtt.c
```diff
@@ -113,19 +115,31 @@ void AwsIotDefenderInternal_DeleteTopicsNames( void )
 
 /*-----------------------------------------------------------*/
 
-IotMqttError_t AwsIotDefenderInternal_MqttConnect( void )
+void AwsIotDefenderInternal_SetMqttConnection( IotMqttConnection_t mqttConnection )
```
gordonwang0: This internal function doesn't seem to have much use.  It's setting an internal library variable, but you could just declare the variable in the library's internal header.

lib/source/defender/aws_iot_defender_mqtt.c
```diff
@@ -113,19 +115,31 @@ void AwsIotDefenderInternal_DeleteTopicsNames( void )
...
 IotMqttError_t AwsIotDefenderInternal_MqttSubscribe( IotMqttCallbackInfo_t acceptCallback,
-                                                     IotMqttCallbackInfo_t rejectCallback )
+                                                     IotMqttCallbackInfo_t rejectCallback,
+                                                     IotMqttCallbackInfo_t * pMqttCallbackInfo )
 {
+    IOT_FUNCTION_ENTRY( IotMqttError_t, IOT_MQTT_SUCCESS );
+    IotMqttOperation_t subscribeOperation = IOT_MQTT_OPERATION_INITIALIZER;
+
+    /* If it is mocked mqtt connection return success */
+    if( _mqttConnection == DEFENDER_MQTT_CONNECTION_MOCK )
```
gordonwang0: See the already-implemented MQTT mocks... Library code should not have test-specific code in it!

lib/source/defender/aws_iot_defender_mqtt.c
```diff
@@ -139,35 +153,100 @@ IotMqttError_t AwsIotDefenderInternal_MqttSubscribe( IotMqttCallbackInfo_t accep
     subscriptions[ 1 ].topicFilterLength = ( uint16_t ) strlen( _pRejectTopic );
     subscriptions[ 1 ].callback = rejectCallback;
 
-    return IotMqtt_SubscribeSync( _mqttConnection,
-                                  subscriptions,
-                                  2,
-                                  0,
-                                  _defenderToMilliseconds( AWS_IOT_DEFENDER_MQTT_SUBSCRIBE_TIMEOUT_SECONDS ) );
+    /* Call the asynchronous SUBSCRIBE function. */
+    status = IotMqtt_SubscribeAsync( _mqttConnection,
```
gordonwang0: Consider using the "+" wildcard so you don't have to make 2 subscriptions.  This halves the memory usage for subscriptions.  You will need to parse the topic to find what the wildcard represents, but there's already a function in aws_iot.h to do this.

There's also a function already in aws_iot.h that handles subscriptions.

lib/source/defender/aws_iot_defender_mqtt.c
```diff
@@ -139,35 +153,100 @@ IotMqttError_t AwsIotDefenderInternal_MqttSubscribe( IotMqttCallbackInfo_t accep
...
     publishInfo.topicNameLength = ( uint16_t ) strlen( _pPublishTopic );
     publishInfo.pPayload = pData;
     publishInfo.payloadLength = dataLength;
-
-    return IotMqtt_PublishSync( _mqttConnection,
-                                &publishInfo,
-                                0,
-                                _defenderToMilliseconds( AWS_IOT_DEFENDER_MQTT_PUBLISH_TIMEOUT_SECONDS ) );
+    /* PublishSync is non blocking for IOT_MQTT_QOS_0, therefore execute callback function after return */
+    status = IotMqtt_PublishSync( _mqttConnection,
```
gordonwang0: It may be better to allow the user to specify QoS 0 or 1

demos/source/aws_iot_demo_defender.c
abhidixi11: Ok will do and make changes if necessary.

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -179,26 +196,24 @@ AwsIotDefenderError_t AwsIotDefender_Start( AwsIotDefenderStartInfo_t * pStartIn
...
-            taskPoolError = IotTaskPool_CreateJob( _disconnectRoutine, NULL, &_disconnectJobStorage, &_disconnectJob );
-            AwsIotDefender_Assert( taskPoolError == IOT_TASKPOOL_SUCCESS );
+            /*  Subscribe to Metrics Publish topic */
+            _metricsSubscribeRoutine();
 
-            /* Create metrics job. */
+            /* Create metrics publish job, which will periodically publish metrics */
             taskPoolError = IotTaskPool_CreateJob( _metricsPublishRoutine, NULL, &_metricsPublishJobStorage, &_metricsPublishJob );
+            /* Silence warnigns when asserts are disabled. */
+            ( void ) taskPoolError;
```
aggarw13: If the `taskPoolError` value is not consumed at all, then it can just be removed. Instead every `IotTaskPool_*()` API call be `(void)`_ed_ to improve expressivity

demos/source/aws_iot_demo_defender.c
```diff
@@ -41,35 +41,142 @@
...
 /*-----------------------------------------------------------*/
 
+void _defenderCallback( void * pCallbackContext,
+                        AwsIotDefenderCallbackInfo_t * const pCallbackInfo )
+{
+    ( void ) pCallbackContext;
+
+    IotLogInfo( "User's callback is invoked on event: %s.", AwsIotDefender_EventType( pCallbackInfo->eventType ) );
+
+    /* Print out the sent metrics report if there is. */
```
gordonwang0: This comment isn't accurate, please update

demos/source/aws_iot_demo_defender.c
```diff
@@ -41,35 +41,142 @@
...
 /**
  * @brief Callback used to get notification of defender's events.
  */
-static void _defenderCallback( void * param1,
+static void _defenderCallback( void * pCallbackContext,
                                AwsIotDefenderCallbackInfo_t * const pCallbackInfo );
 
 /*-----------------------------------------------------------*/
 
+void _defenderCallback( void * pCallbackContext,
```
gordonwang0: Documentation is needed for this function and parameters.

demos/source/aws_iot_demo_defender.c
```diff
@@ -41,35 +41,142 @@
...
-                                             void * pNetworkServerInfo,
-                                             void * pNetworkCredentialInfo,
-                                             const IotNetworkInterface_t * pNetworkInterface );
+#define TIMEOUT_MS            ( 5000 )
 
 /**
  * @brief Callback used to get notification of defender's events.
  */
-static void _defenderCallback( void * param1,
+static void _defenderCallback( void * pCallbackContext,
```
gordonwang0: No declaration needed.

tests/jobs/unit/aws_iot_tests_jobs_api.c
```diff
@@ -805,7 +805,7 @@ TEST( Jobs_Unit_API, SetCallback )
  */
 TEST( Jobs_Unit_API, SetCallbackMultiple )
 {
-    int32_t i = 0;
+    intptr_t i = 0;
```
aggarw13: (Hygiene) Can this be changed to something more descriptive?

abhidixi11: That's jobs warning suppress, not related to this PR. I agree though that it should be changed for readability.

lib/source/defender/aws_iot_defender_mqtt.c
```diff
@@ -168,7 +180,26 @@ IotMqttError_t AwsIotDefenderInternal_MqttPublish( uint8_t * pData,
 
 /*-----------------------------------------------------------*/
 
-void AwsIotDefenderInternal_MqttDisconnect( void )
+IotMqttError_t AwsIotDefenderInternal_MqttUnSubscribe( void )
```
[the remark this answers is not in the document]
abhidixi11: These functions will change and we will update it later.

lib/source/defender/aws_iot_defender_mqtt.c
```diff
@@ -35,25 +35,32 @@
...
+extern IotMqttCallbackInfo_t _acceptCallbackInfo;
+extern IotMqttCallbackInfo_t _rejectCallbackInfo;
 extern AwsIotDefenderStartInfo_t _startInfo;
 
-/* defender internally manages network and mqtt connection */
+/* MQTT connection will be set by  application */
 static IotMqttConnection_t _mqttConnection = IOT_MQTT_CONNECTION_INITIALIZER;
 
 static char * _pPublishTopic = NULL;
+static size_t _publishTopicLength = 0;
```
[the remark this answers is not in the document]
abhidixi11: Since we are already computing the length, so instead of computing  it again, just use the length.

lib/include/aws_iot_defender.h
```diff
@@ -188,9 +199,10 @@ typedef struct AwsIotDefenderCallback
  */
 typedef struct AwsIotDefenderStartInfo
 {
-    IotMqttNetworkInfo_t mqttNetworkInfo;    /**< MQTT Network info used by defender (required). */
-    IotMqttConnectInfo_t mqttConnectionInfo; /**< MQTT connection info used by defender (required). */
-    AwsIotDefenderCallback_t callback;       /**< Callback function parameter (optional). */
+    const char * pClientIdentifier;     /**< @brief MQTT client identifier. */
+    uint16_t clientIdentifierLength;    /**< @brief Length of #IotMqttConnectInfo_t.pClientIdentifier. */
```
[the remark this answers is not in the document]
abhidixi11: We are typecasting it to uint16_t in demo example.  I don't see it as big issue, but good point. (It is coming from original AwsIotDefenderInternal_BuildTopicsNames() which uses uint16_t.)

lib/source/defender/aws_iot_defender_mqtt.c
```diff
@@ -153,13 +164,14 @@ IotMqttError_t AwsIotDefenderInternal_MqttPublish( uint8_t * pData,
                                                    size_t dataLength )
 {
     IotMqttPublishInfo_t publishInfo = IOT_MQTT_PUBLISH_INFO_INITIALIZER;
+    IotMqttCallbackParam_t callbackParams;
```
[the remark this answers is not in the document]
abhidixi11: Good catch, will remove it.

lib/include/private/aws_iot_defender_internal.h
```diff
@@ -321,23 +321,28 @@ AwsIotDefenderError_t AwsIotDefenderInternal_BuildTopicsNames( void );
...
 /**
  * Publish metrics data to defender topic.
  */
 IotMqttError_t AwsIotDefenderInternal_MqttPublish( uint8_t * pData,
                                                    size_t dataLength );
 
+/**
+ * Set shared MQTT connection
+ */
+void AwsIotDefenderInternal_SetMqttConnection( IotMqttConnection_t _mqttConnection );
```
aggarw13: Is this function valid anymore (now that the MQTT connection handle is _externed_)?

tests/defender/system/aws_iot_tests_defender_system.c
```diff
@@ -67,10 +68,11 @@
 #define NO_EVENT                             10000
 
 /* Empty callback structure passed to startInfo. */
-static const AwsIotDefenderCallback_t _EMPTY_CALLBACK = { .function = NULL, .param1 = NULL };
+static const AwsIotDefenderCallback_t _EMPTY_CALLBACK = { .function = NULL, .pCallbackContext = NULL };
```
aggarw13: The style guide mentions about use of capitals only for `#define`s and **enumerations**. (I had recently given a proposal of updating the style rule to do the same for `const` objects as well, however, that was turned down)

tests/defender/system/aws_iot_tests_defender_system.c
```diff
@@ -168,17 +174,8 @@ TEST_SETUP( Defender_System )
...
-    _startInfo.mqttNetworkInfo.u.setup.pNetworkServerInfo = &_serverInfo;
-    _startInfo.mqttNetworkInfo.u.setup.pNetworkCredentialInfo = &_credential;
-
-    _startInfo.mqttNetworkInfo.pNetworkInterface = IotNetworkMetrics_GetInterface();
-
-    _startInfo.mqttConnectionInfo = ( IotMqttConnectInfo_t ) IOT_MQTT_CONNECT_INFO_INITIALIZER;
-    _startInfo.mqttConnectionInfo.pClientIdentifier = AWS_IOT_TEST_SHADOW_THING_NAME;
-    _startInfo.mqttConnectionInfo.clientIdentifierLength = ( uint16_t ) strlen( AWS_IOT_TEST_SHADOW_THING_NAME );
-
+    _startInfo.pClientIdentifier = AWS_IOT_TEST_SHADOW_THING_NAME;
```
aggarw13: `SHADOW_THING_NAME` in a defender test (I know that it does not affect the tests!)? Can you define a `DEFENDER` thing name to avoid any confusion?

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -552,18 +556,39 @@ void _rejectCallback( void * pArgument,
...
+        callbackInfo.eventType = event;
 
         callbackInfo.pMetricsReport = AwsIotDefenderInternal_GetReportBuffer();
         callbackInfo.metricsReportLength = AwsIotDefenderInternal_GetReportBufferSize();
 
-        callbackInfo.pPayload = pPublish->u.message.info.pPayload;
-        callbackInfo.payloadLength = pPublish->u.message.info.payloadLength;
+        if( pPublish == NULL )
+        {
+            callbackInfo.pPayload = NULL;
```
aggarw13: (Minor) If there is an initializer macro for the `AwsIotDefenderCallbackInfo_t` type, then the values assigned in this branch can just be set at initialization

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -237,57 +271,58 @@ void AwsIotDefender_Stop( void )
...
+        
+        _unsubscribeMqtt();
+       
+        /* Destroy metrics' mutex. */
+        IotMutex_Destroy( &_AwsIotDefenderMetrics.mutex );
 
-    /* Destroy metrics' mutex. */
-    IotMutex_Destroy( &_AwsIotDefenderMetrics.mutex );
+        /* Post so that taken semaphore is returned */
+        IotSemaphore_Post( &_doneSem );
```
aggarw13: What's the advantage of **posting** to the semaphore when it is supposed to get destroyed in the next statement?

tests/defender/system/aws_iot_tests_defender_system.c
abhidixi11: That's correct. The earlier style guide had '_' in the beginning, so this define carried over from earlier code and therefore you are seeing styling guide mismatch.

lib/source/defender/aws_iot_defender_api.c
```diff
@@ -237,57 +271,58 @@ void AwsIotDefender_Stop( void )
...
-
-        return;
     }
-
-    /* Wait for all the metrics processing to be done, if there are outstanding requests */
-    if( IotSemaphore_GetCount( &_doneSem ) > MAX_DEFENDER_OUTSTANDING_PUBLISH_REQ )
+    else
     {
+        /* Wait for all the metrics processing to be done, if there are outstanding requests */
         IotSemaphore_Wait( &_doneSem );
```
aggarw13: Why is this a blocking call? Wouldn't this call be vulnerable of making the application thread go into an indefinite wait state?

lib/source/defender/aws_iot_defender_api.c
abhidixi11: Destroy will fail (and trigger crash in macOS) if semaphore count is not same as initial count, which is set 1 when it this semaphore is created.

lib/source/defender/aws_iot_defender_api.c
abhidixi11: When application calls stop, library needs to wait till taskpool thread has finished executing publish job, otherwise connection will be destroyed while it is being used.

PR #604, opened 2019-10-17
Remove macOS and Windows ports.

sarenameas:
*Issue #, if available:* N/A

*Description of changes:* Since the macOS and Windows ports are not officially supported, for this next release we need to temporarily remove them.

For internal employees: 
Please see: https://code.amazon.com/packages/AwsIotCsdkMacOsWindowsPorts/trees/mainline


muneebahmed10:
`scripts/ci_test_mqtt.sh` contains code for the Windows build on Travis, and `scripts/setup/` still contains files for Travis' macOS and Windows builds. They should be removed for neatness.

sarenameas:
> `scripts/ci_test_mqtt.sh` contains code for the Windows build on Travis, and `scripts/setup/` still contains files for Travis' macOS and Windows builds. They should be removed for neatness.

Thanks, I neglected the scripts folder

.gitignore
```diff
@@ -4,6 +4,3 @@ doc/tag/*
 
 # Ignore CMake build directory.
 build/
-
-# Ignore macOS metadata files.
-.DS_Store
```
muneebahmed10: I think this should be left here, as we can still use Macs to develop even if we don't test on them.

scripts/ci_test_build.sh
```diff
@@ -8,31 +8,13 @@ set -e
 # Treat warnings as errors.
 if [ "$TRAVIS_COMPILER" = "clang" ]; then
     COMPILER_OPTIONS+=" -Werror"
```
muneebahmed10: Nit: $TRAVIS_COMPILER should always be clang for this file and $TRAVIS_OS_NAME will always be "linux" since we're removing the ports. Might be neater to just remove the ifs.

.gitignore
sarenameas: Might as well remove all references.

scripts/ci_test_build.sh
sarenameas: It will be easier to add back in the ports if these stay untouched.

sarenameas:
Still in progress. Missed things in the doc/ folder

third_party/tinycbor/CMakeLists.txt
```diff
@@ -38,11 +38,6 @@ add_library( tinycbor
 target_include_directories( tinycbor SYSTEM
                             PUBLIC tinycbor/src )
 
-# Link math library. This is not needed on Windows.
-if( NOT ${CMAKE_SYSTEM_NAME} STREQUAL "Windows" )
-    target_link_libraries( tinycbor PRIVATE m )
-endif()
```
muneebahmed10: If this is needed when not on Windows, shouldn't it be left in? Or is it linked elsewhere?

.gitignore
muneebahmed10: I disagree since .DS_Store files are created by Finder accessing directories and are unrelated to the Mac port. It's a minor annoyance having them show up in git status, and a customer developing on a Mac might feel the same.

third_party/tinycbor/CMakeLists.txt
sarenameas: oops thanks

.gitignore
sarenameas: Then this would be a problem in every git repo, right? You should probably add it to a global .gitignore: git config --global core.excludesfile ~/.gitignore

CMakeLists.txt
```diff
@@ -50,18 +45,6 @@ if( NOT DEFINED IOT_PLATFORM_NAME )
         option( IOT_NETWORK_USE_OPENSSL
```
gordonwang0: Put an else() to exclude other systems now.

PR #603, opened 2019-10-15
Documentation cleanup/improved signature for IotNetworkInterface_t::create

gkwicker:
- Documentation cleanup
- Improved signature for IotNetworkInterface_t::create

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #607, opened 2019-10-18
Revert signature change to IotNetworkInterface_t::create, doc updates

gkwicker:
PR #603 changed the signature of `IotNetworkInterface_t::create()`.  After review we've decided to revert this and keep it as it was to reduce the chance of incorrect configuration when multiple network interfaces exist on a system.  This change also addresses some documentation issues that were raised for PR #603 after it had been merged.


libraries/standard/mqtt/include/types/iot_mqtt_types.h
```diff
@@ -998,7 +998,7 @@ typedef struct IotMqttNetworkInfo
              * interface when creating a new network connection. It is only valid when
              * #IotMqttNetworkInfo_t::createNetworkConnection is `true`.
              */
-            IotNetworkServerInfo_t * pNetworkServerInfo;
```
abhidixi11: Line 948-957 are not needed and should be removed.

libraries/standard/mqtt/include/iot_mqtt.h
```diff
@@ -807,8 +807,7 @@ const char * IotMqtt_OperationType( IotMqttOperationType_t operation );
  * @param[in] pTopicFilter The topic filter to check.
  * @param[in] topicFilterLength Length of `pTopicFilter`.
  * @param[out] pCurrentSubscription If a subscription is found, its details are
- * copied here. This output parameter is only valid if this function returns `true`.
- * Pass `NULL` to ignore.
+ * copied here. This output parameter is only valid if this function returns `true` (NULL to disable)
```
gordonwang0: Nit: period at end of comment, use code syntax for NULL: (`NULL` to disable).

libraries/standard/mqtt/include/types/iot_mqtt_types.h
gkwicker: Good catch, removed.

libraries/standard/mqtt/include/iot_mqtt.h
gkwicker: Added period here and changed to `NULL` throughout.

PR #608, opened 2019-10-18
Use #defines for max client ID & "remove all subscriptions" constants

gkwicker:
Use a #define for the MQTT maximum client ID length.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #610, opened 2019-10-21
Simplify IotMqtt_Connect(); fix max client ID length check.

gkwicker:
Simplify IotMqtt_Connect(); fix max client ID length check.


libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1036,86 +1089,19 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
...
-        else
-        {
-            EMPTY_ELSE_MARKER;
-        }
-    }
-    else
-    {
-        EMPTY_ELSE_MARKER;
-    }
+    networkCreated = _initializeNetworkConnection( pNetworkInfo,
```
gkwicker: LWT info is now validated in `_IotMqtt_ValidateConnect()`.

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -93,10 +108,11 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
     {
         if( pConnectInfo->pPreviousSubscriptions != NULL )
         {
-            if( pConnectInfo->previousSubscriptionCount == 0 )
+            if( _IotMqtt_ValidateSubscriptionList( IOT_MQTT_SUBSCRIBE,
```
gkwicker: The subscription count is already validated in [`_IotMqtt_ValidateSubscriptionList()`](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/iot_mqtt_validate.c#L344).

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -137,6 +171,7 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
                          "the maximum supported length of %hu.",
                          pConnectInfo->clientIdentifierLength,
                          maxClientIdLength );
+            IOT_SET_AND_GOTO_CLEANUP( false );
```
gkwicker: This was removed in PR #608, adding it back.

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -483,6 +486,18 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo );
...
                                const IotMqttPublishInfo_t * pPublishInfo );
 
+/**
+ * @brief Check that an #IotMqttPublishInfo_t is valid for an LWT publish
+ *
+ * @param[in] awsIotMqttMode Specifies if this PUBLISH packet is being sent to
+ * an AWS IoT MQTT server.
+ * @param[in] pPublishInfo The #IotMqttPublishInfo_t to validate.
+ *
+ * @return `true` if `pPublishInfo` is valid; `false` otherwise.
```
muneebahmed10: Parameter is named `pLwtPublishInfo` in the declaration

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -404,6 +418,55 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
 
 /*-----------------------------------------------------------*/
 
+static bool _initializeNetworkConnection( const IotMqttNetworkInfo_t * pNetworkInfo,
+                                          void ** pNetworkConnection,
+                                          bool * pCreatedNetworkConnection )
+{
+    IOT_FUNCTION_ENTRY( bool, true );
+    bool networkStatus = false;
```
muneebahmed10: Is `bool` the best type for this given that `pNetworkInterface->create()` returns an `IotNetworkError_t`?

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -34,6 +34,21 @@
 /* MQTT internal include. */
 #include "private/iot_mqtt_internal.h"
 
+/**
+ * @brief Check that an #IotMqttPublishInfo_t is valid
+ *
+ * @param[in] awsIotMqttMode Specifies if this PUBLISH packet is being sent to
+ * an AWS IoT MQTT server.
+ * @param[in] pPublishTypeDescription String describing the publish type.
```
muneebahmed10: Missing documentation for `maximumPayloadLength`

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
gkwicker: Fixed

libraries/standard/mqtt/src/iot_mqtt_validate.c
gkwicker: Fixed

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: It is not, good catch.  Reworked.

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -34,6 +34,22 @@
 /* MQTT internal include. */
 #include "private/iot_mqtt_internal.h"
 
+/**
+ * @brief Check that an #IotMqttPublishInfo_t is valid
```
muneebahmed10: Nit: period here too.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -111,6 +111,20 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
...
+ * @brief Initialize a network connection, creating it if necessary
+ *
+ * @param[in] pNetworkInfo User-provided network information for the connection
+ * connection.
+ * @param[out] pNetworkConnection On success, the created and/or initialized network connection.
+ * @param[out] pCreatedNetworkConnection On success, `true` if the network connection was created; `false` if it was only initialized.
+ *
+ * @return Any #IotNetworkError_t, as defined by the network stack.
+ */
+static IotNetworkError_t _initializeNetworkConnection( const IotMqttNetworkInfo_t * pNetworkInfo,
```
gordonwang0: Do you need the pCreatedNetworkConnection output variable?  Can't you determine if a network connection was created based on the return value?

Also nit: This function should be called `_createNetworkConnection`.  We use the terms `_create/_destroy` for objects, while `Init/Cleanup` refer to library initialization sequences (such as IotMqtt_Init/IotMqtt_Cleanup).

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -245,9 +247,10 @@
  *
  * Used to validate parameters if when connecting to an AWS IoT MQTT server.
  */
-#define AWS_IOT_MQTT_SERVER_MAX_CLIENTID_LENGTH                ( ( uint16_t ) 128 ) /**< @brief Maximum length of client identifier accepted by AWS IoT. */
-#define AWS_IOT_MQTT_SERVER_MAX_TOPIC_LENGTH                   ( 256 )              /**< @brief Maximum length of topic names or filters accepted by AWS IoT. */
-#define AWS_IOT_MQTT_SERVER_MAX_TOPIC_FILTERS_PER_SUBSCRIBE    ( 8 )                /**< @brief Maximum number of topic filters in a single SUBSCRIBE packet. */
+#define AWS_IOT_MQTT_SERVER_MAX_CLIENTID_LENGTH                ( ( uint16_t ) 128 )         /**< @brief Maximum length of client identifier accepted by AWS IoT. */
+#define AWS_IOT_MQTT_SERVER_MAX_TOPIC_LENGTH                   ( 256 )                      /**< @brief Maximum length of topic names or filters accepted by AWS IoT. */
+#define AWS_IOT_MQTT_SERVER_MAX_TOPIC_FILTERS_PER_SUBSCRIBE    ( 8 )                        /**< @brief Maximum number of topic filters in a single SUBSCRIBE packet. */
```
abhidixi11: Should we add data type for consistency since we are modifying the lines?
( ( uint16_t ) 256 ) and ( (uint16_t) 8)

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -404,6 +418,55 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
 
 /*-----------------------------------------------------------*/
 
+static IotNetworkError_t _initializeNetworkConnection( const IotMqttNetworkInfo_t * pNetworkInfo,
+                                                       void ** pNetworkConnection,
+                                                       bool * pCreatedNetworkConnection )
+{
+    IOT_FUNCTION_ENTRY( IotNetworkError_t, IOT_NETWORK_SUCCESS );
+    IotNetworkError_t networkStatus = IOT_NETWORK_SUCCESS;
```
gordonwang0: Why have a separate networkStatus variable?  One should be sufficient.

libraries/standard/mqtt/src/iot_mqtt_validate.c
gordonwang0: Nit: period at end of comment

libraries/standard/mqtt/src/iot_mqtt_validate.c
gordonwang0: Add newline before 175

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -137,6 +172,7 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
                          "the maximum supported length of %hu.",
                          pConnectInfo->clientIdentifierLength,
                          maxClientIdLength );
+            IOT_SET_AND_GOTO_CLEANUP( false );
         }
```
gordonwang0: Nit: add EMPTY_ELSE_MARKER where needed

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -148,9 +184,10 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
 }
 
 /*-----------------------------------------------------------*/
-
-bool _IotMqtt_ValidatePublish( bool awsIotMqttMode,
-                               const IotMqttPublishInfo_t * pPublishInfo )
+static bool _mqtt_ValidatePublish( bool awsIotMqttMode,
```
gordonwang0: This function does not follow any naming convention.

It should be called _validatePublish.

Also extra newline needed before 187.

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -282,6 +330,39 @@ bool _IotMqtt_ValidatePublish( bool awsIotMqttMode,
 
 /*-----------------------------------------------------------*/
 
+bool _IotMqtt_ValidatePublish( bool awsIotMqttMode,
```
gordonwang0: Only one function for validating publishes is needed.  The first parameter is `awsIotMqttMode` anyways, so the maximum payload length check should be done there.

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -235,7 +235,9 @@
     #define EMPTY_ELSE_MARKER
 #endif
 
-#define IOT_MQTT_SERVER_MAX_CLIENTID_LENGTH                    ( ( uint16_t ) 23 ) /**< @brief Optional maximum length of client identifier specified by MQTT 3.1.1. */
+#define IOT_MQTT_SERVER_MAX_CLIENTID_LENGTH                    ( ( uint16_t ) 23 )          /**< @brief Optional maximum length of client identifier specified by MQTT 3.1.1. */
+#define IOT_MQTT_SERVER_MAX_PUBLISH_PAYLOAD_LENGTH             ( ( uint32_t ) ( 1 << 28 ) ) /**< @brief Maximum publish payload length supported by MQTT 3.1.1. */
```
gordonwang0: Could you change this to `268435456`.  It's easier for the reader than `1<<28`.

Same for `1<<17` below.

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -235,7 +235,9 @@
     #define EMPTY_ELSE_MARKER
 #endif
 
-#define IOT_MQTT_SERVER_MAX_CLIENTID_LENGTH                    ( ( uint16_t ) 23 ) /**< @brief Optional maximum length of client identifier specified by MQTT 3.1.1. */
+#define IOT_MQTT_SERVER_MAX_CLIENTID_LENGTH                    ( ( uint16_t ) 23 )          /**< @brief Optional maximum length of client identifier specified by MQTT 3.1.1. */
```
gordonwang0: I think we can shorten these macro names.

So just `MQTT_MAX_CLIENTID_LENGTH`, `MQTT_MAX_PUBLISH_LENGTH`, etc.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -404,6 +418,55 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
...
+
+        if( networkStatus == IOT_NETWORK_SUCCESS )
+        {
+            /* This MQTT connection owns the network connection it created and
+             * should destroy it on cleanup. */
+            *pCreatedNetworkConnection = true;
+        }
+        else
+        {
+            IOT_SET_AND_GOTO_CLEANUP( networkStatus );
```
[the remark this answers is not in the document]
abhidixi11: +1 when _initializeNetworkConnection() fails we are not logging an error either (line 1098), we should log error here for easier debugging.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -404,6 +418,55 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
...
+                                                       void ** pNetworkConnection,
+                                                       bool * pCreatedNetworkConnection )
+{
+    IOT_FUNCTION_ENTRY( IotNetworkError_t, IOT_NETWORK_SUCCESS );
+    IotNetworkError_t networkStatus = IOT_NETWORK_SUCCESS;
+
+    /* Network info must not be NULL. */
+    if( pNetworkInfo == NULL )
+    {
+        IOT_SET_AND_GOTO_CLEANUP( false );
```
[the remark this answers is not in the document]
abhidixi11: May be log error here as well.

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
gkwicker: Done.   Also, `AWS_IOT_MQTT_SERVER_MAX_TOPIC_FILTERS_PER_SUBSCRIBE` & `AWS_IOT_MQTT_SERVER_MAX_TOPIC_FILTERS_PER_SUBSCRIBE` are better represented with `size_t`.

libraries/standard/mqtt/src/iot_mqtt_validate.c
gkwicker: Fixed

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -137,6 +172,7 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
                          "the maximum supported length of %hu.",
                          pConnectInfo->clientIdentifierLength,
                          maxClientIdLength );
+            IOT_SET_AND_GOTO_CLEANUP( false );
```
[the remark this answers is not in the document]
gkwicker: Fixed, looks like uncrustify doesn't do that.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -111,6 +111,20 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
                                        uint16_t keepAliveSeconds,
                                        _mqttConnection_t * pMqttConnection );
 
+/**
+ * @brief Initialize a network connection, creating it if necessary
```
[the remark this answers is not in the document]
gkwicker: Fixed

libraries/standard/mqtt/src/iot_mqtt_api.c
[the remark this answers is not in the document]
gkwicker: Added `IotLogError()` calls.

libraries/standard/mqtt/src/iot_mqtt_api.c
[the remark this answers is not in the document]
gkwicker: Fixed

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: We can't eliminate the out parameter because this function handles 2 cases; the case in which the connection is created (`pNetworkInfo->createNetworkConnection == true`), _and_ the case in which an existing connection is re-used (`pNetworkInfo->createNetworkConnection == false`).  The out parameter lets the caller know if they need to clean up the connection.

Changed the name of this function to `_createNetworkConnection()`.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Removed `networkStatus` and used `status` instead.

libraries/standard/mqtt/src/iot_mqtt_validate.c
gkwicker: Renamed to `_validatePublish()`, newline added.

libraries/standard/mqtt/src/iot_mqtt_validate.c
gkwicker: That doesn't handle the LWT case, which requires a different maximum length.  It is true that we could modify `_IotMqtt_ValidatePublish()` by adding a third parameter `IsLwtTopic`, and then modifying its logic with additional tests for LWT to select the appropriate size.  This would make calls to `_IotMqtt_ValidatePublish()` less readable in the publish/receive code, which is why the additional function `_IotMqtt_ValidateLwtPublish()` was added for the connect case instead.

libraries/standard/mqtt/src/iot_mqtt_api.c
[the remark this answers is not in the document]
gkwicker: `status` can be used instead, I've removed `networkStatus`.

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
gkwicker: Changed

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
gkwicker: Changed

libraries/standard/mqtt/src/iot_mqtt_validate.c
gkwicker: Done

libraries/standard/mqtt/src/iot_mqtt_api.c
muneebahmed10: I think having a `pCreatedNetworkConnection` variable in a function named `_createNetworkConnection` might be a source of confusion since its value isn't tied to the function's success. Perhaps renaming it to something like `pOwnNetworkConnection` instead?

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -439,19 +440,21 @@ static IotNetworkError_t _initializeNetworkConnection( const IotMqttNetworkInfo_
...
             /* This MQTT connection owns the network connection it created and
              * should destroy it on cleanup. */
             *pCreatedNetworkConnection = true;
         }
         else
         {
-            IOT_SET_AND_GOTO_CLEANUP( networkStatus );
+            IotLogError( "Failed to create network connection: %d", status );
+
+            IOT_SET_AND_GOTO_CLEANUP( status );
```
muneebahmed10: Just `IOT_GOTO_CLEANUP();` since `status` is what gets set.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -111,6 +111,20 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
...
 
+/**
+ * @brief Initialize a network connection, creating it if necessary.
+ *
+ * @param[in] pNetworkInfo User-provided network information for the connection
+ * connection.
+ * @param[out] pNetworkConnection On success, the created and/or initialized network connection.
+ * @param[out] pCreatedNetworkConnection On success, `true` if the network connection was created; `false` if it was only initialized.
+ *
+ * @return Any #IotNetworkError_t, as defined by the network stack.
```
muneebahmed10: These comments seems less clear to me now that the function is renamed. Perhaps rewording them to clarify the difference between initializing and creating, e.g. "true if a new network connection was created; false if an existing one will be used"

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: After another go-round in `IotMqtt_Connect()` I've been able to eliminate this out parameter.  Agree that it was confusing.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Changed

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: I was able to eliminate this parameter, agreed that it is confusing.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1262,18 +1250,25 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
         IotLogError( "Failed to establish new MQTT connection, error %s.",
                      IotMqtt_strerror( status ) );
 
-        /* The network connection must be closed if it was created. */
-        if( ownNetworkConnection == true )
+        if( status != IOT_MQTT_NETWORK_ERROR )
```
muneebahmed10: Line 1130 also sets an `IOT_MQTT_NETWORK_ERROR` if the receive callback was not set. In this case, wouldn't the connection have been created but not destroyed by the cleanup code? If you kept the ownNetworkConnection bool, then changing line 1093's `EMPTY_ELSE_MARKER` to `ownNetworkConnection = pNetworkInfo->createNetworkConnection;` would allow most of this code block to stay as it was.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -117,20 +117,18 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
...
+                                                   void ** pNetworkConnection );
 
 /**
  * @brief Creates a new MQTT connection and initializes its members.
  *
  * @param[in] awsIotMqttMode Specifies if this connection is to an AWS IoT MQTT server.
- * @param[in] pNetworkInfo User-provided network information for the new
- * connection.
+ * @param[in] pNetworkInfo User-provided network information for the new connection
+ * MQTT connection.
```
muneebahmed10: `connection` appears twice

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -137,6 +172,8 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
                          "the maximum supported length of %hu.",
```
muneebahmed10: Nit: While you're touching this file, you might as well fix the "the" twice in a row on 171-172

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -404,6 +416,50 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
...
+        status = pNetworkInfo->pNetworkInterface->create( pNetworkInfo->u.setup.pNetworkServerInfo,
+                                                          pNetworkInfo->u.setup.pNetworkCredentialInfo,
+                                                          pNetworkConnection );
+
+        if( status != IOT_NETWORK_SUCCESS )
+        {
+            IotLogError( "Failed to create network connection: %d", status );
+
+            IOT_GOTO_CLEANUP();
+        }
```
muneebahmed10: Nit: EMPTY_ELSE_MARKER after the if block

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: I've reverted the change which eliminated `ownNetworkConnection`.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Removing this parameter has caused more problems than it fixed, so I've restored it.  Will update documentation as suggested.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Changed

libraries/standard/mqtt/src/iot_mqtt_validate.c
gkwicker: Fixed

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Eliminated after reverting to earlier code which sets `pCreatedNetworkConnection`.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -117,18 +117,20 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
  * @param[in] pNetworkInfo User-provided network information for the connection
  * connection.
  * @param[out] pNetworkConnection On success, the created and/or initialized network connection.
+ * @param[out] pCreatedNetworkConnection On success, true if a new network connection was created; false if an existing one will be used
```
muneebahmed10: Nit: period at end of comment and backticks around true and false for consistency. Are we sure we want to keep the `pCreatedNetworkConnection` name?

gkwicker: Fixed backticks & period.  Changed parameter name to `pCreatedNewNetworkConnection`.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -111,6 +111,20 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
                                        uint16_t keepAliveSeconds,
                                        _mqttConnection_t * pMqttConnection );
 
+/**
+ * @brief Initialize a network connection, creating it if necessary.
+ *
+ * @param[in] pNetworkInfo User-provided network information for the connection
+ * connection.
```
muneebahmed10: Nit: connection twice here

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -111,6 +111,20 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
...
+ * @param[out] pNetworkConnection On success, the created and/or initialized network connection.
+ * @param[out] pCreatedNewNetworkConnection On success, `true` if a new network connection was created; `false` if an existing one will be used.
```
muneebahmed10: Split this line before merging, it's a bit long, and I don't believe uncrustify does this for you.

PR #613, opened 2019-10-23
Guard against multiple calls to IotMqtt_Init/IotMqtt_Cleanup

gkwicker:


libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -994,36 +1007,29 @@ void _IotMqtt_DecrementConnectionReferences( _mqttConnection_t * pMqttConnection
 IotMqttError_t IotMqtt_Init( void )
 {
     IotMqttError_t status = IOT_MQTT_SUCCESS;
+    uint32_t initializationStatus = Atomic_CompareAndSwap_u32( &_initCalled,
+                                                               MQTT_LIBRARY_INITIALIZED,
+                                                               MQTT_LIBRARY_UNINITIALIZED );
 
-    if( _initCalled == 0 )
+    if( initializationStatus != 0 )
```
muneebahmed10: Should the `0` be `MQTT_LIBRARY_UNINITIALIZED` as it is below, or is `MQTT_LIBRARY_UNINITIALIZED` only meant to be assigned to `_initCalled`?

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -994,36 +1007,29 @@ void _IotMqtt_DecrementConnectionReferences( _mqttConnection_t * pMqttConnection
...
-        if( status != IOT_MQTT_SUCCESS )
-        {
-            IotLogError( "Failed to initialize MQTT library serializer. " );
-        }
-        else
-        {
-            /* Set the flag that specifies initialization is complete. */
-            _initCalled = 1;
-
-            IotLogInfo( "MQTT library successfully initialized." );
```
muneebahmed10: This function doesn't log success anymore

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1037,10 +1043,12 @@ IotMqttError_t IotMqtt_Init( void )
...
 void IotMqtt_Cleanup( void )
 {
-    if( _initCalled == 1 )
-    {
-        _initCalled = 0;
+    uint32_t initializationStatus = Atomic_CompareAndSwap_u32( &_initCalled,
+                                                               MQTT_LIBRARY_UNINITIALIZED,
+                                                               MQTT_LIBRARY_INITIALIZED );
 
+    if( initializationStatus == MQTT_LIBRARY_UNINITIALIZED )
```
muneebahmed10: Is this correct? Since `Atomic_CompareAndSwap_u32` returns `1` when successful, if `initializationStatus` were to equal `MQTT_LIBRARY_UNINITIALIZED`, won't that mean the swap failed, and `_initCalled` was NOT equal to `MQTT_LIBRARY_INITIALIZED` beforehand?

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -238,20 +241,30 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+ */
+#define MQTT_LIBRARY_INITIALIZED      ( ( uint32_t ) 1 )
+
 /**
  * @brief Tracks whether @ref mqtt_function_init has been called.
  *
  * API functions will fail if @ref mqtt_function_init was not called.
  */
-static uint32_t _initCalled = 0;
+static volatile uint32_t _initCalled = MQTT_LIBRARY_UNINITIALIZED;
```
gordonwang0: Does this need to be declared volatile?  Probably not, it's not intended to be a variable shared between threads.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: `initializationStatus` is the return value from `Atomic_CompareAndSwap_u32()`, so it will be `1` if the CAS was successful and `0` otherwise.  This test is correct, the one below is not.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1066,8 +1074,10 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
 {
     IOT_FUNCTION_ENTRY( IotMqttError_t, IOT_MQTT_SUCCESS );
     bool ownNetworkConnection = false;
-    IotNetworkError_t networkStatus = IOT_NETWORK_SUCCESS;
-    IotTaskPoolError_t taskPoolStatus = IOT_TASKPOOL_SUCCESS;
+    /* Initialize to failure status */
+    IotNetworkError_t networkStatus = IOT_NETWORK_FAILURE;
```
gordonwang0: What was the reason for changing these initial values?

I realize that in most cases the initial value of these variables are unused.  The only reason I initialized them is so that they are not garbage values when debugging.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Good catch, re-added.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Good catch, that should be comparing `!= 0` just like the init case.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: It has to be `volatile` in order to [avoid a compiler warning.](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/ports/common/include/atomic/iot_atomic_generic.h#L98)

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: See the internal review CR for context/comments.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1066,7 +1076,9 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
 {
     IOT_FUNCTION_ENTRY( IotMqttError_t, IOT_MQTT_SUCCESS );
     bool ownNetworkConnection = false;
+    /* Initialize to failure status */
     IotNetworkError_t networkStatus = IOT_NETWORK_SUCCESS;
+    /* Initialize to failure status */
```
muneebahmed10: These comments are inaccurate now.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -238,20 +241,30 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
+/**
+ * @brief Uninitialized value for @ref _initCalled.
+ */
+#define MQTT_LIBRARY_UNINITIALIZED    ( ( uint32_t ) 0 )
+
+/**
+ * @brief Initialized value for @ref _initCalled.
+ */
+#define MQTT_LIBRARY_INITIALIZED      ( ( uint32_t ) 1 )
+
```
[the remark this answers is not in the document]
gkwicker: Will move.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -994,36 +1007,30 @@ void _IotMqtt_DecrementConnectionReferences( _mqttConnection_t * pMqttConnection
...
          * overrides are enabled. */
         #if IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES == 1
             #ifdef _IotMqtt_InitSerializeAdditional
                 if( _IotMqtt_InitSerializeAdditional() == false )
                 {
+                    /* Log initialization status. */
+                    IotLogError( "Failed to initialize MQTT library serializer. " );
+
                     status = IOT_MQTT_INIT_FAILED;
                 }
```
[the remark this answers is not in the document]
gkwicker: This is true, however if we try to make this not be the case we'll raise other issues.  For example, if `_IotMqtt_InitSerializeAdditional()` fails, do we allow a call to `IotMqtt_Cleanup()`?  What will `_IotMqtt_CleanupSerializeAdditional()` do in that case?  Will it fail?  If it does, we will have no way of knowing that it did because it [doesn't return a status.](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/doc/lib/mqtt.txt#L190)

I think the safest course of action for the time being is to simply say that calling any MQTT library API after a failed call to `IotMqtt_Init()` will result in undefined behavior.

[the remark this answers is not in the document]
muneebahmed10: I'm fine with that. However, nit: if the serializer init fails, then the log will look like:
```
[ERROR] Failed to initialize MQTT library serializer. 
[INFO] MQTT library successfully initialized.
```
and the function will return `IOT_MQTT_INIT_FAILED`. This could be a bit confusing, so I think either the info log could be in an if block, or setting the status could be in a set-and-goto.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1037,10 +1044,12 @@ IotMqttError_t IotMqtt_Init( void )
 
 void IotMqtt_Cleanup( void )
 {
-    if( _initCalled == 1 )
-    {
-        _initCalled = 0;
+    uint32_t initializationStatus = Atomic_CompareAndSwap_u32( &_initCalled,
```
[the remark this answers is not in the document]
gkwicker: Agreed, I've changed these names to `allowInitialization` and `allowCleanup`.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Fixed.

libraries/standard/mqtt/src/iot_mqtt_api.c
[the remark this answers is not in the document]
gkwicker: Good point, fixed.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -994,53 +1009,52 @@ void _IotMqtt_DecrementConnectionReferences( _mqttConnection_t * pMqttConnection
 IotMqttError_t IotMqtt_Init( void )
 {
     IotMqttError_t status = IOT_MQTT_SUCCESS;
```
muneebahmed10: Change to `IOT_FUNCTION_ENTRY( IotMqttError_t, IOT_MQTT_SUCCESS );`

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -994,53 +1009,52 @@ void _IotMqtt_DecrementConnectionReferences( _mqttConnection_t * pMqttConnection
...
         #if IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES == 1
             #ifdef _IotMqtt_InitSerializeAdditional
                 if( _IotMqtt_InitSerializeAdditional() == false )
                 {
+                    /* Log initialization status. */
+                    IotLogError( "Failed to initialize MQTT library serializer. " );
+
                     status = IOT_MQTT_INIT_FAILED;
+
+                    IOT_GOTO_CLEANUP();
```
muneebahmed10: Combining these to `IOT_SET_AND_GOTO_CLEANUP( IOT_MQTT_INIT_FAILED );` would be cleaner.

PR #627, opened 2019-11-06
Fix buffer overflows in json parser

markrtuttle:
Fixed buffer overflows in json parser:
* Return false immediately after incrementing index i past jsonDocumentLength to avoid buffer overflow in pJsonDocument[i].  Copied existing check to several other locations in the code.
* Return false immediately when key has length 0 to avoid buffer overflow in comparison pJsonDocument[i+1] == pJsonKey[0].

Fixed error in json key lookup:  Check for lead quotation mark when comparing key with json key to avoid key "b" matching json key "ab".

Simplified some code with IS_QUOTE and IS_WHITESPACE defines.

CBMC currently demonstrates memory safety on json documents and keys of length at most 10.

### DIFF
--- a/libraries/standard/serializer/src/iot_json_utils.c
+++ b/libraries/standard/serializer/src/iot_json_utils.c
@@ -75,6 +75,12 @@ bool IotJsonUtils_FindJsonValue( const char * pJsonDocument,
             /* Skip the characters in the JSON key and closing double quote. */
             i += jsonKeyLength + 1;
 
+	    /* If the end of the document is reached, this isn't a match. */
+	    if( i >= jsonDocumentLength )
+	    {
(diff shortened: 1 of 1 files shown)


gordonwang0:
This seems to cause all of the JSON parsing tests to fail.  Will have to take a look.

muneebahmed10:
Uncrustify: tabs in the file should be replaced with spaces.

leegeth:
> Uncrustify: tabs in the file should be replaced with spaces.

Yea the formatting needs to be corrected, otherwise the changes look good.

gordonwang0:
Looks like Lee formatted and merged this in a separate PR.  Will close this one.

PR #631, opened 2019-11-09
Shadow code restructure and documentation changes

leegeth:
Shadow library changes made do code restructure and add documentation.

Tests
Verified Shadow demo
Verified Shadow unit and system tests

### DIFF
--- a/libraries/aws/common/src/aws_iot_parser.c
+++ b/libraries/aws/common/src/aws_iot_parser.c
@@ -145,8 +145,17 @@ AwsIotStatus_t AwsIot_ParseStatus( const char * pTopicName,
     IOT_FUNCTION_ENTRY( AwsIotStatus_t, AWS_IOT_UNKNOWN );
     const char * pSuffixStart = NULL;
 
+    /* Both 'accepted' and  'rejected' topics are of the same length
+     * The below is a defensive check at compile time to ensure that.
+     * static_assert is only available from C11 and above*/
+    #if ( __STDC_VERSION__ >= 201112L )
+        static_assert( AWS_IOT_ACCEPTED_SUFFIX_LENGTH == AWS_IOT_REJECTED_SUFFIX_LENGTH,
+                   "accepted and rejected identifiers must be the same length!" );
+    #endif
+
     /* Check that the status topic name is at least as long as the
-     * "accepted" suffix. */
+     * "accepted" suffix. This length check will be good for rejected also
+     * as both are of 8 characters in length. */
     if( topicNameLength > AWS_IOT_ACCEPTED_SUFFIX_LENGTH )
     {
         /* Calculate where the "accepted" suffix should start. */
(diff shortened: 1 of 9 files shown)


libraries/aws/common/src/aws_iot_parser.c
```diff
@@ -145,8 +145,17 @@ AwsIotStatus_t AwsIot_ParseStatus( const char * pTopicName,
     IOT_FUNCTION_ENTRY( AwsIotStatus_t, AWS_IOT_UNKNOWN );
     const char * pSuffixStart = NULL;
 
+    /* Both 'accepted' and  'rejected' topics are of the same length
+     * The below is a defensive check at compile time to ensure that.
+     * static_assert is only available from C11 and above*/
+    #if ( __STDC_VERSION__ >= 201112L )
```
gordonwang0: We're targeting C99 with these libraries.  I don't think this block is useful.

leegeth: yea. makes sense. Thanks

libraries/aws/shadow/src/private/aws_iot_shadow_internal.h
```diff
@@ -258,6 +258,18 @@
...
 
+/**
+ * @brief The macro to convert MQTT error codes to Shadow error codes.
+ * Below are the conversions happening.
+ * IOT_MQTT_SUCCESS to AWS_IOT_SHADOW_SUCCESS
+ * IOT_MQTT_NO_MEMORY to AWS_IOT_SHADOW_NO_MEMORY
+ * all other error codes to AWS_IOT_SHADOW_MQTT_ERROR
+ */
+#define SHADOW_CONVERT_STATUS_CODE_MQTT_TO_SHADOW( X ) \
+    ( X == IOT_MQTT_SUCCESS ) ? AWS_IOT_SHADOW_SUCCESS : \
```
gkwicker: If you want to use a macro for this, the argument needs `(` `)` around it whenever it's used in the body.  More info on why [here](https://wiki.sei.cmu.edu/confluence/display/c/PRE01-C.+Use+parentheses+within+macros+around+parameter+names).

libraries/aws/common/src/aws_iot_subscription.c
```diff
@@ -77,7 +77,11 @@ static IotMqttError_t _modifySubscriptions( AwsIotMqttFunction_t mqttOperation,
     subscription.callback.pCallbackContext = NULL;
     subscription.callback.function = pSubscriptionInfo->callbackFunction;
 
-    /* Call the MQTT operation function. */
+    /* Call the MQTT operation function.
+     * Subscription count is 1 in this case.
+     * None of the flags are set in this call. Hence 0 is passed for flags.
+     * Please refer to documentation for AwsIotMqttFunction_t for more details.
```
gkwicker: Note that [`AwsIotMqttFunction_t` is not currently documented](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/aws/common/include/aws_iot.h#L111), and that even if it was we don't have a reference to the doc entry here.  Those should be added, *and* we should also add comments for the 3rd and 4th arguments here explaining that these are the # of subscription and flags.

leegeth: It is already handled in this PR

Issue #524, opened 2019-07-17
The server actively disconnects, the SDK cannot perceive(SDK v4.0.4beta)


MrZhuGe:
Hello, I found a problem during the use. I use the SDK connection service on one device, and use the same device information connection service on the other device. The server will actively disconnect the first device, and then the SDK does not sense it and does not give it. Any callback information is blocked on the heartbeat. In previous versions of the SDK, it was able to detect that the server was actively disconnected.

MrZhuGe:
@gordonwang0 I think that the server actively closes the connection of the device, there should be a disconnect callback, right?

gordonwang0:
Hi @MrZhuGe,

You are correct, currently server-to-client disconnects are not reported through the disconnect callback.  We will implement this shortly.

gordonwang0:
By the way -- it is expected for your application to block until the MQTT keep-alive, as that is how MQTT can detect the server-side closed connection.  Any other ways to detect a server-closed connection, while possible, are outside of the MQTT spec.

MrZhuGe:
It is convenient to ask about the handling of this problem. We are using it. This issue has a relatively large impact on us, thank you.

abhidixi11:
I have created internal issue to track this feature request.

MrZhuGe:
@abhidixi11 @gordonwang0 Excuse me, is there a solution to this problem?

lundinc2:
Hello @MrZhuGe,

We currently have not implemented this feature yet. In the meantime, would it be possible for you to use the MQTT keep alive mechanism described earlier by @godonwang0 in your application?

gkwicker:
Hi @MrZhuGe, PR #634 adds a close callback to the network interface, which you should be able to use to detect the server close condition.  I'll leave this issue open for now, but if that works for you please let me know.

MrZhuGe:
Hi @gkwicker, This callback is useful and has been verified, thank you very much.

abhidixi11:
Hello MrZhuGe,
Thanks for your feedback. I'm closing this issue. If you have any further questions, please feel free to open it again, or open a new issue.
Thanks.

PR #634, opened 2019-11-12
Add close callback to network interface

gkwicker:
*Issue #524

*Description of changes:*

- Add a close callback to the network interface
- mbedTLS & OpenSSL implementations

### DIFF
--- a/libraries/platform/iot_network.h
+++ b/libraries/platform/iot_network.h
@@ -54,11 +54,11 @@ typedef enum IotNetworkError
  */
 typedef enum IotNetworkCloseReason
 {
-    IOT_NETWORK_NOT_CLOSED = 0,     /**< Not closed, still open */
-    IOT_NETWORK_SERVER_CLOSED,      /**< Server closed connection. */
-    IOT_NETWORK_TRANSPORT_FAILURE,  /**< Transport failed. */
-    IOT_NETWORK_CLIENT_CLOSED,      /**< Client closed connection. */
-    IOT_NETWORK_UNKNOWN_CLOSED      /**< Unknown close reason. */
+    IOT_NETWORK_NOT_CLOSED = 0,    /**< Not closed, still open */
+    IOT_NETWORK_SERVER_CLOSED,     /**< Server closed connection. */
+    IOT_NETWORK_TRANSPORT_FAILURE, /**< Transport failed. */
+    IOT_NETWORK_CLIENT_CLOSED,     /**< Client closed connection. */
+    IOT_NETWORK_UNKNOWN_CLOSED     /**< Unknown close reason. */
 } IotNetworkCloseReason_t;
 
 /**
@@ -127,6 +127,7 @@ typedef void ( * IotNetworkCloseCallback_t )( IotNetworkConnection_t pConnection
                                               IotNetworkCloseReason_t reason,
                                               void * pContext );
 /* @[declare_platform_network_closecallback] */
+
 /**
  * @brief Create a new network connection.
  *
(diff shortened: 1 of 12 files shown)


ports/common/src/iot_network_openssl.c
```diff
@@ -781,6 +794,26 @@ IotNetworkError_t IotNetworkOpenssl_SetReceiveCallback( IotNetworkConnection_t p
...
+    if( closeCallback != NULL )
+    {
+        /* Set the callback and parameter. */
+        pConnection->closeCallback = closeCallback;
+        pConnection->pCloseContext = pContext;
+
+        status = IOT_NETWORK_SUCCESS;
+    }
+
+    IOT_FUNCTION_CLEANUP_END();
```
muneebahmed10: Should be `IOT_FUNCTION_EXIT_NO_CLEANUP()` to be consistent with the way we use these macros. Same for the mbedtls file.

gkwicker: Fixed

libraries/platform/iot_network.h
```diff
@@ -94,6 +111,178 @@ typedef void ( * IotNetworkReceiveCallback_t )( IotNetworkConnection_t pConnecti
                                                 void * pContext );
 /* @[declare_platform_network_receivecallback] */
 
+/**
+ * @brief Provide an asynchronous notification of network closing
+ *
+ * A function with this signature may be set with @ref platform_network_function_setclosecallback
+ * to be invoked when data is available on the network.
```
abhidixi11: This should read "to be invoked when network connection is closed"

libraries/platform/iot_network.h
```diff
@@ -94,6 +111,178 @@ typedef void ( * IotNetworkReceiveCallback_t )( IotNetworkConnection_t pConnecti
...
+ *
+ * @param[in] pConnection The connection on which data is available, defined by
+ * the network stack.
+ * @param[in] reason The reason the connection was closed
+ * @param[in] pContext The third argument passed to @ref platform_network_function_setreceivecallback.
```
abhidixi11: should be platform_network_function_setclosecallback().

libraries/platform/iot_network.h
```diff
@@ -48,6 +48,19 @@ typedef enum IotNetworkError
     IOT_NETWORK_SYSTEM_ERROR   /**< An error occurred when calling a system API. */
 } IotNetworkError_t;
 
+/**
+ * @ingroup platform_datatypes_enums
+ * @brief Disconnect reasons for [the network close callback](@ref platform_network_function_closecallback).
+ */
+typedef enum IotNetworkCloseReason
+{
+    IOT_NETWORK_NOT_CLOSED = 0,    /**< Not closed, still open */
```
abhidixi11: IOT_NETWORK_NOT_CLOSED :  I'm assuming this is the status of client initiated close ( to report close failure).  It does not look like the reason for the close of the connection. ( The enum is called "IotNetworkCloseReason_t"). So should this be  part of the reason?

gkwicker: I had a hard time naming this one.  I was thinking we needed a reason code which represented "not closed" in the same way that a function can return a 0 value which means success.  Once we finish implementing the ports such that they can actually populate the reason code, we can investigate removing this value if it is never needed.

libraries/platform/iot_network.h
gkwicker: good catch, fixed

libraries/platform/iot_network.h
gkwicker: Fixed

ports/common/src/iot_network_openssl.c
```diff
@@ -752,28 +766,53 @@ IotNetworkError_t IotNetworkOpenssl_SetReceiveCallback( IotNetworkConnection_t p
...
+
+/*-----------------------------------------------------------*/
+
+IotNetworkError_t IotNetworkOpenssl_SetCloseCallback( IotNetworkConnection_t pConnection,
+                                                      IotNetworkCloseCallback_t closeCallback,
+                                                      void * pContext )
+{
+    IotNetworkError_t status = IOT_NETWORK_BAD_PARAMETER;
+
+    if( closeCallback != NULL )
```
abhidixi11: Should we check value of if (pConnection  != NULL) here  (and in IotNetworkOpenssl_SetReceiveCallback )?

gkwicker: I didn't add those here since they hadn't been here before, but this is a fair point.  Also, `assert()` might be more appropriate for both of these checks.

libraries/platform/iot_network.h
```diff
@@ -94,6 +111,178 @@ typedef void ( * IotNetworkReceiveCallback_t )( IotNetworkConnection_t pConnecti
                                                 void * pContext );
 /* @[declare_platform_network_receivecallback] */
 
+/**
+ * @brief Provide an asynchronous notification of network closing
+ *
+ * A function with this signature may be set with @ref platform_network_function_setclosecallback
+ * to be invoked when the network connection is closed.
+ *
+ * @param[in] pConnection The connection on which data is available, defined by
```
abhidixi11: Should read: The connection that is being closed...

ports/common/src/iot_network_openssl.c
abhidixi11: I'm pretty sure there will be many such places, which we have overlooked earlier. I just happen to notice.  I'm not sure if CBMC proofs are written for interface and if that should have caught the problem.

libraries/platform/iot_network.h
gkwicker: good catch, fixed.

PR #640, opened 2019-11-14
Close MQTT connection on network close/disconnect.

gkwicker:
Issue #524 

Close MQTT connection on network close/disconnect.

### DIFF
--- a/libraries/standard/mqtt/include/iot_mqtt.h
+++ b/libraries/standard/mqtt/include/iot_mqtt.h
@@ -127,6 +127,25 @@ void IotMqtt_ReceiveCallback( IotNetworkConnection_t pNetworkConnection,
 /* @[declare_mqtt_receivecallback] */
 
 /**
+ * @brief Network close callback for the MQTT library.
+ *
+ * This function should be called by the system whenever the network connection
+ * has been closed
+ *
+ * @param[in] pNetworkConnection The network connection associated with the MQTT
+ * connection, passed by the network stack.
+ * @param[in] pNetworkConnection The network connection associated with the MQTT
+ * connection, passed by the network stack.
+ * @param[in] pCloseContext A pointer to the MQTT connection handle on which the
+ * network connection has closed.
+ */
+/* @[declare_mqtt_closecallback] */
+void IotMqtt_CloseCallback( IotNetworkConnection_t pNetworkConnection,
+                            IotNetworkCloseReason_t reason,
+                            void * pCloseContext );
+/* @[declare_mqtt_closecallback] */
+
+/**
  * @brief Establish a new MQTT connection.
  *
  * This function opens a connection between a new MQTT client and an MQTT server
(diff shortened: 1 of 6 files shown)


libraries/standard/mqtt/include/iot_mqtt.h
```diff
@@ -126,6 +126,25 @@ void IotMqtt_ReceiveCallback( IotNetworkConnection_t pNetworkConnection,
...
+/**
+ * @brief Network close callback for the MQTT library.
+ *
+ * This function should be called by the system whenever the network connection
+ * has been closed
+ *
+ * @param[in] pNetworkConnection The network connection associated with the MQTT
+ * connection, passed by the network stack.
+ * @param[in] pNetworkConnection The network connection associated with the MQTT
+ * connection, passed by the network stack.
```
muneebahmed10: Should be documenting `reason` instead of `pNetworkConnection` here.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1165,6 +1165,22 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
         EMPTY_ELSE_MARKER;
     }
 
+    /* Set the MQTT close callback. */
+    networkStatus = pNewMqttConnection->pNetworkInterface->setCloseCallback( pNetworkConnection,
```
muneebahmed10: This line is causing segfaults in CI, as part of `TEST(MQTT_Unit_API, ConnectParameters)`.

Possible fix:
In `iot_tests_mqtt_api.c`, define function:
```
/**
 * @brief A function for setting the close callback that just returns success.
 */
static IotNetworkError_t _setCloseCallback( IotNetworkConnection_t pConnection,
                                              IotNetworkCloseCallback_t closeCallback,
                                              void * pReceiveContext )
{
    /* Silence warnings about unused parameters. */
    ( void ) pConnection;
    ( void ) closeCallback;
    ( void ) pReceiveContext;

    return IOT_NETWORK_SUCCESS;
}
```
and add the line
`_networkInterface.setCloseCallback = _setCloseCallback;`
to `TEST_SETUP( MQTT_Unit_API )`

libraries/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -843,3 +843,16 @@ void IotMqtt_ReceiveCallback( IotNetworkConnection_t pNetworkConnection,
 }
 
 /*-----------------------------------------------------------*/
+
+void IotMqtt_CloseCallback( IotNetworkConnection_t pNetworkConnection,
+                            IotNetworkCloseReason_t reason,
```
muneebahmed10: Doesn't look like these parameters are used. Should be voided to get rid of unused parameter warnings.

gkwicker: Good catch, fixed.

libraries/standard/mqtt/src/iot_mqtt_api.c
gkwicker: Good catch, fixed.

libraries/standard/mqtt/include/iot_mqtt.h
gkwicker: Good catch, fixed.

gkwicker:
After further investigation, it will take more design to propagate the server close/network fail condition up through the MQTT API, so we'll cancel this PR.  The new network abstraction close callback may be sufficient to address issue #524 for now.

Issue #641, opened 2019-11-14
Support MQTT username/password authentication


dcgaws:
Customers are requesting Embedded C SDK support for existing devices that use username and password to authenticate to their MQTT broker. That capability is part of the MQTT open standard. Related, there is an IANA-standardized protocol name, "mqtt", that should be used as an ALPN extension for TLS when negotiating the above.

gordonwang0:
Completed with merged PR.

PR #642, opened 2019-11-14
Add support for MQTT username/password and IANA protocol string.

dcgaws:
The purpose of this change is to allow standard-based username/password device authentication to an MQTT broker (#641).


demos/app/iot_demo.c
```diff
@@ -118,19 +118,55 @@ int main( int argc,
     /* For a secured connection, set the members of the credentials. */
     if( demoArguments.securedConnection == true )
     {
-        /* Set credential paths. */
+        /* Set credential information. */
         credentials.pClientCert = demoArguments.pClientCertPath;
```
gkwicker: If `demoArguments.pUserName` and `demoArguments.pPassword` are non-NULL and the broker supports authenticating with a username and password, can these cert/private key/root CA fields be set to `NULL`?

demos/app/iot_demo.c
```diff
@@ -118,19 +118,55 @@ int main( int argc,
...
         credentials.pPrivateKey = demoArguments.pPrivateKeyPath;
         credentials.pRootCa = demoArguments.pRootCaPath;
+
```
gkwicker: If you add the following here, lines 134-137 and 148-151 can be deleted.

```c
credentials.pUserName = NULL;
credentials.pPassword = NULL;
```

demos/app/iot_demo.c
```diff
@@ -118,19 +118,55 @@ int main( int argc,
...
+        /* Set credential information. */
         credentials.pClientCert = demoArguments.pClientCertPath;
         credentials.pPrivateKey = demoArguments.pPrivateKeyPath;
         credentials.pRootCa = demoArguments.pRootCaPath;
+        
+        /* Set the MQTT username, as long as it's not empty or NULL. */
+        if( demoArguments.pUserName != NULL )
+        {
+            credentials.userNameSize = strlen( demoArguments.pUserName );
+            if( credentials.userNameSize > 0 )
```
gkwicker: Do we need to enforce a maximum size for `credentials.userNameSize`?

demos/app/iot_demo.c
```diff
@@ -118,19 +118,55 @@ int main( int argc,
...
+            {
+                credentials.pUserName = NULL;
+            }
+        }
+
+        /* Set the MQTT password, as long as it's not empty or NULL. */
+        if( demoArguments.pPassword != NULL )
+        {
+            credentials.passwordSize = strlen( demoArguments.pPassword );
+            if( credentials.passwordSize > 0 )
```
gkwicker: Do we need to enforce a maximum size for `credentials.passwordSize`?  The MQTT spec specifies that these [can't exceed 65535 bytes.](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718031)

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -692,7 +692,8 @@ void _serializeConnect( const IotMqttConnectInfo_t * pConnectInfo,
     }
 
     /* Username and password depend on MQTT mode. */
-    if( pConnectInfo->awsIotMqttMode == true )
+    if( ( pConnectInfo->pUserName == NULL ) && 
+        ( pConnectInfo->awsIotMqttMode == true ) )
     {
         /* Set the username flag for AWS IoT metrics. The AWS IoT MQTT server
```
gkwicker: If `pConnectInfo->awsIotMqttMode == true` and `pConnectInfo->pUserName != NULL` and `AWS_IOT_MQTT_ENABLE_METRICS == 1` is defined, we should log an error since the metrics will not be able to use the username area.

demos/app/iot_demo_arguments.c
```diff
@@ -133,22 +143,37 @@ static bool _validateArguments( const IotDemoArguments_t * pArguments )
...
-            IotLogError( "Client certificate private key not set. Exiting." );
+                /* Check that a client certificate private key was set. */
+                if( ( pArguments->pPrivateKeyPath == NULL ) ||
+                    ( strlen( pArguments->pPrivateKeyPath ) == 0 ) )
+                {
+                    IotLogError( "Either username/password or private key path must be set. Exiting." );
 
-            IOT_SET_AND_GOTO_CLEANUP( false );
+                    IOT_SET_AND_GOTO_CLEANUP( false );
+                }
```
gkwicker: Is a valid `pArguments->pRootCaPath` also required to connect to AWS IoT with a client certificate/private key?

demos/app/iot_demo.c
dcgaws: Yes. They don't have to be, though. That is, enforcement of client certificate authentication over TLS, and username/password authentication at the MQTT broker, are orthogonal.

demos/app/iot_demo.c
dcgaws: Good catch, thank you, fix is pending.

demos/app/iot_demo.c
dcgaws: Good catch, thank you, fix is pending.

demos/app/iot_demo.c
dcgaws: Good catch, thank you, fix is pending.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
dcgaws: This limitation got corrected based on offline conversation. I've updated the code.

demos/app/iot_demo_arguments.c
dcgaws: Yes, the proposed validation rules for the CLI options are that "secure mode" requires an explicit root certificate for validating the server chain. Beyond that, if you're also in "AWS IoT mode," then a client credential is also required: username/password or certificate/private key.

demos/app/iot_demo.c
```diff
@@ -118,19 +118,63 @@ int main( int argc,
...
         {
             credentials.pAlpnProtos = NULL;
         }
 
+        /* Per IANA standard:
+         * https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml. */
+        if( ( demoArguments.pUserName != NULL ) &&
+            ( demoArguments.awsIotMqttMode == true ) )
+        {
+            credentials.pAlpnProtos = "mqtt";
```
leegeth: Is the value of credentials.pAlpnProtos need to be "mqtt" even if demoArguments.port is 443?

demos/app/iot_demo_arguments.c
```diff
@@ -133,22 +143,37 @@ static bool _validateArguments( const IotDemoArguments_t * pArguments )
...
-        /* Check that a client certificate path was set. */
-        if( ( pArguments->pClientCertPath == NULL ) ||
-            ( strlen( pArguments->pClientCertPath ) == 0 ) )
+        /* If the host is connecting to the MQTT broker hosted by AWS IoT Core,
+         * there must either be a set of X.509 credentials or a
+         * username/password. Therefore, check that here in order to facilitate
+         * debugging. For other MQTT brokers, assume that the CLI arguments are
+         * as intended.
+         */
+        if( pArguments->awsIotMqttMode == true )
```
leegeth: Why are we not checking for the validity of the non AWS IoT MQTT Broker arguments?

libraries/platform/iot_network.h
```diff
@@ -356,6 +356,10 @@ struct IotNetworkCredentials
     size_t clientCertSize;    /**< @brief Size associated with #IotNetworkCredentials.pClientCert. */
     const char * pPrivateKey; /**< @brief String representing the client certificate's private key. */
     size_t privateKeySize;    /**< @brief Size associated with #IotNetworkCredentials.pPrivateKey. */
+    const char * pUserName;   /**< @brief String representing the username for MQTT. */
+    size_t userNameSize;      /**< @brief Size associated with #IotNetworkCredentials.pUserName. */
+    const char * pPassword;   /**< @brief String representing the password for MQTT. */
+    size_t passwordSize;      /**< @brief Size associated with #IotNetworkCredentials.pPassword. */
```
leegeth: Is there a case in which both credentials and user/name password will be used? If not can we make this a union instead?

demos/app/iot_demo.c
dcgaws: Yes, the proposal is to adopt the official application protocol string for MQTT as per IANA standard. Depending on the MQTT broker, and the gateway that sits in front of it, the app protocol specification may be required regardless of the port.

libraries/platform/iot_network.h
dcgaws: Yes, the settings are orthogonal. The web gateway may or may not require client certificate authentication, and the MQTT broker may or may not require username/password.

leegeth: ok. thanks

demos/app/iot_demo_arguments.c
dcgaws: Please see my comment below. Because of that, in the non-AWS case, we don't know what the server-side policies will be with respect to web server and MQTT broker authentication, so we can't be as specific in our CLI argument checks.

demos/app/iot_demo.c
leegeth: Thanks. Is this already supported in AWS IoT MQTT Broker? The documentation I was referring to was specific to AWS IoT MQTT broker

demos/src/iot_demo_mqtt.c
```diff
@@ -416,6 +417,10 @@ static int _establishMqttConnection( bool awsIotMqttMode,
     connectInfo.cleanSession = true;
     connectInfo.keepAliveSeconds = KEEP_ALIVE_SECONDS;
     connectInfo.pWillInfo = &willInfo;
+    connectInfo.pUserName = pCredentials->pUserName;
+    connectInfo.userNameLength = ( uint16_t ) pCredentials->userNameSize;
+    connectInfo.pPassword = pCredentials->pPassword;
+    connectInfo.passwordLength = ( uint16_t ) pCredentials->passwordSize;
```
muneebahmed10: If `demoArguments.securedConnection == true` in `iot_demo.c` (as is in CI), this will attempt to dereference a null pointer, as the network credential info pointer won't be set. `pCredentials != NULL` needs to be checked before executing these lines.

dcgaws: Yes, good catch, I had missed that. Fix appears to have worked, per latest CI results.

demos/iot_config.h
```diff
@@ -34,6 +34,8 @@
 #define IOT_DEMO_ROOT_CA               "" /* Command line: -r */
 #define IOT_DEMO_CLIENT_CERT           "" /* Command line: -c */
 #define IOT_DEMO_PRIVATE_KEY           "" /* Command line: -k */
+#define IOT_DEMO_USER_NAME             "" /* Command line: -m */
+#define IOT_DEMO_PASSWORD              "" /* Command line: -w */
```
muneebahmed10: If the demo config defines these macros, should `tests/iot_config.h` also be changed to define `IOT_TEST_USER_NAME` and `IOT_TEST_PASSWORD`?

dcgaws: Good catch. Fixed.

demos/src/iot_demo_mqtt.c
```diff
@@ -417,6 +418,21 @@ static int _establishMqttConnection( bool awsIotMqttMode,
...
+        connectInfo.passwordLength = ( uint16_t ) pCredentials->passwordSize;
+    }
+    else
+    {
+        connectInfo.pUserName = NULL;
+        connectInfo.userNameLength = 0;
+        connectInfo.pPassword = NULL;
+        connectInfo.passwordLength = 0;
+    }
+
```
muneebahmed10: Is this something that should be added to `_establishMqttConnection()` in the other AWS demos (shadow, jobs, defender), or just for this one?

dcgaws: Good question. My understanding is that the MQTT CLI demo - plus the library itself - is sufficient for the near-term deadline.

demos/app/iot_demo.c
```diff
@@ -118,19 +119,63 @@ int main( int argc,
...
-        /* Set credential paths. */
+        /* Set credential information. */
         credentials.pClientCert = demoArguments.pClientCertPath;
         credentials.pPrivateKey = demoArguments.pPrivateKeyPath;
         credentials.pRootCa = demoArguments.pRootCaPath;
+        credentials.pUserName = NULL;
+        credentials.pPassword = NULL;
+
+        /* Set the MQTT username, as long as it's not empty or NULL. */
+        if( demoArguments.pUserName != NULL )
```
gordonwang0: Should move these checks to iot_demo_arguments.c (other checks are done there).

demos/src/iot_demo_mqtt.c
```diff
@@ -417,6 +418,21 @@ static int _establishMqttConnection( bool awsIotMqttMode,
...
+    if( pCredentials != NULL )
+    {
+        connectInfo.pUserName = pCredentials->pUserName;
+        connectInfo.userNameLength = ( uint16_t ) pCredentials->userNameSize;
+        connectInfo.pPassword = pCredentials->pPassword;
+        connectInfo.passwordLength = ( uint16_t ) pCredentials->passwordSize;
+    }
+    else
+    {
+        connectInfo.pUserName = NULL;
```
gordonwang0: The default initializer should set these all to 0, so this block isn't necessary.  May still be helpful as a demo though, so I understand if you want to keep it.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -786,42 +790,79 @@ void _serializeConnect( const IotMqttConnectInfo_t * pConnectInfo,
...
 
-            pBuffer = _encodeString( pBuffer,
-                                     AWS_IOT_METRICS_USERNAME,
-                                     AWS_IOT_METRICS_USERNAME_LENGTH );
-        #endif
+            /* Determine if the Connect packet should use a combination of the username
+             * for authentication plus the SDK version string. */
+            if( pConnectInfo->pUserName != NULL )
+            {
+                pcUserName = Iot_DefaultMalloc( pConnectInfo->userNameLength +
```
gordonwang0: Another buffer shouldn't be allocated for this.  The username can be encoded directly into `pBuffer`; you just can't use the `_encodeString` function as currently implemented to do this.  Or you could modify that function for this use case.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -487,30 +488,30 @@ static bool _connectPacketSize( const IotMqttConnectInfo_t * pConnectInfo,
     if( pConnectInfo->awsIotMqttMode == true )
     {
         #if AWS_IOT_MQTT_ENABLE_METRICS == 1
-            connectPacketSize += AWS_IOT_METRICS_USERNAME_LENGTH + sizeof( uint16_t );
+            connectPacketSize += AWS_IOT_METRICS_USERNAME_LENGTH +
+                                 pConnectInfo->userNameLength + sizeof( uint16_t );
+            encodedUserName = true;
         #endif
     }
+
```
gkwicker: Can the SDK metrics and username for authentication coexist in the same MQTT Connect message?  If so what is the layout?  Which comes first and how much space is reserved for each?  If this is the case, does the maximum username length check on line 135 need to be adjusted?

demos/app/iot_demo.c
dcgaws: Done, thank you.

demos/app/iot_demo.c
dcgaws: Fixed.

demos/app/iot_demo.c
dcgaws: Fixed.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
dcgaws: Yes, good catch, thank you. I've updated our length checks to reflect that metrics + logon name may be combined.

demos/app/iot_demo.c
gordonwang0: Consider also moving this (and the length check of user name above) to iot_demo_arguments.c too just so all checks are in one function.

demos/app/iot_demo_arguments.c
```diff
@@ -30,6 +30,7 @@
 /* Standard includes. */
 #include <stdlib.h>
 #include <string.h>
+#include <limits.h>
```
gordonwang0: Is this include needed now?  Maybe stdint.h should be used instead, I'm not sure how portable limits.h is.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -30,6 +30,7 @@
 
 /* Standard includes. */
 #include <string.h>
+#include <limits.h>
```
gordonwang0: Same comment about this header as above.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -786,42 +790,81 @@ void _serializeConnect( const IotMqttConnectInfo_t * pConnectInfo,
...
-                                     AWS_IOT_METRICS_USERNAME,
-                                     AWS_IOT_METRICS_USERNAME_LENGTH );
-        #endif
+            /* Determine if the Connect packet should use a combination of the username
+             * for authentication plus the SDK version string. */
+            if( pConnectInfo->pUserName != NULL )
+            {
+                /* Only include metrics if it will fit within the encoding
+                 * standard. */
+                if( ( pConnectInfo->userNameLength + AWS_IOT_METRICS_USERNAME_LENGTH ) <= UINT16_MAX )
```
gordonwang0: This check should be moved to _connectPacketSize.  It could potentially fail silently here and generate an invalid MQTT packet.

ports/common/include/iot_network_mbedtls.h
```diff
@@ -51,17 +51,24 @@
...
+ * @brief Initialize an #IotNetworkCredentials for AWS IoT for using TLS mutual
+ * authentication with certificates, TCP port 443, and mbedTLS.
  *
  * @note This initializer may change at any time in future versions, but its
  * name will remain the same.
  */
-#define AWS_IOT_NETWORK_CREDENTIALS_MBEDTLS_INITIALIZER \
-    {                                                   \
-        .pAlpnProtos = "x-amzn-mqtt-ca"                 \
+#define AWS_IOT_NETWORK_CREDENTIALS_MBEDTLS_INITIALIZER_FOR_CLIENT_CERTIFICATE \
```
gordonwang0: Can't change the name of this (per documentation note line 57)

Same goes for OpenSSL below.

demos/app/iot_demo.c
```diff
@@ -118,19 +120,51 @@ int main( int argc,
...
         {
             credentials.pAlpnProtos = NULL;
         }
 
+        /* Per IANA standard:
+         * https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml. */
+        if( ( demoArguments.pUserName != NULL ) &&
+            ( demoArguments.awsIotMqttMode == true ) )
+        {
+            credentials.pAlpnProtos = IOT_DEMO_ALPN_FOR_PASSWORD_AUTHENTICATION;
```
leegeth: If the demoArguments.port is 443, and demoArguments.pUserName != NULL  &&  demoArguments.awsIotMqttMode == true , would IOT_DEMO_ALPN_FOR_PASSWORD_AUTHENTICATION for credentials.pAlpnProtos work with AWS IoT MQTT broker?  Is the support for both become available at the same time in Broker?

PR #662, opened 2019-11-29
Fleet provisioning beta

aggarw13:
*Description of changes:*
Release support for **Fleet Provisioning<sup>BETA</sup>** feature of the AWS IoT Core service


tests/iot_config.h
```diff
@@ -124,17 +124,46 @@
...
 #if defined( IOT_TEST_MQTT_CLIENT_IDENTIFIER )
     #define IOT_DEMO_IDENTIFIER            IOT_TEST_MQTT_CLIENT_IDENTIFIER
 #elif defined( AWS_IOT_TEST_SHADOW_THING_NAME )
     #define IOT_DEMO_IDENTIFIER            AWS_IOT_TEST_SHADOW_THING_NAME
 #endif
 
+/* Set defines required for Provisioning demo when building demo target with IOT_BUILD_TESTS set to 1. */
+#ifndef AWS_IOT_DEMO_PROVISIONING_TEMPLATE_NAME
+    #define AWS_IOT_DEMO_PROVISIONING_TEMPLATE_NAME                           ""
+#endif
```
muneebahmed10: You're already defining `AWS_IOT_TEST_PROVISIONING_TEMPLATE_NAME` in this file. These lines should be:

```
#ifdef AWS_IOT_TEST_PROVISIONING_TEMPLATE_NAME
    #define AWS_IOT_DEMO_PROVISIONING_TEMPLATE_NAME    AWS_IOT_TEST_PROVISIONING_TEMPLATE_NAME
#endif

File: libraries/standard/serializer/lexicon.txt [hunk 2]
```diff
@@ -38,3 +38,7 @@ sizeof
 tinycbor
 todo
 www
+getsizeof
+https
+ietf
+iterable
```
muneebahmed10: These should be in alphabetical order in this file

demos/src/aws_iot_demo_provisioning.c
```diff
@@ -415,10 +415,10 @@ int RunProvisioningDemo( bool awsIotMqttMode,
 
     /* The list of parameters that will be used as "context" for provisioning the demo application.
      * This demo will pass exactly 2 different parameter entries for requesting provisioning. */
-    AwsIotProvisioningRequestParameterEntry_t provisioningParameters[ 2 ] = { 0 };
+    AwsIotProvisioningRequestParameterEntry_t provisioningParameters[ 2 ] = { { 0 } };
 
     /* Determine if a provisioning template name has been specified. */
-    if( ( pTemplateName == NULL ) || ( strlen( pTemplateName ) == 0 ) )
+    if( ( &pTemplateName[ 0 ] == NULL ) || ( strlen( pTemplateName ) == 0 ) )
```
muneebahmed10: & and [0] seems kind of like a hack, especially for a static variable. Since the default initialization is to "", is the first condition even necessary?

aggarw13: I agree that it is a hack. Both Dan Griffin and Cobus wanted the NULL check to exist as the value of the #define is customer configurable. (In a negative use case, one can remove the default `""` with nothing)

muneebahmed10: Adding to it, if the pointer value were NULL, wouldn't [ 0 ] cause an error as it dereferences the pointer? I don't think this part needs to be here.

muneebahmed10: Would casting the array to ( void * ) before the comparison work?

README.md
```diff
@@ -1,31 +1,19 @@
-# AWS IoT Device SDK C v4.0.0
+# AWS IoT Device SDK C with Fleet Provisioning<sup>BETA</sup> support
 
 **[Link to API documentation](https://docs.aws.amazon.com/freertos/latest/lib-ref/c-sdk/main/index.html)**
 
-[![Build Status](https://travis-ci.org/aws/aws-iot-device-sdk-embedded-C.svg?branch=v4_beta)](https://travis-ci.org/aws/aws-iot-device-sdk-embedded-C)
-[![Coverage Status](https://coveralls.io/repos/github/aws/aws-iot-device-sdk-embedded-C/badge.svg?branch=v4_beta)](https://coveralls.io/github/aws/aws-iot-device-sdk-embedded-C?branch=v4_beta)
+[![Build Status](https://travis-ci.org/aws/aws-iot-device-sdk-embedded-C.svg?branch=FleetProvisioning_beta)](https://travis-ci.org/aws/aws-iot-device-sdk-embedded-C)
+[![Coverage Status](https://coveralls.io/repos/github/aws/aws-iot-device-sdk-embedded-C/badge.svg?branch=FleetProvisioning_beta)](https://coveralls.io/github/aws/aws-iot-device-sdk-embedded-C?branch=FleetProvisioning_beta)
```
leegeth: is coverage supported yet? If so is it a good plan to keep it as a readme message

PR #677, opened 2019-12-06
fix rendering of path in install steps

jbpratt78:
*Description of changes:*
It seems the path was not being rendered properly so I swapped the character to use the HTML entities instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ This section explains the individual steps to retrieve the necessary files and b
 
 Steps:
 
- * Create a directory to hold your application e.g. (/home/<user>/aws_iot/my_app)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +59,7 @@ This section explains the individual steps to retrieve the necessary files and b
 
 Steps:
 
- * Create a directory to hold your application e.g. (/home/<user>/aws_iot/my_app)
+ * Create a directory to hold your application e.g. (/home/&lt;user&gt;/aws_iot/my_app)
```
dan4thewin: Would putting in a backslash before the brackets work? i.e., `/home/\<user>/aws_iot/my_app` => /home/\<user>/aws_iot/my_app

PR #682, opened 2019-12-13
Set priority and stack size for posix threads

gerardoti:
Set priority and stack size for posix threads port layer.

### DIFF
--- a/ports/posix/src/iot_threads_posix.c
+++ b/ports/posix/src/iot_threads_posix.c
@@ -160,10 +160,7 @@ bool Iot_CreateDetachedThread( IotThreadRoutine_t threadRoutine,
     _threadInfo_t * pThreadInfo = NULL;
     pthread_t newThread;
     pthread_attr_t threadAttributes;
-
-    /* Ignore priority and stack size. */
-    ( void ) priority;
-    ( void ) stackSize;
+    struct sched_param priorityParam;
 
     /* Allocate memory for the new thread info. */
     pThreadInfo = IotThreads_Malloc( sizeof( _threadInfo_t ) );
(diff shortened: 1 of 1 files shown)


ports/posix/src/iot_threads_posix.c
```diff
@@ -200,6 +197,34 @@ bool Iot_CreateDetachedThread( IotThreadRoutine_t threadRoutine,
         IOT_SET_AND_GOTO_CLEANUP( false );
     }
 
+    if( stackSize != 0 )
```
gordonwang0: Could you use the constants `IOT_THREAD_DEFAULT_PRIORITY` and `IOT_THREAD_DEFAULT_STACK_SIZE` in iot_platform_types.h instead of hardcoded 0?  Thanks.

gerardoti: Making that change would essentially cancel out these changes. We set `IOT_THREAD_DEFAULT_STACK_SIZE` & `IOT_THREAD_DEFAULT_PRIORITY` to the default value we want to be used to create threads by the SDK, such as for thread pools. The SDK internally will use this value when calling `Iot_CreateDetachedThread()`, if inside this function we are checking for this value and ignoring it if they are the same these would never be set as they will always match.

The only reason for checking for 0 is to see if the value of `IOT_THREAD_DEFAULT_STACK_SIZE` & `IOT_THREAD_DEFAULT_PRIORITY` were set in the configuration. If they were never set then these two macros will default to a value of 0, therefore the threads stack size and priority shouldn’t be set so that the system’s defaults are used.

gordonwang0: Understood, thank you.  I may change the priority comparison to `priority != -1` just because `0` may be a valid priority and this code would prevent that priority from being used.  Nothing further that you need to do though, just wanted to let you know.

PR #687, opened 2019-12-16
Refactor to reduce complexity

gordonwang0:

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_operation.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_operation.c
@@ -976,20 +976,11 @@ _mqttOperation_t * _IotMqtt_FindOperation( _mqttConnection_t * pMqttConnection,
     operationMatchParams.type = type;
     operationMatchParams.pPacketIdentifier = pPacketIdentifier;
 
-    if( pPacketIdentifier != NULL )
-    {
-        IotLogDebug( "(MQTT connection %p) Searching for operation %s pending response "
-                     "with packet identifier %hu.",
-                     pMqttConnection,
-                     IotMqtt_OperationType( type ),
-                     *pPacketIdentifier );
-    }
-    else
-    {
-        IotLogDebug( "(MQTT connection %p) Searching for operation %s pending response.",
-                     pMqttConnection,
-                     IotMqtt_OperationType( type ) );
-    }
+    IotLogDebug( "(MQTT connection %p) Searching for operation %s "
+                 "with packet identifier %hu.",
+                 pMqttConnection,
+                 IotMqtt_OperationType( type ),
+                 ( pPacketIdentifier == NULL ) ? 0 : *pPacketIdentifier );
 
     /* Find and remove the first matching element in the list. */
     IotMutex_Lock( &( pMqttConnection->referencesMutex ) );
@@ -1018,12 +1009,6 @@ _mqttOperation_t * _IotMqtt_FindOperation( _mqttConnection_t * pMqttConnection,
             if( taskPoolStatus != IOT_TASKPOOL_SUCCESS )
             {
                 pResult = NULL;
-            }
-            else
-            {
-                /* Check job reference counts. A waitable operation should have a
-                 * count of 2; a non-waitable operation should have a count of 1. */
-                IotMqtt_Assert( pResult->u.operation.jobReference == ( 1 + ( waitable == true ) ) );
             }
         }
         else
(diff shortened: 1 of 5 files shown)


libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -1019,12 +1010,6 @@ _mqttOperation_t * _IotMqtt_FindOperation( _mqttConnection_t * pMqttConnection,
             {
                 pResult = NULL;
             }
-            else
-            {
-                /* Check job reference counts. A waitable operation should have a
-                 * count of 2; a non-waitable operation should have a count of 1. */
-                IotMqtt_Assert( pResult->u.operation.jobReference == ( 1 + ( waitable == true ) ) );
```
leegeth: Is this assert not required anymore?

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1007,6 +1020,68 @@ static void _serializeUnsubscribe( const IotMqttSubscription_t * pSubscriptionLi
...
 /*-----------------------------------------------------------*/
 
+static IotMqttError_t _decodeSubackStatus( size_t statusCount,
+                                           const uint8_t * pStatusStart,
+                                           _mqttPacket_t * pSuback )
+{
+    IotMqttError_t status = IOT_MQTT_SUCCESS;
+    uint8_t subscriptionStatus = 0;
+    size_t i = 0;
+
```
leegeth: validating arguments for being non null would be good.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1007,6 +1020,68 @@ static void _serializeUnsubscribe( const IotMqttSubscription_t * pSubscriptionLi
...
+    {
+        /* Read a single status byte in SUBACK. */
+        subscriptionStatus = *( pStatusStart + i );
+
+        /* MQTT 3.1.1 defines the following values as status codes. */
+        switch( subscriptionStatus )
+        {
+            case 0x00:
+            case 0x01:
+            case 0x02:
```
leegeth: I know it is refactoring code, but would it be better to consider using constants/enum values for this case statements?

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1007,6 +1020,68 @@ static void _serializeUnsubscribe( const IotMqttSubscription_t * pSubscriptionLi
...
+                IotLog( IOT_LOG_DEBUG,
+                        &_logHideAll,
+                        "Topic filter %lu accepted, max QoS %hhu.",
+                        ( unsigned long ) i, subscriptionStatus );
+                break;
+
+            case 0x80:
```
leegeth: same as above

libraries/standard/mqtt/src/iot_mqtt_serialize.c
gordonwang0: This is an internal function, so we can guarantee its arguments will not be NULL.

libraries/standard/mqtt/src/iot_mqtt_operation.c
gordonwang0: Yes, it's not required.

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -252,6 +260,302 @@ static bool _validatePublish( bool awsIotMqttMode,
...
+
+/*-----------------------------------------------------------*/
+
+static bool _validateSubscription( bool awsIotMqttMode,
+                                   IotMqttOperationType_t operation,
+                                   const IotMqttSubscription_t * pSubscription )
+{
+    bool status = true;
+    uint16_t i = 0;
+
```
leegeth: need to validate pSubscription to be non null as well

gordonwang0: It's not possible for pSubscription to be NULL, since it's the address of an array element.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
leegeth: Already discussed about it with Gordon and I think I agree with him that having the raw values would give more visibility into what the code is doing.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1007,6 +1020,68 @@ static void _serializeUnsubscribe( const IotMqttSubscription_t * pSubscriptionLi
...
+    /* Iterate through each status byte in the SUBACK packet. */
+    for( i = 0; i < statusCount; i++ )
+    {
+        /* Read a single status byte in SUBACK. */
+        subscriptionStatus = *( pStatusStart + i );
+
+        /* MQTT 3.1.1 defines the following values as status codes. */
+        switch( subscriptionStatus )
+        {
+            case 0x00:
```
gkwicker: In general I'd prefer to see `enum` values here rather than integer values, but we won't hold this up for that since the code was already this way.

PR #689, opened 2019-12-17
Refactor subscription topic parsing

gordonwang0:


libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -61,6 +61,60 @@ typedef struct _packetMatchParams
 
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Handle special corner cases regarding terminal wildcards, as
```
dcgaws: Terminal meaning at the end of the string?

Is there a specific section of (a specific version of) the MQTT standard that we could reference? I have found these feature/standard references to be tough to audit after the fact.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -88,108 +142,123 @@ static bool _packetMatch( const IotLink_t * pSubscriptionLink,
...
-    bool status = false;
-    uint16_t nameIndex = 0, filterIndex = 0;
+    bool status = false, terminalChar = false;
 
-    /* Because this function is called from a container function, the given link
-     * must never be NULL. */
-    IotMqtt_Assert( pSubscriptionLink != NULL );
+    /* Determine if the terminal character is reached for both topic name and topic
+     * filter for the '#' wildcard. */
+    terminalChar = ( nameIndex == topicNameLength - 1 ) && ( filterIndex == topicFilterLength - 3 );
```
dcgaws: Why "- 1" and "- 3"?

Also, what if topicFilterLength is less than three?

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -88,108 +142,123 @@ static bool _packetMatch( const IotLink_t * pSubscriptionLink,
...
+                                     uint16_t topicFilterLength )
 {
-    bool status = false;
-    uint16_t nameIndex = 0, filterIndex = 0;
+    bool status = false, terminalChar = false;
 
-    /* Because this function is called from a container function, the given link
-     * must never be NULL. */
-    IotMqtt_Assert( pSubscriptionLink != NULL );
+    /* Determine if the terminal character is reached for both topic name and topic
```
dcgaws: Again, terminal seems to be an ambiguous term in this context, but maybe that's because I don't have the knowledge of the MQTT specification.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -88,108 +142,123 @@ static bool _packetMatch( const IotLink_t * pSubscriptionLink,
...
+    terminalChar = ( nameIndex == topicNameLength - 1 ) && ( filterIndex == topicFilterLength - 3 );
 
-    _mqttSubscription_t * pSubscription = IotLink_Container( _mqttSubscription_t,
-                                                             pSubscriptionLink,
-                                                             link );
-    _topicMatchParams_t * pParam = ( _topicMatchParams_t * ) pMatch;
+    if( terminalChar == true )
+    {
+        /* Determine if the topic filter ends with the '#' wildcard. */
+        status = ( pTopicFilter[ filterIndex + 1 ] == '/' ) && ( pTopicFilter[ filterIndex + 2 ] == '#' );
```
dcgaws: Are you guaranteed that filterIndex + 2 is inbounds for pTopicFilter?

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -88,108 +142,123 @@ static bool _packetMatch( const IotLink_t * pSubscriptionLink,
...
-        status = false;
-        goto cleanup;
+        /* Move topic name index to the end of the current level.
+         * This is identified by '/'. */
+        while( ( *pNameIndex < topicNameLength ) && ( pTopicName[ *pNameIndex ] != '/' ) )
+        {
+            ( *pNameIndex )++;
+        }
+
+        ( *pNameIndex )--;
```
dcgaws: Is there any chance that pNameIndex was zero?

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -88,108 +142,123 @@ static bool _packetMatch( const IotLink_t * pSubscriptionLink,
...
-                    nameIndex++;
-                }
-
-                /* Increment filter index to skip '/'. */
-                filterIndex++;
-                continue;
-            }
-            else if( pTopicFilter[ filterIndex ] == '#' )
+            /* Check for matching wildcards. */
+            if( _matchWildcards( pTopicFilter,
```
dcgaws: This is different code structure from above, where you assign the function output to a local, then test it.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -203,10 +272,56 @@ static bool _topicMatch( const IotLink_t * pSubscriptionLink,
...
+/*-----------------------------------------------------------*/
+
+static bool _topicMatch( const IotLink_t * pSubscriptionLink,
+                         void * pMatch )
+{
+    bool status = false;
+
+    /* Because this function is called from a container function, the given link
+     * must never be NULL. */
+    IotMqtt_Assert( pSubscriptionLink != NULL );
```
dcgaws: Seems like this should be a runtime check rather than an assert then.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -203,10 +272,56 @@ static bool _topicMatch( const IotLink_t * pSubscriptionLink,
...
+
+    _mqttSubscription_t * pSubscription = IotLink_Container( _mqttSubscription_t,
+                                                             pSubscriptionLink,
+                                                             link );
+    _topicMatchParams_t * pParam = ( _topicMatchParams_t * ) pMatch;
+
+    /* Extract the relevant strings and lengths from parameters. */
```
dcgaws: Will we run afoul of any older embedded compilers by placing these variable declarations mid-block?

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: There is: 
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/csprd02/mqtt-v3.1.1-csprd02.html#_Toc385349843

The specific requirements this function checks are:
>  “sport/#” also matches the singular “sport”, since # includes the parent level. 
and
> “sport/+” does not match “sport” but it does match “sport/”

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: `topicNameLength - 1` refers to the end of the topic name.  This is the index of the last character.

`topicFilterLength - 3` refers to the character before the wildcard and separator, so before "/#" (see line 160).  The index of this character is 3 before the end of the topic filter.

If `topicFilterLength` is less than 3, then `topicFilterLength - 3` will underflow to about 16535 (`topicFilterLength` is a u16).  This will cause the comparison `( filterIndex == topicFilterLength - 3 )` to evaluate to `false` (`filterIndex` is always in range `[0,topicFilterLength-1]`) and cause this check to be skipped.  That is the correct behavior as this is checking:
> “sport/#” also matches the singular “sport”, since # includes the parent level.

Which requires at least 3 characters ('/#' and another before the separator)

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: I'll update this comment to be clearer.  "terminal" just refers to the last character in the string.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: Yes.  The check at line 155 `( filterIndex == topicFilterLength - 3 )` requires a length of at least 3 to pass.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: Yes, if comparing against the topic name '/'.  That would cause this line to decrement nameIndex to 65535.  But line 267 accounts for this and would bring it back to 0.  The end behavior is still correct: the topic match comparison would fail.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: The behavior of these functions is slightly different:
- `_matchTerminalWildcards` only verifies positive matches, so it can exit on all true values.
- `_matchWildcards` may determine a positive or negative match; it needs a return value to determine whether it should exit, and the status for exit.

I'll think about how to make these more similar...

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: Perhaps that comment should be reworded.  It means that the caller is this function is guaranteed to never pass NULL, as it's an internal function we implemented.  So the assert for not NULL is supposed to pass, unless the library is implemented incorrectly.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: I believe this should still qualify as the top of the block?  The only lines separating it from the other declarations are comments.

PR #694, opened 2019-12-19
Refactor _serializeConnect for complexity

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Moved some if blocks in MQTT's _serializeConnect to their own functions.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_serialize.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_serialize.c
@@ -217,6 +217,27 @@ static uint8_t * _encodeString( uint8_t * pDestination,
                                 uint16_t sourceLength );
 
 /**
+ * @brief Get the CONNECT flags based on the given parameters.
+ * 
+ * @param[in] pConnectInfo User-provided CONNECT information.
+ * 
+ * @return CONNECT flags field for an MQTT packet.
+ */
+static uint8_t _getConnectFlags( const IotMqttConnectInfo_t * pConnectInfo );
+
+/**
+ * @brief Encode a username into a CONNECT packet, if necessary.
+ * 
+ * @param[out] pBuffer Buffer for the CONNECT packet.
+ * @param[in] pConnectInfo User-provided CONNECT information.
+ * 
+ * @return Pointer to the end of the encoded string, which will be identical to
+ * `pBuffer` if nothing was encoded.
+ */
+static uint8_t * _encodeUserName( uint8_t * pBuffer,
+                                  const IotMqttConnectInfo_t * pConnectInfo );
+
+/**
  * @brief Calculate the size and "Remaining length" of a CONNECT packet generated
  * from the given parameters.
  *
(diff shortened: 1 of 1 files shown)


libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -465,6 +486,141 @@ static uint8_t * _encodeString( uint8_t * pDestination,
...
+                 * standard. */
+                if( ( pConnectInfo->userNameLength + AWS_IOT_METRICS_USERNAME_LENGTH ) <= UINT16_MAX )
+                {
+                    /* Write the high byte of the combined length. */
+                    *( pBuffer++ ) = UINT16_HIGH_BYTE( ( pConnectInfo->userNameLength +
+                                                         AWS_IOT_METRICS_USERNAME_LENGTH ) );
+
+                    /* Write the low byte of the combined length. */
+                    *( pBuffer++ ) = UINT16_LOW_BYTE( ( pConnectInfo->userNameLength +
+                                                        AWS_IOT_METRICS_USERNAME_LENGTH ) );
```
muneebahmed10: Originally this was:
```
*pBuffer = ...
*( pBuffer + sizeof( uint8_t ) ) = ...
pBuffer += sizeof( uint16_t );
```

The `sizeof`s can be changed to `1` and `2`, but the complexity of this function would still be 8. Changing those three lines to two `*(pBuffer++)`'s reduced the complexity to 7, probably since there are less lines of code.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -216,6 +216,27 @@ static uint8_t * _encodeString( uint8_t * pDestination,
...
                                 uint16_t sourceLength );
 
+/**
+ * @brief Get the CONNECT flags based on the given parameters.
+ * 
+ * @param[in] pConnectInfo User-provided CONNECT information.
+ * 
+ * @return CONNECT flags field for an MQTT packet.
+ */
+static uint8_t _getConnectFlags( const IotMqttConnectInfo_t * pConnectInfo );
```
gordonwang0: Should rename to `setConnectFlags`, this function is setting the flags in a CONNECT packet.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
gordonwang0: Actually this looks rather confusing since UINT16_* are macros already.  Maybe change it back.

muneebahmed10: I'm not sure I understand. The only changes were:
`*pBuffer` -> `*( pBuffer++ )`
`*( pBuffer + sizeof( uint8_t ) )` -> `*( pBuffer++ )`
`pBuffer += sizeof( uint16_t );` -> deleted.

The UINT16_* were there already. Not making this change leaves the complexity at 8.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
muneebahmed10: I think this function could just be removed, as after testing again it seems most of the complexity from this _serializeConnect came from the username portion. Looks like leaving the connectFlags doesn't increase the complexity beyond 8.

PR #696, opened 2019-12-20
Reduce complexity for _deserializeIncomingPacket

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Creates helper functions to de-serialize each type of MQTT packet.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_network.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_network.c
@@ -66,6 +66,47 @@ static IotMqttError_t _getIncomingPacket( void * pNetworkConnection,
                                           _mqttPacket_t * pIncomingPacket );
 
 /**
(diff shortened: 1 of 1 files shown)


libraries/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -274,11 +315,155 @@ static IotMqttError_t _getIncomingPacket( void * pNetworkConnection,
...
+                                         pPacketIdentifier );
+
+    if( pOperation != NULL )
+    {
+        pOperation->u.operation.status = status;
+        _IotMqtt_Notify( pOperation );
+    }
+
+    return status;
+}
```
muneebahmed10: This function was abstracted out since four very similar blocks of code existed in `_deserializeIncomingPacket`, and so that each case in the `switch` statement would look more consistent. However, most of the complexity came from the publish and pingresp cases. Even without this function, the complexity of `_deserializeIncomingPacket` is still below 8, so it isn't absolutely necessary.

PR #700, opened 2019-12-24
Refactor MQTT operations for complexity

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Refactors _IotMqtt_ProcessSend, _IotMqtt_Notify, and _IotMqtt_AddSubscriptions to reduce GNU complexity below 8.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_operation.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_operation.c
@@ -116,7 +116,7 @@ static bool _scheduleNextRetry( _mqttOperation_t * pOperation );
  *
  * @return `true` if the operation was successful; `false` otherwise.
  */
-static bool _sendKeepAlivePingReq( _mqttConnection_t * pMqttConnection );
+static bool _sendPingReq( _mqttConnection_t * pMqttConnection );
 
 /**
  * @brief Schedule a callback for a completed MQTT operation.
@@ -332,7 +332,7 @@ static bool _scheduleNextRetry( _mqttOperation_t * pOperation )
 
 /*-----------------------------------------------------------*/
 
-static bool _sendKeepAlivePingReq( _mqttConnection_t * pMqttConnection )
+static bool _sendPingReq( _mqttConnection_t * pMqttConnection )
 {
     bool status = true;
     uint32_t swapStatus = 0;
@@ -363,12 +363,10 @@ static bool _sendKeepAlivePingReq( _mqttConnection_t * pMqttConnection )
                                                 0 );
         IotMqtt_Assert( swapStatus == 1 );
 
-        /* Schedule a check for PINGRESP. */
+        /* Set the period for scheduling a PINGRESP check. */
         pPingreqOperation->u.operation.periodic.ping.nextPeriodMs = IOT_MQTT_RESPONSE_WAIT_MS;
 
-        IotLogDebug( "(MQTT connection %p) PINGREQ sent. Scheduling check for PINGRESP in %d ms.",
-                     pMqttConnection,
-                     IOT_MQTT_RESPONSE_WAIT_MS );
+        IotLogDebug( "(MQTT connection %p) PINGREQ sent.", pMqttConnection );
     }
 
     return status;
@@ -779,7 +777,7 @@ void _IotMqtt_ProcessKeepAlive( IotTaskPool_t pTaskPool,
     {
         IotLogDebug( "(MQTT connection %p) Sending PINGREQ.", pMqttConnection );
 
-        status = _sendKeepAlivePingReq( pMqttConnection );
+        status = _sendPingReq( pMqttConnection );
     }
     else
     {
--- a/libraries/standard/mqtt/src/iot_mqtt_operation.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_operation.c
@@ -109,6 +109,37 @@ static bool _checkRetryLimit( _mqttOperation_t * pOperation );
  */
 static bool _scheduleNextRetry( _mqttOperation_t * pOperation );
 
+/**
+ * @brief Send a PINGREQ packet for a keep-alive operation.
+ *
+ * @param[in] pMqttConnection The associated MQTT connection.
+ *
+ * @return `true` if the operation was successful; `false` otherwise.
+ */
+static bool _sendKeepAlivePingReq( _mqttConnection_t * pMqttConnection );
+
+/**
+ * @brief Schedule a callback for a completed MQTT operation.
+ *
+ * @param[in] pOperation The completed MQTT operation.
+ *
+ * @return `IOT_MQTT_SUCCESS` if the schedule was successful;
+ * `IOT_MQTT_SCHEDULING_ERROR` otherwise.
+ */
+static IotMqttError_t _scheduleCallback( _mqttOperation_t * pOperation );
+
+/**
+ * @brief Complete a pending send operation.
+ *
+ * @param[in] pOperation The pending MQTT send operation.
+ * @param[out] pDestroyOperation Whether the operation should be destroyed afterwards.
+ *
+ * @return `true if the operation is awaiting a response from the network;
+ * `false` if not.
+ */
+static bool _completePendingSend( _mqttOperation_t * pOperation,
+                                  bool * pDestroyOperation );
+
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -791,74 +942,28 @@ void _IotMqtt_ProcessSend( IotTaskPool_t pTaskPool,
...
+        }
+        /* Non-waitable operations with no callback are also considered successful. */
+        else if( waitable == false )
+        {
+            if( pOperation->u.operation.notify.callback.function == NULL )
             {
-                if( pOperation->u.operation.notify.callback.function == NULL )
-                {
-                    pOperation->u.operation.status = IOT_MQTT_SUCCESS;
-                }
```
muneebahmed10: This if block had a nesting level of 5, so the outermost `else` was refactored into `else if`s to reduce the nesting depth.

libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -109,6 +109,37 @@ static bool _checkRetryLimit( _mqttOperation_t * pOperation );
...
+ */
+static IotMqttError_t _scheduleCallback( _mqttOperation_t * pOperation );
+
+/**
+ * @brief Complete a pending send operation.
+ *
+ * @param[in] pOperation The pending MQTT send operation.
+ * @param[out] pDestroyOperation Whether the operation should be destroyed afterwards.
+ *
+ * @return `true if the operation is awaiting a response from the network;
```
dcgaws: You're missing the trailing backtick for true. I don't think the backticks are necessary in general, though.

libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -301,6 +332,158 @@ static bool _scheduleNextRetry( _mqttOperation_t * pOperation )
...
+        status = false;
+    }
+    else
+    {
+        /* Assume the keep-alive will fail. The network receive callback will
+         * clear the failure flag upon receiving a PINGRESP. */
+        swapStatus = Atomic_CompareAndSwap_u32( &( pPingreqOperation->u.operation.periodic.ping.failure ),
+                                                1,
+                                                0 );
+        IotMqtt_Assert( swapStatus == 1 );
```
dcgaws: This assert doesn't make sense to me. Isn't the purpose of this block to safely handle a known/expected race condition?

libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -301,6 +332,158 @@ static bool _scheduleNextRetry( _mqttOperation_t * pOperation )
...
+static IotMqttError_t _scheduleCallback( _mqttOperation_t * pOperation )
+{
+    IotMqttError_t status = IOT_MQTT_SCHEDULING_ERROR;
+    _mqttConnection_t * pMqttConnection = pOperation->pMqttConnection;
+
+    /* Non-waitable operation should have job reference of 1. */
+    IotMqtt_Assert( pOperation->u.operation.jobReference == 1 );
+
+    /* Schedule an invocation of the callback. */
+    if( pOperation->u.operation.notify.callback.function != NULL )
```
dcgaws: Is it allowed for the app to specify a NULL pointer here? That seems like it should be a hard error.

muneebahmed10: Yes, the user-provided callback function isn't required, and the callback struct is set to 0 in the [initializer](https://docs.aws.amazon.com/freertos/latest/lib-ref/c-sdk/mqtt/iot__mqtt__types_8h.html#a9776f60b74d9f1a80dd4036085ab1569)

libraries/standard/mqtt/src/iot_mqtt_operation.c
muneebahmed10: Thanks, will fix. I'm not sure about the backticks, but I want to be consistent with the other comments in this file.

libraries/standard/mqtt/src/iot_mqtt_operation.c
muneebahmed10: I'm not entirely sure why the assert was part of the code if it's possible for the CAS to fail. Will update once I have more info.

libraries/standard/mqtt/src/iot_mqtt_operation.c
gordonwang0: The backticks are used so that variable names show up `like this` in the generated documentation.

libraries/standard/mqtt/src/iot_mqtt_operation.c
gordonwang0: The swap here is expected to always succeed.  The logic around the `failure` flag is:

1. Create connection with `failure=0`
1. When a ping request is sent, set `failure=1` (what this line of code is doing)
1. Schedule a task to check for ping response some time in the future.
1. If a response is received, set `failure=0`.
1. Task for checking response runs. If it sees `failure=0`, this loop repeats.  If it sees `failure=1`, that means no response was received and the connection terminates.

The only reason this CAS can fail is if `failure=1` here.  That shouldn't be possible because the connection would have been terminated by the lack of ping response and this code should never run.  That's what the assert checks.

libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -301,6 +332,158 @@ static bool _scheduleNextRetry( _mqttOperation_t * pOperation )
 
 /*-----------------------------------------------------------*/
 
+static bool _sendKeepAlivePingReq( _mqttConnection_t * pMqttConnection )
```
gordonwang0: Should just call this `_sendKeepAlive` or `_sendPingreq`.

libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -301,6 +332,158 @@ static bool _scheduleNextRetry( _mqttOperation_t * pOperation )
...
+         * clear the failure flag upon receiving a PINGRESP. */
+        swapStatus = Atomic_CompareAndSwap_u32( &( pPingreqOperation->u.operation.periodic.ping.failure ),
+                                                1,
+                                                0 );
+        IotMqtt_Assert( swapStatus == 1 );
+
+        /* Schedule a check for PINGRESP. */
+        pPingreqOperation->u.operation.periodic.ping.nextPeriodMs = IOT_MQTT_RESPONSE_WAIT_MS;
+
+        IotLogDebug( "(MQTT connection %p) PINGREQ sent. Scheduling check for PINGRESP in %d ms.",
```
gordonwang0: Where is the call to task pool Schedule?  This log message seems misplaced.

muneebahmed10: The scheduling happens at line 820 in `_IotMqtt_ProcessKeepAlive`, from which the code for this function was taken. Since that also logs `"(MQTT connection %p) Next keep-alive job in %lu ms."`, I think the reference to scheduling in this statement can just be removed to make more sense.

libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -791,74 +940,28 @@ void _IotMqtt_ProcessSend( IotTaskPool_t pTaskPool,
...
+        {
+            if( pOperation->u.operation.notify.callback.function == NULL )
             {
-                if( pOperation->u.operation.notify.callback.function == NULL )
-                {
-                    pOperation->u.operation.status = IOT_MQTT_SUCCESS;
-                }
+                pOperation->u.operation.status = IOT_MQTT_SUCCESS;
             }
         }
```
gkwicker: Note that [MISRA 2012 15.7 states](https://rules.sonarsource.com/c/RSPEC-126) _"All if...__else if__ constructs shall be terminated with an else statement"_.  It looks like those may have been pulled out in a previous revision as I don't see them there either.  No need to change anything for this review but this will need to be revisited for MISRA compliance later.

libraries/standard/mqtt/src/iot_mqtt_operation.c
dcgaws: Per our in-person discussion, it seems that swapStatus=0 could happen, however unlikely. While that's a pre-existing issue not directly related to this PR, there is a chance that this refactoring change could cause a latent bug of that nature to occur more frequently.

PR #706, opened 2020-01-03
Update/reduce complexity of provisioning api functions

aggarw13:
**Issue** - https://issues.amazon.com/issues/TS-10581

*Description of changes:*
* Simplify serialization logic in the Provisioning API functions to reduce their complexity scores by moving the logic to the serializer module 
* Update unit tests 

### DIFF
--- a/libraries/aws/provisioning/src/aws_iot_provisioning_api.c
+++ b/libraries/aws/provisioning/src/aws_iot_provisioning_api.c
@@ -145,6 +145,13 @@ static void _resetActiveOperationData();
  */
 static AwsIotProvisioningError_t _timedWaitForServerResponse( uint32_t timeoutMs );
 
+/**
+ * @brief Checks whether the data that is provided to send along with the provisioning device request is valid.
+ * @param pRequestData The data for the RegisterThing service API request whose validity will be checked.
+ * @return Returns #true if data is valid; #false otherwise.
+ */
+static bool _isDataForRegisterThingRequestValid( const AwsIotProvisioningRegisterThingRequestInfo_t * pRequestData );
+
 /*------------------------------------------------------------------*/
 
 static bool _checkInit( void )
@@ -316,6 +323,48 @@ static void _setActiveOperation( const _provisioningCallbackInfo_t * pUserCallba
 
     /* Decrement the reference count as we have released the mutex. */
     ( void ) Atomic_Decrement_u32( &_activeOperation.mutexReferenceCount );
+}
+
+/*-----------------------------------------------------------*/
+
+static bool _isDataForRegisterThingRequestValid( const AwsIotProvisioningRegisterThingRequestInfo_t * pRequestData )
+{
+    IOT_FUNCTION_ENTRY( bool, true );
+
+    if( pRequestData == NULL )
+    {
+        IotLogError( "Invalid request data passed for provisioning device." );
+
+        IOT_SET_AND_GOTO_CLEANUP( false );
+    }
+
+    if( ( pRequestData->pDeviceCertificateId == NULL ) ||
+        ( pRequestData->deviceCertificateIdLength == 0 ) )
+    {
+        IotLogError( "Invalid certificate ID data passed for device provisioning request." );
+
+        IOT_SET_AND_GOTO_CLEANUP( false );
+    }
+
+    if( ( pRequestData->pCertificateOwnershipToken == NULL ) ||
+        ( pRequestData->ownershipTokenLength == 0 ) )
+    {
+        IotLogError( "Invalid certificate ownership token data passed for device provisioning request." );
+
+        IOT_SET_AND_GOTO_CLEANUP( false );
+    }
+
+    /* Check that the provided template name is valid. */
+    if( ( pRequestData->pTemplateName == NULL ) ||
+        ( pRequestData->templateNameLength == 0 ) ||
+        ( pRequestData->templateNameLength > PROVISIONING_MAX_TEMPLATE_NAME_LENGTH ) )
+    {
+        IotLogError( "Invalid template name information passed for device provisioning request." );
+
+        IOT_SET_AND_GOTO_CLEANUP( false );
+    }
+
+    IOT_FUNCTION_EXIT_NO_CLEANUP();
 }
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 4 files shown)


libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -392,6 +401,37 @@ AwsIotProvisioningError_t AwsIotProvisioning_Init( uint32_t mqttTimeoutMs )
...
+    {
+        status = AWS_IOT_PROVISIONING_TIMEOUT;
+    }
+
+    /* There can be an edge case of receiving the server response right after the timeout has occurred.
+     * If such an edge case has occurred and causes the subscription callback to finish its execution after we timeout,
+     * we use the outcome of the callback execution instead of treating this case as a "timeout".
+     * Rationale - This API is not intended to have an "exact" implementation of server timeout, as the completion of the
+     * the operation is more valuable for the customer than having an exact behavior of server timeout. */
+    IotMutex_Lock( &_activeOperation.lock );
```
cobusve: This does not seem right, it seems like it is just reducing the probability of the race condition and not handling it in a reliable way? What happens if you take the lock, check the status, the response has not arrived, you release the lock and THEN the response arrives?

The only way in a case like this to reliably handle the race condition is to change state inside the lock (not just the local variable "status", but the state which will cause the response to be dropped if it arrives after this check instead of normally processing it.

This is probably better done by taking the lock inside the if statement above, checking the active operation inside that lock and then updating the local status AND the active operation status and then releasing the lock.

libraries/aws/provisioning/src/aws_iot_provisioning_serializer.c
```diff
@@ -294,6 +318,82 @@ AwsIotProvisioningError_t _AwsIotProvisioning_SerializeRegisterThingRequestPaylo
...
+
+    /* Get the calculated required size. */
+    *pBufferSize = _pAwsIotProvisioningEncoder->getExtraBufferSizeNeeded( &outermostPayloadEncoder );
+    AwsIotProvisioning_Assert( *pBufferSize != 0 );
+
+    /* Clean the encoder object handle. */
+    _pAwsIotProvisioningEncoder->destroy( &outermostPayloadEncoder );
+
+    /* Allocate memory for the request payload based on the size required from the dry-run of serialization */
+    *pSerializationBuffer = AwsIotProvisioning_MallocPayload( *pBufferSize * sizeof( uint8_t ) );
```
cobusve: Curious when this memory gets freed and how we ensure that it never leaks? I prefer the pattern where the one allocating the memory is responsible for also freeing it. There are many patterns you can use to ensure you alloc and free symmetrically. What I do not like is when the library allocates and the application frees, this is not symmetrical and also btw. makes it impossible for the application to use the stack or statically allocated memory for this. If you let the app allocate and free then it is easy to do just that.

libraries/aws/provisioning/test/unit/aws_iot_tests_provisioning_serializer.c
```diff
@@ -114,45 +114,37 @@ TEST_GROUP_RUNNER( Provisioning_Unit_Serializer )
...
+    /* Test the serializer function. */
     TEST_ASSERT_EQUAL( AWS_IOT_PROVISIONING_SUCCESS,
-                       _AwsIotProvisioning_SerializeCreateKeysAndCertificateRequestPayload( &testEncoder,
-                                                                                            pSerializationBuffer,
-                                                                                            sizeof( pSerializationBuffer ) ) );
-
-    _pAwsIotProvisioningEncoder->destroy( &testEncoder );
+                       _AwsIotProvisioning_SerializeCreateKeysAndCertificateRequestPayload( &pSerializationBuffer,
+                                                                                            &bufferSize ) );
+    TEST_ASSERT_NOT_NULL( pSerializationBuffer );
```
cobusve: When will this be NULL in a test like this? Seems like you should mock Malloc and force it to fail to get this to fail, which would be a good test, but at this point the test should always pass?

aggarw13: This is testing that the serializer function under test is allocating memory for the buffer pointer that was passed from the test to it as an output parameter. The initialized value of the buffer pointer in the test  is `NULL`.
I'm not sure what your suggestion of making `malloc` fail with the mock is intended for...

aggarw13: Is it to test that the serializer function will return an error code if the memory allocation fails?

libraries/aws/provisioning/src/aws_iot_provisioning_serializer.c
aggarw13: I understand that concern, and yes, having this asymmetry of memory allocation is not ideal. 
The responsibility of memory allocation and release was being handled by the caller initially, but that was inter-twined with the extraneous logic of calling the serialize function **twice**, **one** for a dry serialization run with the encoder which would be used to calculate the amount of buffer required for storing the serialized data, and **the other** for performing the _actual_ serialization. 
The above details are related to the functioning (or limitation) of the serializer which, I think, does not need to be exposed in the API function implementation. Thus, I decided to encapsulate this complexity in the serializer module so that the API implementation only contains a single call for serialization to improve readability and complexity score for the API function as the goal of this PR. 

I had discussed this with @gkwicker yesterday to make sure that the memory management asymmetry is a pain we are fine with taking (_for-now_) to reduce the complexity score (or improve readability) of the API function...

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: Discussed on Chime, and Cobus provided the suggestion to remove the race-condition handling logic as it does not fully resolve the race condition problem. 
Thus, we decided to remove the logic that attempts to partially solve the race condition case. The cost of not solving the race-condition is (just) a re-attempt of the request to the service API.

libraries/aws/provisioning/test/unit/aws_iot_tests_provisioning_serializer.c
cobusve: Just a general comment about testing. You should answer that question, if we do not know why we added this assert then I would be worried. I was just pointing out that it seems like this will never assert under test conditions, maybe I am wrong :)

PR #718, opened 2020-01-14
fix: MISRA 15.1 (goto) violations - part 2

abhidixi11:

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_api.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_api.c
@@ -511,7 +511,7 @@ static _mqttConnection_t * _createMqttConnection( bool awsIotMqttMode,
         }
     }
 
-    if( status != false )
+    if( status == true )
     {
         /* Create the subscription mutex for a new connection. */
         subscriptionMutexCreated = IotMutex_Create( &( pMqttConnection->subscriptionMutex ), false );
@@ -532,12 +532,9 @@ static _mqttConnection_t * _createMqttConnection( bool awsIotMqttMode,
             /* Check if keep-alive is active for this connection. */
             if( keepAliveSeconds != 0 )
             {
-                if( _createKeepAliveOperation( pNetworkInfo,
-                                               keepAliveSeconds,
-                                               pMqttConnection ) == false )
-                {
-                    status = false;
-                }
+                status = _createKeepAliveOperation( pNetworkInfo,
+                                                    keepAliveSeconds,
+                                                    pMqttConnection );
             }
         }
     }
@@ -773,28 +770,27 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
         {
             _IotMqtt_ProcessSend( IOT_SYSTEM_TASKPOOL, pSubscriptionOperation->job, pSubscriptionOperation );
         }
+        else if( ( status = _IotMqtt_ScheduleOperation( pSubscriptionOperation,
+                                                        _IotMqtt_ProcessSend,
+                                                        0 ) ) != IOT_MQTT_SUCCESS )
+        {
+            IotLogError( "(MQTT connection %p) Failed to schedule %s for sending.",
+                         mqttConnection,
+                         IotMqtt_OperationType( operation ) );
+
+            if( operation == IOT_MQTT_SUBSCRIBE )
+            {
+                _IotMqtt_RemoveSubscriptionByPacket( mqttConnection,
+                                                     pSubscriptionOperation->u.operation.packetIdentifier,
+                                                     MQTT_REMOVE_ALL_SUBSCRIPTIONS );
+            }
+
+            /* Clear the previously set (and now invalid) reference. */
+            _setOperationReference( pOperationReference, IOT_MQTT_OPERATION_INITIALIZER );
+        }
         else
         {
-            status = _IotMqtt_ScheduleOperation( pSubscriptionOperation,
-                                                 _IotMqtt_ProcessSend,
-                                                 0 );
-
-            if( status != IOT_MQTT_SUCCESS )
-            {
-                IotLogError( "(MQTT connection %p) Failed to schedule %s for sending.",
-                             mqttConnection,
-                             IotMqtt_OperationType( operation ) );
-
-                if( operation == IOT_MQTT_SUBSCRIBE )
-                {
-                    _IotMqtt_RemoveSubscriptionByPacket( mqttConnection,
-                                                         pSubscriptionOperation->u.operation.packetIdentifier,
-                                                         MQTT_REMOVE_ALL_SUBSCRIPTIONS );
-                }
-
-                /* Clear the previously set (and now invalid) reference. */
-                _setOperationReference( pOperationReference, IOT_MQTT_OPERATION_INITIALIZER );
-            }
+            /* Empty else MISRA 15.7 */
         }
     }
 
@@ -1066,24 +1062,22 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
         {
             status = IOT_MQTT_NO_MEMORY;
         }
-        else
-        {
-            /* Set the network connection associated with the MQTT connection. */
-            pNewMqttConnection->pNetworkConnection = pNetworkConnection;
-            pNewMqttConnection->ownNetworkConnection = ownNetworkConnection;
-
-            /* Set the MQTT packet serializer overrides. */
-            #if IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES == 1
-                pNewMqttConnection->pSerializer = pNetworkInfo->pMqttSerializer;
-            #else
-                pNewMqttConnection->pSerializer = NULL;
-            #endif
-        }
     }
 
     /* Set the MQTT receive callback. */
     if( status == IOT_MQTT_SUCCESS )
     {
+        /* Set the network connection associated with the MQTT connection. */
+        pNewMqttConnection->pNetworkConnection = pNetworkConnection;
+        pNewMqttConnection->ownNetworkConnection = ownNetworkConnection;
+
+        /* Set the MQTT packet serializer overrides. */
+        #if IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES == 1
+            pNewMqttConnection->pSerializer = pNetworkInfo->pMqttSerializer;
+        #else
+            pNewMqttConnection->pSerializer = NULL;
+        #endif
+
         networkStatus = pNewMqttConnection->pNetworkInterface->setReceiveCallback( pNetworkConnection,
                                                                                    IotMqtt_ReceiveCallback,
                                                                                    pNewMqttConnection );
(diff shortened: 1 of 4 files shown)


abhidixi11:
> Removing the gotos increases the complexity of functions. Please fix the following:
> 
Yes, I'm aware of that and we have a task to fix complexity after all gotos are removed and all MISRA work is done.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1054,126 +1035,124 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
...
+            pNewMqttConnection->pNetworkConnection = pNetworkConnection;
+            pNewMqttConnection->ownNetworkConnection = ownNetworkConnection;
+
+            /* Set the MQTT packet serializer overrides. */
+            #if IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES == 1
+                pNewMqttConnection->pSerializer = pNetworkInfo->pMqttSerializer;
+            #else
+                pNewMqttConnection->pSerializer = NULL;
+            #endif
+        }
```
muneebahmed10: The contents of this `else` block could also be at the beginning of the very next `if`. It makes no difference for this PR, but it would help in reducing complexity down the line due to a lower nesting level.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1054,126 +1035,124 @@ IotMqttError_t IotMqtt_Connect( const IotMqttNetworkInfo_t * pNetworkInfo,
...
-    }
+        /* Convert the connect info and will info objects to an MQTT CONNECT packet. */
+        status = _getMqttConnectSerializer( pNetworkInfo->pMqttSerializer )( pConnectInfo,
+                                                                             &( pOperation->u.operation.pMqttPacket ),
+                                                                             &( pOperation->u.operation.packetSize ) );
 
-    /* Check the serialized MQTT packet. */
-    IotMqtt_Assert( pOperation->u.operation.pMqttPacket != NULL );
-    IotMqtt_Assert( pOperation->u.operation.packetSize > 0 );
+        if( status == IOT_MQTT_SUCCESS )
```
muneebahmed10: Having an `if( status == IOT_MQTT_SUCCESS )` within another `if( status == IOT_MQTT_SUCCESS )` looks a bit weird to me. Maybe move it to its own block and reduce nesting complexity at the same time.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -273,14 +270,15 @@ static bool _topicFilterMatch( const char * pTopicName,
         filterIndex++;
     }
 
-    /* If the end of both strings has been reached, they match. */
-    if( ( nameIndex == topicNameLength ) && ( filterIndex == topicFilterLength ) )
+    if( status == false )
```
muneebahmed10: Is this condition necessary? Is there any harm in setting status to `true` in the event when status is already `true`?

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1711,19 +1686,21 @@ IotMqttError_t IotMqtt_Wait( IotMqttOperation_t operation,
...
+        /* Validate the given operation reference. */
+        if( _IotMqtt_ValidateOperation( operation ) == false )
+        {
+            status = IOT_MQTT_BAD_PARAMETER;
+        }
+        else
+        {
+            /* Check the MQTT connection status. */
+            pMqttConnection = operation->pMqttConnection;
+        }
```
muneebahmed10: Looks like the outermost `else` block can be removed if this `if` were changed to an `else if`

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1541,102 +1521,97 @@ IotMqttError_t IotMqtt_PublishAsync( IotMqttConnection_t mqttConnection,
...
+            status = IOT_MQTT_BAD_PARAMETER;
+        }
+        else
+        {
+            /* Create a PUBLISH operation. */
+            status = _IotMqtt_CreateOperation( mqttConnection,
+                                               flags,
+                                               pCallbackInfo,
+                                               &pOperation );
+        }
```
muneebahmed10: Same here: This can be turned into `if`-`else if`-`else` to reduce nesting.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -505,50 +499,50 @@ static _mqttConnection_t * _createMqttConnection( bool awsIotMqttMode,
...
-        goto cleanup;
+            status = false;
+        }
     }
 
-    /* Create the subscription mutex for a new connection. */
-    subscriptionMutexCreated = IotMutex_Create( &( pMqttConnection->subscriptionMutex ), false );
-
-    if( subscriptionMutexCreated == false )
+    if( status != false )
```
muneebahmed10: Nit: `if( status == true )` is more straightforward.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
abhidixi11: There is no harm, but if status is true, then we don't want to do unnecessary  comparison.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1541,85 +1519,76 @@ IotMqttError_t IotMqtt_PublishAsync( IotMqttConnection_t mqttConnection,
...
-                                             _IotMqtt_ProcessSend,
-                                             0 );
+        /* Send the PUBLISH packet. */
+        if( ( flags & MQTT_INTERNAL_FLAG_BLOCK_ON_SEND ) == MQTT_INTERNAL_FLAG_BLOCK_ON_SEND )
+        {
+            _IotMqtt_ProcessSend( IOT_SYSTEM_TASKPOOL, pOperation->job, pOperation );
+        }
+        else if( ( ( status = _IotMqtt_ScheduleOperation( pOperation,
+                                                        _IotMqtt_ProcessSend,
+                                                        0 ) )!= IOT_MQTT_SUCCESS ) )
```
muneebahmed10: Needs a space between `)` and `!=`. Better run uncrustify on all these files again to be safe.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -273,14 +270,15 @@ static bool _topicFilterMatch( const char * pTopicName,
...
     {
-        status = true;
+        /* If the end of both strings has been reached, they match. */
+        if( ( nameIndex == topicNameLength ) && ( filterIndex == topicFilterLength ) )
+        {
+            status = true;
+        }
```
muneebahmed10: Nit: I like reducing the number of `if`s whenever possible, and it looks like `status = ( nameIndex == topicNameLength ) && ( filterIndex == topicFilterLength );` would be equivalent to this block. Is there a disadvantage to this?

abhidixi11: Boolean expression in C returns 0 or 1. It does not return bool datatype.  I believe that's the reason we don't use boolDataType = expression.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -762,62 +751,54 @@ static IotMqttError_t _subscriptionCommon( IotMqttOperationType_t operation,
...
         {
-            IotLogError( "(MQTT connection %p) Failed to schedule %s for sending.",
-                         mqttConnection,
-                         IotMqtt_OperationType( operation ) );
+            status = _IotMqtt_ScheduleOperation( pSubscriptionOperation,
+                                                 _IotMqtt_ProcessSend,
+                                                 0 );
 
-            if( operation == IOT_MQTT_SUBSCRIBE )
+            if( status != IOT_MQTT_SUCCESS )
```
muneebahmed10: The same thing as line 1588 can be done here by moving the assignment and comparison into an `else if` to reduce a nesting level. Not necessary for this PR though. In both cases, we'd need an empty `else` to comply with MISRA [15.7](https://rules.sonarsource.com/c/RSPEC-126)

libraries/standard/mqtt/src/iot_mqtt_subscription.c
abhidixi11: Since it is used elsewhere I'll make the change.

libraries/standard/mqtt/src/iot_mqtt_api.c
abhidixi11: I avoided empty else statements ,  unless it is  necessary to  reduce complexity we should not use it. Since this function has higher complexity, I'll try to make the change.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -704,37 +696,31 @@ static IotMqttError_t _subscriptionCreateAndSerialize( IotMqttOperationType_t op
...
-    IotMqtt_Assert( pSubscriptionOperation->u.operation.periodic.retry.limit == 0 );
-    pSubscriptionOperation->u.operation.type = operation;
 
-    /* Generate a subscription packet from the subscription list. */
-    status = serializeSubscription( pSubscriptionList,
-                                    subscriptionCount,
-                                    &( pSubscriptionOperation->u.operation.pMqttPacket ),
-                                    &( pSubscriptionOperation->u.operation.packetSize ),
-                                    &( pSubscriptionOperation->u.operation.packetIdentifier ) );
+        /* Check the subscription operation data and set the operation type. */
```
dcgaws: This change is reordering this logic in addition to removing the goto. What's the rationale for the former?

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -911,9 +888,9 @@ bool _IotMqtt_IncrementConnectionReferences( _mqttConnection_t * pMqttConnection
     {
         ( pMqttConnection->references )++;
 
-        /* In some implementations IotLog() maps to C standard printing API 
-         * that need specific primitive types for format specifiers. Also, 
-         * inttypes.h may not be available on some C99 compilers, despite 
+        /* In some implementations IotLog() maps to C standard printing API
+         * that need specific primitive types for format specifiers. Also,
```
dcgaws: Should this be "needs"? It's not clear whether API is plural in this sentence.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -911,9 +888,9 @@ bool _IotMqtt_IncrementConnectionReferences( _mqttConnection_t * pMqttConnection
     {
         ( pMqttConnection->references )++;
 
-        /* In some implementations IotLog() maps to C standard printing API 
-         * that need specific primitive types for format specifiers. Also, 
-         * inttypes.h may not be available on some C99 compilers, despite 
+        /* In some implementations IotLog() maps to C standard printing API
```
dcgaws: The call is to IoTLogDebug.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1541,85 +1515,76 @@ IotMqttError_t IotMqtt_PublishAsync( IotMqttConnection_t mqttConnection,
...
-    {
-        status = _IotMqtt_ScheduleOperation( pOperation,
-                                             _IotMqtt_ProcessSend,
-                                             0 );
+        /* Send the PUBLISH packet. */
+        if( ( flags & MQTT_INTERNAL_FLAG_BLOCK_ON_SEND ) == MQTT_INTERNAL_FLAG_BLOCK_ON_SEND )
+        {
+            _IotMqtt_ProcessSend( IOT_SYSTEM_TASKPOOL, pOperation->job, pOperation );
+        }
+        else if( ( ( status = _IotMqtt_ScheduleOperation( pOperation,
```
dcgaws: Is this combined assign/compare also a MISRA violation?

libraries/standard/mqtt/src/iot_mqtt_operation.c
```diff
@@ -270,9 +270,9 @@ static bool _scheduleNextRetry( _mqttOperation_t * pOperation )
             pOperation->u.operation.periodic.retry.nextPeriodMs = IOT_MQTT_RETRY_MS_CEILING;
         }
 
-        /* In some implementations IotLog() maps to C standard printing API 
-         * that need specific primitive types for format specifiers. Also 
-         * inttypes.h may not be available on some C99 compilers, despite 
+        /* In some implementations IotLog() maps to C standard printing API
+         * that need specific primitive types for format specifiers. Also
+         * inttypes.h may not be available on some C99 compilers, despite
```
dcgaws: Could these debug output lines be rewritten so that the MISRA exception isn't required?

libraries/standard/mqtt/src/iot_mqtt_api.c
abhidixi11: Yes, it is (13.4) . Along with goto removal, we have to also make sure complexity is below 8. This is one of those functions where we have complexity score more than 8, So I opted for using assign and compare, in order to keep it less. I'll add annotation for these for now. I'll revisit when I take another shot at complexity reduction.

libraries/standard/mqtt/src/iot_mqtt_operation.c
abhidixi11: That change was part of another PR and we agreed upon this exception for log messages only. We need to have this due to format specifiers in log messages. For example, uint32_t may not map to unsigned long in some platforms.

libraries/standard/mqtt/src/iot_mqtt_api.c
dcgaws: Agreed, thank you.

PR #722, opened 2020-01-15
Check for a NULL pOperation before cleaning up.

sarenameas:
- Before _deserializePublish would immediately goto cleanup if the MallocOperation failed.
- In commit f1a8b698b1d2fd5b8475b11a16d399181c1d867f the goto was deleted for the MISRA Rule 15.1 effort.


libraries/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -425,7 +425,7 @@ static IotMqttError_t _deserializePublish( _mqttConnection_t * pMqttConnection,
     if( status != IOT_MQTT_SUCCESS )
     {
         /* Check ownership of the received MQTT packet. */
-        if( pOperation->u.publish.pReceivedData != NULL )
+        if( ( pOperation != NULL ) && ( pOperation->u.publish.pReceivedData != NULL ) )
```
dcgaws: This does not appear to be sufficient. Could you change line 425 to skip this block if status indicates an allocation failure?

muneebahmed10: +1. Line 438 also dereferences `pOperation`. It appears that `status` can only be `IOT_MQTT_NO_MEMORY` if the malloc failed, so that check might be sufficient.

abhidixi11: This block of code will not execute if pOperation is NULL. On line 364 we set the status to IOT_MQTT_NO_MEMORY, so the status will not be IOT_MQTT_SUCCESS. Once we set the status to something other than IOT_MQTT_SUCCESS, it should not enter any of the other blocks.

sarenameas: Right I see now.  But Misra also likes this change. Maybe I can just add a comment or we can keep it as a defensive?

abhidixi11: Just to add to above comment for the context - this MISRA violation got introduced when I removed 'goto' after malloc failure.

sarenameas:
Adding misra comment instead.

libraries/standard/mqtt/src/iot_mqtt_network.c
abhidixi11: I take back my comment, the failure happens when it not IOT_MQTT_SUCCESS. Original cleanup label was pointing directly above return. Thanks Muneeb.

sarenameas: Let me go back and fix this more carefully.

libraries/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -422,7 +422,7 @@ static IotMqttError_t _deserializePublish( _mqttConnection_t * pMqttConnection,
     }
 
     /* Free PUBLISH operation on error. */
-    if( status != IOT_MQTT_SUCCESS )
+    if( ( status != IOT_MQTT_NO_MEMORY ) && ( status != IOT_MQTT_SUCCESS ) )
```
abhidixi11: This will increase complexity, but we will address it later.

dcgaws: Is there any other reason that status could have been set to IOT_MQTT_NO_MEMORY? There are other API calls above that return status. It might be safer to instead test for pOperation being NULL.

sarenameas: Good point

PR #726, opened 2020-01-16
Address MISRA 12.1, 13, and 14

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Addresses violations of MISRA rules 12.1, 13.2, 13.3, 13.6, and 14.3 found by coverity.


libraries/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -483,13 +483,18 @@ static IotMqttError_t _deserializeIncomingPacket( _mqttConnection_t * pMqttConne
...
+
+    /* Packet validity is not checked when asserts are disabled. */
+    ( void ) packetValidity;
 
     /* A buffer for remaining data must be allocated if remaining length is not 0. */
     IotMqtt_Assert( ( pIncomingPacket->remainingLength > 0 ) ==
                     ( pIncomingPacket->pRemainingData != NULL ) );
 
     /* Only valid packets should be given to this function. */
-    IotMqtt_Assert( _incomingPacketValid( pIncomingPacket->type ) == true );
```
muneebahmed10: The 13.6 violation here was "The operand of the sizeof operator contains potential side effect: calling function _incomingPacketValid(pIncomingPacket->type)". This seems to be a false positive as the `sizeof` operator shouldn't be getting called. I introduced a boolean here to reduce the number of coverity exceptions we have in comments, but that can be changed if it's more desirable.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -522,12 +522,14 @@ static uint8_t * _encodeUserName( uint8_t * pBuffer,
...
+                    *pBuffer = UINT16_HIGH_BYTE( ( pConnectInfo->userNameLength +
+                                                   AWS_IOT_METRICS_USERNAME_LENGTH ) );
+                    pBuffer++;
 
                     /* Write the low byte of the combined length. */
-                    *( pBuffer++ ) = UINT16_LOW_BYTE( ( pConnectInfo->userNameLength +
-                                                        AWS_IOT_METRICS_USERNAME_LENGTH ) );
+                    *pBuffer = UINT16_LOW_BYTE( ( pConnectInfo->userNameLength +
+                                                  AWS_IOT_METRICS_USERNAME_LENGTH ) );
+                    pBuffer++;
```
muneebahmed10: This change does increase the complexity to 8, but this is acceptable as the requirement is <= 8 and not < 8.

libraries/standard/mqtt/src/iot_mqtt_network.c
```diff
@@ -483,13 +483,18 @@ static IotMqttError_t _deserializeIncomingPacket( _mqttConnection_t * pMqttConne
                                                   _mqttPacket_t * pIncomingPacket )
 {
     IotMqttError_t status = IOT_MQTT_STATUS_PENDING;
+    bool packetValidity = false;
+
+    /* Packet validity is not checked when asserts are disabled. */
+    ( void ) packetValidity;
```
abhidixi11: packetValidity is used on line 496, which is not part of an assert statement, so (void) may not be required.

muneebahmed10: I was following the pattern I observed [here](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/iot_mqtt_operation.c#L717), but after testing with Travis, it doesn't look like there are any build warnings without the voids. I will remove them.

libraries/standard/common/include/iot_linear_containers.h
```diff
@@ -496,6 +496,11 @@ static inline void IotListDouble_Remove( IotLink_t * const pLink )
     IotContainers_Assert( pLink != NULL );
 
     /* This function must be called on a linked element. */
+    /* Coverity finds a false positive in the below assert.
+     * As this is only an assert, the sizeof operator should
+     * not be called. Regardless, as IotLink_IsLinked does not
+     * mutate any data, there is no side effect in calling it. */
+    /* coverity[misra_c_2012_rule_13_6_violation] */
```
sarenameas: Does this violation still come up in the release type build?

muneebahmed10: No, it doesn't. It also doesn't come up if asserts are disabled.

gordonwang0: Let's remove the annotation then and run Coverity with asserts disabled.

libraries/standard/mqtt/src/iot_mqtt_network.c
gordonwang0: Should change it back.  `packetValidity` is not used when asserts are disabled.

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -526,7 +527,8 @@ void _IotMqtt_InvokeSubscriptionCallback( _mqttConnection_t * pMqttConnection,
         if( pSubscription->unsubscribed == true )
         {
             /* An unsubscribed subscription should have been removed from the list. */
-            IotMqtt_Assert( IotLink_IsLinked( &( pSubscription->link ) ) == false );
+            linkStatus = IotLink_IsLinked( &( pSubscription->link ) );
```
gordonwang0: Similarly, linkStatus will not be used when asserts are disabled.  Strange that the build check does not fail this...

libraries/standard/common/include/iot_linear_containers.h
muneebahmed10: Reverted

libraries/standard/mqtt/src/iot_mqtt_network.c
muneebahmed10: Reverted

PR #728, opened 2020-01-17
Comply with MISRA Rules 11.1, 15.7, and 16.4

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Addition of const modifiers and empty else blocks

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_serialize.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_serialize.c
@@ -838,6 +838,7 @@ static void _serializeConnect( const IotMqttConnectInfo_t * pConnectInfo,
                 break;
 
             default:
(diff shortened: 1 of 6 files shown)


libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -825,6 +825,7 @@ static void _serializeConnect( const IotMqttConnectInfo_t * pConnectInfo,
                 break;
 
             default:
+                /* Empty default MISRA 16.4 */
```
abhidixi11: Do we need this comment here?

abhidixi11: It is interesting that a comment just sufficient here.

PR #729, opened 2020-01-17
Fix MISRA 15.1 part 3

abhidixi11:
Note: Please only review Part 3 changes.  Part 2 changes have been already committed. 
*Issue #, if available:*

*Description of changes:*


libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1375,88 +1429,87 @@ IotMqttError_t _IotMqtt_DeserializeConnack( _mqttPacket_t * pConnack )
                 pConnack->type );
 
         status = IOT_MQTT_BAD_RESPONSE;
-        goto cleanup;
     }
```
muneebahmed10: Remove newline on 1433 so there's less space between the `if` and `else if`

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1572,127 +1625,101 @@ IotMqttError_t _IotMqtt_DeserializePublish( _mqttPacket_t * pPublish )
...
+    if( status == IOT_MQTT_SUCCESS )
     {
-        /* PUBLISH packet is invalid if both QoS 1 and QoS 2 bits are set. */
-        if( UINT8_CHECK_BIT( publishFlags, MQTT_PUBLISH_FLAG_QOS1 ) == true )
+        IotLog( IOT_LOG_DEBUG,
+                &_logHideAll,
+                "QoS is %d.", pOutput->qos );
+
+        /* Parse the DUP bit. */
+        if( UINT8_CHECK_BIT( publishFlags, MQTT_PUBLISH_FLAG_DUP ) == true )
```
muneebahmed10: Perhaps this and line 1622 could be in `_getIncomingPublishQOS`, so that all `UINT8_CHECK_BIT`s were in one function called `_checkPublishFlags`?

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -2041,23 +2066,18 @@ IotMqttError_t _IotMqtt_DeserializePingresp( _mqttPacket_t * pPingresp )
...
+    else if( pPingresp->remainingLength != MQTT_PACKET_PINGRESP_REMAINING_LENGTH )
     {
         IotLog( IOT_LOG_ERROR,
                 &_logHideAll,
                 "PINGRESP does not have remaining length of %d.",
                 MQTT_PACKET_PINGRESP_REMAINING_LENGTH );
 
         status = IOT_MQTT_BAD_RESPONSE;
-        goto cleanup;
     }
```
muneebahmed10: Empty `else` here

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -2519,36 +2500,33 @@ IotMqttError_t IotMqtt_DeserializePublish( IotMqttPacketInfo_t * pMqttPacket )
...
 
-    if( status == IOT_MQTT_SUCCESS )
-    {
-        pMqttPacket->pubInfo = mqttOperation.u.publish.publishInfo;
-        pMqttPacket->packetIdentifier = mqttPacket.packetIdentifier;
+        if( status == IOT_MQTT_SUCCESS )
+        {
+            pMqttPacket->pubInfo = mqttOperation.u.publish.publishInfo;
+            pMqttPacket->packetIdentifier = mqttPacket.packetIdentifier;
+        }
```
muneebahmed10: This doesn't need to be part of this `else` block, but this function is simple enough that this additional nesting doesn't matter.

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -115,6 +129,14 @@ static bool _validateWildcardPlus( uint16_t index,
 static bool _validateWildcardHash( uint16_t index,
                                    const IotMqttSubscription_t * pSubscription );
 
+/**
+ * @brief Check the MQTT clientId length does not exceed.
+ *
+ * @param[in] pConnectInfo The #IotMqttConnectInfo_t to validate.
+ *
+ */
```
muneebahmed10: `@return`?

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -202,65 +209,100 @@ static bool _validatePublish( bool awsIotMqttMode,
...
+            status = false;
         }
         else
         {
-            if( pPublishInfo->pPayload == NULL )
-            {
-                IotLogError( "Nonzero payload length cannot have a NULL payload." );
+            /* Empty else MISRA 15.7 */
+        }
+    }
```
muneebahmed10: This also needs an `else`

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -147,52 +169,37 @@ static bool _validateString( const char * pString,
...
+                                     size_t maximumPayloadLength,
+                                     const char * pPublishTypeDescription )
 {
     bool status = true;
 
     /* This parameter is not used when logging is disabled. */
     ( void ) pPublishTypeDescription;
 
-    /* Check for NULL. */
     if( pPublishInfo == NULL )
```
muneebahmed10: Is this necessary? At line 257, `_validatePublish` only calls `_validatePublishPayload` when `pPublishInfo != NULL`

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -438,57 +475,103 @@ static bool _validateWildcardHash( uint16_t index,
...
 
-bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
+static bool _validateClientIdLength( const IotMqttConnectInfo_t * pConnectInfo )
 {
     bool status = true;
     uint16_t maxClientIdLength = MQTT_SERVER_MAX_CLIENTID_LENGTH;
     bool enforceMaxClientIdLength = false;
 
-    /* Check for NULL. */
     if( pConnectInfo == NULL )
```
muneebahmed10: Same here as above. `_IotMqtt_ValidateConnect` already checks pConnectInfo for `NULL`

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -678,26 +734,22 @@ bool _IotMqtt_ValidateSubscriptionList( IotMqttOperationType_t operation,
                                 pListStart,
                                 listSize );
 
-    if( status == false )
+    if( status != false )
```
muneebahmed10: `status == true` is more straightforward

libraries/standard/mqtt/src/iot_mqtt_serialize.c
abhidixi11: Yup. Agree. I'll move it out in next iteration.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
abhidixi11: OK that's good suggestion, that will reduce complexity as well.

libraries/standard/mqtt/src/iot_mqtt_validate.c
abhidixi11: That's correct, but this is independent function and may be used later somewhere else (and may get flagged since we are checking for NULL)

libraries/standard/mqtt/src/iot_mqtt_validate.c
abhidixi11: Same comment as above, since it is an independent function now, I added this extra check.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -382,15 +382,16 @@ static IotMqttError_t _checkRemainingLength( _mqttPacket_t * pPublish,
                                              size_t qos0Minimum );
 
 /**
- * @brief Determine QOS for incoming publish flags.
+ * @brief Process incoming Publish publish flags.
  *
- * @param[in] publishFlags Incoming publish flags where QOS is encoded.
- * @param[in, out] pOutput Pointer to IotMqttPublishInfo_t struct where QOS will be written.
+ * @param[in] publishFlags Incoming publish flags.
+ * @param[in, out] pOutput Pointer to IotMqttPublishInfo_t struct
```
muneebahmed10: Nit: Tag `IotMqttPublishInfo_t` with `#` like other comments

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -382,15 +382,16 @@ static IotMqttError_t _checkRemainingLength( _mqttPacket_t * pPublish,
...
+ * @param[in] publishFlags Incoming publish flags.
+ * @param[in, out] pOutput Pointer to IotMqttPublishInfo_t struct 
+ * where output will be written.
  *
  * @return #IOT_MQTT_SUCCESS, #IOT_MQTT_BAD_RESPONSE.
  */
 
-static IotMqttError_t _getIncomingPublishQOS( const uint8_t publishFlags,
+static IotMqttError_t _processIncomingPublishFlags( const uint8_t publishFlags,
                                               IotMqttPublishInfo_t * pOutput );
```
muneebahmed10: Uncrustify this

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -382,15 +382,16 @@ static IotMqttError_t _checkRemainingLength( _mqttPacket_t * pPublish,
                                              size_t qos0Minimum );
 
 /**
- * @brief Determine QOS for incoming publish flags.
+ * @brief Process incoming Publish publish flags.
```
muneebahmed10: Remove first `Publish`

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1178,6 +1191,74 @@ static IotMqttError_t _checkRemainingLength( _mqttPacket_t * pPublish,
...
+                    &_logHideAll,
+                    "DUP is 1." );
+        }
+        else
+        {
+            IotLog( IOT_LOG_DEBUG,
+                    &_logHideAll,
+                    "DUP is 0." );
+        }
+    }
```
dan4thewin: Is there a field in pOutput that should be updated here?

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1377,88 +1458,87 @@ IotMqttError_t _IotMqtt_DeserializeConnack( _mqttPacket_t * pConnack )
...
+                status = IOT_MQTT_BAD_RESPONSE;
+            }
+        }
+        else
         {
-            status = IOT_MQTT_BAD_RESPONSE;
-            goto cleanup;
+            IotLog( IOT_LOG_DEBUG,
+                    &_logHideAll,
+                    "CONNACK session present bit not set." );
```
dan4thewin: Is this block solely for protocol verification?  I think we should consider Postel's law.  ( https://en.wikipedia.org/wiki/Robustness_principle )

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1377,88 +1458,87 @@ IotMqttError_t _IotMqtt_DeserializeConnack( _mqttPacket_t * pConnack )
...
+        else
+        {
+            /* Print the appropriate message for the CONNACK response code if logs are
+             * enabled. */
+            #if LIBRARY_LOG_LEVEL > IOT_LOG_NONE
+                IotLog( IOT_LOG_DEBUG,
+                        &_logHideAll,
+                        "%s",
+                        pConnackResponses[ pRemainingData[ 1 ] ] );
+            #endif
```
dan4thewin: Why is this IotLog() wrapped with #if when the preceding ones are not?

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -1567,134 +1647,83 @@ IotMqttError_t _IotMqtt_DeserializePublish( _mqttPacket_t * pPublish )
...
-                "DUP is 1." );
+        /* Extract the topic name starting from the first byte of the variable header.
+         * The topic name string starts at byte 3 in the variable header. */
+        pOutput->topicNameLength = UINT16_DECODE( pVariableHeader );
+
+        /* Sanity checks for topic name length and "Remaining length". The remaining
+         * length must be at least as large as the variable length header. */
+        status = _checkRemainingLength( pPublish,
+                                        pOutput->qos,
+                                        pOutput->topicNameLength + sizeof( uint16_t ) );
```
dan4thewin: I don't see any way in which we communicate to the user that the buffer is incomplete.  Seems like that would make this hard to use with a streams model.  Am I missing something?

PR #730, opened 2020-01-17
Address MISRA Rule 11.2 violations

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
MISRA Rule 11.2 prohibits a cast from an incomplete type - such as a `struct _networkConnection *` - to `void *`. This changes instances of `void *` to `IotNetworkConnection_t` when dealing with pointers to network connections.


abhidixi11:
Was this particular error (use of void*) flagged as part of 11.2 ? Note: This change may break light weight MQTT POC and will need update.  We have to make sure we update doxygen for CSDK and it is an API change, if we decide to go with the change. 
Note: Not all changes are Lightweight API related.
For lightweight MQTTAPI,  IotNetworkConnection_t should be just socket, but for regular MQTT API, it is defined to be struct _networkConnection* . To avoid two separate definitions void * was used.

PR #731, opened 2020-01-21
[Preview Only] Remove goto statements in Jobs library

qiutongs:
*Issue #, if available:* https://sim.amazon.com/issues/TS-10687

*Description of changes:*
Remove goto statements in Jobs library


libraries/aws/jobs/src/aws_iot_jobs_serialize.c
```diff
@@ -1010,104 +1015,107 @@ static AwsIotJobsError_t _parseErrorDocument( const char * pErrorDocument,
...
+                {
+                    status = AWS_IOT_JOBS_INVALID_STATE;
+                }
 
-        default:
-            break;
+                break;
+
+            default:
+                break;
```
muneebahmed10: Unrelated to this PR as it was already there, but having an empty `default` without a comment violates MISRA 16.4. It doesn't need to be fixed for this PR but it's only a one line change to add a comment like `/* Empty default MISRA 16.4 */`.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
```diff
@@ -492,33 +489,35 @@ static AwsIotJobsError_t _generateGetPendingRequest( const AwsIotJobsRequestInfo
...
-    /* Ensure offsets are valid. */
-    AwsIotJobs_Assert( copyOffset == requestLength );
-    AwsIotJobs_Assert( pOperation->pClientToken > pOperation->pJobsRequest );
-    AwsIotJobs_Assert( pOperation->pClientToken <
-                       pOperation->pJobsRequest + pOperation->jobsRequestLength );
-
-    IotLogDebug( "Jobs GET PENDING request: %.*s",
-                 pOperation->jobsRequestLength,
-                 pOperation->pJobsRequest );
+    if( status == AWS_IOT_JOBS_SUCCESS )
```
muneebahmed10: This `if`, along with several others in this PR, could just be an `else` since `status` is only changed in the previous `if` block.

qiutongs: That is a better idea. I will do.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
qiutongs: No problem. I can fix it in this PR. Does that comment tell user that the empty default is intended? 
I got a better idea. Put following line to defaut:

```
status = AWS_IOT_JOBS_BAD_RESPONSE;
```

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
```diff
@@ -1010,104 +1011,106 @@ static AwsIotJobsError_t _parseErrorDocument( const char * pErrorDocument,
                                    &pCode,
                                    &codeLength ) == false )
     {
-        IOT_SET_AND_GOTO_CLEANUP( AWS_IOT_JOBS_BAD_RESPONSE );
+        status = AWS_IOT_JOBS_BAD_RESPONSE;
     }
-
-    /* Match the JSON error code to a Jobs return value. Assume invalid status
-     * unless matched.*/
-    status = AWS_IOT_JOBS_BAD_RESPONSE;
```
muneebahmed10: I believe deleting this line changes the behavior of the function, as now if the switch does not flow into the default, then `AWS_IOT_JOBS_STATUS_PENDING` can be returned instead of `AWS_IOT_JOBS_BAD_RESPONSE`.

You could save a few lines by initializing `status` to `AWS_IOT_JOBS_BAD_RESPONSE`, changing the `if` block to `== true`, and move the `switch` into the `if`

qiutongs: You are totally right. It seems the AWS_IOT_JOBS_STATUS_PENDING will never return in original code.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
```diff
@@ -695,55 +694,56 @@ static AwsIotJobsError_t _generateDescribeRequest( const AwsIotJobsRequestInfo_t
...
-    if( includeJobDocument == false )
-    {
-        copyOffset = _appendFlag( pJobsRequest,
-                                  copyOffset,
-                                  INCLUDE_JOB_DOCUMENT_KEY,
-                                  INCLUDE_JOB_DOCUMENT_KEY_LENGTH,
-                                  false );
-    }
+        /* Add the "include job document" flag if false. */
+        if( includeJobDocument == false )
```
dcgaws: It looks like we're appending the job document only if the caller has told us not to.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
```diff
@@ -897,108 +897,109 @@ static AwsIotJobsError_t _generateUpdateRequest( const AwsIotJobsRequestInfo_t *
...
-                       EXPECTED_VERSION_KEY,
-                       EXPECTED_VERSION_KEY_LENGTH );
+        /* Construct the request JSON. */
+        APPEND_STRING( pJobsRequest, copyOffset, "{\"", 2 );
+
+        /* Add the status. */
+        APPEND_STRING( pJobsRequest, copyOffset, STATUS_KEY, STATUS_KEY_LENGTH );
         APPEND_STRING( pJobsRequest, copyOffset, "\":\"", 3 );
-        APPEND_STRING( pJobsRequest, copyOffset, pExpectedVersion, expectedVersionLength );
+        APPEND_STRING( pJobsRequest, copyOffset, pStatus, statusLength );
```
dcgaws: This change, and those below, appear to be above and beyond the scope of just removing goto statements. Should these be a separate PR with a title and description that better captures the rationale?

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
```diff
@@ -1008,106 +1009,102 @@ static AwsIotJobsError_t _parseErrorDocument( const char * pErrorDocument,
...
-        IOT_SET_AND_GOTO_CLEANUP( AWS_IOT_JOBS_BAD_RESPONSE );
-    }
-
-    /* Match the JSON error code to a Jobs return value. Assume invalid status
-     * unless matched.*/
-    status = AWS_IOT_JOBS_BAD_RESPONSE;
+        switch( codeLength )
+        {
+            /* InvalidJson */
+            case 13:
```
dcgaws: These should be #defines instead of magic numbers.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
qiutongs: I didn't change it actually. That line is same as the line 913 in original file. 
It is just the diff view problem. If you click "hide white space changes" , it will be clearer.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
qiutongs: Not sure about this. If you doubt the behavior, I can talk to the library owner.

muneebahmed10: This block is adding just the 'include job document' flag, not the actual job document. From the comments on lines 652-653, it appears that the default value of this flag is `true`, so it does not need to be added in that case.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
abhidixi11: Yes, it is due to uncrustify, or code shifting due to additional if/else block.

libraries/aws/jobs/src/aws_iot_jobs_serialize.c
abhidixi11: +1 Although this PR is not addressing updating  code other than gotos, it should be easy fix.

qiutongs: I thought they are error code but actually not. They are the string length. For example, 13 is for "InvalidJson" (including quote).

The way it does is not elegant. I think ideally a "dictionary" data structure is perfect for such case, which can be implemented by a Trie. For the purpose of this PR, I suggest we leave it.

qiutongs: Not easy fix because what it does here is to match a string to a set of per-defined ones.

PR #734, opened 2020-01-21
Shadow/misra directive 4 6

nateglims:
*Description of changes:* Add exception for `long int` required by the printf format specifier.


gordonwang0:
long int is not guaranteed to be int32_t.  The %ld specifier requires long int.  In other instances like this, we have just suppressed the error with an annotation and explanation.

sarenameas:
> long int is not guaranteed to be int32_t. The %ld specifier requires long int. In other instances like this, we have just suppressed the error with an annotation and explanation.

See https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/iot_mqtt_api.c#L892 for an example and approved annotation text.

PR #735, opened 2020-01-22
misra shadow: resolve 10.1 rule violations

nateglims:
*Description of changes:*
* Add unsigned annotation to flag literals
* Change variable used for shifting to unsigned
* Explicitly cast enum value used in arithmetic expressions

### DIFF
--- a/libraries/aws/shadow/src/aws_iot_shadow_api.c
+++ b/libraries/aws/shadow/src/aws_iot_shadow_api.c
@@ -349,7 +349,7 @@ static AwsIotShadowError_t _setCallbackCommon( IotMqttConnection_t mqttConnectio
     }
 
     /* Check parameters. */
-    status = _validateThingNameFlags( ( _shadowOperationType_t ) ( type + SHADOW_OPERATION_COUNT ),
+    status = _validateThingNameFlags( ( _shadowOperationType_t ) ( ( int32_t ) type + SHADOW_OPERATION_COUNT ),
                                       pThingName,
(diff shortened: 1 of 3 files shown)


libraries/aws/shadow/src/aws_iot_shadow_api.c
```diff
@@ -349,7 +349,7 @@ static AwsIotShadowError_t _setCallbackCommon( IotMqttConnection_t mqttConnectio
     }
 
     /* Check parameters. */
-    status = _validateThingNameFlags( ( _shadowOperationType_t ) ( type + SHADOW_OPERATION_COUNT ),
+    status = _validateThingNameFlags( ( _shadowOperationType_t ) ( ( int32_t ) type + SHADOW_OPERATION_COUNT ),
```
cobusve: Is your intent to typecast "type" and then let the Integer promotion take care of the rest, or is your intent to cast the result of the opertation? I think brackets make the intent clearer in these cases.

cobusve:
Also you will need to refresh this branch is out of date now

libraries/aws/shadow/src/aws_iot_shadow_api.c
nateglims: The specific violation is that we are doing arithmetic with an enum (`type`).

cobusve: Then I expect (( int32_t) type ) + SHADOW_OPERATION_COUNT

PR #738, opened 2020-01-22
Remove goto statements in aws_iot_shadow_parser.c

qiutongs:
*Issue #, if available:* https://sim.amazon.com/issues/TS-10929

*Description of changes:*
Remove goto statements in aws_iot_shadow_parser.c

### DIFF
--- a/libraries/aws/shadow/src/aws_iot_shadow_parser.c
+++ b/libraries/aws/shadow/src/aws_iot_shadow_parser.c
@@ -35,9 +35,6 @@
 /* Shadow internal include. */
 #include "private/aws_iot_shadow_internal.h"
 
-/* Error handling include. */
-#include "iot_error.h"
-
 /* AWS Parser include. */
 #include "aws_iot_doc_parser.h"
 
@@ -133,7 +130,7 @@ static AwsIotShadowError_t _codeToShadowStatus( uint32_t code )
 AwsIotShadowError_t _AwsIotShadow_ParseErrorDocument( const char * pErrorDocument,
                                                       size_t errorDocumentLength )
 {
-    IOT_FUNCTION_ENTRY( AwsIotShadowError_t, AWS_IOT_SHADOW_STATUS_PENDING );
+    AwsIotShadowError_t status = AWS_IOT_SHADOW_STATUS_PENDING;
     const char * pCode = NULL, * pMessage = NULL;
     size_t codeLength = 0, messageLength = 0;
     uint32_t code = 0;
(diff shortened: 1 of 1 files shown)


libraries/aws/shadow/src/aws_iot_shadow_parser.c
```diff
@@ -144,52 +141,53 @@ AwsIotShadowError_t _AwsIotShadow_ParseErrorDocument( const char * pErrorDocumen
...
-                                   &pMessage,
-                                   &messageLength ) == true )
+                                   &codeLength ) == true )
     {
-        IotLogWarn( "Code %lu: %.*s.",
-                    code,
-                    messageLength,
-                    pMessage );
+        /* Code must be in error document. */
+        AwsIotShadow_Assert( ( pCode > pErrorDocument ) &&
```
aggarw13: (Minor) Not sure if this assert check is really necessary. The first expression checks if the code is present after the starting address of the error document. Do you know why that is a requirement? If this function is carrying the knowledge of the JSON encoding format, and is thus, checking that the binary error code's location exists after the starting location, I'm not a fan of this parser to carry serialization format level information.

(I understand that this is out of the scope of the PR, but I would like it to be removed if not required)

libraries/aws/shadow/src/aws_iot_shadow_parser.c
```diff
@@ -144,52 +141,53 @@ AwsIotShadowError_t _AwsIotShadow_ParseErrorDocument( const char * pErrorDocumen
...
+                             ( pCode + codeLength < pErrorDocument + errorDocumentLength ) );
+
+        /* Convert the code to an unsigned integer value. */
+        code = ( uint32_t ) strtoul( pCode, NULL, 10 );
```
aggarw13: Not sure if the string to integer conversion is actually required here, as it seems like the error code will only be logged, which can be done with the string version stored in `pCode`

qiutongs: That code is also used here:
```
status = _codeToShadowStatus( code );
```

libraries/aws/shadow/src/aws_iot_shadow_parser.c
qiutongs: I agree with you on this. @gordonwang0 Do you think so?

qiutongs: @leegeth Can you comment on this?

PR #739, opened 2020-01-23
Enable quality check, but allow it to fail

gordonwang0:


scripts/ci_test_quality.sh
```diff
@@ -8,4 +8,4 @@ set -e
 # Get the list of files to check. Only check library files and exclude tests.
 # Run complexity with a threshold of 8.
 find ../libraries/ \( -name '*.c' ! -name *tests*.c \) -type f | \
-xargs complexity --scores --threshold=8 --horrid-threshold=8
+xargs complexity --scores --threshold=8 #--horrid-threshold=8
```
muneebahmed10: Per the man page, we should note that this test will still fail if there's a function with complexity > 100. I don't think we'll be increasing the complexity of any function that high though.

gordonwang0: Yeah, I think a failure in that case is fine

muneebahmed10:
Nit: also this file seems to be the only script whose permissions aren't `755`. Can we update that now?

scripts/ci_test_quality.sh
gkwicker: Why do we have the commented-out `#--horrid-threshold=8`?  If it doesn't add value, remove it.

gordonwang0: It's temporarily commented out to allow the check to pass.  At some point, we will begin enforcing it; I left it there so it's easy to re-enable.

PR #741, opened 2020-01-23
Address MISRA 17.7, 17.8, and 18.7 in MQTT

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Addresses MISRA violations for 17.7 and 17.8, and suppresses 18.7.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_api.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_api.c
@@ -1646,25 +1646,23 @@ IotMqttError_t IotMqtt_PublishSync( IotMqttConnection_t mqttConnection,
     IotMqttError_t status = IOT_MQTT_STATUS_PENDING;
     IotMqttOperation_t publishOperation = IOT_MQTT_OPERATION_INITIALIZER,
                        * pPublishOperation = NULL;
-    uint32_t asyncFlags = flags;
+    /* Set only the "serial" flag. */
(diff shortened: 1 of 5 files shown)


libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1636,21 +1636,25 @@ IotMqttError_t IotMqtt_PublishSync( IotMqttConnection_t mqttConnection,
     IotMqttError_t status = IOT_MQTT_STATUS_PENDING;
     IotMqttOperation_t publishOperation = IOT_MQTT_OPERATION_INITIALIZER,
                        * pPublishOperation = NULL;
+    uint32_t asyncFlags = flags;
```
gordonwang0: Consider rename, these are flags to make the async function do a sync operation.

gkwicker: +1.  Also don't assign it the value of `flags` here since the comment on line 1641 specifies it as ignored.  Instead, just assign it directly, e.g.:

```c
/* Set only the "serial" flag. */
uint32_t syncFlags = MQTT_INTERNAL_FLAG_BLOCK_ON_SEND;
```

...and then you can delete lines 1644/1645.

muneebahmed10: Will do

PR #742, opened 2020-01-24
Fix: Reduce code complexity below 9 Part 2

abhidixi11:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_api.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_api.c
@@ -511,7 +511,7 @@ static _mqttConnection_t * _createMqttConnection( bool awsIotMqttMode,
         }
     }
 
-    if( status != false )
+    if( status == true )
     {
         /* Create the subscription mutex for a new connection. */
         subscriptionMutexCreated = IotMutex_Create( &( pMqttConnection->subscriptionMutex ), false );
@@ -532,12 +532,9 @@ static _mqttConnection_t * _createMqttConnection( bool awsIotMqttMode,
             /* Check if keep-alive is active for this connection. */
             if( keepAliveSeconds != 0 )
             {
-                if( _createKeepAliveOperation( pNetworkInfo,
-                                               keepAliveSeconds,
-                                               pMqttConnection ) == false )
-                {
-                    status = false;
-                }
+                status = _createKeepAliveOperation( pNetworkInfo,
+                                                    keepAliveSeconds,
+                                                    pMqttConnection );
             }
         }
     }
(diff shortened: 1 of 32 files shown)


libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -726,6 +802,46 @@ static IotMqttError_t _subscriptionCreateAndSerialize( IotMqttOperationType_t op
...
+                         mqttConnection,
+                         IotMqtt_OperationType( operation ) );
+
+            if( operation == IOT_MQTT_SUBSCRIBE )
+            {
+                _IotMqtt_RemoveSubscriptionByPacket( mqttConnection,
+                                                     pSubscriptionOperation->u.operation.packetIdentifier,
+                                                     MQTT_REMOVE_ALL_SUBSCRIPTIONS );
+            }
+        }
```
muneebahmed10: Except for this block, this function looks identical to `_sendPublish`. Perhaps they could be combined into one function and this logic could be moved back to `_subscriptionCommon`?

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -876,6 +975,162 @@ static IotMqttError_t _waitForOperation( IotMqttOperation_t operation,
...
+         *
+         * This error is triggered by a dereference of 'pNewMqttConnection' in
+         * '_destroyMqttConnection'. Coverity assumes that 'pNewMqttConnection'
+         * was freed in '_IotMqtt_CreateOperation' above, where cleanup code will
+         * free 'pNewMqttConnection' upon allocation failure.
+         *
+         * This will never happen as a valid MQTT connection passed to this
+         * function always has a positive reference count; therefore,
+         * '_IotMqtt_CreateOperation' will not free it. Only unreferenced MQTT
+         * connections will be freed.
```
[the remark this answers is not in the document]
abhidixi11: I no longer see 'USE AFTER FREE' violation after removing the coverity_ignore. So I'll get rid the comment completely.

libraries/standard/mqtt/src/iot_mqtt_api.c
abhidixi11: _IotMqtt_ProcessSend() is common for all send functions, so it is already factored out. I'm not sure if complexity of _subscriptionCommon() will increase by doing that. I'll leave it that way right now, we can revisit later if needed.

abhidixi11: I have updated the function and and the complexity is still below 9,  so we are good.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -876,6 +952,125 @@ static IotMqttError_t _waitForOperation( IotMqttOperation_t operation,
...
+    pOperation->u.operation.type = IOT_MQTT_CONNECT;
+
+    /* Add previous session subscriptions. */
+    if( pConnectInfo->pPreviousSubscriptions != NULL )
+    {
+        /* Previous subscription count should have been validated as nonzero. */
+        IotMqtt_Assert( pConnectInfo->previousSubscriptionCount > 0U );
+
+        status = _IotMqtt_AddSubscriptions( pMqttConnection,
+                                            2,
```
gkwicker: This should be a `#define` or `enum` rather than a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)), especially since this is part of the MQTT spec.  It looks like the closest match in the current code is [`MQTT_PACKET_TYPE_SUBSCRIBE`](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/private/iot_mqtt_internal.h#L256), however this is `0x82` (that's also [the value in the spec](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718063))rather than 2 as specified here.  Please add a `#define` or `enum` for this value.

abhidixi11: This is not  setting 'packet type' , but setting packet identifier. It can have any unique value to identify a packet.  This function is adding existing subscriptions to the managed subscription internal list, therefore therefore it can be any value.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -267,6 +328,14 @@ static IotMqttError_t _waitForOperation( IotMqttOperation_t operation,
 
 /*-----------------------------------------------------------*/
 
+/**
+ * @brief Place holder packet identifier used when
+ * _IotMqtt_AddSubscriptions is called with previous subscriptions lists.
```
gkwicker: I would add a comment clarifying if there are any restrictions on this value; why 2 and not 3?  If there are no restrictions then 0 might be a more obvious choice.

abhidixi11: Agree. There is no restriction and it is not used so 0 should work.  I'll add the clarification in the next PR, and update it to 0 if that works.

PR #752, opened 2020-01-29
Annotate mqtt source files to suppress MISRA 20.7

qiutongs:
*Issue #, if available:* https://sim.amazon.com/issues/TS-10728

*Description of changes:*
Annotate mqtt source files to suppress MISRA 20.7

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_api.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_api.c
@@ -299,6 +299,8 @@ static bool _mqttSubscription_setUnsubscribe( const IotLink_t * const pSubscript
      * must never be NULL. */
     IotMqtt_Assert( pSubscriptionLink != NULL );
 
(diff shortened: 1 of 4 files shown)


libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -299,6 +299,8 @@ static bool _mqttSubscription_setUnsubscribe( const IotLink_t * const pSubscript
      * must never be NULL. */
     IotMqtt_Assert( pSubscriptionLink != NULL );
 
+    /* coverity[misra_c_2012_rule_20_7_violation] */
+    /* coverity[caretline] */
```
muneebahmed10: Can we have a comment above this about why these suppressions are necessary? It can be copy pasted to the other eight as well.

qiutongs: ok, let me add them

PR #753, opened 2020-01-29
Address MQTT MISRA 11.5, 11.8, and 11.9

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Fixes for some violations of MISRA 11.8 and 11.9 and suppressions for other 11.8 violations. Disables 11.5 (cast from void * to object pointer) as `void *` is used for abstractions and passing context.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_serialize.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_serialize.c
@@ -2060,7 +2060,11 @@ IotMqttError_t _IotMqtt_SerializePingreq( uint8_t ** pPingreqPacket,
                                           size_t * pPacketSize )
 {
     /* PINGREQ packets are always the same. */
-    static const uint8_t pPingreq[ MQTT_PACKET_PINGREQ_SIZE ] =
+    /* It is not necessary to make this array const. Since there are other
+     * types of MQTT packets that are not constant, this array would be
+     * cast to remove the const qualifier anyway. */
+    /* coverity[misra_c_2012_rule_8_13_violation] */
+    static uint8_t pPingreq[ MQTT_PACKET_PINGREQ_SIZE ] =
     {
         MQTT_PACKET_TYPE_PINGREQ,
         0x00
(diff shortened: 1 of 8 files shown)


libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -2059,7 +2059,11 @@ IotMqttError_t _IotMqtt_SerializePingreq( uint8_t ** pPingreqPacket,
                                           size_t * pPacketSize )
 {
     /* PINGREQ packets are always the same. */
-    static const uint8_t pPingreq[ MQTT_PACKET_PINGREQ_SIZE ] =
+    /* It is not necessary to make this array const. Since there are other
+     * types of MQTT packets that are not constant, this array would be
+     * cast to remove the const qualifier anyway. */
+    /* coverity[misra_c_2012_rule_8_13_violation] */
+    static uint8_t pPingreq[ MQTT_PACKET_PINGREQ_SIZE ] =
```
muneebahmed10: The 11.8 violation occurred on line 2073 ( removing `const`-ness of `pPingreq` by casting to `uint8_t *` ). After some offline discussion, I decided the best way to deal with this would be to remove the `const` from the array declaration, since adding `const` anywhere to `pPingreqPacket` would require changing the function signature, which the serializer overrides rely on.

scripts/coverity_misra.config
```diff
@@ -26,6 +26,10 @@
             deviation: "Rule 2.5",
             reason: "Allow unused macros. Library headers may define macros intended for the application's use, but not used by a specific file."
         },
+        {
+            deviation: "Rule 11.5",
+            reason: "Allow casts from void *. Casts to and from void * are used for abstractions."
+        },
```
muneebahmed10: After offline discussion with both @gordonwang0 and @gkwicker , disabling this rule seemed to be the way forward, for the reason stated.

gordonwang0: Suggest reword to "Allow casts from void *. Contexts are passed as void * and must be cast to the correct data type before use."

PR #755, opened 2020-01-29
Shadow Misra 8.3 and 10.5 Violations

nateglims:
*Description of changes:*
- Fix 8.3 and 10.5 MISRA violations
- Fix mismatched parameter names between function declaration and definitions.
- Convert enums that derived their value with arithmetic expressions or casting to macros and literal numbers.

### DIFF
--- a/libraries/aws/shadow/include/types/aws_iot_shadow_types.h
+++ b/libraries/aws/shadow/include/types/aws_iot_shadow_types.h
@@ -332,6 +332,21 @@ typedef enum AwsIotShadowError
     AWS_IOT_SHADOW_SERVER_ERROR = 500,
 } AwsIotShadowError_t;
 
+/**
+ * @ingroup shadow_datatypes_enums
+ * @brief Types of Shadow library callbacks.
+ *
+ * One of these values will be placed in #AwsIotShadowCallbackParam_t.callbackType
+ * to identify the reason for invoking a callback function.
+ */
+typedef enum AwsIotShadowCallbackType
+{
+    AWS_IOT_SHADOW_DELETE_COMPLETE, /**< Callback invoked because a [Shadow delete](@ref shadow_function_deleteasync) completed. */
+    AWS_IOT_SHADOW_GET_COMPLETE,    /**< Callback invoked because a [Shadow get](@ref shadow_function_getasync) completed. */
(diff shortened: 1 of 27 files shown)


libraries/aws/shadow/include/types/aws_iot_shadow_types.h
```diff
@@ -381,7 +366,7 @@ typedef enum AwsIotShadowCallbackType
  */
 typedef struct AwsIotShadowCallbackParam
 {
-    AwsIotShadowCallbackType_t callbackType; /**< @brief Reason for invoking the Shadow callback function to provide context. */
+    uint32_t callbackType; /**< @brief Reason for invoking the Shadow callback function to provide context. */
```
gkwicker: I would prefer `enum` to `uint32_t` here.  Is the underlying problem that this library was using the old `enum` in arithmetic?  If so, the *best* solution would be make that not be the case so that we could still use an `enum`; however if we don't have time to do that then I suppose this could be acceptable.

gordonwang0: Yes, also for consistency with other libraries (which use enum).

nateglims: My comment wasn't clear. Casting other essential types into enums is forbidden in MISRA 10.5. 
The original solution was this
https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/743/files 

I believe the concern was the lookup was possibly too expensive.

libraries/aws/shadow/src/aws_iot_shadow_operation.c
```diff
@@ -184,8 +184,80 @@ IotMutex_t _AwsIotShadowPendingOperationsMutex;
 
 /*-----------------------------------------------------------*/
 
-static bool _shadowOperationMatch( const IotLink_t * pOperationLink,
-                                   void * pMatch )
+_shadowOperationType_t _AwsIotShadow_IntToShadowOperationType( int32_t n )
```
gkwicker: Why take a signed parameter here?  If you used `uint32_t n` here, then line 214 would just be:

```c
AwsIotShadow_Assert( n < 5 );
```

Same comment applies on `_AwsIotShadow_IntToShadowCallbackType()`

PR #756, opened 2020-01-29
Suppress MISRA 19.2 in MQTT files

qiutongs:
*Issue #, if available:* https://sim.amazon.com/issues/TS-10727

*Description of changes:*
Suppress MISRA 19.2 in MQTT files

### DIFF
--- a/libraries/standard/mqtt/include/types/iot_mqtt_types.h
+++ b/libraries/standard/mqtt/include/types/iot_mqtt_types.h
@@ -445,6 +445,7 @@ typedef struct IotMqttCallbackParam
      */
     IotMqttConnection_t mqttConnection;
 
+    /* coverity[misra_c_2012_rule_19_2_violation] */
     union
     {
         /* Valid for completed operations. */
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/include/types/iot_mqtt_types.h
```diff
@@ -445,6 +445,7 @@ typedef struct IotMqttCallbackParam
      */
     IotMqttConnection_t mqttConnection;
 
+    /* coverity[misra_c_2012_rule_19_2_violation] */
```
muneebahmed10: We should have a comment justifying this suppression, something about how we are using unions here to reduce struct size

qiutongs: I didn't add that because I think that is a simple well-known general C knowledge, not specific to our code.  If you insist, I can add.

muneebahmed10: We should at least have something about unions here, as otherwise whoever is reading this code would have to be familiar with or look up MISRA 19.2. Since 19.2 specifically says unions should not be used, we should say why we are using them here.

qiutongs:
> Please add a comment about what this annotation is for; not all readers may be familiar with the MISRA spec.

OK.

PR #757, opened 2020-01-29
Complete coverage of iot_mqtt_validate.c

gordonwang0:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_validate.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_validate.c
@@ -517,7 +517,7 @@ static bool _validateClientId( const IotMqttConnectInfo_t * pConnectInfo )
     }
 
     /* Check for a zero-length client identifier. Zero-length client identifiers
-     * are not allowed with clean sessions. */
+     * are not allowed with persistent sessions. */
     if( status == true )
     {
         if( pConnectInfo->clientIdentifierLength == 0U )
--- a/libraries/standard/mqtt/src/iot_mqtt_validate.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_validate.c
@@ -217,10 +217,6 @@ static bool _validatePublishPayload( const IotMqttPublishInfo_t * pPublishInfo,
             /* Empty else MISRA 15.7 */
         }
     }
-    else
-    {
-        /* Empty else MISRA 15.7 */
-    }
 
     return status;
 }
@@ -539,10 +535,6 @@ static bool _validateClientId( const IotMqttConnectInfo_t * pConnectInfo )
         maxClientIdLength = AWS_IOT_MQTT_SERVER_MAX_CLIENTID_LENGTH;
         enforceMaxClientIdLength = true;
     }
-    else
-    {
-        /* Empty else MISRA 15.7 */
-    }
 
     if( status == true )
     {
@@ -584,10 +576,9 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
 
         status = false;
     }
-
-    /* Check client identifier. */
-    if( status == true )
+    else
     {
+        /* Check client identifier.*/
         status = _validateClientId( pConnectInfo );
     }
 
--- a/libraries/standard/mqtt/src/iot_mqtt_validate.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_validate.c
@@ -130,13 +130,14 @@ static bool _validateWildcardHash( uint16_t index,
                                    const IotMqttSubscription_t * pSubscription );
 
 /**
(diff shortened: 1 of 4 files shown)


libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -194,13 +195,7 @@ static bool _validatePublishPayload( const IotMqttPublishInfo_t * pPublishInfo,
...
     ( void ) pPublishTypeDescription;
 
-    if( pPublishInfo == NULL )
-    {
-        IotLogError( "Publish information cannot be NULL." );
-
-        status = false;
-    }
-    else if( pPublishInfo->payloadLength != 0U )
+    if( pPublishInfo->payloadLength != 0U )
```
muneebahmed10: Since the `else if` was turned into an `if`, it seems the `else` block at line 220 could be removed.

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -508,20 +502,39 @@ static bool _validateWildcardHash( uint16_t index,
...
+                IotLogError( "A zero-length client identifier cannot be used with a persistent session." );
+
+                status = false;
+            }
+        }
+    }
+
     /* The AWS IoT MQTT service enforces a client ID length limit. */
-    else if( pConnectInfo->awsIotMqttMode == true )
+    if( pConnectInfo->awsIotMqttMode == true )
```
muneebahmed10: Same for the `else` at line 542

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -571,30 +584,11 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
...
-        if( pConnectInfo->cleanSession == true )
-        {
-            IotLogError( "A zero-length client identifier cannot be used with a clean session." );
 
-            status = false;
-        }
-    }
-    else
+    /* Check client identifier. */
+    if( status == true )
```
muneebahmed10: This could just be an `else`

tests/iot_config.h
```diff
@@ -121,7 +121,7 @@
 
 /* Enable asserts in the libraries. */
 #define IOT_CONTAINERS_ENABLE_ASSERTS           ( 1 )
-#define IOT_MQTT_ENABLE_ASSERTS                 ( 1 )
+#define IOT_MQTT_ENABLE_ASSERTS                 ( ! IOT_TEST_COVERAGE )
```
abhidixi11: So is this change just disabling asserts for unit test ?

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -508,20 +502,39 @@ static bool _validateWildcardHash( uint16_t index,
...
 
-static bool _validateClientIdLength( const IotMqttConnectInfo_t * pConnectInfo )
+static bool _validateClientId( const IotMqttConnectInfo_t * pConnectInfo )
 {
     bool status = true;
     uint16_t maxClientIdLength = MQTT_SERVER_MAX_CLIENTID_LENGTH;
     bool enforceMaxClientIdLength = false;
 
-    if( pConnectInfo == NULL )
+    /* Check that a client identifier was set. */
```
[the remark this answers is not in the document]
gordonwang0: No, it's already done by the caller.  It's not necessary to do this twice.

libraries/standard/mqtt/src/iot_mqtt_validate.c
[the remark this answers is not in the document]
gordonwang0: No, it's already done by the caller.  It's not necessary to do this twice.

tests/iot_config.h
gordonwang0: Only when building for coverage.

libraries/standard/mqtt/test/access/iot_test_access_mqtt_api.c
```diff
@@ -53,7 +53,7 @@ _mqttConnection_t * IotTestMqtt_createMqttConnection( bool awsIotMqttMode,
 
 void IotTestMqtt_mqttOperation_tryDestroy( void * pData )
 {
-    return _mqttOperation_tryDestroy( pData );
+    _mqttOperation_tryDestroy( pData );
```
gkwicker: Not a blocker, but the verb 'try' here isn't reassuring.  How do I know if it did anything if it doesn't return a status?

gordonwang0: It sets a flag in the parameter to let other users know that the subscription couldn't be destroyed.

PR #761, opened 2020-01-30
Address MISRA 21.15, reduce complexity

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
The process of reducing the complexity of _encodeUserName also resolved some MISRA violations, so two identical violations are also fixed as part of this PR

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_serialize.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_serialize.c
@@ -533,11 +533,9 @@ static uint8_t * _encodeString( uint8_t * pDestination,
     *pBuffer = UINT16_LOW_BYTE( sourceLength );
     pBuffer++;
 
-    /* Copy the string into pBuffer. */
-    ( void ) memcpy( pBuffer, source, sourceLength );
-
-    /* Return the pointer to the end of the encoded string. */
-    pBuffer += sourceLength;
+    /* Copy the string into pBuffer and return the pointer to the end of the
+     * encoded string. */
+    pBuffer = _copyBytes( pBuffer, source, sourceLength );
 
     return pBuffer;
 }
@@ -999,8 +997,7 @@ static void _serializePublish( const IotMqttPublishInfo_t * pPublishInfo,
     /* The payload is placed after the packet identifier. */
     if( pPublishInfo->payloadLength > 0U )
     {
-        ( void ) memcpy( pBuffer, pPublishInfo->pPayload, pPublishInfo->payloadLength );
-        pBuffer += pPublishInfo->payloadLength;
+        pBuffer = _copyBytes( pBuffer, pPublishInfo->pPayload, pPublishInfo->payloadLength );
     }
 
     /* Ensure that the difference between the end and beginning of the buffer
--- a/libraries/standard/mqtt/src/iot_mqtt_serialize.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_serialize.c
@@ -535,7 +535,7 @@ static uint8_t * _encodeString( uint8_t * pDestination,
 
     /* Copy the string into pBuffer and return the pointer to the end of the
      * encoded string. */
-    pBuffer = _copyBytes( pBuffer, source, sourceLength );
+    pBuffer = _copyBytes( pBuffer, ( const uint8_t * ) source, sourceLength );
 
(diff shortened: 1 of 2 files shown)


gordonwang0:
_encodeUserName is still showing up in ci_test_quality as complexity 9.

muneebahmed10:
> I don't see how this is an improvement to this code. The memcpy/advance-buffer idiom is incredibly common in C. Hiding it with _copyBytes makes the pattern less recognizable to customers and future maintainers.

This was done to address the MISRA 21.15 violation from coverity, which triggered because one argument to the memcpy was a `uint8_t *` and the other was a `const char *`. Since casting from `uint8_t *` to `char *` or from `const char *` to `const uint8_t *` did not resolve this, having `_copyBytes` avoids having to suppress the warning with more coverity annotations.

muneebahmed10:
Merging this should also close #765

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -591,6 +581,56 @@ static uint8_t * _encodeUserName( uint8_t * pDestination,
 
 /*-----------------------------------------------------------*/
 
+static uint8_t * _encodeUserNameAndMetrics( uint8_t * pDestination,
```
gkwicker: I know this PR is for reorganization, however I want to make sure we take all opportunities to improve robustness of this code.  This function isn't taking a parameter for how much space is available after `pDestination`, yet that [information is available if we go up the call stack a bit](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/iot_mqtt_serialize.c#L1414).  Note that there is already a bounds check on line 595, if you had just carried the available packet size down to this function you could use that and be totally safe, it wouldn't be hard to add.  Lines 610/617 are frightening without it.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
```diff
@@ -973,6 +1013,9 @@ static void _serializePublish( const IotMqttPublishInfo_t * pPublishInfo,
     /* The payload is placed after the packet identifier. */
     if( pPublishInfo->payloadLength > 0U )
     {
+        /* This memcpy intentionally copies bytes from a void * buffer into
```
gkwicker: Similar to previous comment – we have `publishPacketSize` but we're not paying attention to it.  Let's compare against it to make sure we won't overrun after `pBuffer` on line 1019.  This makes the code a little safer and a little less scary for any future reviewers.  That will also allow you to delete 958.

muneebahmed10: I'm not sure checking against `publishPacketSize` is necessary here since the buffer and its size are always checked whenever `_serializePublish` is called. Also, the buffer is written into several times before this line, so checking that `( ( ( size_t ) ( pBuffer - pPacket ) ) + pPublishInfo->payloadLength ) <= publishPacketSize` seems out of place here as it's not checked anywhere else in the function. Refactoring the function to make this check every time seems outside the scope of this PR.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
muneebahmed10: This function and `_encodeUserName` are similar to `_encodeString` and `_encodeRemainingLength` in that the size of the buffer is always checked before they are called, and so don't check it themselves. However, both of those functions have some sort of warning:
```
 @warning This function does not check the size of `pDestination`! Ensure that
 `pDestination` is large enough to hold `sourceLength+2` bytes to avoid a buffer
 overflow.
```
Would adding a similar warning to these functions suffice?

libraries/standard/mqtt/src/iot_mqtt_serialize.c
gkwicker: These are good points, I agree and don't think we should touch this for now.  At some point in the future however, we should think about an abstraction for working with these buffers rather than the direct approach used in this code.

libraries/standard/mqtt/src/iot_mqtt_serialize.c
gkwicker: At a minimum let's go ahead and add that warning to be consistent.

PR #762, opened 2020-01-30
Complete coverage of iot_mqtt_subscription.c

gordonwang0:


libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -354,41 +354,41 @@ static bool _validateSubscription( bool awsIotMqttMode,
     /* Check for a valid QoS and callback function when subscribing. */
     if( operation == IOT_MQTT_SUBSCRIBE )
     {
-        status = _validateQos( pSubscription->qos );
+        if( pSubscription->callback.function == NULL )
```
muneebahmed10: Due to this change, line 364 could be changed to an `else`

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -161,7 +161,7 @@ static bool _matchEndWildcards( const char * pTopicFilter,
     if( endChar == true )
     {
         /* Determine if the topic filter ends with the '#' wildcard. */
-        status = ( pTopicFilter[ filterIndex + 1U ] == '/' ) && ( pTopicFilter[ filterIndex + 2U ] == '#' );
```
muneebahmed10: Is `pTopicFilter[filterIndex+1] == '/'` not checked due to being validated elsewhere?

libraries/standard/mqtt/src/iot_mqtt_validate.c
gordonwang0: changed

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: Correct

PR #763, opened 2020-01-30
Fix MISRA 20.9 by providing default macro definitions

qiutongs:
*Issue #, if available:* https://sim.amazon.com/issues/TS-10729

*Description of changes:*
Fix MISRA 20.9 by providing default macro definitions

### DIFF
--- a/demos/iot_config.h
+++ b/demos/iot_config.h
@@ -24,6 +24,11 @@
 #ifndef IOT_CONFIG_H_
 #define IOT_CONFIG_H_
 
+/* Disable test builds by default. */
+#ifndef IOT_BUILD_TESTS
+    #define IOT_BUILD_TESTS    ( 0 )
+#endif
+
 /* Server endpoints used for the demos. May be overridden with command line
  * options at runtime. */
 #define IOT_DEMO_SECURED_CONNECTION    ( true ) /* Command line: -s (secured) or -u (unsecured) */
@@ -45,8 +50,13 @@
  * #define IOT_DEMO_IDENTIFIER         "" */
 
 /* MQTT demo configuration. The demo publishes bursts of messages. */
-#define IOT_DEMO_MQTT_PUBLISH_BURST_COUNT       ( 10 ) /* Number of message bursts. */
-#define IOT_DEMO_MQTT_PUBLISH_BURST_SIZE        ( 10 ) /* Number of messages published in each burst. */
+#define IOT_DEMO_MQTT_PUBLISH_BURST_COUNT    ( 10 )    /* Number of message bursts. */
+#define IOT_DEMO_MQTT_PUBLISH_BURST_SIZE     ( 10 )    /* Number of messages published in each burst. */
+
+/* MQTT library configuration. */
+#ifndef IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES
+    #define IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES    ( 0 )
+#endif
 
 /* Shadow demo configuration. The demo publishes periodic Shadow updates and responds
  * to changing Shadows. */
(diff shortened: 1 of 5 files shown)


demos/iot_config.h
```diff
@@ -24,6 +24,11 @@
 #ifndef IOT_CONFIG_H_
 #define IOT_CONFIG_H_
 
+/* Disable test builds by default. */
+#ifndef IOT_BUILD_TESTS
```
gordonwang0: This is a demo config file intended to be used by applications.  There should be no references to test-specific constants in this file.

qiutongs: ok, I kind of agree. Then do you suggest I put it to iot_mqtt_internal.h?
We reference it in source file here: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/iot_mqtt_subscription.c#L700

dan4thewin: Can you define via compiler argument in the test build script?

qiutongs: That is already there https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/scripts/ci_test_build.sh

But when coverity analyzes the code, it doesn't see  IOT_BUILD_TESTS is defined, which violates MISRA 20.9

libraries/standard/mqtt/src/private/iot_mqtt_internal.h
```diff
@@ -42,6 +42,11 @@
 /* Task pool include. */
 #include "iot_taskpool.h"
 
+/* Disable test builds by default. */
+#ifndef IOT_BUILD_TESTS
+    #define IOT_BUILD_TESTS    ( 0 )
+#endif
```
muneebahmed10: This should go somewhere around line 226, in the DOXYGEN_IGNORE block

demos/iot_config.h
```diff
@@ -45,8 +45,13 @@
...
 /* MQTT demo configuration. The demo publishes bursts of messages. */
-#define IOT_DEMO_MQTT_PUBLISH_BURST_COUNT       ( 10 ) /* Number of message bursts. */
-#define IOT_DEMO_MQTT_PUBLISH_BURST_SIZE        ( 10 ) /* Number of messages published in each burst. */
+#define IOT_DEMO_MQTT_PUBLISH_BURST_COUNT    ( 10 )    /* Number of message bursts. */
+#define IOT_DEMO_MQTT_PUBLISH_BURST_SIZE     ( 10 )    /* Number of messages published in each burst. */
+
+/* MQTT library configuration. */
+#ifndef IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES
+    #define IOT_MQTT_ENABLE_SERIALIZER_OVERRIDES    ( 0 )
+#endif
```
muneebahmed10: `iot_mqtt_internal.h` already has this `#ifndef`. Does putting `IOT_BUILD_TESTS` in that file succeed in eliminating the MISRA 20.9 warning?

qiutongs: Yes, iot_mqtt_internal.h already has this. I added it here for iot_mqtt_types.h. I know this is not perfect solution.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1983,6 +1983,8 @@ const char * IotMqtt_OperationType( IotMqttOperationType_t operation )
 /*-----------------------------------------------------------*/
 
 /* Provide access to internal functions and variables if testing. */
+/* IOT_BUILD_TESTS is defined outside the codebase, e.g. passed in by CMake command. */
+/* coverity[misra_c_2012_rule_20_9_violation] */
```
muneebahmed10: I believe this also needs a `/* coverity[caretline] */` to suppress this specific violation.

qiutongs: You are right. Fixed it.

PR #768, opened 2020-01-31
Re-enable checks for resource allocation failure

gordonwang0:


libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -469,8 +469,8 @@ static bool _createKeepAliveOperation( const IotMqttNetworkInfo_t * pNetworkInfo
     pMqttConnection->pingreq.u.operation.periodic.ping.nextPeriodMs = ( uint32_t ) keepAliveSeconds * 1000U;
 
     /* Generate a PINGREQ packet. */
-    serializeStatus = _getMqttPingreqSerializer( pMqttConnection->pSerializer )( &( pMqttConnection->pingreq.u.operation.pMqttPacket ),
-                                                                                 &( pMqttConnection->pingreq.u.operation.packetSize ) );
+    serializeStatus = _getMqttPingreqSerializer( pNetworkInfo->pMqttSerializer )( &( pMqttConnection->pingreq.u.operation.pMqttPacket ),
```
muneebahmed10: Wondering why some of these in this file use `pNetworkInfo->pMqttSerializer` while others use `pMqttConnection->pSerializer`

gordonwang0: This was a bug.  This is during the initialization of the MQTT connection, before the MQTT connection has been set up.  Therefore, the serializer needs to be read from the parameter, pNetworkInfo.

libraries/standard/mqtt/include/types/iot_mqtt_types.h
```diff
@@ -797,7 +797,7 @@ typedef IotMqttError_t ( * IotMqttSerializeSubscribe_t )( const IotMqttSubscript
  * @param[out] uint8_t** Where the DISCONNECT packet is written.
  * @param[out] size_t* Size of the DISCONNECT packet.
  */
-typedef IotMqttError_t ( * IotMqttSerializeDisconnect_t )( uint8_t ** ppDisconnectPacket,
+typedef IotMqttError_t ( * IotMqttSerializeDisconnect_t )( uint8_t ** pDisconnectPacket,
```
dcgaws: We're using "p" Hungarian, but not "pp"?

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
```diff
@@ -999,6 +992,9 @@ TEST( MQTT_Unit_API, ConnectMallocFail )
     int32_t i = 0;
     IotMqttError_t status = IOT_MQTT_STATUS_PENDING;
     IotMqttConnectInfo_t connectInfo = IOT_MQTT_CONNECT_INFO_INITIALIZER;
+    IotMqttSerializer_t serializer = IOT_MQTT_SERIALIZER_INITIALIZER;
+
+    _pAllocatedPingreq = NULL;
```
dcgaws: Why does this need to be re-initialized?

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
```diff
@@ -165,6 +165,11 @@ static IotMqttNetworkInfo_t _networkInfo = IOT_MQTT_NETWORK_INFO_INITIALIZER;
  */
 static IotNetworkInterface_t _networkInterface = { 0 };
 
+/**
+ * @brief A packet allocated by _serializePingreq.
+ */
+static uint8_t * _pAllocatedPingreq = NULL;
```
dcgaws: Why does this need to be global?

libraries/standard/mqtt/include/types/iot_mqtt_types.h
gordonwang0: Yes.  This change is just for consistency.

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
gordonwang0: It doesn't

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
gordonwang0: It's shared between _serializePingreq and the test function.  It's set by _serializePingreq so the test function knows what to free to clean up.

PR #772, opened 2020-02-04
Feat: Refactor MQTT Light Weight API code

abhidixi11:
**I'm fixing build errors  and will update PR shortly ** 
*Issue #, if available:*

*Description of changes:*
Light Weight API code has been factored out and put into separate files.  
After this PR, there will not be any common code between managed and Light-weight API. A separate PR will be created after this PR is merged to make managed API use common code.

Following new files have been added as part of this PR:

**iot_mqtt_helper.h
iot_mqtt_helper.c**

Helper files contain static functions from iot_mqtt_serializer.c. These function can be used by managed MQTT as well as Light Weight MQTT.  This PR does not have changes to update managed code to use these functions, but it will be done the next PR if common code is desirable.

**iot_mqtt_protocol.h**

This file contains protocol specific defines which were scattered across bunch of header files. Currently only used by LW API but can be used by managed API.


libraries/standard/mqtt/include/types/iot_mqtt_common_types.h
```diff
@@ -0,0 +1,557 @@
...
+
+#ifndef IOT_MQTT_MANAGED_API
+
+/**
+ * @brief The type used to represent network connections.
+ *
+ * For the light weight MQTT API, it is application's responsibility to define and 
+ * interpret IotNetworkConnection_t, therefore it is defined as void * here.
+ */
+    typedef void * IotNetworkConnection_t;
```
gordonwang0: Include `iot_config.h` at the top of this file, and you don't need to do this.

libraries/standard/mqtt/include/types/iot_mqtt_common_types.h
```diff
@@ -0,0 +1,557 @@
+/*
+ * IoT MQTT V2.1.0
```
gordonwang0: I do not think we should split the types file.  It seems like types in this header will be different based on whether the lightweight or full API is used, but retain the same name across both.  It is confusing to have the same name refer to different types.

I would prefer all types in 1 file.  The lightweight API should just ignore the full API's types.

abhidixi11: The common types from iot_mqtt_types.h will be removed and iot_mqtt_types.h will include iot_mqtt_common_types.h.  The reason I want to separate those two is to avoid including taskpool and other data types not needed by LW API. Also I'm not creating separate data types for LW API to avoid code duplication.

libraries/standard/mqtt/include/types/iot_mqtt_common_types.h
abhidixi11: Talked to Gordon, since iot_config.h is going to be platform and demo specific, we should able to add platform specific network connection there.

abhidixi11:
> The file names aren't clear to me.
> 
>     * `iot_mqtt_helper.h`: These appear to be the public serializer functions.  If so, I suggest renaming this to `iot_mqtt_serialize.h`
> 
These are not public functions, there are internal functions. 


>     * `iot_mqtt_protocol.h`: Defines relating to the MQTT protocol.  Why is a separate header needed?  Could these just be placed `iot_mqtt_helper.h` or the .c file that uses them?
> 

Could be, but these definitions will be used by application as well for light weight demo. I was planning of using protocol.h in managed API as well, removing protocol related definitions scattered across various header files.

>     * The functions formerly in `iot_mqtt_serialize.c` are now split between 3 files: helper.c, lightweight.c, and serialize.c.  We should reduce the number of files here.

lightweight.c has all Public API functions for LW usage. helper.c contains common functions between managed AP and LW API, serialize.c contains function used by managed API only. helper.c is basically common functions for serializing. There is duplication in helper.c and serialize.c, which will be removed in next PR. We can rename helper.c to serialize_common.c and serialize.h will be renamed to  lightweight.h 

> 
>     * Please make sure all functions follow the current naming convention.

OK I'm sticking the current coding convention in next PR.

libraries/standard/mqtt/include/iot_mqtt_helper.h
```diff
@@ -22,7 +22,7 @@
 
 /**
  * @file iot_mqtt_helper.c
- * @brief Implements helper functions for the MQTT library.
+ * @brief Implements internal helper functions for the MQTT library.
```
gordonwang0: Maybe rename this iot_mqtt_serialize_internal.h, as all the functions are for the serializer?

Also should be moved to src/private. include/ is for public headers only.

libraries/standard/mqtt/include/iot_mqtt_lightweight.h
```diff
@@ -21,14 +21,14 @@
...
- * @file iot_mqtt_serialize.h
+ * @file iot_mqtt_lightweight.h
  * @brief User-facing functions for serializing MQTT 3.1.1 packets. This header should
  * be included for building a single threaded light-weight MQTT client bypassing
  * stateful CSDK MQTT library.
  */
 
-#ifndef _IOT_MQTT_SERIALIZE_H_
-#define _IOT_MQTT_SERIALIZE_H_
+#ifndef _IOT_MQTT_LIGHTWEIGHT_H_
```
gordonwang0: Include guards should not start with _

libraries/standard/mqtt/include/iot_mqtt_helper.h
abhidixi11: I have moved it to private folder. I'll rename it in the next PR since by then we it will be clear which code resides where.

PR #777, opened 2020-02-05
Address MISRA for LW API, sync changes

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Addresses new MISRA violations as a result of #772, syncs changes from #761 to copied files, and uncrustify.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_helper.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_helper.c
@@ -161,6 +161,12 @@ static uint8_t * _encodeUserNameAndMetrics( uint8_t * pDestination,
 
             *pEncodedUserName = true;
         }
+        else
+        {
+            IotLogWarn( "Username length of %lu is larger than maximum %lu.",
+                        ( pConnectInfo->userNameLength + AWS_IOT_METRICS_USERNAME_LENGTH ),
+                        UINT16_MAX );
+        }
     #else /* if AWS_IOT_MQTT_ENABLE_METRICS == 1 */
         /* Avoid unused variable warnings when AWS_IOT_MQTT_ENABLE_METRICS is set to 0. */
         ( void ) pBuffer;
@@ -176,10 +182,6 @@ uint8_t * _IotMqtt_EncodeUserName( uint8_t * pDestination,
 {
     bool encodedUserName = false;
     uint8_t * pBuffer = pDestination;
-    const char * pMetricsUserName = NULL;
-
-    /* Avoid unused variable warning when AWS_IOT_MQTT_ENABLE_METRICS is set to 0 */
-    ( void ) pMetricsUserName;
 
     /* If metrics are enabled, write the metrics username into the CONNECT packet.
      * Otherwise, write the username and password only when not connecting to the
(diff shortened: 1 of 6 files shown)


libraries/standard/mqtt/src/iot_mqtt_helper.c
```diff
@@ -25,6 +25,9 @@
  * @brief Implements helper functions for the MQTT library.
  */
 
+/* The config header is always included first. */
+#include "iot_config.h"
```
abhidixi11: Why do we need iot_config.h in this file ( since you have added AWS_IOT_MQTT_ENABLE_METRICS below )?

muneebahmed10: I included it here since logging and assert macros were used in this file, which are configured in iot_config.h. Including it has no impact on MISRA though, so I will remove it

libraries/standard/mqtt/src/private/iot_mqtt_helper.h
```diff
@@ -277,7 +277,32 @@ IotMqttError_t _IotMqtt_ProcessIncomingPublishFlags( uint8_t publishFlags,
  *
  * @return Pointer to the end of the encoded string, which will be identical to
  * `pDestination` if nothing was encoded.
+ *
+ * @warning This function does not check the size of `pDestination`! To avoid a buffer
```
dcgaws: The caller appears to have this value, though. Why don't we add a parameter to the below to take it in, then add a check for it?

muneebahmed10: This warning is a sync from the changes of #761 to functions that were duplicated as part of #772. This was the [reasoning](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/761/#discussion_r374288528) for not adding a similar check right now, and to add the [warning](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/761/#discussion_r374293230) instead. The buffer size is checked before `_IotMqtt_SerializeConnectCommon` is called, which calls this function after others that also write into the buffer. Checking the buffer a second time only for this function seems out of place

gkwicker: Let's defer this for a future revision since it's outside of the scope of this PR and a larger topic.

libraries/standard/mqtt/src/iot_mqtt_helper.c
```diff
@@ -151,17 +147,68 @@ uint8_t * _IotMqtt_EncodeUserName( uint8_t * pDestination,
...
+
+            /* Write the metrics portion of the username.
+             * As the types of char and uint8_t are of the same size, this memcpy
+             * is acceptable. */
+            /* coverity[misra_c_2012_rule_21_15_violation] */
+            ( void ) memcpy( pBuffer, pMetricsUserName, AWS_IOT_METRICS_USERNAME_LENGTH );
+            pBuffer += AWS_IOT_METRICS_USERNAME_LENGTH;
+
+            *pEncodedUserName = true;
+        }
```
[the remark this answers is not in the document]
muneebahmed10: Added a warning log statement.

libraries/standard/mqtt/src/iot_mqtt_helper.c
```diff
@@ -173,7 +220,10 @@ uint8_t * _IotMqtt_EncodeString( uint8_t * pDestination,
     *pBuffer = UINT16_LOW_BYTE( sourceLength );
     pBuffer++;
 
-    /* Copy the string into pBuffer. */
+    /* Copy the string into pBuffer.
+     * As the types of char and uint8_t are of the same size, this memcpy
```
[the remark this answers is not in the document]
muneebahmed10: Reworded. However, some of our code does still assume 8 bit bytes. Should a header check be added as part of this PR?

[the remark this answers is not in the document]
gkwicker: I think we are okay to defer this since we don't support any targets which use a differently sized `char`.

PR #783, opened 2020-02-08
Fix platform test failure

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Fixes the reentrancy test failures introduced in #782 due to not initializing the network


libraries/standard/mqtt/test/unit/iot_tests_mqtt_platform.c
```diff
@@ -1179,6 +1178,8 @@ TEST( MQTT_Unit_Platform, SubscribeCompleteReentrancy )
     IotMqttSubscription_t subscription = IOT_MQTT_SUBSCRIPTION_INITIALIZER;
     IotMqttCallbackInfo_t callbackInfo = IOT_MQTT_CALLBACK_INFO_INITIALIZER;
 
+    TEST_ASSERT_EQUAL( IOT_NETWORK_SUCCESS, IotTestNetwork_Init() );
```
cobusve: By the Unity design test setup (like init of the network) should happen in the test setup phase and test teardown should happen in the TEST_TEAR_DOWN phase. It looks like this PR is moving away from the Unity design. Can you please explain the reasoning for this?

libraries/standard/mqtt/test/unit/iot_tests_mqtt_platform.c
```diff
@@ -639,7 +639,6 @@ TEST_SETUP( MQTT_Unit_Platform )
 TEST_TEAR_DOWN( MQTT_Unit_Platform )
 {
     IotMqtt_Cleanup();
-    IotTestNetwork_Cleanup();
```
cobusve: Please see my other comment on Unity framework architecture

libraries/standard/mqtt/test/unit/iot_tests_mqtt_platform.c
muneebahmed10: The init was added to the test setup in this PR as part of commit d307888; however, that caused another test, PublishRetryScheduleFailure (which doesn't need the network init), to report a leak on every run. The two tests that need the network were moved to the platform group as part of #782. Since they are the only tests in this group that need the network, I added the inits to only those two.

abhidixi11: Yes that's correct. These tests were moved from another file where network init and cleanup were called in correct places.  For platform file  when network init was added in TEST_SETUP  it caused memory leak error for other unrelated test. We will investigate it , but this was a quick fix, so others are not blocked.   (Also these tests are needed in platform so we can disable them on FreeRTOSOrg release.

cobusve: Seem like the best way to handle this is to create 2 test groups, one that requires the network and one that does not ... That way you can do the setup and teardown in the right place.

muneebahmed10: Moved to new test group

muneebahmed10:
> What was the issue with the tests? Why does moving them to a new test group fix them?

The tests that were failing did so because `IotTestNetwork_Init()` wasn't being called in test setup. However, doing so for all tests caused PublishRetryScheduleFailure to consistently report a leak when using openssl. So, a new test group was created that initialized the network for only the tests that needed it

libraries/standard/mqtt/test/unit/iot_tests_mqtt_platform.c
gkwicker: It looks like this is being done correctly now, the network init/cleanup is only being done in the `MQTT_System_Platform` setup/teardown & unit tests are using the `MQTT_Unit_Platform` group.

Ideally unit tests should be in a different file, invoked using a different make target, etc. than functional/integration tests but that's beyond the scope of this review.

PR #784, opened 2020-02-08
chore: Changed managed API to use common code.

abhidixi11:
Use helper functions in iot_mqtt_serialize.c.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_helper.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_helper.c
@@ -83,6 +83,94 @@
 
 /*-----------------------------------------------------------*/
 
+/* Username for metrics with AWS IoT. */
+#if ( AWS_IOT_MQTT_ENABLE_METRICS == 1 ) || ( DOXYGEN == 1 )
+    #ifndef AWS_IOT_METRICS_USERNAME
+
+/**
+ * @brief Specify C SDK and version.
+ */
+        #define AWS_IOT_METRICS_USERNAME           "?SDK=C&Version=4.0.0"
+
+/**
+ * @brief The length of #AWS_IOT_METRICS_USERNAME.
+ */
+        #define AWS_IOT_METRICS_USERNAME_LENGTH    ( ( uint16_t ) sizeof( AWS_IOT_METRICS_USERNAME ) - 1U )
+    #endif /* ifndef AWS_IOT_METRICS_USERNAME */
+#endif /* if AWS_IOT_MQTT_ENABLE_METRICS == 1 || DOXYGEN == 1 */
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Encode both connection and metrics username into a buffer,
+ * if they will fit.
+ *
+ * @param[in] pDestination Buffer to write username into.
+ * @param[in] pConnectInfo User-provided CONNECT information.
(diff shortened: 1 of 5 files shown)


libraries/standard/mqtt/src/iot_mqtt_helper.c
```diff
@@ -80,8 +83,96 @@
...
+
+/*-----------------------------------------------------------*/
+
+static uint8_t * _encodeUserNameAndMetrics( uint8_t * pDestination,
+                                            const IotMqttConnectInfo_t * pConnectInfo,
+                                            bool * pEncodedUserName )
+{
+    uint8_t * pBuffer = pDestination;
+
+    #if AWS_IOT_MQTT_ENABLE_METRICS == 1
```
nateglims: Do we test with both metrics enabled and disabled?

libraries/standard/mqtt/src/private/iot_mqtt_helper.h
```diff
@@ -71,7 +71,17 @@
  * @param[in] x The unsigned int to check.
  * @param[in] position Which bit to check.
  */
-#define UINT8_CHECK_BIT( x, position )    ( ( ( x ) & ( 0x01U << ( position ) ) ) == ( 0x01U << ( position ) ) )
+#define UINT8_CHECK_BIT( x, position )    ( ( ( x ) &( 0x01U << ( position ) ) ) == ( 0x01U << ( position ) ) )
```
nateglims: I think this formatting is slightly off.

libraries/standard/mqtt/src/private/iot_mqtt_helper.h
```diff
@@ -278,6 +288,6 @@ IotMqttError_t _IotMqtt_ProcessIncomingPublishFlags( uint8_t publishFlags,
  * @return Pointer to the end of the encoded string, which will be identical to
  * `pDestination` if nothing was encoded.
  */
- uint8_t * _IotMqtt_EncodeUserName( uint8_t * pDestination,
-                                  const IotMqttConnectInfo_t * pConnectInfo );
-#endif
\ No newline at end of file
+uint8_t * _IotMqtt_EncodeUserName( uint8_t * pDestination,
+                                   const IotMqttConnectInfo_t * pConnectInfo );
+#endif /* ifndef IOT_MQTT_SERIALIZE_INTERNAL_H_ */
```
nateglims: It looks like it was this way before the change, but is the header guard intentionally different from the file name?

libraries/standard/mqtt/src/iot_mqtt_helper.c
abhidixi11: It depends on the configuration. I think we test with enabled , but my local setup it was disabled. So didn't this problem.

libraries/standard/mqtt/src/private/iot_mqtt_helper.h
abhidixi11: Good catch. It is cut and paste error. Will fix it.

libraries/standard/mqtt/src/private/iot_mqtt_helper.h
abhidixi11: I'll run uncrustify again.

PR #787, opened 2020-02-11
Complete coverage of iot_mqtt_serialize.c

gordonwang0:


libraries/standard/mqtt/test/unit/iot_tests_mqtt_receive.c
```diff
@@ -1266,6 +1266,33 @@ TEST( MQTT_Unit_Receive, PublishResourceFailure )
...
     _publishMallocFail( IOT_MQTT_QOS_1 );
 
+    /* Test PUBACK generation when malloc fails. */
+    {
+        IotMqttError_t status = IOT_MQTT_STATUS_PENDING;
+        int32_t i = 0;
+        uint8_t * pPubackPacket = NULL;
+        size_t pubackSize = 0;
+
+        for( i = 0; ; i++ )
```
gkwicker: Why are we using a loop here?  Why is there no test on line 1276?  Also, why pass `-1` on line 1293?  I think this can be written in a much more straightforward fashion with no loop, e.g:

1. Test Success Case
1. Test Failure Case

PR #788, opened 2020-02-11
Adds a proof harness for the IotMqtt_Init function

feliperodri:
*Description of changes:*

This PR depends on https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/780!

- Adds `.gitignore` file on cbmc directory;
- Updates CBMC flags in the `Makefile.common`;
- Adds a new proof harness for the `IotMqtt_Init` function.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #789, opened 2020-02-11
Adds a proof harness for the IotMqtt_Cleanup function

feliperodri:
*Description of changes:*

This PR depends on https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/788!

- Adds a new proof harness for the `IotMqtt_Cleanup` function.

### DIFF
--- a/cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
+++ b/cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
@@ -1,0 +1,14 @@
+#include "iot_config.h"
+#include "private/iot_mqtt_internal.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+void harness()
+{
+  IotMqtt_Cleanup();
+}
+
+
+
+
--- a/cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
+++ b/cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
@@ -6,9 +6,6 @@
 
 void harness()
 {
+  IotMqtt_Init();
   IotMqtt_Cleanup();
 }
-
-
-
-
--- a/cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
+++ b/cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
@@ -6,6 +6,18 @@
 
 void harness()
 {
+  /*
+   * The code specification says that IotMqtt_Cleanup must be
+   * called after IotMqtt_Init, since it is supposed to undo
+   * anything that IotMqtt_Init does. Both IotMqtt_Cleanup and
+   * IotMqtt_Init basically update the status of the same global
+   * variable _initCalled. This variable is declared as volatile,
+   * so CBMC will always treat it as non-deterministic.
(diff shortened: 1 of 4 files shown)


cbmc/proofs/IotMqtt_Cleanup/Makefile
```diff
@@ -0,0 +1,10 @@
+ENTRY=IotMqtt_Cleanup
+
+ABSTRACTIONS += --remove-function-body IotLog_Generic
```
markrtuttle: You had a nice comment you added to the _init makefile in response to DSN, maybe we should use that everywhere.

cbmc/proofs/Makefile.common
```diff
@@ -62,11 +64,24 @@ $(ENTRY).goto: $(ENTRY)2.goto
 # Run cbmc and build html report
 
 CBMCFLAGS += \
+	--object-bits $(CBMC_OBJECT_BITS) \
 	$(UNWINDING) \
 	--bounds-check \
 	--pointer-check \
 	--unwinding-assertions \
```
markrtuttle: Your init PR merged these flags into a single block.

cbmc/proofs/IotMqtt_Cleanup/Makefile
feliperodri: Done!

cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
```diff
@@ -0,0 +1,14 @@
+#include "iot_config.h"
+#include "private/iot_mqtt_internal.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+void harness()
+{
+  IotMqtt_Cleanup();
```
gordonwang0: You should call IotMqtt_Init() before IotMqtt_Cleanup.  Cleanup is meant to undo what Init does, and may act on uninitialized resources if Init was not called beforehand.  (Although neither Init nor Cleanup do much right now, we may change that in the future).

cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
```diff
@@ -0,0 +1,14 @@
...
+}
+
+
+
+
```
gordonwang0: Remove unnecessary newlines at the end of this file.

cbmc/proofs/IotMqtt_Cleanup/IotMqtt_Cleanup_harness.c
feliperodri: Both `IotMqtt_Cleanup` and `IotMqtt_Init` basically update the status of the same global variable `_initCalled`. However, this variable is declared as volatile, so CBMC will always treat it as non-deterministic.
```c
static volatile uint32_t _initCalled = MQTT_LIBRARY_UNINITIALIZED;
```
Therefore, even if we called `IotMqtt_Init` first, it wouldn't make a significant impact in this proof harness. In any case, since the developers may change the implementation of both functions in the future, I added it to the proof. Thanks!

markrtuttle: I'd add a comment about this to the harness, then, because in your style of writing proof harnesses, this is certainly unusual.

feliperodri: Done @markrtuttle

PR #793, opened 2020-02-11
Set proper state in disconnect handler

qiutongs:
*Issue #, if available:*

*Description of changes:*
Set proper state in disconnect handler

Even if aws_iot_mqtt_attempt_reconnect() is successful in the disconnect handler, the client state is set to CLIENT_STATE_DISCONNECTED_ERROR and this result in the next aws_iot_mqtt_yield() call returns NETWORK_DISCONNECTED_ERROR.

### DIFF
--- a/samples/linux/subscribe_publish_sample/subscribe_publish_sample.c
+++ b/samples/linux/subscribe_publish_sample/subscribe_publish_sample.c
@@ -78,16 +78,13 @@ static void disconnectCallbackHandler(AWS_IoT_Client *pClient, void *data) {
 
 	if(aws_iot_is_autoreconnect_enabled(pClient)) {
 		IOT_INFO("Auto Reconnect is enabled, Reconnecting attempt will start now");
-		pClient->clientStatus.clientState = CLIENT_STATE_PENDING_RECONNECT;
 	} else {
 		IOT_WARN("Auto Reconnect not enabled. Starting manual reconnect...");
 		rc = aws_iot_mqtt_attempt_reconnect(pClient);
 		if(NETWORK_RECONNECTED == rc) {
 			IOT_WARN("Manual Reconnect Successful");
-			pClient->clientStatus.clientState = CLIENT_STATE_CONNECTED_IDLE;
 		} else {
 			IOT_WARN("Manual Reconnect Failed - %d", rc);
-			pClient->clientStatus.clientState = CLIENT_STATE_DISCONNECTED_ERROR;
 		}
 	}
 }
--- a/samples/linux/subscribe_publish_sample/subscribe_publish_sample.c
+++ b/samples/linux/subscribe_publish_sample/subscribe_publish_sample.c
@@ -75,18 +75,6 @@ static void disconnectCallbackHandler(AWS_IoT_Client *pClient, void *data) {
 	}
 
 	IOT_UNUSED(data);
-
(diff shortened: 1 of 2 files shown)


qiutongs:
> I think you should change how the state is set instead of removing the reconnect attempt from the demo.
> 
> Set CLIENT_STATE_DISCONNECTED_ERROR before invoking the disconnect handler. Then, the disconnect handler may change the state if they want.

I misunderstood what we want. Now I see. Let me make the change.

samples/linux/subscribe_publish_sample/subscribe_publish_sample.c
```diff
@@ -78,13 +78,16 @@ static void disconnectCallbackHandler(AWS_IoT_Client *pClient, void *data) {
 
 	if(aws_iot_is_autoreconnect_enabled(pClient)) {
 		IOT_INFO("Auto Reconnect is enabled, Reconnecting attempt will start now");
+		pClient->clientStatus.clientState = CLIENT_STATE_PENDING_RECONNECT;
```
gordonwang0: You should not make any changes to the state in this file.  State changes will be handled by API functions.

src/aws_iot_mqtt_client_yield.c
```diff
@@ -61,12 +61,12 @@ static IoT_Error_t _aws_iot_mqtt_handle_disconnect(AWS_IoT_Client *pClient) {
...
 
+	pClient->clientStatus.clientState = CLIENT_STATE_DISCONNECTED_ERROR;
+
 	if(NULL != pClient->clientData.disconnectHandler) {
 		pClient->clientData.disconnectHandler(pClient, pClient->clientData.disconnectHandlerData);
 	}
 
-	/* Reset to 0 since this was not a manual disconnect */
-	pClient->clientStatus.clientState = CLIENT_STATE_DISCONNECTED_ERROR;
 	FUNC_EXIT_RC(NETWORK_DISCONNECTED_ERROR);
```
gordonwang0: Don't return NETWORK_DISCONNECTED_ERROR if the client reconnected.

samples/linux/subscribe_publish_sample/subscribe_publish_sample.c
qiutongs: ok, I will remove those state changes code.

src/aws_iot_mqtt_client_yield.c
```diff
@@ -61,12 +61,16 @@ static IoT_Error_t _aws_iot_mqtt_handle_disconnect(AWS_IoT_Client *pClient) {
...
+	pClient->clientStatus.clientState = CLIENT_STATE_DISCONNECTED_ERROR;
+
 	if(NULL != pClient->clientData.disconnectHandler) {
 		pClient->clientData.disconnectHandler(pClient, pClient->clientData.disconnectHandlerData);
 	}
 
-	/* Reset to 0 since this was not a manual disconnect */
-	pClient->clientStatus.clientState = CLIENT_STATE_DISCONNECTED_ERROR;
+	if (pClient->clientStatus.clientState == CLIENT_STATE_CONNECTED_IDLE) {
+		FUNC_EXIT_RC(NETWORK_RECONNECTED);
```
gordonwang0: Should return SUCCESS here.

qiutongs: ok

PR #794, opened 2020-02-12
Adds a proof harness for the IotMqtt_IsSubscribed function

feliperodri:
*Description of changes:*

This PR depends on https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/788!!!

- Adds a new proof harness for the `IotMqtt_IsSubscribed` function.
- Add a fix to avoid undefined behavior when calling `strncmp` function.

### DIFF
--- a/cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
+++ b/cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
@@ -26,5 +26,5 @@ void harness()
   IotMqtt_IsSubscribed( mqttConnection,
 			TopicFilter,
 			topicFilterLength,
-                        nondet_boo() ? NULL : &result );
+                        nondet_bool() ? NULL : &result );
 }
--- a/cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
+++ b/cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
@@ -23,7 +23,7 @@ void harness()
 
   IotMqttSubscription_t result;
 
-  IotMqtt_IsSubscribed( nondet_bool() ? NULL : mqttConnection,
+  IotMqtt_IsSubscribed( mqttConnection,
 			TopicFilter,
 			topicFilterLength,
                         nondet_boo() ? NULL : &result );
--- a/cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
+++ b/cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
@@ -1,0 +1,40 @@
+#include "iot_config.h"
+#include "private/iot_mqtt_internal.h"
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "mqtt_state.h"
+
+/*
+
+bool IotMqtt_IsSubscribed( IotMqttConnection_t mqttConnection,
+                           const char * pTopicFilter,
+                           uint16_t topicFilterLength,
+                           IotMqttSubscription_t * pCurrentSubscription );
+
+*/
+
+void harness()
+{
+  IotMqttConnection_t mqttConnection = allocate_IotMqttConnection(NULL);
+  __CPROVER_assume(mqttConnection);
+  ensure_IotMqttConnection_has_lists(mqttConnection);
+  __CPROVER_assume(valid_IotMqttConnection(mqttConnection));
+
+  // Move to allocation and valid connection?
+  allocate_IotMqttSubscriptionList(&(mqttConnection->subscriptionList), 1);
+  __CPROVER_assume(valid_IotMqttSubscriptionList(&(mqttConnection->subscriptionList), 1));
+
+  uint16_t topicFilterLength;
+  __CPROVER_assume(topicFilterLength < TOPIC_LENGTH_MAX);
+
+  char* TopicFilter = malloc_can_fail(topicFilterLength);
+
+  IotMqttSubscription_t result;
+
+  IotMqtt_IsSubscribed( nondet_bool() ? NULL : mqttConnection,
+			TopicFilter,
+			topicFilterLength,
+                        nondet_boo() ? NULL : &result );
+}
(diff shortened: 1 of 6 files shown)


libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -655,6 +660,12 @@ bool IotMqtt_IsSubscribed( IotMqttConnection_t mqttConnection,
                            IotMqttSubscription_t * const pCurrentSubscription )
 {
     bool status = false;
+    
+    if( pTopicFilter == NULL )
+    {
+    	return status;
+    }
```
feliperodri: If we don't check whether `pTopicFilter` is NULL or not, we might call `_topicMatch` with `topicMatchParams.pTopicName`. Then, the `pTopicName` will be passed as an input for `strncmp` function. @gordonwang0

libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -299,14 +299,19 @@ static bool _topicMatch( const IotLink_t * const pSubscriptionLink,
...
 
+    const _topicMatchParams_t * pParam = ( _topicMatchParams_t * ) pMatch;
+ 
     /* Extract the relevant strings and lengths from parameters. */
     const char * pTopicName = pParam->pTopicName;
     const char * pTopicFilter = pSubscription->pTopicFilter;
+    if ( pTopicName == NULL || pTopicFilter == NULL )
+    {
+        return status;
+    }
```
feliperodri: Also check here before we pass them to `strncmp`. @gordonwang0

cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
```diff
@@ -0,0 +1,30 @@
...
+  __CPROVER_assume(topicFilterLength < TOPIC_LENGTH_MAX);
+
+  char* TopicFilter = malloc_can_fail(topicFilterLength);
+
+  IotMqttSubscription_t result;
+
+  IotMqtt_IsSubscribed( mqttConnection,
+			TopicFilter,
+			topicFilterLength,
+                        nondet_boo() ? NULL : &result );
```
danielsn: `bool`

libraries/standard/mqtt/src/iot_mqtt_subscription.c
gordonwang0: The reason there's no NULL check here is because this is a private function (static) that expects its caller to perform parameter checks.  Can you move this check (and the one below) to the IsSubscribed function?

feliperodri: I remove this code from `_topicMatch`. The check bellow is on the `isSubscribed` function and it'll be enough if you assume the caller will always perform the parameter check.

cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
```diff
@@ -0,0 +1,30 @@
...
+  ensure_IotMqttConnection_has_lists(mqttConnection);
+  __CPROVER_assume(valid_IotMqttConnection(mqttConnection));
+
+  allocate_IotMqttSubscriptionList(&(mqttConnection->subscriptionList), 1);
+  __CPROVER_assume(valid_IotMqttSubscriptionList(&(mqttConnection->subscriptionList), 1));
+
+  uint16_t topicFilterLength;
+  __CPROVER_assume(topicFilterLength < TOPIC_LENGTH_MAX);
+
+  char* TopicFilter = malloc_can_fail(topicFilterLength);
```
markrtuttle: I think there is an invariant assumed in MQTT that (length == 0) iff (string == NULL) for strings and their lengths.  But if this works, that's great.

cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
```diff
@@ -0,0 +1,30 @@
...
+
+  IotMqttSubscription_t result;
+
+  IotMqtt_IsSubscribed( mqttConnection,
+			TopicFilter,
+			topicFilterLength,
+                        nondet_bool() ? NULL : &result );
```
markrtuttle: Check that indention is consistent in the source file (diffs sometimes mess that up).

cbmc/proofs/IotMqtt_IsSubscribed/Makefile
```diff
@@ -0,0 +1,43 @@
...
+
+OBJS += $(ENTRY)_harness.goto
+OBJS += $(MQTT)/cbmc/proofs/mqtt_state.goto
+OBJS += $(MQTT)/libraries/standard/mqtt/src/iot_mqtt_api.goto
+OBJS += $(MQTT)/libraries/standard/mqtt/src/iot_mqtt_subscription.goto
+
+# One more than actual number of subscriptions
+SUBSCRIPTION_COUNT_MAX=4
+DEF += -DSUBSCRIPTION_COUNT_MAX=$(SUBSCRIPTION_COUNT_MAX)
+OPERATION_COUNT_MAX=4
```
markrtuttle: Add blank line and comment "One more than actual number of operations in an operation list"

cbmc/proofs/IotMqtt_IsSubscribed/Makefile
```diff
@@ -0,0 +1,43 @@
...
+ABSTRACTIONS += --remove-function-body _packetMatch
+ABSTRACTIONS += --remove-function-body _topicFilterMatch
+
+
+OBJS += $(ENTRY)_harness.goto
+OBJS += $(MQTT)/cbmc/proofs/mqtt_state.goto
+OBJS += $(MQTT)/libraries/standard/mqtt/src/iot_mqtt_api.goto
+OBJS += $(MQTT)/libraries/standard/mqtt/src/iot_mqtt_subscription.goto
+
+# One more than actual number of subscriptions
```
markrtuttle: ... in a subscription list

cbmc/proofs/IotMqtt_IsSubscribed/IotMqtt_IsSubscribed_harness.c
feliperodri: The indentation is correct on the source file.

PR #796, opened 2020-02-12
Apply patches before running proofs

karkhaz:
prepare.py will now apply any patch files in the cbmc/patches directory
before starting the proof. Patches to the codebase are necessary for
some of the proofs to run.

### DIFF
--- a/cbmc/patches/README.md
+++ b/cbmc/patches/README.md
@@ -1,0 +1,6 @@
+Patches for Proofs
+==================
+
+This directory contains patch files that must be applied to the codebase
+before running CBMC proofs. The `cbmc/proofs/prepare.py` script takes
+care of applying these patches.
--- a/cbmc/proofs/prepare.py
+++ b/cbmc/proofs/prepare.py
@@ -24,16 +24,52 @@
 
 import logging
 import os
+import pathlib
+import subprocess
 import sys
 import textwrap
-from subprocess import CalledProcessError
 
 from make_cbmc_batch_files import create_cbmc_yaml_files
+
+
+def apply_patches():
+    patch_dir = pathlib.Path(__file__).resolve().parent.parent / "patches"
+    if not patch_dir.is_dir():
+        logging.error("Patches directory at '%s' is not a directory", patch_dir)
+        sys.exit(1)
+
+    proj_dir = pathlib.Path(__file__).resolve().parent.parent.parent
+    if not proj_dir.is_dir():
+        logging.error("Root of project at '%s' is not a directory", proj_dir)
+        sys.exit(1)
+    if not (proj_dir / "LICENSE").exists():
+        logging.error(
+            "Directory '%s' doesn't seem to be root of project", proj_dir)
+        sys.exit(1)
+
+    for fyle in sorted(patch_dir.glob("*.patch")):
+        logging.info("Checking patch '%s'", fyle.name)
+        cmd = ["git", "apply", "--check", str(fyle)]
+        proc = subprocess.run(
+            cmd, cwd=proj_dir, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+            universal_newlines=True)
+        if proc.returncode:
+            logging.warning(
+                "Patch checking failed. Check output:\n%s",
+                "\n".join(["    %s" % line for line in proc.stdout]))
+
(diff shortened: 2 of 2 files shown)


cbmc/proofs/prepare.py
```diff
@@ -24,16 +24,52 @@
...
+        logging.error(
+            "Directory '%s' doesn't seem to be root of project", proj_dir)
+        sys.exit(1)
+
+    for fyle in sorted(patch_dir.glob("*.patch")):
+        logging.info("Checking patch '%s'", fyle.name)
+        cmd = ["git", "apply", "--check", str(fyle)]
+        proc = subprocess.run(
+            cmd, cwd=proj_dir, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+            text=True)
```
markrtuttle: This is not 3.5.  Consider newlines=True.

cbmc/proofs/prepare.py
```diff
@@ -24,16 +24,52 @@
...
+import subprocess
 import sys
 import textwrap
-from subprocess import CalledProcessError
 
 from make_cbmc_batch_files import create_cbmc_yaml_files
 
+
+def apply_patches():
+    patch_dir = pathlib.Path(__file__).resolve().parent.parent / "patches"
```
markrtuttle: Does this really work?  Creates a string to a patches directory?

cbmc/proofs/prepare.py
```diff
@@ -24,16 +24,52 @@
...
+    if not patch_dir.is_dir():
+        logging.error("Patches directory at '%s' is not a directory", patch_dir)
+        sys.exit(1)
+
+    proj_dir = pathlib.Path(__file__).resolve().parent.parent.parent
+    if not proj_dir.is_dir():
```
markrtuttle: Seems like overkill, but okay.

cbmc/proofs/prepare.py
```diff
@@ -24,16 +24,52 @@
...
+        logging.error("Root of project at '%s' is not a directory", proj_dir)
+        sys.exit(1)
+    if not (proj_dir / "LICENSE").exists():
```
markrtuttle: Seems like overkill, but okay.

cbmc/proofs/prepare.py
```diff
@@ -24,16 +24,52 @@
...
+        sys.exit(1)
+
+    for fyle in sorted(patch_dir.glob("*.patch")):
+        logging.info("Checking patch '%s'", fyle.name)
+        cmd = ["git", "apply", "--check", str(fyle)]
+        proc = subprocess.run(
+            cmd, cwd=proj_dir, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
+            text=True)
+        if proc.returncode:
+            logging.warning(
```
markrtuttle: We want this to appear in the cloud watch log, is warning going to print on stdout?

cbmc/proofs/prepare.py
karkhaz: It creates a Path object, which returns the path as a string if you wrap it in `str()` or print it. Otherwise, it's more convenient than `os.path` because you can directly call methods on it and access its `parent` member etc.

cbmc/proofs/prepare.py
karkhaz: done

cbmc/proofs/prepare.py
karkhaz: It'll print on stderr, doesn't that work?

cbmc/patches/README.md
```diff
@@ -0,0 +1,6 @@
+Patches for Proofs
+==================
+
+This directory contains patch files that must be applied to the codebase
+before running CBMC proofs. The `cbmc/proofs/prepare.py` script takes
+care applying these patches.
```
dcgaws: "care of applying"

karkhaz: done, thanks

cbmc/proofs/prepare.py
markrtuttle: I'm concerned about the / notation.  It appears to be trying to form a string without using a string.

markrtuttle: I've confirmed it does work, but I don't understand the python mechanism that makes it work.

PR #804, opened 2020-02-14
CBMC: Correct type used in mqtt_state for operation list allocation.

markrtuttle:
This patch repairs a type used in operation list allocation for the CBMC proofs.  Since the offset of the link struct in the two types are both 0, this won't change any results.

### DIFF
--- a/cbmc/proofs/mqtt_state.c
+++ b/cbmc/proofs/mqtt_state.c
@@ -182,7 +182,7 @@ bool valid_IotMqttOperationList( const IotListDouble_t *pOp,
   IotListDouble_t *pLink;
   IotContainers_ForEach( pOp, pLink ) {
     IotMqttOperation_t
-      *pElt = IotLink_Container( _mqttSubscription_t, pLink, link );
+      *pElt = IotLink_Container( struct _mqttOperation, pLink, link );
     if (! valid_IotMqttOperation( pElt ) ) return false;
   }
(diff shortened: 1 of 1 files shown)


cbmc/proofs/mqtt_state.c
```diff
@@ -182,7 +182,7 @@ bool valid_IotMqttOperationList( const IotListDouble_t *pOp,
   IotListDouble_t *pLink;
   IotContainers_ForEach( pOp, pLink ) {
     IotMqttOperation_t
-      *pElt = IotLink_Container( _mqttSubscription_t, pLink, link );
+      *pElt = IotLink_Container( struct _mqttOperation, pLink, link );
```
feliperodri: You also need to update line 460.

markrtuttle: Why?  Line 460 is iterating over a subscription list and the type is correct.  This mistake was made iterating over an operation list.

feliperodri: Oh true! Approved!

gordonwang0: Spacing of these 3 lines seems inconsistent.

markrtuttle: I'm not sure how to break the long line, I pushed a revision that doesn't break the line.

PR #815, opened 2020-03-02
Adds --conversion-check flag to CBMC Makefile

feliperodri:

*Description of changes:*

1. Adds the `--conversion-check` flag to all CBMC proof harnesses (and sort them by lexicographical order).
2. Update proofs to avoid conversion errors.
3. Adds loop unwindings for `IotMqtt_DeserializeSuback`proof harness.

### DIFF
--- a/cbmc/proofs/IotMqtt_DeserializeSuback/Makefile
+++ b/cbmc/proofs/IotMqtt_DeserializeSuback/Makefile
@@ -28,10 +28,11 @@ DEF += -DCBMC=1 -DCBMC_PROOF_DESERIALIZE_SUBACK=1
 ################################################################
 # Loop unwinding
 
+LOOP += _decodeSubackStatus.0:$(BUFFER_SIZE)
+LOOP += IotListDouble_FindFirstMatch.0:$(SUBSCRIPTION_COUNT_MAX)
+LOOP += IotListDouble_RemoveAllMatches\$$link1.0:$(SUBSCRIPTION_COUNT_MAX)
 LOOP += valid_IotMqttSubscriptionList.0:$(SUBSCRIPTION_COUNT_MAX)
 LOOP += valid_IotMqttOperationList.0:$(OPERATION_COUNT_MAX)
-
-LOOP += _decodeSubackStatus.0:$(BUFFER_SIZE)
 
 UNWINDING += --unwind 1
 UNWINDING += --unwindset '$(shell echo $(LOOP) | sed 's/ /,/g')'
--- a/cbmc/proofs/Makefile.common
+++ b/cbmc/proofs/Makefile.common
@@ -75,17 +75,18 @@ CBMC_MAX_OBJECT_SIZE = "(SIZE_MAX>>(CBMC_OBJECT_BITS+1))"
 CBMCFLAGS += \
 	$(UNWINDING) \
 	--bounds-check \
-	--pointer-check \
-	--unwinding-assertions \
-	--nondet-static \
+	--conversion-check \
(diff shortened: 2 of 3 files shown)


feliperodri:
Pending on https://github.com/diffblue/cbmc/pull/5250

cbmc/proofs/IotMqtt_DeserializeSuback/Makefile
```diff
@@ -28,11 +28,12 @@ DEF += -DCBMC=1 -DCBMC_PROOF_DESERIALIZE_SUBACK=1
 ################################################################
 # Loop unwinding
 
+LOOP += _decodeSubackStatus.0:$(BUFFER_SIZE)
+LOOP += IotListDouble_FindFirstMatch.0:$(SUBSCRIPTION_COUNT_MAX)
+LOOP += IotListDouble_RemoveAllMatches\$$link1.0:$(SUBSCRIPTION_COUNT_MAX)
```
markrtuttle: This is adding loop unwindings?

cbmc/proofs/Makefile.common
```diff
@@ -75,17 +75,18 @@ CBMC_MAX_OBJECT_SIZE = "(SIZE_MAX>>(CBMC_OBJECT_BITS+1))"
...
 	--nan-check \
+	--nondet-static \
+	--pointer-check \
 	--pointer-overflow-check \
-	--undefined-shift-check \
 	--signed-overflow-check \
+	--undefined-shift-check \
 	--unsigned-overflow-check \
-	--flush \
+	--unwinding-assertions \
```
markrtuttle: Can you confirm this is doing nothing but adding conversion-check and sorting?

cbmc/proofs/mqtt_state.c
```diff
@@ -380,9 +380,7 @@ bool valid_IotMqttSubscriptionArray( const IotMqttSubscription_t *pSub,
 
 _mqttSubscription_t *allocate_IotMqttSubscriptionListElt( _mqttSubscription_t *pElt )
 {
-  size_t length;
-  // Assumption avoids arithmetic overflow in malloc, rechecked in valid_*
-  __CPROVER_assume( length < CBMC_MAX_OBJECT_SIZE - sizeof( *pElt ) );
+  uint16_t length;
```
markrtuttle: This is VERY brittle.  Suppose some day an engineer changes the type of length in the data structure.  I don't think this is worth the change.

markrtuttle:
This is doing more than just adding a flag.

cbmc/proofs/Makefile.common
feliperodri: Yes, it only adding conversion-check and sorting.

cbmc/proofs/mqtt_state.c
feliperodri: This is actually necessary for this function.  `length`  is used to set `topicFilterLength`, which is `uint16_t`. There is a conversion error when assigning the `size_t`non-deterministic value to an `uint16_t` variable.

cbmc/proofs/IotMqtt_DeserializeSuback/Makefile
feliperodri: Yes, there is a path that calls `IotListDouble_RemoveAllMatches`, so we need to add these loop unwindings, otherwise, the proof will fail (see `_decodeSubackStatus` function).

feliperodri:
> This is doing more than just adding a flag.

I updated the PR description to document the additional changes.

PR #825, opened 2020-03-14
Avoid undefined behaviour in IotMqtt_IsSubscribed function

feliperodri:
*Description of changes:*

This PR adds a nullness check for the `pTopicFilter` parameter in `IotMqtt_IsSubscribed` function.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_subscription.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_subscription.c
@@ -655,13 +655,6 @@ bool IotMqtt_IsSubscribed( IotMqttConnection_t mqttConnection,
                            IotMqttSubscription_t * const pCurrentSubscription )
 {
     bool status = false;
-
-    if( pTopicFilter == NULL )
-    {
-        IotLogError( "Topic filter must be set." );
-        return status;
-    }
-
     const _mqttSubscription_t * pSubscription = NULL;
     const IotLink_t * pSubscriptionLink = NULL;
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/src/iot_mqtt_subscription.c
```diff
@@ -655,6 +655,13 @@ bool IotMqtt_IsSubscribed( IotMqttConnection_t mqttConnection,
                            IotMqttSubscription_t * const pCurrentSubscription )
 {
     bool status = false;
+
+    if( pTopicFilter == NULL )
+    {
+        IotLogError( "Topic filter must be set." );
+        return status;
```
muneebahmed10: This return statement adds a second exit point for this function, and adds code before all local variables are declared, contrary to the [style guide](https://docs.aws.amazon.com/freertos/latest/lib-ref/c-sdk/main/guide_developer_styleguide.html). Is it possible to move this null check later, such as to line 679? E.g.

```
if( pTopicFilter == NULL )
{
    IotLogError( "Topic filter must be set." );
}
else
{
    /* Search for a matching subscription. */
    pSubscriptionLink = IotListDouble_FindFirstMatch( &( mqttConnection->subscriptionList ),
                                                      NULL,
                                                      _topicMatch,
                                                      &topicMatchParams );
}
```

feliperodri: Done.

PR #823, opened 2020-03-12
Remove thread-safety from fleet provisioning APIs

aggarw13:
*Issue #, if available:*

*Description of changes:*
Remove all synchronization logic from Fleet Provisioning API functions for simplification and better readability. 
Changes involve:
* Deletion of all code relating to the `lock` and `mutexReferenceCount` members of the `_provisioningOperation` struct
* Updates to the API documentation for non-thread-safe behavior

**Rationale**: The Fleet Provisioning workflow does not requires the API to support concurrency. Synchronization logic was adding unnecessary overhead on the library.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #826, opened 2020-03-14
Csr feat/implement api layer

aggarw13:
*Issue #, if available:*
TS-12335

*Description of changes:*
Implement API layer for CSR support in the Fleet Provisioning library by adding a new API function  `AwsIotProvisioning_CreateCertificateFromCsr`
(This PR stubs out the internal serializer layer dependency)

Link to the service-side specification for the new APIs: https://w.amazon.com/bin/view/AWS/Mobile/IoT/Foundry/Internal/FoundryCSRSupport/#HAPIOverview


libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
```diff
@@ -277,6 +277,41 @@ typedef struct AwsIotProvisioningRejectedResponse
...
+ * The @ref AwsIotProvisioning_CreateKeysAndCertificate library API passes this object to a user-provided callback
+ * function
+ * whenever the operation completes with a response from the server.
+ */
+typedef struct AwsIotProvisioningCreateCertificateFromCsrResponse
+{
+    /** @brief The highest level HTTP based status code sent by the server. */
+    AwsIotProvisioningServerStatusCode_t statusCode;
+
+    union
```
abhidixi11: Generally, we have decided to minimize use of unions  in the new code unless absolutely required.  I believe there MISRA as well doxygen issue.

aggarw13: Will take it as a hygiene story. Removing the union will involve a larger re-factor of the library architecture

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +516,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
+                     "Operation={%s}", CREATE_CERTIFICATE_FROM_CSR_OPERATION_LOG );
+
+        status = AWS_IOT_PROVISIONING_BAD_PARAMETER;
+    }
+    /* Check that a callback function object along with a valid callback functor is provided. */
+    else if( ( pResponseCallback == NULL ) ||
+             ( pResponseCallback->function == NULL ) )
+    {
+        IotLogError(
+            "Bad Paramater: Invalid callback provided: Operation={%s}",
```
muneebahmed10: `Parameter`

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
```diff
@@ -316,6 +351,41 @@ typedef struct AwsIotProvisioningCreateKeysAndCertificateResponse
...
+    /**
+     * @brief User-provided callback function signature.
+     *
+     * @param[in] void* #AwsIotProvisioningCreateCertificateFromCsrCallbackInfo_t.userParam
+     * @param[in] AwsIotProvisioningCallbackParam_t * Parsed server response of either device credentials
+     * or provisioned device information.
+     *
+     * @see #AwsIotProvisioningCreateCertificateFromCsrResponse_t for more information on the second parameter.
+     */
+    void ( * function )( void *,
```
abhidixi11: We specify parameters, not just types.

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
```diff
@@ -277,6 +277,41 @@ typedef struct AwsIotProvisioningRejectedResponse
...
+            size_t certificateIdLength;              /**< The length of the certificate ID.*/
+            const char * pCertificateOwnershipToken; /**< The token that represents ownership of certificate and
+                                                      * associated private key that the device.*/
+            size_t ownershipTokenLength;             /**< The size of the ownership token.*/
+        } acceptedResponse;
+
+        /** @brief Represents the rejected response information received from the server. */
+        AwsIotProvisioningRejectedResponse_t rejectedResponse;
+    } u; /**< @brief Valid member depends on operation status. */
+} AwsIotProvisioningCreateCertificateFromCsrResponse_t;
```
muneebahmed10: These type names seem a bit long. Is there a problem with a shorter name like `AwsIotProvisioningCsrCreateCertResponse_t`?

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +516,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
 
 /*-----------------------------------------------------------*/
 
+AwsIotProvisioningError_t AwsIotProvisioning_CreateCertificateFromCsr( IotMqttConnection_t connection,
```
abhidixi11: Possibly refactor this function to reduce complexity as well as increase readability.  (Another hygiene type task, will eventually get addressed when you run complexity tool. Since this is new code, I'm pointing it out. )

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
```diff
@@ -522,6 +529,22 @@ AwsIotProvisioningError_t _AwsIotProvisioning_ParseRegisterThingResponse( AwsIot
 AwsIotProvisioningError_t _AwsIotProvisioning_SerializeCreateKeysAndCertificateRequestPayload( uint8_t ** pSerializationBuffer,
                                                                                                size_t * pBufferSize );
 
+/**
+ * @brief Serializes payload data in a buffer, if provided, of a request to the CreateCertificateFromCsr API on AWS IoT Core;
+ * otherwise stores the calculated size of the serialized data in the passed paramater, @a pBufferSize.
```
muneebahmed10: `parameter`

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
```diff
@@ -277,6 +277,41 @@ typedef struct AwsIotProvisioningRejectedResponse
     size_t errorMessageLength;  /**< The length of the error messsage sent by the server. */
 } AwsIotProvisioningRejectedResponse_t;
 
+/**
+ * @ingroup provisioning_datatypes_paramstructs
+ * @brief Aggregates the data sent as response from AWS IoT Core for the request to create a new certificate for a CSR request.
```
muneebahmed10: Re-word: `for` -> `to` and delete last `request`. E.g. `Aggregates the data sent as AWS IoT Core's response to the request to create  a new certificate for a CSR.` or even `Aggregates the data sent as AWS IoT Core's response to the CSR certificate creation request.`

libraries/aws/provisioning/src/aws_iot_provisioning_serializer.c
```diff
@@ -355,6 +355,21 @@ AwsIotProvisioningError_t _AwsIotProvisioning_SerializeCreateKeysAndCertificateR
 
 /*------------------------------------------------------------------*/
 
+bool _AwsIotProvisioning_SerializeCreateCertificateFromCsrRequestPayload( const char * pCertificateSigningRequest,
```
abhidixi11: This function just returns true ? Is serialization code going to get added later ?

libraries/aws/provisioning/include/aws_iot_provisioning.h
```diff
@@ -115,12 +116,59 @@ AwsIotProvisioningError_t AwsIotProvisioning_Init( uint32_t mqttTimeout );
...
                                                                        const AwsIotProvisioningCreateKeysAndCertificateCallbackInfo_t * pResponseCallback );
 /* @[declare_provisioning_createkeysandcertificate] */
 
+/**
+ * @brief Requests the AWS IoT Core service for a certificate by sending a Certificate-Signing Request, and
+ * invokes the provided user-defined response handler with the response it receives from the server.
+ *
+ * @note It is advised to use a shared MQTT connection to AWS IoT Core across all API functions.
+ *
+ * @warning This function is NOT thread-safe. Concurrent calls to the library API functions can result in undefined
```
muneebahmed10: I see the other functions also follow this pattern, but it seems weird that the `@note`s and `@warning`s are separated by the `@param` list instead of being all together

libraries/aws/provisioning/src/aws_iot_provisioning_serializer.c
aggarw13: Yes, stubbed out to reduce the scope of this PR.

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
aggarw13: No issue, thanks for the suggestion :)

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
aggarw13: Updated with `Aggregates the data from AWS IoT Core's response to the request of a new certificate with a CSR.`

libraries/aws/provisioning/include/aws_iot_provisioning.h
aggarw13: I had intentionally followed the pattern to keep the information of the API functions separate from its parameter list documentation, as it was more readable to me.

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
aggarw13: Updated all callback types to have parameter names

libraries/aws/provisioning/include/aws_iot_provisioning.h
abhidixi11: OK.

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
abhidixi11: OK.

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
```diff
@@ -279,15 +279,15 @@ typedef struct AwsIotProvisioningRejectedResponse
...
+ * @brief Aggregates the data from AWS IoT Core's response to the certificate creation request with a CSR.
  *
  * @paramfor Response Callback function of @ref provisioning_function_createkeysandcertificate
  *
  * The @ref AwsIotProvisioning_CreateKeysAndCertificate library API passes this object to a user-provided callback
  * function
  * whenever the operation completes with a response from the server.
  */
-typedef struct AwsIotProvisioningCreateCertificateFromCsrResponse
+typedef struct AwsIotProvisioningCreateCertFromCsrResponse
```
muneebahmed10: This change can also apply to `AwsIotProvisioningCreateCertificateFromCsrCallbackInfo_t` and the functions with long names as well.

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
dcgaws: If this violates MISRA, and creates problems for Doxygen, and it's new code, why not fix it now? Once the code is committed, the changes will be riskier to make.

libraries/aws/provisioning/include/aws_iot_provisioning.h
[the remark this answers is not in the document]
aggarw13: The non-thread-safe character of the library is a deliberate change I made recently (in [this PR](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/823) ) to simplify the internal implementation by removing the use of a mutex and an atomic reference counter. 
As the Foundry workflow requires a sequential invocation of these APIs, providing thread-safe APIs was just adding performance overhead and complexity which was unnecessary.

libraries/aws/provisioning/include/aws_iot_provisioning.h
```diff
@@ -115,12 +116,59 @@ AwsIotProvisioningError_t AwsIotProvisioning_Init( uint32_t mqttTimeout );
...
+ * behavior. Device provisioning with this library REQUIRES calling the API functions of this library sequentially.
+ *
+ * @param[in] connection The MQTT connection handle that will be used to communicate with AWS IoT Core for
+ * the Certificate-Signing Request.
+ * @param[in] operationQos The Quality of Service (QoS) level for the MQTT publish/subscribe communication with the server.
+ * @param[in] pCertificateSigningRequest The PEM encoded string for the Certificate-Signing Request.
```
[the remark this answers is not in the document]
aggarw13: I agree, that is something the service-side doesn't currently support. I can initiate a conversation with them to see if they could support it before the GA. 
From the SDK perspective, it would be simple to accommodate both DER and PEM encoding

libraries/aws/provisioning/include/aws_iot_provisioning.h
```diff
@@ -115,12 +116,59 @@ AwsIotProvisioningError_t AwsIotProvisioning_Init( uint32_t mqttTimeout );
...
+ * function returns #AWS_IOT_PROVISIONING_TIMEOUT.
+ * @param[in] pResponseCallback The user-defined callback that will be invoked with the server's response to the CSR request.
+ * The server can respond either with the new certificate information (in case of request acceptance) OR with error (in case of request
+ * rejection).
+ *
+ * @note The callback is expected to store/retain the new certificate information from the server response, which will be required for
+ * registering the device with @ref provisioning_function_registerthing function.
+ *
+ * @warning Do not overwrite the existing certificate (used for making the server connection) with the new certificate that is provided
+ * by the server. It is RECOMMENDED to retain the Provisinoning claim credentials until the device has been provisioned with a new certificate
```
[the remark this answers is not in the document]
aggarw13: I'm referring to the client certificate being replaced as the "claim credentials". In the Foundry workflow, the claim certificate and private key (which is either programmed on the manufacturing floor OR handed to the device through a mobile app) is used to make the first connection to AWS IoT Core to start the provisioning process. 
The customer shouldn't need to store the claim credentials once they have established an IoT Core connection with the new vended certificate. (In the scenario of the claim credentials being given by a mobile app, the claim credential stays valid only for a limited time of some amount of minutes)

libraries/aws/provisioning/include/types/aws_iot_provisioning_types.h
aggarw13: It is new code that fits within the library's architecture of interaction between the API and Parser modules. I agree that the union should be removed, however, removing that will require me to accept 2 separate callbacks (one for success response, and one for rejected response) from the caller, and write custom-code only for this new API.
This would be inconsistent with the existing `CreateKeysAndCertificate` and `RegisterThing` APIs. Thus, I want to tackle this hygiene re-factor in a separate PR to avoid scope creep.

libraries/aws/provisioning/include/aws_iot_provisioning.h
```diff
@@ -115,12 +116,59 @@ AwsIotProvisioningError_t AwsIotProvisioning_Init( uint32_t mqttTimeout );
...
+ *
+ * @param[in] connection The MQTT connection handle that will be used to communicate with AWS IoT Core for
+ * the Certificate-Signing Request.
+ * @param[in] operationQos The Quality of Service (QoS) level for the MQTT publish/subscribe communication with the server.
+ * @param[in] pCertificateSigningRequest The PEM encoded string for the Certificate-Signing Request.
+ * @param[in] csrLength The length of the Certificate-Signing Request string.
+ * @param[in] timeoutMs The timeout (in milliseconds) for a response from the server. If there is a timeout, this
+ * function returns #AWS_IOT_PROVISIONING_TIMEOUT.
+ * @param[in] pResponseCallback The user-defined callback that will be invoked with the server's response to the CSR request.
+ * The server can respond either with the new certificate information (in case of request acceptance) OR with error (in case of request
```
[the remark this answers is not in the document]
aggarw13: Updated the entire file to screen-size readable

libraries/aws/provisioning/include/aws_iot_provisioning.h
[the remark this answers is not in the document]
dcgaws: If the API operates on a context structure that is provided as an explicit input to each call, and it's the responsibility of the caller to protect the sharing of that context, then this API should otherwise be implemented in a way that is thread safe (and commented as such).

libraries/aws/provisioning/include/aws_iot_provisioning.h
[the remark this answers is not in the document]
dcgaws: My understanding of the claim credentials is that we need to allow them to be claimed more than once. For example, if I activate the part in my account, then later deactivate the part in my account, then resell the product to somebody else, then try to activate the part in their account - that workflow should be possible. In that case, the claim credential is retained forever but all of the derived credentials should be wiped when I deprovision the device before reselling it.

libraries/aws/provisioning/include/aws_iot_provisioning.h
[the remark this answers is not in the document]
aggarw13: No context input is provided by the caller for any of the APIs that is shared between them. It is the internal state object of the library that is shared by each of the API functions, which is not protected for concurrent access and modification (as the Foundry provisioning workflow is ordered in the use of its APIs)

libraries/aws/provisioning/include/aws_iot_provisioning.h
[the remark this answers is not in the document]
aggarw13: That's an interesting use-case for "factory-reset". Yes, in that case, the birth certificate installed at the manufacturing floor would need to be always retained. I will update the API comment.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +533,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
+                                    &responseSubscription );
+
+        AwsIotProvisioning_FreePayload( pPayloadBuffer );
+    }
+
+    IotLogInfo( "Operation is complete: Operation={%s}",
+                CREATE_CERT_FROM_CSR_OPERATION_LOG );
+
+    /* Reset the active operation */
+    _resetActiveOperationData();
```
dcgaws: Is this global state? If so, this is an example of something that could be refactered into a caller-maintained context handle.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +533,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
+    {
+        IotLogDebug( "Subscribed to response topics: Operation={%s}",
+                     CREATE_CERT_FROM_CSR_OPERATION_LOG );
+
+        /* Update the operation object to represent an active "Certificate-Signing Request" operation. */
+        _provisioningCallbackInfo_t callbackInfo;
+        callbackInfo.createCertificateFromCsrCallback = *pResponseCallback;
+        _setActiveOperation( &callbackInfo );
+
+        /* Serialization of request payload occurs in a 2-step process, one for calculation of buffer size, and then,
```
dcgaws: This is a common idiom, of course. There is a cost, though, which is that computing the payload length may require the whole request to be serialized and then thrown away.

An alternative approach is to pass the allocator into the function.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -230,6 +236,17 @@ static void _keysAndCertificateResponseReceivedCallback( void * param1,
 
 /*-----------------------------------------------------------*/
 
+static void _csrResponseReceivedCallback( void * param1,
+                                          IotMqttCallbackParam_t * const pPublish )
+{
+    /* Silence warnings about unused variables.*/
+    ( void ) param1;
+
+    _commonServerResponseHandler( pPublish, _AwsIotProvisioning_ParseCsrResponse );
```
dcgaws: Why not pass param1 all the way through to the _common handler?

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
```diff
@@ -522,6 +585,22 @@ AwsIotProvisioningError_t _AwsIotProvisioning_ParseRegisterThingResponse( AwsIot
...
                                                                                                size_t * pBufferSize );
 
+/**
+ * @brief Serializes payload data in a buffer, if provided, of a request to the CreateCertificateFromCsr API on AWS IoT Core;
+ * otherwise stores the calculated size of the serialized data in the passed parameter, @a pBufferSize.
+ *
+ * @param[in] pCertificateSigningRequest The Certificate-Signing Request string to serialize for the request.
+ * @param[in] csrLength The length of the Certificate-Signing Request string.
+ * @param[in] pSerializationBuffer The buffer for storing the serialized payload data.
+ * @param[in, out] pBufferSize If the payload buffer is provided, this should contain the size of the buffer. Otherwise,
```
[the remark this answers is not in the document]
aggarw13: > What is the return value if pSerializationBuffer is NULL?
That is the case for a dry-run of serialization for calculating the required buffer size. The `pBufferSize` output parameter will be populated with the calculated size

> What about if pSerializationBuffer is non-NULL and pBufferSize is too small?
It will return an error as the return value of the call.

Do you prefer that I break down these use-cases in 2 different functions instead?

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +533,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
+        .topicFilterBaseLength = PROVISIONING_CREATE_CERT_FROM_CSR_RESPONSE_TOPIC_FILTER_LENGTH
+    };
+    size_t payloadSize = 0;
+    uint8_t * pPayloadBuffer = NULL;
+    IotMqttPublishInfo_t publishInfo = IOT_MQTT_PUBLISH_INFO_INITIALIZER;
+
+    AwsIotProvisioningError_t status = AWS_IOT_PROVISIONING_SUCCESS;
+
+    /* Verify that library has been initialized. */
+    AwsIotProvisioning_Assert( _checkInit() == true );
```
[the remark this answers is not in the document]
aggarw13: It is somewhat following the kernel-style of error handling to enforce correct usage of API through asserts than adding explicit error handling logic. I think that this misuse of the API _should_ be caught and fixed during development.

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
```diff
@@ -257,6 +261,42 @@
...
+ * It should be utilized in the @ref provisioning_function_registerthing API function.
+ */
+#define PROVISIONING_CREATE_CERT_FROM_CSR_RESPONSE_TOPIC_FILTER \
+    "$aws/certificates/create-from-csr/"PROVISIONING_FORMAT
+
+/**
+ * @brief Length of the response topic filter for the MQTT CreateCertificateFromCsr service API.
+ */
+#define PROVISIONING_CREATE_CERT_FROM_CSR_RESPONSE_TOPIC_FILTER_LENGTH \
+    ( ( uint16_t ) ( sizeof( PROVISIONING_CREATE_CERT_FROM_CSR_RESPONSE_TOPIC_FILTER ) - 1 ) )
```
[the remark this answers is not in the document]
aggarw13: This is a non-customer facing macro.. Why would these need to be re-implemented as functions?

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +533,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
 /*-----------------------------------------------------------*/
 
+AwsIotProvisioningError_t AwsIotProvisioning_CreateCertificateFromCsr( IotMqttConnection_t connection,
+                                                                       IotMqttQos_t operationQos,
+                                                                       const char * pCertificateSigningRequest,
+                                                                       size_t csrLength,
+                                                                       uint32_t timeoutMs,
+                                                                       const AwsIotProvisioningCreateCertFromCsrCallbackInfo_t * pResponseCallback )
+{
+    char responseTopicsBuffer[ PROVISIONING_CREATE_CERT_FROM_CSR_RESPONSE_MAX_TOPIC_LENGTH ] =
```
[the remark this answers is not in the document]
aggarw13: Okay, I will update the initialization to directly store the topic filter string instead of doing a memcpy later.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +533,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
+            "Bad parameter: Invalid callback provided: Operation={%s}",
+            CREATE_CERT_FROM_CSR_OPERATION_LOG );
+
+        status = AWS_IOT_PROVISIONING_BAD_PARAMETER;
+    }
+
+    if( status == AWS_IOT_PROVISIONING_SUCCESS )
+    {
+        /* Copy the response topics in a local buffer for appropriate suffixes to be added. */
+        ( void ) memcpy( responseTopicsBuffer, PROVISIONING_CREATE_KEYS_AND_CERTIFICATE_RESPONSE_TOPIC_FILTER,
```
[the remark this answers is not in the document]
aggarw13: By allocating the buffer with `PROVISIONING_CREATE_CERT_FROM_CSR_RESPONSE_MAX_TOPIC_LENGTH` which is equal to `topic filter base length + max(accepted, rejected) length`
To enforce that these values are not changed, I can add an assert check

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
```diff
@@ -28,6 +28,9 @@
 #ifndef AWS_IOT_PROVISIONING_INTERNAL_H_
 #define AWS_IOT_PROVISIONING_INTERNAL_H_
 
+/* Standard include */
+#include <stdbool.h>
```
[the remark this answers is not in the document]
aggarw13: The Fleet Provisioning library already contains dependency on CRT with the use of string functions.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +533,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
+
+        /* Update the operation object to represent an active "Certificate-Signing Request" operation. */
+        _provisioningCallbackInfo_t callbackInfo;
+        callbackInfo.createCertificateFromCsrCallback = *pResponseCallback;
+        _setActiveOperation( &callbackInfo );
+
+        /* Serialization of request payload occurs in a 2-step process, one for calculation of buffer size, and then,
+         * serialization in allocated buffer. */
+        /* Dry run serialization */
+        if( _AwsIotProvisioning_SerializeCreateCertificateFromCsrRequestPayload( pCertificateSigningRequest,
```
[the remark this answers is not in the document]
aggarw13: As I had thought this function to return only binary status codes (`PROVISIONING_SUCCESS` or `PROVISIONING_INTERNAL_FAILURE`), I thought that using a `bool` type would be more fitting from a readability perspective. Ultimately, the `false` return value would be translated to `PROVISIONING_INTERNAL_FAILURE` by the API function to return as the error code to the app

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: As the `tinycbor` library we are using, requires the buffer to be pre-allocated, we cannot avoid the 2-step process of serialization in any layer. The same is followed by any other library performing CBOR encoding (Defender, BLE (I think))
For this (CSR) feature's serializer, I deviated from the existing serializer functions (of the other APIs in the library) to make it the responsibility of the API function/layer to allocate the memory instead of making the serializer do so to align follow the principle of single-responsibility.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: As it is not required by the `_commonServerResponseHandler()` function. I would have to use the same `(void) param1` logic in the common function.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: But the state is internal to the functioning of the library...Do we want to expose the internal library state details? (Or is your suggestion that the library implementation should be state-less to make it _light-weight_ ? )

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -517,6 +533,195 @@ AwsIotProvisioningError_t AwsIotProvisioning_CreateKeysAndCertificate( IotMqttCo
...
+    size_t payloadSize = 0;
+    uint8_t * pPayloadBuffer = NULL;
+    IotMqttPublishInfo_t publishInfo = IOT_MQTT_PUBLISH_INFO_INITIALIZER;
+
+    AwsIotProvisioningError_t status = AWS_IOT_PROVISIONING_SUCCESS;
+
+    /* Verify that library has been initialized. */
+    AwsIotProvisioning_Assert( _checkInit() == true );
+
+    if( connection == IOT_MQTT_CONNECTION_INITIALIZER )
```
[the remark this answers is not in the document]
aggarw13: The intention here is to check if the passed connection handle is valid. Ideally, if the MQTT API provided a utility for checking the validity of the connection handle, that would suffice the intended parameter checking.
As you have pointed out that this method of verification is not-readable and vulnerable to customer change, would it be fine if I make an explicit verification with `if(connection == NULL)`? The publicly visible type-definition of `IotMqttOperation_t` is [`typedef struct _mqttOperation    * IotMqttOperation_t;`](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/173cf48a8c9687d8dd0b798e0a85c6a86015d89c/libraries/standard/mqtt/include/types/iot_mqtt_types.h#L87), and I want to input-validate against the caller just sending in a NULL value.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
dcgaws: Good point. For internal state, the caller context could either be entirely opaque, or be a public struct that includes a pointer to the opaque data.

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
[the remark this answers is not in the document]
dcgaws: Fewer external dependencies, even in terms of individual headers, helps simplify porting and reduce code size. In this particular case, it’s also not clear why bool is preferable to a status code.

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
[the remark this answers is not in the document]
dcgaws: Are they system topics that can’t change? If so, I agree that it’s not an issue.

libraries/aws/provisioning/include/aws_iot_provisioning.h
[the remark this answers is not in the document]
dcgaws: In other words, the existing library already depends on global state?

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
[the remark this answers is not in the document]
dcgaws: Or at least define and document those two behaviors.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
dcgaws: Is the callee having to do the encoding twice in order to service the size-only call? Or can it compute the required size just based on the input size?

libraries/aws/provisioning/include/aws_iot_provisioning.h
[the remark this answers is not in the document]
aggarw13: Yes.

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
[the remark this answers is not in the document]
aggarw13: Separated the 2 use-cases in different functions `_AwsIotProvisioning_CalculateCertFromCsrPayloadSize` and `_AwsIotProvisioning_SerializeCreateCertFromCsrRequestPayload` in [this commit](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/826/commits/cb41a55110f582211428e5085505223a95d50fad#diff-77f76cfbe42e8353b432bae257edbe4e)

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: There is no exposure of the internal state in the APIs, it is completely encapsulated within the library implementation..

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: Yes, the callee is effectively doing the encoding twice. As mentioned in a previous comment-chain reply, I have broken the serializer function into 2 for better code readability

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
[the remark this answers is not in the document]
aggarw13: Added the (runtime) assert check

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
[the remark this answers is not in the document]
aggarw13: Okay, fair point. Will tackle that in a separate PR to avoid delaying this PR further.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
dcgaws: Understood. It should stay that way. But we could eliminate the lock by having the caller hold our state in caller-supplied memory, and pass that state to us with each API call. A "handle" parameter is how that's frequently done.

libraries/aws/provisioning/src/private/aws_iot_provisioning_internal.h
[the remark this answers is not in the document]
dcgaws: We may be doing the review sequencing incorrectly in this case then. There's no reason not to merge these changes as-is into a feature branch.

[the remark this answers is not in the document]
aggarw13: Okay, have removed the `bool` type dependency in [this commit](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/826/commits/ad87e23a259b933a47206d6b92b035c63ca59d9c)

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -169,51 +202,145 @@ static bool _checkInit( void )
...
+
+            /* Notify the waiting thread about arrival of response from server */
+            /* Increment the number of PUBLISH messages received. */
+            IotSemaphore_Post( &_activeOperation[ operationIndex ].responseReceivedSem );
+
+            /* Invalidate the user-callback information to prevent re-processing the response
+             * if we receive the same response multiple times (which is possible for a QoS 1 publish
+             * from the server). This is done to post on the semaphore ONLY ONCE on receiving the
+             * response from the server.*/
+            memset( &_activeOperation[ operationIndex ].info.userCallback,
```
dcgaws: In the QoS1 multiple responses scenarios, can those responses be posted to multiple task pool threads? If so, could this memset be racing with another one of those response-handling task pool threads?

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -169,51 +202,145 @@ static bool _checkInit( void )
...
-        if( responseStatus == AWS_IOT_UNKNOWN )
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+static void _cleanUpOperationObject( uint8_t operationIndex )
+{
+    /*
+     * Check whether the operation semaphore is being accessed by another thread context,
```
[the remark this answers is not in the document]
aggarw13: Yes, the concurrent access can be from the response callback on the taskpool thread. I will update the comment.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -169,51 +202,145 @@ static bool _checkInit( void )
...
+     * before destroying the resource.
+     * This tackles the RACE CONDITION of concurrent access of the semaphore from multiple
+     * thread contexts.
+     */
+    if( Atomic_CompareAndSwap_u32( &_activeOperation[ operationIndex ].semReferenceCount,
+                                   0u,
+                                   1u ) == 1u )
+    {
+        /* Semaphore is safe to destroy as no other context is accessing it. */
```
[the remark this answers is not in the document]
aggarw13: I have added logic in the MQTT subscription callback to delete the resource in that corner case.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -169,51 +202,145 @@ static bool _checkInit( void )
...
+
+/*-----------------------------------------------------------*/
+
+static void _commonServerResponseHandler( const uint8_t operationIndex,
+                                          IotMqttCallbackParam_t * const pPublishData,
+                                          _provisioningServerResponseParser responseParser )
+{
+    /*
+     * Check whether the operation's sempahore is valid. There is possibility
+     * of a RACE CONDITION between #AwsIotProvisioning_CleanUp destroying the
```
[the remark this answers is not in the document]
aggarw13: I intended the comment to mention that the use of semaphore reference counter is addressing the race condition. 
I will re-phrase the comment

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: Yes, you have rightly pointed at the race condition in the QoS multiple responses case. I will remove this logic.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -169,51 +202,145 @@ static bool _checkInit( void )
...
-        /* Notify the waiting thread about arrival of response from server */
-        /* Increment the number of PUBLISH messages received. */
-        IotSemaphore_Post( &_activeOperation.responseReceivedSem );
-
-        /* Invalidate the user-callback information to prevent re-processing the response
-         * if we receive the same response multiple times (which is possible for a QoS 1 publish
-         * from the server). This is done to post on the semaphore ONLY ONCE on receiving the
-         * response from the server.*/
-        memset( &_activeOperation.info.userCallback, 0, sizeof( _activeOperation.info.userCallback ) );
+    /* Decrement the operation's semaphore reference count,* as we don't need
```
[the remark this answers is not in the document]
aggarw13: Removed

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
```diff
@@ -169,51 +202,145 @@ static bool _checkInit( void )
...
+     *  the mutex anymore.
+     * If reference count is lower than the count at which we incremented at the
+     * beginning of the callback, it means that the the #AwsIotProvisioning_CleanUp
+     * function was called. In that case, we destroy of semaphore to clean the memory.
```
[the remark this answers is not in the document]
aggarw13: Removed this section as pointed out to be vulnerable to a race-condition in QoS 1 multiple responses case.

libraries/aws/provisioning/src/aws_iot_provisioning_api.c
aggarw13: Purged it 🔥

PR #842, opened 2020-03-27
Csr feat/bugfix for rejected response callback invokation

aggarw13:
*Issue #, if available:*

*Description of changes:*
* Fix bug in Provisioning parser that was not calling user-callback for rejected server response case
* Update parser unit tests to check for callback invocations to prevent future un-detection

### DIFF
--- a/libraries/aws/provisioning/src/aws_iot_provisioning_parser.c
+++ b/libraries/aws/provisioning/src/aws_iot_provisioning_parser.c
@@ -120,6 +120,7 @@ static AwsIotProvisioningError_t _parseRejectedResponse( IotSerializerDecoderObj
     AwsIotProvisioning_Assert( pResponseData != NULL );
     AwsIotProvisioning_Assert( pStatusCode != NULL );
 
+    IotSerializerError_t decoderResult = IOT_SERIALIZER_SUCCESS;
     IotSerializerDecoderObject_t statusCodeDecoder = IOT_SERIALIZER_DECODER_OBJECT_INITIALIZER;
     IotSerializerDecoderObject_t errorCodeDecoder = IOT_SERIALIZER_DECODER_OBJECT_INITIALIZER;
     IotSerializerDecoderObject_t errorMessageDecoder = IOT_SERIALIZER_DECODER_OBJECT_INITIALIZER;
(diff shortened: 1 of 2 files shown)


libraries/aws/provisioning/src/aws_iot_provisioning_parser.c
```diff
@@ -120,88 +120,91 @@ static AwsIotProvisioningError_t _parseRejectedResponse( IotSerializerDecoderObj
...
     IotSerializerDecoderObject_t statusCodeDecoder = IOT_SERIALIZER_DECODER_OBJECT_INITIALIZER;
     IotSerializerDecoderObject_t errorCodeDecoder = IOT_SERIALIZER_DECODER_OBJECT_INITIALIZER;
     IotSerializerDecoderObject_t errorMessageDecoder = IOT_SERIALIZER_DECODER_OBJECT_INITIALIZER;
 
     /* Suppress warning when parameter is unused in non-Debug mode. */
     ( void ) pOperationName;
 
-    /* Initialize the status as "server refused" as we are parsing the "rejected" response. */
-    IOT_FUNCTION_ENTRY( AwsIotProvisioningError_t, AWS_IOT_PROVISIONING_SERVER_REFUSED );
+    AwsIotProvisioningError_t status = AWS_IOT_PROVISIONING_SUCCESS;
```
dan4thewin: What does it mean when a function named _parseRejectedResponse() returns provisioning success?  Does that mean parse success?

libraries/aws/provisioning/src/aws_iot_provisioning_parser.c
```diff
@@ -406,8 +409,13 @@ AwsIotProvisioningError_t _AwsIotProvisioning_ParseKeysAndCertificateResponse( A
             /* Invoke the user-provided callback with the parsed rejected data, if parsing was successful . */
             if( status == AWS_IOT_PROVISIONING_SUCCESS )
             {
+                IotLogInfo( "parser: About to call user-callback with rejected server response: Operation={\"%s\"}",
```
dan4thewin: "About to call" -> "calling" ?

libraries/aws/provisioning/src/aws_iot_provisioning_parser.c
aggarw13: Yes, as the goal of the function is to only parse the payload, my thought is to not add more responsibility/information in its return value than what its purpose warranties :)

dan4thewin: OK.  Maybe in the future we can make these constants less confusing.

PR #846, opened 2020-03-31
Update Provisioning CI script to facilitate CSR system tests (in Gamma)

aggarw13:
…run AWS CLI commands against gamma endpoint

*Issue #, if available:*

*Description of changes:*
Update Provisioning CI script to run against Gamma endpoint


scripts/ci_test_provisioning.sh
```diff
@@ -137,8 +156,12 @@ if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
         } \
     }"
 
-    # Compiler flags for integration and units tests.
-    COMMON_CMAKE_C_FLAGS="$AWS_IOT_CREDENTIAL_DEFINES -DAWS_IOT_TEST_PROVISIONING_TEMPLATE_NAME=\"\\\"$TEMPLATE_NAME\\\"\" -DAWS_IOT_TEST_PROVISIONING_TEMPLATE_PARAMETERS=\"$PROVISIONING_PARAMETERS\" $COMPILER_OPTIONS"
+    # Save the generated CSR for testing.
+    CSR_PEM_DATA=$(echo -E $(cat $CSR_FILE))
```
muneebahmed10: Why does the `echo -E` need to be here? Does `CSR_PEM_DATA=$(< $CSR_FILE)` cause the backslash escapes to be interpreted?

aggarw13: The `-E` flag enables interpretation of the escape characters

aggarw13: `CSR_PEM_DATA` is to be passed as a string for a macro definition in the cmake command, and thus, cannot easily pass a multiline string. 
With `$(eho -E $(cat <file-name>))`, I'm able to collapse the entire CSR string into a single line without the newline characters to pass a single-line string in the CMake command

PR #851, opened 2020-04-03
Add security mitigation documentatio to API

aggarw13:
*Issue #, if available:*

*Description of changes:*
Update API documentation of Provisioning library with security mitigation recommendations

### DIFF
--- a/libraries/aws/provisioning/include/aws_iot_provisioning.h
+++ b/libraries/aws/provisioning/include/aws_iot_provisioning.h
@@ -103,9 +103,9 @@ AwsIotProvisioningError_t AwsIotProvisioning_Init( uint32_t mqttTimeout );
  * For the threat of an unauthorized reuse of the private key, that is received in
  * the server response, in order to clone the device, an effective way to mitigate that
  * is to audit, in the cloud, the use of each device private key. For example, if a
- * device private key is reused, your cloud app could log an audit event for operator
- * follow-up, and/or initiate a workflow for revoking the previous certificate(s)
- * issued to that key.
+ * device private key is reused, your cloud app (through Fleet Provisioning Hooks) could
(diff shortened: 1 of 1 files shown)


libraries/aws/provisioning/include/aws_iot_provisioning.h
```diff
@@ -92,6 +92,21 @@ AwsIotProvisioningError_t AwsIotProvisioning_Init( uint32_t mqttTimeout );
  * functions can result in undefined behavior. Device provisioning with this library
  * REQUIRES calling the API functions of this library sequentially.
  *
+ * @warning Do not overwrite the existing Provisioning claim credentials with the new
+ * credentials provided by the server, at least until the device has been provisioned
+ * with a new certificate using the @ref provisioning_function_registerthing function.
+ * It is also recommended to always retain the Provisioning claim credentials,
+ * if the claim certificate is long-lived.
```
leegeth, line 99: What does long-lived claim certificate mean? Is that a standard?

leegeth: I think if we don't provide a clear definition of what a long lived certificate means, that would be confusing to the reader. For an example a reader could interpret the long -lived certificate as certificate that is valid for 30 days. What do you think?

aggarw13: In this use-case of "Provisioning By Claim", the claim credential can be created by the customer and registered on AWS IoT, thus, specifying a duration for "long-lived" certificate is not suitable here. 
The intent of the statement is to suggest that customer should retain the Provisioning Claim if they have the product use-case for re-provisioning the device with the same certificate (which implies that the certificate be valid to support re-provisioning). Maybe, I can re-word the sentence to say that "It is also recommended to always retain the Provisioning claim credentials, if your product use-case supports re-provisioning of the device"

PR #854, opened 2020-04-03
Fix doxygen errors in v3 MQTT documentation

gordonwang0:
Resolves doxygen issues in v3 MQTT documentation

### DIFF
--- a/include/aws_iot_mqtt_client.h
+++ b/include/aws_iot_mqtt_client.h
@@ -61,6 +61,7 @@ extern "C" {
 #include "threads_interface.h"
 #endif
 
+/** Greatest packet identifier, per MQTT spec */
 #define MAX_PACKET_ID 65535
 
 typedef struct _Client AWS_IoT_Client;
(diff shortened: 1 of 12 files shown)


include/aws_iot_mqtt_client.h
```diff
@@ -61,6 +61,7 @@ extern "C" {
 #include "threads_interface.h"
 #endif
 
+/** Greatest packet identifier, per MQTT spec */
```
sarenameas, line 64: Do we need periods here?

gordonwang0: The documentation for this version does not put periods at the end of lines.  So for consistency, no.

sarenameas: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/include/types/iot_mqtt_types.h#L1129

gordonwang0: That's v4 though.  v3 does not do that.

PR #859, opened 2020-04-09
Move the source and include variables in CMakeLists.txt to a common file

leegeth:
*Description of changes:*
Move the source and include variables in CMakeLists.txt to a common file for MQTT.

This change will help reuse the sources and include paths in a Cmake based build system in a different repository when using these libraries.


libraries/standard/mqtt/CMakeLists.txt
```diff
@@ -46,41 +37,30 @@ if( ${IOT_BUILD_TESTS} )
 
     # This test library is used to mock MQTT connections.
     add_library( iot_mqtt_mock
-                 test/mock/iot_tests_mqtt_mock.c
+                 ${MQTT_TEST_MOCK_SOURCES}
                  test/mock/iot_tests_mqtt_mock.h
```
gkwicker, line 41: Why do we use a variable for the .c but not the .h on line 41?

libraries/standard/mqtt/CMakeLists.txt
```diff
@@ -46,41 +37,30 @@ if( ${IOT_BUILD_TESTS} )
...
                  ${CONFIG_HEADER_PATH}/iot_config.h )
 
     # Organization of MQTT mock library in folders.
     set_property( TARGET iot_mqtt_mock PROPERTY FOLDER tests )
     source_group( "" FILES
-                  test/mock/iot_tests_mqtt_mock.c
+                  ${MQTT_TEST_MOCK_SOURCES}
```
gkwicker, line 47: Same comment as line 40.

libraries/standard/mqtt/filePaths.cmake
```diff
@@ -0,0 +1,47 @@
...
+
+# MQTT mock source files.
+set( MQTT_TEST_MOCK_SOURCES
+     test/mock/iot_tests_mqtt_mock.c )
+
+# MQTT mock include directories.
+set( MQTT_TEST_MOCK_INCLUDE_PRIVATE_DIRS
+     src )
+set( MQTT_TEST_MOCK_INCLUDE_PUBLIC_DIRS
+     test/mock )
```
gkwicker, line 47: No newline at the end of the file, please add one.

leegeth: Will do thanks.

libraries/standard/mqtt/CMakeLists.txt
leegeth: It is because the header files can be included in the repository by adding them in the include paths. It is already covered by the variable MQTT_TEST_MOCK_INCLUDE_PUBLIC_DIRS. That is the reason why I have not included .h file in the MQTT_TEST_MOCK_SOURCES variable.

PR #864, opened 2020-04-13
Avoid undefined behavior in IotMqtt_Connect

feliperodri:

*Description of changes:*

Update `_IotMqtt_ValidateConnect` function to only accept a nonull payload.

### DIFF
--- a/libraries/standard/mqtt/src/iot_mqtt_validate.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_validate.c
@@ -600,7 +600,8 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
         /* If will info is provided, check that it is valid. */
         if( pConnectInfo->pWillInfo != NULL )
         {
-            status = _IotMqtt_ValidateLwtPublish( pConnectInfo->awsIotMqttMode,
+            status = pConnectInfo->pWillInfo->pPayload != NULL &&
+                    _IotMqtt_ValidateLwtPublish( pConnectInfo->awsIotMqttMode,
                                                   pConnectInfo->pWillInfo );
         }
     }
--- a/libraries/standard/mqtt/src/iot_mqtt_validate.c
+++ b/libraries/standard/mqtt/src/iot_mqtt_validate.c
@@ -600,9 +600,9 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
         /* If will info is provided, check that it is valid. */
         if( pConnectInfo->pWillInfo != NULL )
         {
-            status = ( pConnectInfo->pWillInfo->pPayload != NULL &&
-                       _IotMqtt_ValidateLwtPublish( pConnectInfo->awsIotMqttMode,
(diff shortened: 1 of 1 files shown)


libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -600,7 +600,8 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
         /* If will info is provided, check that it is valid. */
         if( pConnectInfo->pWillInfo != NULL )
         {
-            status = _IotMqtt_ValidateLwtPublish( pConnectInfo->awsIotMqttMode,
+            status = pConnectInfo->pWillInfo->pPayload != NULL &&
+                    _IotMqtt_ValidateLwtPublish( pConnectInfo->awsIotMqttMode,
```
cobusve, line 604: Looks from the description that we are trying to avoid incorrect parameters but we are still makng the Validate publish call even when we think the data is wrong?

Also here it seems the Validate function is supposed to return some result which we should be checking? Are we not doubling up on the check now?

libraries/standard/mqtt/src/iot_mqtt_validate.c
```diff
@@ -600,7 +600,8 @@ bool _IotMqtt_ValidateConnect( const IotMqttConnectInfo_t * pConnectInfo )
         /* If will info is provided, check that it is valid. */
         if( pConnectInfo->pWillInfo != NULL )
```
cobusve, line 601: Why are we not updating the status when pWillInfo == NULL?

feliperodri: We do not update the status if `pWillInfo == NULL`, because it’s valid to have a ConnectInfo with `pWillInfo == NULL`; however, if pWillInfo is nonnull, we must check whether it's valid.

libraries/standard/mqtt/src/iot_mqtt_validate.c
feliperodri: Instead of anding the result from publish into the result, I could: 
```c
if(pConnectInfo->pWillInfo->pPayload != NULL)
{
    status = _IotMqtt_ValidateLwtPublish( pConnectInfo->awsIotMqttMode,
                                          pConnectInfo->pWillInfo );
}
```
We should not update the `_validatePublish`, because pPayload != NULL is specific to ConnectInfo. There are other functions that might accept pPayload == NULL iff payloadLength == 0.

libraries/standard/mqtt/src/iot_mqtt_validate.c
cobusve: Ok so you are saying that if the LWT is not being used it is allowed to be null and we only need to validate it if supplied. That makes sense thanks!

libraries/standard/mqtt/src/iot_mqtt_validate.c
cobusve: Ok makes sense. Can we please just put brackets in that status asignment as it will confuse people.

feliperodri: Done.

PR #869, opened 2020-04-14
Move common types to types.h

gordonwang0:
Moves shared types to interfaces.h


libraries/standard/common/include/types.h
```diff
@@ -0,0 +1,59 @@
+/*
```
sarenameas, line 1: Do we really want a common header that is shared by multiple libraries like this? It looks like an abstraction layer and adds to what needs to be consumed if taking MQTT by itself out of the CSDK.

gordonwang0: I thought we decided not to duplicate the types?  It does make it harder to separate the libraries though.  I'll close this.

sarenameas: I think my concern is more of a discussion point on how we want libraries consumed. I think this opens up some unanswered questions on how libraries are to be consumed. I am not sure if everyone shares what my interpretation. We should talk about it in a team meeting.

abhidixi11: We must move fixed buffer to MQTT specific header where the context would reside.  But I'm fine with sharing the transport interface as long as the signature is the same. Do you see any problem in sharing the transport interface ?

PR #874, opened 2020-04-16
Add state engine

muneebahmed10:
*Description of changes:*
Add functions to update states in the record of QoS 1/2 publishes.

### DIFF
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -24,34 +24,98 @@
 
 #include "mqtt.h"
 
+#define MQTT_STATE_CURSOR_INITIALIZER ( size_t ) 0
+
+/**
+ * @brief Value indicating either send or receive.
+ */
 typedef enum MQTTStateOperation
 {
     MQTT_SEND,
     MQTT_RECEIVE,
 } MQTTStateOperation_t;
 
+/**
+ * @brief Cursor for iterating through state records.
+ */
+typedef size_t MQTTStateCursor_t;
+
+/**
+ * @brief Reserve an entry for an outgoing QoS 1/2 publish.
+ *
+ * @param[in] pMqttContext Initialized MQTT context.
+ * @param[in] packetId The ID of the publish packet.
+ * @param[in] qos 1 or 2.
+ *
+ * @return MQTTSuccess, MQTTNoMemory, or MQTTStateCollision.
+ */
 MQTTStatus_t MQTT_ReserveState( MQTTContext_t * pMqttContext,
                                 uint16_t packetId,
                                 MQTTQoS_t qos );
 
+/**
+ * @brief Calculate the new state for a publish from its qos and operation type.
+ *
+ * @param[in] opType Send or Receive.
+ * @param[in] qos 0, 1, or 2.
+ *
+ * @return The calculated state.
+ */
 MQTTPublishState_t MQTT_CalculateStatePublish( MQTTStateOperation_t opType,
                                                MQTTQoS_t qos );
 
+/**
+ * @brief Update the state record for a PUBLISH packet.
+ *
+ * @param[in] pMqttContext Initialized MQTT context.
+ * @param[in] packetId ID of the PUBLISH packet.
+ * @param[in] opType Send or Receive.
+ * @param[in] qos 0, 1, or 2.
+ *
+ * @return The new state of the publish.
+ */
 MQTTPublishState_t MQTT_UpdateStatePublish( MQTTContext_t * pMqttContext,
                                             uint16_t packetId,
                                             MQTTStateOperation_t opType,
                                             MQTTQoS_t qos );
 
+/**
+ * @brief Calculate the state from a PUBACK, PUBREC, PUBREL, or PUBCOMP.
+ *
+ * @param[in] packetType PUBACK, PUBREC, PUBREL, or PUBCOMP.
+ * @param[in] opType Send or Receive.
+ * @param[in] qos 1 or 2.
+ *
+ * @return The calculated state.
+ */
 MQTTPublishState_t MQTT_CalculateStateAck( MQTTPubAckType_t packetType,
                                            MQTTStateOperation_t opType,
                                            MQTTQoS_t qos );
 
+/**
+ * @brief Update the state record for an ACKed publish.
+ *
+ * @param[in] pMqttContext Initialized MQTT context.
+ * @param[in] packetId ID of the ack packet.
+ * @param[in] packetType PUBACK, PUBREC, PUBREL, or PUBCOMP.
+ * @param[in] opType Send or Receive.
+ *
+ * @return The new state of the publish.
+ */
 MQTTPublishState_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
                                         uint16_t packetId,
                                         MQTTPubAckType_t packetType,
                                         MQTTStateOperation_t opType );
 
+/**
+ * @brief Get the packet ID and index of a publish in a specified state.
+ *
+ * @param[in] pMqttContext Initialized MQTT context.
+ * @param[in] searchState The state to search for.
+ * @param[in,out] pCursor Index at which to start searching.
+ */
 uint16_t MQTT_StateSelect( MQTTContext_t * pMqttContext,
-                           MQTTPublishState_t searchState );
+                           MQTTPublishState_t searchState,
+                           MQTTStateCursor_t * pCursor );
 
 #endif /* ifndef MQTT_STATE_H */
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -40,7 +40,7 @@ MQTTPublishState_t MQTT_CalculateStatePublish( MQTTStateOperation_t opType,
 MQTTPublishState_t MQTT_UpdateStatePublish( MQTTContext_t * pMqttContext,
                                             uint16_t packetId,
                                             MQTTStateOperation_t opType,
-                                            MQTTQoS_t receivedQos );
+                                            MQTTQoS_t receivedQoS );
 
 MQTTPublishState_t MQTT_CalculateStateAck( MQTTPubAckType_t packetType,
                                            MQTTStateOperation_t opType,
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -51,4 +51,7 @@ MQTTPublishState_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
                                         MQTTPubAckType_t packetType,
                                         MQTTStateOperation_t opType );
 
+uint16_t MQTT_StateSelect( MQTTContext_t * pMqttContext,
+                           MQTTPublishState_t searchState );
+
 #endif /* ifndef MQTT_STATE_H */
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -30,9 +30,9 @@ typedef enum MQTTStateOperation
     MQTT_RECEIVE,
 } MQTTStateOperation_t;
 
-MQTTPublishState_t MQTT_ReserveState( MQTTContext_t * pMqttContext,
-                                      uint16_t packetId,
-                                      MQTTQoS_t qos );
+MQTTStatus_t MQTT_ReserveState( MQTTContext_t * pMqttContext,
+                                uint16_t packetId,
+                                MQTTQoS_t qos );
 
 MQTTPublishState_t MQTT_CalculateStatePublish( MQTTStateOperation_t opType,
                                                MQTTQoS_t qos );
@@ -40,7 +40,7 @@ MQTTPublishState_t MQTT_CalculateStatePublish( MQTTStateOperation_t opType,
 MQTTPublishState_t MQTT_UpdateStatePublish( MQTTContext_t * pMqttContext,
                                             uint16_t packetId,
                                             MQTTStateOperation_t opType,
-                                            MQTTQoS_t receivedQoS );
+                                            MQTTQoS_t qos );
 
 MQTTPublishState_t MQTT_CalculateStateAck( MQTTPubAckType_t packetType,
                                            MQTTStateOperation_t opType,
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -1,0 +1,52 @@
+/*
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#ifndef MQTT_STATE_H
+#define MQTT_STATE_H
+
+#include "mqtt.h"
+
+typedef enum MQTTPacketIDState
+{
+    ID_INVALID = 0,
+    ID_UNKNOWN,
+    ID_KNOWN,
+} MQTTPacketIDState_t;
+
+typedef enum MQTTStateOperation
+{
+    MQTT_RESERVE,
+    MQTT_SEND,
+    MQTT_RECEIVE,
+} MQTTStateOperation_t;
+
+MQTTPublishState_t _MQTT_CalculateState( MQTTPublishType_t packetType,
+                                         MQTTStateOperation_t opType,
+                                         MQTTQoS_t qos,
+                                         MQTTPacketIDState_t idStatus );
+
+MQTTStatus_t _MQTT_UpdateState( MQTTContext_t * pMqttContext,
+                                uint16_t packetId,
+                                MQTTPublishType_t packetType,
+                                MQTTStateOperation_t opType,
+                                MQTTQoS_t qos );
+
+#endif /* ifndef MQTT_STATE_H */
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -24,29 +24,31 @@
 
 #include "mqtt.h"
 
-typedef enum MQTTPacketIDState
-{
-    ID_INVALID = 0,
-    ID_UNKNOWN,
-    ID_KNOWN,
-} MQTTPacketIDState_t;
-
 typedef enum MQTTStateOperation
 {
-    MQTT_RESERVE,
     MQTT_SEND,
     MQTT_RECEIVE,
 } MQTTStateOperation_t;
 
-MQTTPublishState_t _MQTT_CalculateState( MQTTPublishType_t packetType,
-                                         MQTTStateOperation_t opType,
-                                         MQTTQoS_t qos,
-                                         MQTTPacketIDState_t idStatus );
+MQTTPublishState_t MQTT_ReserveState( MQTTContext_t * pMqttContext,
+                                      uint16_t packetId,
+                                      MQTTQoS_t qos );
 
-MQTTStatus_t _MQTT_UpdateState( MQTTContext_t * pMqttContext,
-                                uint16_t packetId,
-                                MQTTPublishType_t packetType,
-                                MQTTStateOperation_t opType,
-                                MQTTQoS_t qos );
+MQTTPublishState_t MQTT_CalculateStatePublish( MQTTStateOperation_t opType,
+                                               MQTTQoS_t qos );
+
+MQTTPublishState_t MQTT_UpdateStatePublish( MQTTContext_t * pMqttContext,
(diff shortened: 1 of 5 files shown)


libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+ *
+ * @return `true` if transition is possible, else `false`
+ */
+static bool _validateTransition( MQTTPublishState_t currentState,
+                                 MQTTPublishState_t newState,
+                                 MQTTStateOperation_t opType,
+                                 MQTTQoS_t qos );
+
+/**
+ * @brief Test is a publish is outgoing or incoming.
```
gkwicker, line 90: grammar: "Test __if__ a publish is outgoing or incoming."

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+ *
+ * @return MQTTBadParameter if error; else MQTTSuccess.
+ */
+static MQTTStatus_t _resolveQoS( MQTTQoS_t inputQoS,
+                                 MQTTQoS_t * pFoundQoS );
+
+/**
+ * @brief Find a packet ID in the state record.
+ *
+ * @param[in] records State record array.
```
gkwicker, line 114: The pattern of array+length is so common in our code that we should have a way to represent it without requiring 2 parameters.  This pattern is also extremely error-prone; bugs in which the wrong length gets associated with an array are very easy to write.  C++ addresses this with [`std::span`](https://en.cppreference.com/w/cpp/container/span), we should be able to come up with a simplified C version.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+}
+
+static MQTTStatus_t _updateRecord( MQTTPubAckInfo_t * records,
+                                   size_t recordLen,
+                                   size_t recordIndex,
+                                   uint16_t packetId,
+                                   MQTTQoS_t qos,
+                                   MQTTPublishState_t newState,
+                                   bool shouldStore,
+                                   bool shouldDelete )
```
gkwicker, line 557: Having separate parameters for `shouldStore` and `shouldDelete` means that someone can construct a call to this function which doesn't make sense, e.g. `shouldStore == 1 && shouldDelete == 1`.  Why not create an `enum` for action type, and pass that here?

```c
typedef enum
{
   NO_ACTION = 0,
   SHOULD_STORE = 1,
   SHOULD_DELETE =2
}   MQTTRecordAction_t
```
This would also let you rewrite this logic with a more readable `switch` pattern.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+MQTTPublishState_t _MQTT_CalculateState( MQTTPublishType_t packetType,
+                                         MQTTStateOperation_t opType,
+                                         MQTTQoS_t qos,
+                                         MQTTPacketIDState_t idStatus )
+{
+    MQTTPublishState_t calculatedState = MQTTInvalidState;
+    switch( qos )
+    {
+        case MQTTQoS0:
+            if( idStatus == ID_UNKNOWN && packetType == MQTT_PUBLISH )
```
gkwicker, line 608: Add parentheses around the sub-expressions in this compound statement.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+                if( opType == MQTT_RESERVE )
+                {
+                    calculatedState = MQTTPublishSend;
+                }
+                else /* Sent or received. */
+                {
+                    calculatedState = MQTTPublishDone;
+                }
+            }
+            /* Anything else is an error. */
```
gkwicker, line 619: "error" or "invalid state"?

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+    return calculatedState;
+}
+
+MQTTStatus_t _MQTT_UpdateState( MQTTContext_t * pMqttContext,
+                                uint16_t packetId,
+                                MQTTPublishType_t packetType,
+                                MQTTStateOperation_t opType,
+                                MQTTQoS_t qos )
+{
+    MQTTStatus_t status = MQTTSuccess;
```
gkwicker, line 639: Prefer to initialize this to an invalid/collision state rather than `MQTTSuccess`.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+        case MQTTQoS2:
+            calculatedState = _calculateQoS2( packetType, opType, idStatus );
+            break;
+        default:
+            break;
+    }
+    return calculatedState;
+}
+
+MQTTStatus_t _MQTT_UpdateState( MQTTContext_t * pMqttContext,
```
gkwicker, line 633: This function is at complexity == 11, I'd refactor it to reduce that since you'll have to eventually.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,747 @@
...
+/**
+ * @brief Calculate a state when QoS is 1.
+ *
+ * @param[in] packetType PUBLISH or PUBACK.
+ * @param[in] opType Reserve, Send, or Receive.
+ * @param[in] idStatus Invalid, Unknown, or Known.
+ *
+ * @return The calculated state.
+ */
+static MQTTPublishState_t _calculateQoS1( MQTTPublishType_t packetType,
```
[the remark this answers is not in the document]
muneebahmed10: Changed

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Fixed

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Fixed

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Updated wording

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Right now complexity shows that this function is at 7, but _validateTransition is at 10. I'll refactor that one

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Thank you for the suggestion, and I will update it. Originally this was part of _MQTT_UpdateState(), but was moved out to reduce complexity, hence the two booleans that were meant to be mutually exclusive. I agree an enum would be better here.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,719 @@
...
+                status = MQTTStateCollision;
+            }
+        }
+    }
+    if( status == MQTTSuccess )
+    {
+        if( foundQoS != MQTTQoS0 )
+        {
+            shouldStoreRecord = ( newState == MQTTPublishSend ) ||
+                                ( newState == MQTTPubAckSend ) ||
```
aggarw13, line 689: Is storing a record required for incoming QoS 1 PUBLISH?

aggarw13: Similar comment for the outgoing QoS 1 PUBLISH. If the stored record update is not notified to the user on state transition, then the behavior is the same as just notifying a PUBACK (without maintaining any record).

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,719 @@
...
+ * @return MQTTBadParameter if error; else MQTTSuccess.
+ */
+static MQTTStatus_t _resolveQoS( MQTTQoS_t inputQoS,
+                                 MQTTQoS_t * pFoundQoS );
+
+/**
+ * @brief Find a packet ID in the state record.
+ *
+ * @param[in] records State record array.
+ * @param[in] recordLen Length of record.
```
dan4thewin, line 115: I see recordLen and I think sizeof(MQTTPubAckInfo_t).  How about recordCount (like argc for argv)?

libraries/standard/mqtt/src/mqtt_state.c
dan4thewin: I think the `sizeof(array)/sizeof(array_element)` approach is safe.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,719 @@
...
+#include "mqtt_state.h"
+
+#define MQTT_PACKET_ID_INVALID ( uint16_t ) 0
+
+/**
+ * @brief Set the appropriate state when the packet is one of PUBACK,
+ * PUBREC, PUBREL, or PUBCOMP.
+ * 
+ * @param[in] opType Reserve, Send, or Receive.
+ * @param[in] idStatus Invalid, Unknown, or Known.
```
dan4thewin, line 31: I think there needs to be separation of concerns.  The reserve operation is strictly associated with adding an outbound record, so it need not be present in all these mapping functions.  Likewise, the only legal case for an unknown (barring the trivial reserve) is a receive operation, so again the mapping functions can be simplified further.  Since what I'm suggesting means quite a few changes, let's talk it over together.

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Refactored to prevent this function from doing too many things.

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Split this function into several more specialized functions.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,719 @@
...
+    MQTTStatus_t status = MQTTIllegalState;
+    bool isPublishOutgoing = _isPublishOutgoing( packetType, opType );
+    bool shouldStoreRecord = false;
+    bool shouldDeleteRecord = false;
+    bool isTransitionValid = false;
+    MQTTPacketIDState_t idStatus = ID_INVALID;
+    MQTTPublishState_t currentState = MQTTInvalidState;
+    MQTTPublishState_t newState = MQTTInvalidState;
+    MQTTQoS_t foundQoS = MQTTInvalidQoS;
+    uint16_t recordIndex = MQTT_MAX_QUEUED_PUBLISH_MESSAGES;
```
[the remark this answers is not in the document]
muneebahmed10: Renamed

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Renamed

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,636 @@
...
+ * @param[in] qos 0, 1, or 2.
+ *
+ * @note This function does not validate the current state, or the new state
+ * based on either the operation type or QoS. It assumes the new state is valid
+ * given the opType and QoS, which will be the case if calculated by
+ * MQTT_CalculateStatePublish().
+ *
+ * @return `true` if transition is possible, else `false`
+ */
+static bool _validateTransitionPublish( MQTTPublishState_t currentState,
```
aggarw13, line 41: Use of `bool` type is C89 incompatible. We are not using it in the HTTP implementation

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,636 @@
...
+    return isOutgoing;
+}
+
+static size_t _findInRecord( const MQTTPubAckInfo_t * records,
+                             size_t recordCount,
+                             uint16_t packetId,
+                             MQTTQoS_t * pQos,
+                             MQTTPublishState_t * pCurrentState )
+{
+    size_t index = 0;
```
aggarw13, line 254: Can you add assert checks for input pointer parameters? 
Asserts are a expressing function's assumptions about the input parameters. As these functions are internal to the library, we have control over their calls, thus, the function can carry the assumption that the input parameters are valid (instead of adding error handling logic for invalid parameters )

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: We had previously included stdbool.h in mqtt_lightweight.h. I updated its definition.

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: I was unsure if we wanted to include assert.h since we're not using it anywhere, so I had left it alone. I'll bring it up and will include it if desired.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+    else
+    {
+        records[ recordIndex ].publishState = newState;
+    }
+
+}
+
+MQTTPublishState_t MQTT_ReserveState( MQTTContext_t * pMqttContext,
+                                      uint16_t packetId,
+                                      MQTTQoS_t qos )
```
muneebahmed10, line 324: I realize there's no check for QoS 0. I will update this function for that case.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+static bool _validateTransitionPublish( MQTTPublishState_t currentState,
+                                        MQTTPublishState_t newState,
+                                        MQTTStateOperation_t opType,
+                                        MQTTQoS_t qos )
+{
+    bool isValid = false;
+    switch( currentState )
+    {
+        case MQTTStateNull:
+            /* Transitions from invalid occur when storing a new entry into the record. */
```
dan4thewin, line 128: invalid -> null

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+            if( opType == MQTT_RECEIVE )
+            {
+                isValid = ( newState == MQTTPubAckSend ) || ( newState == MQTTPubRecSend );
+            }
+            break;
+        case MQTTPublishSend:
+            /* Outgoing publish. */
```
dan4thewin, line 135: Perhaps add to the comment that all outgoing publishes start in this state due to the reserve operation.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+        else
+        {
+            newState = MQTTStateNull;
+        }
+        
+    }
+    return newState;
+}
+
+uint16_t MQTT_StateSelect( MQTTContext_t * pMqttContext,
```
dan4thewin, line 547: Seem to be missing the cursor.  You can use an in-out style argument for your recordIndex - just be sure to increment in the success case.

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Updated comment.

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Fixed.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+         * we ignore it. */
+        if( opType == MQTT_RECEIVE )
+        {
+            qos = receivedQoS;
+        }
+        newState = MQTT_CalculateStatePublish( opType, qos );
+        isTransitionValid = _validateTransitionPublish( currentState, newState, opType, qos );
+        if( opType == MQTT_SEND )
+        {
+            if( isTransitionValid && ( recordIndex < MQTT_STATE_ARRAY_MAX_COUNT ) )
```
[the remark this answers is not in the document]
muneebahmed10: Rearranged.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+                                     MQTT_STATE_ARRAY_MAX_COUNT,
+                                     packetId,
+                                     &qos,
+                                     &currentState );
+        /* When the record does not exist, a QoS must be supplied. Otherwise,
+         * we ignore it. */
+        if( opType == MQTT_RECEIVE )
+        {
+            qos = receivedQoS;
+        }
```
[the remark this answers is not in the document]
muneebahmed10: Reordered function

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+        records[ availableIndex ].qos = qos;
+        records[ availableIndex ].publishState = publishState;
+        status = MQTTSuccess;
+    }
+
+    return status;
+}
+
+static void _updateRecord( MQTTPubAckInfo_t * records,
+                           size_t recordCount,
```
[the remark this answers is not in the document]
muneebahmed10: Removed

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,588 @@
...
+        records[ recordIndex ].publishState = MQTTStateNull;
+    }
+    else
+    {
+        records[ recordIndex ].publishState = newState;
+    }
+
+}
+
+MQTTPublishState_t MQTT_ReserveState( MQTTContext_t * pMqttContext,
```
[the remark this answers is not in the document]
muneebahmed10: Changed to `MQTTStatus_t`

libraries/standard/mqtt/include/mqtt_state.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "mqtt.h"
+
+typedef enum MQTTStateOperation
+{
+    MQTT_SEND,
+    MQTT_RECEIVE,
+} MQTTStateOperation_t;
+
+MQTTPublishState_t MQTT_ReserveState( MQTTContext_t * pMqttContext,
```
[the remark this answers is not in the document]
muneebahmed10: Added

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Added cursor argument and type

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,611 @@
...
+    MQTTQoS_t tempQoS = MQTTQoS0;
+    MQTTPublishState_t tempState = MQTTStateNull;
+
+    recordIndex = _findInRecord( pMqttContext->outgoingPublishRecords,
+                                 MQTT_STATE_ARRAY_MAX_COUNT,
+                                 packetId,
+                                 &tempQoS,
+                                 &tempState );
+
+    if( qos == MQTTQoS0 )
```
abhidixi11, line 332: QOS0 records are not added. (assert in _addRecord () ), so should we even check for those. Wondering if the first two conditions can be checked before _findInRecord() to make this function more efficient.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,611 @@
...
+        records = pMqttContext->incomingPublishRecords;
+    }
+
+    recordIndex = _findInRecord( records,
+                                 MQTT_STATE_ARRAY_MAX_COUNT,
+                                 packetId,
+                                 &foundQoS,
+                                 &currentState );
+    
+    if( qos == MQTTQoS0 )
```
abhidixi11, line 409: Same comment as above (line 332).  First two conditions could be checked before _findInRecord. The second condition could be an assert.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -0,0 +1,611 @@
...
+    {
+        /* QoS 0 publish. Do nothing */
+        newState = MQTTPublishDone;
+    }
+    else if( packetId == MQTT_PACKET_ID_INVALID )
+    {
+        /* Publishes > QoS 0 need a valid packet ID. */
+        mqttStatus = MQTTBadParameter;
+    }
+    else if( foundQoS == MQTTQoS0 )
```
abhidixi11, line 419: Is it possible ? Should this be an assert()

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Although these functions are intended to be internal to the library, I didn't want to add asserts here since the functions are not static. Although these functions should not be called for QoS 0, I added this check in case the caller did not perform a check for it. The only difference is this `if` block, all other parts of the function should be the same.

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Yes, it means the ID was not found in the record, so `_findInRecord` was not able to update `foundQoS`.

libraries/standard/mqtt/src/mqtt_state.c
abhidixi11: I'm just wondering if this check should happen before _findInRecord() is called.

libraries/standard/mqtt/src/mqtt_state.c
abhidixi11: Earlier on line 341, it is checked by condition: if( recordIndex >= MQTT_STATE_ARRAY_MAX_COUNT ) ,  if( foundQoS == MQTTQoS0 ) looks bit confusing.

muneebahmed10: They're equivalent for this purpose, since both indicate that the record was not found. I chose to use `foundQoS` since that's the variable that gets updated in the block. Do you think it should be changed?

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Now that it's brought up, I realize the call to `_findInRecord()` could just be removed. Its only purpose is collision detection which `_addRecord()` does anyway.

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Refactored

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: After discussion offline, yes, it's desired to store the record for QoS 1 so that a session can be resumed.

libraries/standard/mqtt/include/mqtt_state.h
```diff
@@ -0,0 +1,52 @@
...
+} MQTTPacketIDState_t;
+
+typedef enum MQTTStateOperation
+{
+    MQTT_RESERVE,
+    MQTT_SEND,
+    MQTT_RECEIVE,
+} MQTTStateOperation_t;
+
+MQTTPublishState_t _MQTT_CalculateState( MQTTPublishType_t packetType,
```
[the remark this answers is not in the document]
muneebahmed10: Removed underscores

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: Rearranged so this check happens before `findInRecord`

PR #884, opened 2020-04-20
Implement and test request header methods for HTTPClient

yourslab:
This PR implements the following request header methods for HTTPClient:

- HTTPClient_InitializeRequestHeaders
- HTTPClient_AddHeader
- HTTPClient_AddRangeHeader

It also contains a unit test for the happy path of HTTPClient_AddHeader using mock4thewin. Note that this will continue to be a draft until all unit tests are completed. I also haven't run the necessary checks yet.


libraries/standard/http/include/http_client.h
```diff
@@ -1,347 +1,434 @@
 #ifndef HTTP_CLIENT_H_
 #define HTTP_CLIENT_H_
 
-#include <stdint.h>
-#include <stdbool.h>
+#include "private/http_client_internal.h"
```
sarenameas, line 4: We shouldn't expose the private header in the public header.

libraries/standard/http/include/http_client.h
```diff
@@ -1,347 +1,434 @@
...
+
+/**
+ * @brief The HTTP protocol version of this library is HTTP/1.1.
+ */
+#define HTTP_PROTOCOL_VERSION              "HTTP/1.1"
+
+/**
+ * @brief Default value when pRequestInfo->pPath == NULL.
+ */
+#define HTTP_EMPTY_PATH                    "/"
```
sarenameas, line 44: I feel like this to line 113 should go to the private header.

libraries/standard/http/include/private/http_client_internal.h
```diff
@@ -0,0 +1,23 @@
...
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "http_client.h"
+
+#define STRLEN_LITERAL( x )    ( ( sizeof( x ) / sizeof( char ) ) - 1 )
+#define HTTP_SUCCEEDED( x )    ( ( x ) != HTTP_SUCCESS )
+#define HTTP_FAILED( x )       ( ( x ) != HTTP_SUCCESS )
+    HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
```
sarenameas, line 15: I think all internal functions should be static and prototyped+defined in the C file.

libraries/standard/http/include/http_client.h
```diff
@@ -1,347 +1,434 @@
...
- * 
+ *
  * @param[in] pRequestHeaders Request header buffer information.
  * @param[in] pRequestInfo Initial request header configurations.
  * @return #HTTP_SUCCESS if successful, an error code otherwise.
  * TODO: Update for exact error codes returned.
  */
-HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pRequestHeaders, 
-                                                  const HTTPRequestInfo_t * pRequestInfo );
+    HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pRequestHeaders,
```
sarenameas, line 420: It is my intention that when someone finishes a function they also expand on the documentation.

libraries/standard/http/src/http_client.c
```diff
@@ -1,42 +1,311 @@
...
+
+    /* Check if header value is long enough for length to overflow. */
+    if( valueLen > ( UINT32_MAX >> 2 ) )
+    {
+        /* TODO: Add log. */
+        status = HTTP_INVALID_PARAMETER;
+    }
+
+    /* "Content-Length" header must not be set by user if
+     * HTTP_REQUEST_DISABLE_CONTENT_LENGTH_FLAG is deactivated. */
```
sarenameas, line 220: How are we going to check for that flag here? Looks like I made an oversight on how to carry that state to other public function.

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,54 @@
 #include "common.h"
 
+/* mock _addHeader */
+HTTPStatus_t status;
+#define _addHeader( a, b, c, d, e )    ( status )
```
sarenameas, line 5: We don't mock library private functions.

libraries/standard/http/src/http_client.c
```diff
@@ -1,42 +1,311 @@
...
 
-HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders, 
-                                       const char* pName, 
-                                       size_t nameLen, 
-                                       const char * pValue, 
-                                       size_t valueLen )
+    return length;
+}
+
+HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
```
sarenameas, line 22: static + docs

libraries/standard/http/src/http_client.c
yourslab: EXACTLY! I realized that we needed pRequestInfo to check for that flag but that would be quite inconvenient for the user.

libraries/standard/http/src/http_client.c
```diff
@@ -1,42 +1,311 @@
...
+    /* Write "bytes=<start>:<end>" for Range header value. */
+    memcpy( rangeValueStr,
+            HTTP_RANGE_BYTES_PREFIX_VALUE, HTTP_RANGE_BYTES_PREFIX_VALUE_LEN );
+    pRangeValueCur += HTTP_RANGE_BYTES_PREFIX_VALUE_LEN;
+    rangeValueStrActualLength += HTTP_RANGE_BYTES_PREFIX_VALUE_LEN;
+    memcpy( rangeValueStr, EQUAL_CHARACTER, EQUAL_CHARACTER_LEN );
+    pRangeValueCur += EQUAL_CHARACTER_LEN;
+    rangeValueStrActualLength += EQUAL_CHARACTER_LEN;
+    memcpy( rangeValueStr, itoa( rangeStart ), itoaLength( rangeStart ) );
+    pRangeValueCur += itoaLength( rangeStart );
```
yourslab, line 281: @sarenameas I was unsure about using this itoaLength private method. Is this the best way to do this consider we can't use sprintf?

libraries/standard/http/src/http_client.c
sarenameas: We could add the flags to the RequestHeaders_t structure.

libraries/standard/http/src/http_client.c
sarenameas: we can use snprintf(), that is allowed and will return the number of bytes written (forgot if this include the null terminator you will have to look at the docs).

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,54 @@
 #include "common.h"
```
sarenameas, line 1: You will have to make new files for the other functions. (you probably already know this)

libraries/standard/http/src/http_client.c
```diff
@@ -1,42 +1,311 @@
...
+        _isNullParam( pRequestInfo->pHost ) ||
+        _isNullParam( pRequestInfo->pPath ) )
+    {
+        status = HTTP_INVALID_PARAMETER;
+    }
+
+    /* Check if buffer can fit "<METHOD> <PATH> HTTP/1.1\r\n". */
+    toAddLen = pRequestInfo->methodLen +                 \
+               SPACE_CHARACTER_LEN +                     \
+               pRequestInfo->pathLen +                   \
```
sarenameas, line 89: If path was null and we added the "/" we will want to account for that? I think if unit tests are complete you will have caught this.

libraries/standard/http/include/http_client.h
```diff
@@ -1,347 +1,371 @@
...
-#define HTTP_METHOD_HEAD                "HEAD" /**< HTTP Method HEAD. */
+#define HTTP_METHOD_GET          "GET"             /**< HTTP Method GET. */
+#define HTTP_METHOD_PUT          "PUT"             /**< HTTP Method PUT. */
+#define HTTP_METHOD_POST         "POST"            /**< HTTP Method POST. */
+#define HTTP_METHOD_HEAD         "HEAD"            /**< HTTP Method HEAD. */
+
+/**
+ * @brief The HTTP protocol version of this library is HTTP/1.1.
+ */
+#define HTTP_PROTOCOL_VERSION    "HTTP/1.1"
```
aggarw13, line 37: This can be in the source file instead

libraries/standard/http/include/http_client.h
```diff
@@ -1,347 +1,371 @@
...
+ *     Connection: close
+ *
+ * Note that Connection header value can be changed to keep-alive by setting
+ * the HTTP_REQUEST_KEEP_ALIVE_FLAG in #HTTPRequestInfo_t.flags.
+ *
  * @param[in] pRequestHeaders Request header buffer information.
  * @param[in] pRequestInfo Initial request header configurations.
  * @return #HTTP_SUCCESS if successful, an error code otherwise.
- * TODO: Update for exact error codes returned.
+ * @return #HTTP_INVALID_PARAMETER if any provided parameters and their members are NULL.
```
aggarw13, line 354: All return text should be within the same `@return` tag. Refer here for an example: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/aws/jobs/include/aws_iot_jobs.h#L146-L150

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -0,0 +1,85 @@
+/* Standard Includes. */
+#include <stdint.h>
+#include <stdbool.h>
```
aggarw13, line 3: We shouldn't have a dependency on `stdbool.h` from our libraries as FreeRTOS codebase suggests against it (Some vendors are not C99 compliant)

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -0,0 +1,85 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#include "http_client.h"
```
aggarw13, line 7: This header shouldn't include the public-facing header file if there is no direct dependency in this file

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -0,0 +1,85 @@
...
+
+
+/**
+ * @brief Constants for header fields added automatically during the request initialization.
+ */
+#define HTTP_USER_AGENT_FIELD                   "User-Agent"
+#define HTTP_HOST_FIELD                         "Host"
+#define HTTP_USER_AGENT_FIELD_LEN               ( 10 )
+#define HTTP_HOST_FIELD_LEN                     ( 5 )
+#define HTTP_CONNECTION_FIELD_LEN               ( 10 )
```
aggarw13, line 48: Please keep the string and its associated length constants together. So, this should be like
```

#define HTTP_USER_AGENT_FIELD        "User-Agent"
#define HTTP_USER_AGENT_FIELD_LEN    ( 10 )
#define HTTP_HOST_FIELD              "Host"
#define HTTP_HOST_FIELD_LEN          ( 5 )
```

aggarw13: The connection length constant should appear in the next section

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -0,0 +1,85 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#include "http_client.h"
+
+/**
+ * @brief Default value when pRequestInfo->pPath == NULL.
+ */
+#define HTTP_EMPTY_PATH                    "/"
+#define HTTP_EMPTY_PATH_LEN                ( 1 )
```
aggarw13, line 13: We use the `sizeof()` utility for length constants to avoid manual errors.
An example is
`#define HTTP_EMPTY_PATH_LEN                (uint8_t) (sizeof(HTTP_EMPTY_PATH) - 1)

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -0,0 +1,85 @@
...
+#define HTTP_RANGE_BYTES_PREFIX_VALUE           "bytes"
+#define HTTP_RANGE_BYTES_PREFIX_VALUE_LEN       ( 5 )
+
+/**
+ * @brief Longest possible string for "Range" header value field.
+ */
+#define HTTP_RANGE_BYTES_VALUE_MAX_LEN          ( 27 )
+
+#define STRLEN_LITERAL( x )    ( ( sizeof( x ) / sizeof( char ) ) - 1 )
+#define HTTP_SUCCEEDED( x )    ( ( x ) == HTTP_SUCCESS )
```
aggarw13, line 84: These macros are not necessarily needed. Keeping the `== HTTP_SUCCESS` logic in the source is still readable. (We want to limit use of function-like macros as much as possible)

libraries/standard/http/src/private/http_client_internal.h
yourslab: So do we use 1 or 0 in place of true or false?

libraries/standard/http/src/private/http_client_internal.h
yourslab: You sir are a life saver

libraries/standard/http/include/http_client.h
yourslab: Oh but user might wanna define different HTTP_PROTOCOL_VERSION?

sarenameas: Users will not be allowed to change the protocol version. This should go into the internal.h header.

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -0,0 +1,74 @@
...
+#define HTTP_RANGE_FIELD_LEN                    ( 5 )
+
+/**
+ * @brief Constants for header value prefix added for "Range" header.
+ */
+#define HTTP_RANGE_BYTES_PREFIX_VALUE           "bytes"
+#define HTTP_RANGE_BYTES_PREFIX_VALUE_LEN       ( 5 )
+
+/**
+ * @brief Longest possible string for "Range" header value field.
```
sarenameas, line 70: Can write briefly about how this number was calculated?

libraries/standard/http/src/http_client.c
```diff
@@ -1,42 +1,362 @@
...
+    pRangeValueCur += itoaLength( rangeStart );
+    rangeValueStrActualLength += itoaLength( rangeStart );
+    memcpy( rangeValueStr, DASH_CHARACTER, DASH_CHARACTER_LEN );
+    pRangeValueCur += DASH_CHARACTER_LEN;
+    rangeValueStrActualLength += DASH_CHARACTER_LEN;
+    bytesWritten = snprintf( rangeValueStr,
+                             HTTP_RANGE_BYTES_VALUE_MAX_LEN - rangeValueStrActualLength,
+                             "%d", rangeStart );
+    pRangeValueCur += bytesWritten;
+    rangeValueStrActualLength += bytesWritten;
```
sarenameas, line 334: I believe lines 317 to 344 can be reduced to: 
```
snprintf( rangeValueStr,
          sizeof("rangeValueStr"),
          "%s=%d-",
          HTTP_RANGE_BYTES_PREFIX_VALUE,
          rangeStart)    
```
Also we may not write the "-" character if the startRange is negative and we write rangeEnd depending on rules here: https://w.amazon.com/bin/view/AWS/Mobile/IoT/Treadstone/HTTPDesign/Redesign/LowLevel/#HReceivingaresponse

So my example will be in some if/else block

libraries/standard/http/test/test-_addHeader.c
```diff
@@ -0,0 +1,50 @@
+#include "common.h"
```
sarenameas, line 1: We should no be testing private functions.

PR #886, opened 2020-04-20
Adds CBMC prefix to all proof-related macros

feliperodri:

*Description of changes:*

To avoid possible conflicts, add CBMC prefix to all proof-related macros. Only a few of them didn't follow this pattern. 


markrtuttle:
Looks good.  I've looked over this and it seems basically fine, but I'm going to need to study it on a bigger screen.  My main concern is concern that the TaskPool and Semaphore methods may have been abstracted differently in different proofs.

cbmc/patches/0001-Bound-packet-remaining-length.patch
```diff
@@ -0,0 +1,28 @@
+From c785aa73d869777a8a0ee4f8ff488397a70e0e29 Mon Sep 17 00:00:00 2001
```
markrtuttle, line 1: Is this patch an addition?  Not just adding a prefix?  Do some proofs change because of this addition?

markrtuttle: Or does this simply replace the original path with a differently named file?

feliperodri: It's not an addition, it simply replace the original path with a differently named file.

PR #891, opened 2020-04-23
Fix Makefile build errors

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Add logging to include path

### DIFF
--- a/demos/include/config.h
+++ b/demos/include/config.h
@@ -29,7 +29,7 @@ typedef int MQTTNetworkContext_t;
 #define MQTT_MAX_QUEUED_PUBLISH_MESSAGES    10
 
 /* Include file for POSIX reference implementation. */
-#include "platform/include/iot_logging.h"
+#include "iot_logging.h"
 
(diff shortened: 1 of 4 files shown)


demos/include/config.h
```diff
@@ -29,7 +29,7 @@ typedef int MQTTNetworkContext_t;
 #define MQTT_MAX_QUEUED_PUBLISH_MESSAGES    10
 
 /* Include file for POSIX reference implementation. */
-#include "platform/include/iot_logging.h"
+#include "iot_logging.h"
```
sarenameas, line 32: Since you are chancing this include path do we need to update:
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/platform/posix/iot_logging.c#L42
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/platform/posix/iot_logging.c#L36

muneebahmed10: It builds without them, but yeah I think that makes sense.

PR #892, opened 2020-04-24
Implement and test HTTPClient_AddHeader method

yourslab:
## Info

This PR contains a complete implementation of HTTPClient_AddHeader() along with a unit test.

## Code Quality
There are no MISRA violations
GNU code complexity found zero violations for -t9.
Building the Tests finds 100% branch coverage:


libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -28,4 +28,84 @@
...
+#define SPACE_CHARACTER_LEN                ( uint8_t ) ( sizeof( SPACE_CHARACTER ) - 1 )
+#define EQUAL_CHARACTER_LEN                ( uint8_t ) ( sizeof( EQUAL_CHARACTER ) - 1 )
+#define DASH_CHARACTER_LEN                 ( uint8_t ) ( sizeof( DASH_CHARACTER ) - 1 )
+
+/**
+ * @brief The maximum length for a 32-bit integer when converted to a string.
+ *
+ * This is used to initialize a local array for the final headers to send.
+ */
+#define INT32_STRING_MAX_LEN               ( uint8_t ) ( sizeof( INT32_STRING_MAX ) - 1 )
```
aggarw13, line 65: `sizeof(INT32_STRING_MAX)` is the size of a 32 bit value i.e. 4 bytes.  This value will resolve to 3. 
I think you want to calculate the number of digits in the max value, whose value is **10**.
You cannot calculate it at compile-time, you will have to just store the value directly.

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -28,4 +28,84 @@
...
+/**
+ * @brief The HTTP protocol version of this library is HTTP/1.1.
+ */
+#define HTTP_PROTOCOL_VERSION              "HTTP/1.1"
+
+/**
+ * @brief Default value when pRequestInfo->pPath == NULL.
+ */
+#define HTTP_EMPTY_PATH                    "/"
+#define HTTP_EMPTY_PATH_LEN                ( uint8_t ) ( sizeof( HTTP_EMPTY_PATH ) - 1 )
```
aggarw13, line 40: You can remove the `(uint8_t)` typecasts from the `sizeof()` calls as that will raise compiler warnings of reducing type width. (I know that I had included the typecast in a different PR's comment, my bad)

libraries/standard/http/src/private/http_client_internal.h
yourslab: Oh my bad. This is what happens when you use regex search and replace lol. You are right, this was originally set to the 10.

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -28,4 +28,84 @@
...
+ */
+#define HTTP_HEADER_LINE_SEPARATOR         "\r\n"
+#define CARRIAGE_RETURN_CHARACTER          '\r'
+#define NEWLINE_CHARACTER                  '\n'
+#define HTTP_HEADER_FIELD_SEPARATOR        ": "
+#define COLON_CHARACTER                    ':'
+#define SPACE_CHARACTER                    ' '
+#define EQUAL_CHARACTER                    '='
+#define DASH_CHARACTER                     '-'
+#define HTTP_HEADER_LINE_SEPARATOR_LEN     ( uint8_t ) ( sizeof( HTTP_HEADER_LINE_SEPARATOR ) - 1 )
```
aggarw13, line 53: Please group the length macros along with the data they represent the length of

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -28,4 +28,84 @@
...
+/**
+ * @brief Constants for header value prefix added for "Range" header.
+ */
+#define HTTP_RANGE_BYTES_PREFIX_VALUE           "bytes"
+#define HTTP_RANGE_BYTES_PREFIX_VALUE_LEN       ( uint8_t ) ( sizeof( HTTP_RANGE_BYTES_PREFIX_VALUE ) - 1 )
+
+/**
+ * @brief Longest possible string for "Range" header value field.
+ */
+#define HTTP_RANGE_BYTES_VALUE_MAX_LEN          ( 27 )
```
aggarw13, line 107: How was this calculated?

libraries/standard/http/src/private/http_client_internal.h
yourslab: No worries! Search and replace to the rescue.

libraries/standard/http/src/http_client.c
```diff
@@ -40,12 +116,128 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
+    {
+        IotLogError( "Parameter check failed: pField is NULL." );
+        returnStatus = HTTP_INVALID_PARAMETER;
+    }
+    else if( ( pValue == NULL ) )
+    {
+        IotLogError( "Parameter check failed: pValue is NULL." );
+        returnStatus = HTTP_INVALID_PARAMETER;
+    }
+    else
```
aggarw13, line 158: You can group all the input-validation if-checks within the same if-else-if construct. 
Thus the checks for the headerLen and valueLen can be continued here,

libraries/standard/http/src/http_client.c
```diff
@@ -40,12 +116,128 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
+    {
+        /* Empty else MISRA 15.7 */
+    }
+
+    /* Check if header field is long enough for length to overflow. */
+    if( fieldLen > ( UINT32_MAX >> 2 ) )
```
aggarw13, line 164: Is the header field size requirement part of the HTTP standard?

yourslab: @sarenameas I just copied this from AFR

libraries/standard/http/src/http_client.c
yourslab: Okay I was thinking we might wanna print both errors but I don't see any point in that now

libraries/standard/http/src/http_client.c
```diff
@@ -40,12 +116,128 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
+        returnStatus = HTTP_INVALID_PARAMETER;
+    }
+    else
+    {
+        /* Empty else MISRA 15.7 */
+    }
+
+    if( returnStatus != HTTP_INVALID_PARAMETER )
+    {
+        /* "Content-Length" header must not be set by user if
```
aggarw13, line 199: Why prevent the user from adding these headers? The users can very well skip the `InitializeRequestHeaders` call, and choose to add them through this function. 
Having this logic couples this function with knowledge of the `InitializeRequestHeader` function, which is not ideal.
Also, having the string comparison operations adds unnecessary performance cost every time a header is added through this function.

libraries/standard/http/src/private/http_client_internal.h
yourslab: HTTP_RANGE_FIELD_LEN + HTTP_HEADER_FIELD_SEPARATOR_LEN +  \
HTTP_RANGE_BYTES_PREFIX_VALUE_LEN + EQUAL_CHARACTER_LEN + \
INT32_STRING_MAX_LEN + DASH_CHARACTER_LEN + INT32_STRING_MAX_LEN;

yourslab: but i'm not working on that anymore so i'll just remove it

libraries/standard/http/src/http_client.c
```diff
@@ -73,7 +265,7 @@ static HTTPStatus_t _sendHttpHeaders( const HTTPTransportInterface_t * pTranspor
     if( transportStatus < 0 )
     {
         IotLogErrorWithArgs( "Error in sending the HTTP headers over the transport "
-                             "interface: Transport status %d.",
+                             "interface: Transport returnStatus %d.",
```
aggarw13: If this log message is not clear, I suggest changing it to:
"Failed to send HTTP headers: Transport send() returned error: TransportStatus=%d" as it clearly mentions the _<Failure Outcome>:<Cause>:<Data>_

libraries/standard/http/src/http_client.c
yourslab: You have a point there. Maybe @sarenameas can provide more insight.

libraries/standard/http/src/http_client.c
yourslab: I just changed that by accident. Added back the original again.

libraries/standard/http/src/http_client.c
sarenameas: Yes let's go with archit's comment then we can remove the flags in the request headers.

libraries/standard/http/src/http_client.c
sarenameas: This is from CBMC testing.

libraries/standard/http/src/http_client.c
sarenameas: yea let's put all in one if/else block

libraries/standard/http/include/http_client.h
```diff
@@ -27,12 +27,12 @@
...
 
 /**
- * Supported HTTP request methods.
+ * @brief Supported HTTP request methods.
  */
-#define HTTP_METHOD_GET     "GET"              /**< HTTP Method GET. */
-#define HTTP_METHOD_PUT     "PUT"              /**< HTTP Method PUT. */
-#define HTTP_METHOD_POST    "POST"             /**< HTTP Method POST. */
-#define HTTP_METHOD_HEAD    "HEAD"             /**< HTTP Method HEAD. */
+#define HTTP_METHOD_GET     "GET"                  /**< HTTP Method GET. */
```
sarenameas, line 32: uncrustify is different on everyone's computer lol!

libraries/standard/http/include/http_client.h
```diff
@@ -178,7 +187,7 @@ typedef enum HTTPStatus
  * @brief Represents header data that will be sent in an HTTP request.
  *
  * The memory for the header data buffer is supplied by the user. Information in
- * the buffer will be filled by calling #HTTPClient_InitializeHeaders.
+ * the buffer will be filled by calling #HTTPClient_InitializeRequestHeaders.
```
sarenameas, line 190: Can you also add "and #HTTPClient_AdddHeader."

libraries/standard/http/include/http_client.h
```diff
@@ -205,6 +214,11 @@ typedef struct HTTPRequestHeaders
      * and #HTTPClient_InitializeRequestHeaders.
      */
     size_t headersLen;
+
+    /**
+     * @brief Flags to activate other request header configurations.
+     */
+    uint32_t flags;
```
sarenameas, line 221: Given we won't couple InitialzieRequestHeaders and Addheader let's take this out.

libraries/standard/http/include/http_client.h
yourslab: I'm using uncrustify 0.67 so I wonder why that is.

libraries/standard/http/include/http_client.h
yourslab: Sure

libraries/standard/http/include/http_client.h
```diff
@@ -381,20 +399,31 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
  * Upon return, pRequestHeaders->headersLen will be updated with the number of
  * bytes written.
  *
- * TODO: Expand documentation.
+ * The maximum value for valueLen and nameLen is ( UINT32_MAX >> 2 ) and also
```
sarenameas, line 402: Let's leave this sentence out per last comment.

libraries/standard/http/include/http_client.h
```diff
@@ -381,20 +399,31 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
+ * cannot be set to 0. Otherwise, #HTTP_INVALID_PARAMETER is returned.
+ *
+ * If the following header fields are set by this method,
```
sarenameas, line 405: This block needs to be deleted per recent change to decouple with HTTPClient_InitializeRequestHeaders().

libraries/standard/http/src/http_client.c
```diff
@@ -31,21 +36,204 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                                const char * pValue,
+                                size_t valueLen );
+
+static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
+                                const char * pField,
+                                size_t fieldLen,
+                                const char * pValue,
+                                size_t valueLen )
+{
+    HTTPStatus_t returnStatus = HTTP_INTERNAL_ERROR;
```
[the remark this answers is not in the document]
yourslab: I initially set it to HTTP_SUCCESS actually, but I saw a comment from Gary in MQTT PR that he prefers to set it to some error code.

libraries/standard/http/include/http_client.h
```diff
@@ -381,20 +399,20 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
  * Upon return, pRequestHeaders->headersLen will be updated with the number of
  * bytes written.
  *
- * TODO: Expand documentation.
- *
```
sarenameas, line 385: Can you add this?

```
Headers are written in the following format:
<pName>: <pfield>\r\n\r\n

The trailing \r\n that denotes the end of the header lines is overwritten, if it already exists in the buffer.
```

sarenameas: Also since you changed name to field can you update the parameter docs as well

libraries/standard/http/include/http_client.h
```diff
@@ -363,13 +373,21 @@ typedef struct HTTPResponse
...
+ * Note that Connection header value can be changed to keep-alive by setting
+ * the HTTP_REQUEST_KEEP_ALIVE_FLAG in #HTTPRequestInfo_t.flags.
  *
  * @param[in] pRequestHeaders Request header buffer information.
  * @param[in] pRequestInfo Initial request header configurations.
- *
- * @return #HTTP_SUCCESS if successful, an error code otherwise.
- * TODO: Update for exact error codes returned.
+ * @return One of the following:
+ * - #HTTP_SUCCESS (If successful, an error code otherwise.)
```
aggarw13, line 388: (Nit) "_an error code otherwise_" is not required as the rest of the status codes call them out

libraries/standard/http/include/http_client.h
```diff
@@ -363,13 +373,21 @@ typedef struct HTTPResponse
...
+ * - #HTTP_INVALID_PARAMETER (If any provided parameters and their members are NULL.)
```
aggarw13, line 389: (Nit): "any provided parameters _or_~~and~~ their members are invalid~~NULL~~"

aggarw13: Same suggestion for the `AddHeader` function doc

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -28,4 +28,70 @@
...
+/**
+ * @brief Constants for header values added based on flags.
+ */
+#define HTTP_CONNECTION_KEEP_ALIVE_VALUE        "keep-alive"
+#define HTTP_CONNECTION_CLOSE_VALUE             "close"
+#define HTTP_CONNECTION_KEEP_ALIVE_VALUE_LEN    ( sizeof( HTTP_CONNECTION_KEEP_ALIVE_VALUE ) - 1 )
+#define HTTP_CONNECTION_CLOSE_VALUE_LEN         ( sizeof( HTTP_CONNECTION_CLOSE_VALUE ) - 1 )
+
+/**
+ * @brief Returns the length of a macro string.
```
sarenameas, line 93: Returns the length of a string literal?

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,88 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    }
+
+    /* Check if there is enough space in buffer for additional header. */
+    toAddLen = fieldLen + HTTP_HEADER_FIELD_SEPARATOR_LEN + valueLen + \
+               HTTP_HEADER_LINE_SEPARATOR_LEN +                        \
+               HTTP_HEADER_LINE_SEPARATOR_LEN;
+
+    if( ( pRequestHeaders->headersLen + toAddLen ) > pRequestHeaders->bufferLen )
+    {
+        IotLogError( "Insufficient memory: Provided buffer size too small " \
```
aggarw13, line 88: Should be
```
"Unable to add header in buffer: Buffer has insufficient memory: RequiredBytes=%d, RemainingBufferSize=%d", toAddLen, (pRequestHeaders->bufferLen - pRequestHeaders->headersLen));
```

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,88 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                     "to add new header." );
+        returnStatus = HTTP_INSUFFICIENT_MEMORY;
+    }
+
+    /* Set header length to original if previously backtracked. */
+    if( ( returnStatus == HTTP_INSUFFICIENT_MEMORY ) && hasTrailingLine )
```
aggarw13, line 94: We avoid using `&&` condition containing unrelated operands (here `returnStatus` and `hasTrainlingLine` because the branch path get hidden (slightly from code expressivity).
Also, please explicitly check as `hasTrailingLine == 1u`, which you can add within the previous if-block as
```
   if( ( pRequestHeaders->headersLen + toAddLen ) <= pRequestHeaders->bufferLen )
    {
        if( hasTrailingLine == 1u )
        {
            pRequestHeaders->headersLen += HTTP_HEADER_LINE_SEPARATOR_LEN;
        }
        else
        {
            /* empty else marker */
        }
    }
    else
    {
        pRequestHeaders->headersLen += HTTP_HEADER_LINE_SEPARATOR_LEN;
        IotLogError( "Insufficient memory: Provided buffer size too small " \
                     "to add new header." );
        returnStatus = HTTP_INSUFFICIENT_MEMORY;
    }

```

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,88 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    {
+        pRequestHeaders->headersLen += HTTP_HEADER_LINE_SEPARATOR_LEN;
```
aggarw13, line 96: Why is the `headersLen` value incremented here?

libraries/standard/http/src/http_client.c
sarenameas: Archit's comment should be:
```
if( ( pRequestHeaders->headersLen + toAddLen ) <= pRequestHeaders->bufferLen )
    {
        /* Write "Field: Value \r\n\r\n" to headers. */
        memcpy( pBufferCur, pField, fieldLen );
        pBufferCur += fieldLen;
        memcpy( pBufferCur, HTTP_HEADER_FIELD_SEPARATOR,
                HTTP_HEADER_FIELD_SEPARATOR_LEN );
        pBufferCur += HTTP_HEADER_FIELD_SEPARATOR_LEN;
        memcpy( pBufferCur, pValue, valueLen );
        pBufferCur += valueLen;
        memcpy( pBufferCur, HTTP_HEADER_LINE_SEPARATOR, HTTP_HEADER_LINE_SEPARATOR_LEN );
        pBufferCur += HTTP_HEADER_LINE_SEPARATOR_LEN;
        memcpy( pBufferCur, HTTP_HEADER_LINE_SEPARATOR, HTTP_HEADER_LINE_SEPARATOR_LEN );
        pRequestHeaders->headersLen += toAddLen;
        returnStatus = HTTP_SUCCESS;
    }
    else
    {
	    /* Restore the length of the trailing line separator. */
        if( hasTrailingLine == 1u )
        {
            pRequestHeaders->headersLen += HTTP_HEADER_LINE_SEPARATOR_LEN;
        }
        IotLogError( "Insufficient memory: Provided buffer size too small " \
                     "to add new header." );
        returnStatus = HTTP_INSUFFICIENT_MEMORY;
    }
```

libraries/standard/http/src/http_client.c
yourslab: If memory is insufficient, we undo the backtrack of trailing \r\n

libraries/standard/http/include/http_client.h
```diff
@@ -399,15 +399,20 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
  * @param[in] pRequestHeaders Request header buffer information.
- * @param[in] pName The header field name to write.
- * @param[in] nameLen The byte length of the header field name.
+ * @param[in] pField The header field name to write.
+ * @param[in] fieldLen The byte length of the header field name.
  * @param[in] pValue The header value to write.
  * @param[in] valueLen The byte length of the header field value.
  *
  * @return One of the following:
  * - #HTTP_SUCCESS (If successful, an error code otherwise.)
```
aggarw13, line 414: (Nit) ~~an error code otherwise~~

libraries/standard/http/src/http_client.c
```diff
@@ -112,6 +100,21 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
     }
+    else
+    {
+        /* Restore the length of the trailing line separator. */
+        if( hasTrailingLine == 1u )
+        {
+            pRequestHeaders->headersLen += HTTP_HEADER_LINE_SEPARATOR_LEN;
+        }
+
+        IotLogErrorWithArgs( "Unable to add header in buffer: " \
```
aggarw13, line 111: You shouldn't need the trailing backslashes.

libraries/standard/http/src/http_client.c
```diff
@@ -65,15 +65,15 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
     /* Backtrack before trailing "\r\n" (HTTP header end) if it's already written.
      * Note that this method also writes trailing "\r\n" before returning. */
     if( strncmp( ( char * ) pBufferCur - ( 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ),
                  "\r\n\r\n", 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ) == 0 )
     {
         /* Set this flag to backtrack in case of HTTP_INSUFFICIENT_MEMORY. */
-        hasTrailingLine = 1;
+        hasTrailingLine = 1u;
         pBufferCur -= HTTP_HEADER_LINE_SEPARATOR_LEN;
         pRequestHeaders->headersLen -= HTTP_HEADER_LINE_SEPARATOR_LEN;
```
aggarw13, line 78: It is risky to change the value of `pRequestHeaders->headersLen` without actually changing the contents of the buffer. I suggest that you account for this overwriting in the happy-path branch that writes to the buffer.

libraries/standard/http/src/http_client.c
```diff
@@ -112,6 +100,21 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
         pRequestHeaders->headersLen += toAddLen;
         returnStatus = HTTP_SUCCESS;
     }
+    else
+    {
+        /* Restore the length of the trailing line separator. */
+        if( hasTrailingLine == 1u )
```
aggarw13, line 106: This update operation wouldn't be required if the `headersLen` update is only dealt within in the happy-path branch above (as I suggested in [this comment](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/892/files/b12faa800ad2f117faf5f2318797173a02a4d31d..4acdde70d7b7409251a1c9fd7d2645915c7b9735#r414869709))

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -28,4 +28,65 @@
...
+#define HTTP_HEADER_FIELD_SEPARATOR_LEN    ( sizeof( HTTP_HEADER_FIELD_SEPARATOR ) - 1 )
+#define COLON_CHARACTER_LEN                ( sizeof( COLON_CHARACTER ) - 1 )
+#define SPACE_CHARACTER_LEN                ( sizeof( SPACE_CHARACTER ) - 1 )
+#define EQUAL_CHARACTER_LEN                ( sizeof( EQUAL_CHARACTER ) - 1 )
+#define DASH_CHARACTER_LEN                 ( sizeof( DASH_CHARACTER ) - 1 )
+
+/**
+ * @brief The maximum length for a 32-bit integer when converted to a string.
+ *
+ * This is used to initialize a local array for the final headers to send.
```
aggarw13, line 63: Please add the INT32_MAX value in the comments for evidence of the number of digits

libraries/standard/http/src/http_client.c
yourslab: It builds

libraries/standard/http/src/private/http_client_internal.h
yourslab: I don't use this value so might as well remote it actually

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,91 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    /* Check if there is enough space in buffer for additional header. */
+    toAddLen = fieldLen + HTTP_HEADER_FIELD_SEPARATOR_LEN + valueLen + \
+               HTTP_HEADER_LINE_SEPARATOR_LEN +                        \
+               HTTP_HEADER_LINE_SEPARATOR_LEN;
+
+    /* If we have enough room for the new header line, then write it to the header buffer. */
+    if( ( pRequestHeaders->headersLen + toAddLen ) <= pRequestHeaders->bufferLen )
+    {
+        /* Write "Field: Value \r\n\r\n" to headers. */
+        memcpy( pBufferCur, pField, fieldLen );
```
aggarw13, line 90: Have you considered using `sprintf` here instead of `memcpy`? (I have checked with Richard that we are fine using `sprintf` for now as we can possibly employ underlying optimization strategies later)

yourslab: I would actually prefer that. @sarenameas?

aggarw13: Nevermind, I don't think that `sprintf` is a candidate for use here as the `fieldLen` and `valueLen` parameters cannot be passed to it, and the user is not required to pass NULL-terminated strings

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,91 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+        pBufferCur -= HTTP_HEADER_LINE_SEPARATOR_LEN;
+    }
+
+    /* Check if there is enough space in buffer for additional header. */
+    toAddLen = fieldLen + HTTP_HEADER_FIELD_SEPARATOR_LEN + valueLen +
+               HTTP_HEADER_LINE_SEPARATOR_LEN +
+               HTTP_HEADER_LINE_SEPARATOR_LEN;
+
+    /* If we have enough room for the new header line, then write it to the header buffer. */
+    if( ( initHeadersLen + toAddLen ) <= pRequestHeaders->bufferLen )
```
sarenameas, line 87: It appears that this initHeadersLen includes the trailing \r\n.  And we could erroneously say there is not enough room.

libraries/standard/http/src/http_client.c
```diff
@@ -75,41 +76,40 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
     /* Check if there is enough space in buffer for additional header. */
-    toAddLen = fieldLen + HTTP_HEADER_FIELD_SEPARATOR_LEN + valueLen + \
-               HTTP_HEADER_LINE_SEPARATOR_LEN +                        \
+    toAddLen = fieldLen + HTTP_HEADER_FIELD_SEPARATOR_LEN + valueLen +
+               HTTP_HEADER_LINE_SEPARATOR_LEN +
                HTTP_HEADER_LINE_SEPARATOR_LEN;
 
     /* If we have enough room for the new header line, then write it to the header buffer. */
-    if( ( pRequestHeaders->headersLen + toAddLen ) <= pRequestHeaders->bufferLen )
+    if( ( initHeadersLen + toAddLen ) <= pRequestHeaders->bufferLen )
```
aggarw13, line 87: Why cache the `pRequestHeaders->headersLen` locally instead of using it here directly?

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,91 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    uint8_t hasTrailingLine = 0u;
+
+    /* Backtrack before trailing "\r\n" (HTTP header end) if it's already written.
+     * Note that this method also writes trailing "\r\n" before returning. */
+    if( strncmp( ( char * ) pBufferCur - ( 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ),
+                 "\r\n\r\n", 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ) == 0 )
+    {
+        /* Set this flag to backtrack in case of HTTP_INSUFFICIENT_MEMORY. */
+        hasTrailingLine = 1u;
+        pBufferCur -= HTTP_HEADER_LINE_SEPARATOR_LEN;
```
sarenameas, line 78: initHeadersLen -= HTTP_HEADER_LINE_SEPARATOR_LEN; ?

I think we need a test where we try to write some headers but we are two characters to large from the current bufferLen. This test should succeed.

libraries/standard/http/src/http_client.c
```diff
@@ -75,41 +76,40 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+               HTTP_HEADER_LINE_SEPARATOR_LEN +
                HTTP_HEADER_LINE_SEPARATOR_LEN;
 
     /* If we have enough room for the new header line, then write it to the header buffer. */
-    if( ( pRequestHeaders->headersLen + toAddLen ) <= pRequestHeaders->bufferLen )
+    if( ( initHeadersLen + toAddLen ) <= pRequestHeaders->bufferLen )
     {
+        if( hasTrailingLine == 1u )
+        {
+            pRequestHeaders->headersLen = initHeadersLen - HTTP_HEADER_LINE_SEPARATOR_LEN;
```
aggarw13, line 91: Line 78 (`   pBufferCur -= HTTP_HEADER_LINE_SEPARATOR_LEN`) can be moved here instead as this is the happy-path branch

libraries/standard/http/src/http_client.c
```diff
@@ -75,41 +76,40 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
-                HTTP_HEADER_FIELD_SEPARATOR_LEN );
-        pBufferCur += HTTP_HEADER_FIELD_SEPARATOR_LEN;
-        memcpy( pBufferCur, pValue, valueLen );
-        pBufferCur += valueLen;
-        memcpy( pBufferCur, HTTP_HEADER_LINE_SEPARATOR, HTTP_HEADER_LINE_SEPARATOR_LEN );
-        pBufferCur += HTTP_HEADER_LINE_SEPARATOR_LEN;
+        pBufferCur += snprintf( ( char * ) pBufferCur,
+                                toAddLen,
+                                "%.*s%s%.*s%s",
+                                ( int ) fieldLen, pField,
```
aggarw13, line 98: Are there MISRA warnings requiring these casts?

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,91 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+
+    /* Check if there is enough space in buffer for additional header. */
+    toAddLen = fieldLen + HTTP_HEADER_FIELD_SEPARATOR_LEN + valueLen +
+               HTTP_HEADER_LINE_SEPARATOR_LEN +
+               HTTP_HEADER_LINE_SEPARATOR_LEN;
+
+    /* If we have enough room for the new header line, then write it to the header buffer. */
+    if( ( initHeadersLen + toAddLen ) <= pRequestHeaders->bufferLen )
+    {
+        if( hasTrailingLine == 1u )
```
sarenameas, line 89: If you decreased the init headers after line 78 you wont need to do this here.

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,91 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
+                                const char * pField,
+                                size_t fieldLen,
+                                const char * pValue,
+                                size_t valueLen )
+{
+    HTTPStatus_t returnStatus = HTTP_SUCCESS;
+    uint8_t * pBufferCur = pRequestHeaders->pBuffer + pRequestHeaders->headersLen;
+    size_t toAddLen = 0;
+    size_t initHeadersLen = pRequestHeaders->headersLen;
```
sarenameas, line 68: I feel like this should be renamed to currHeadersLen

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,91 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                                "%.*s%s%.*s%s",
+                                ( int ) fieldLen, pField,
+                                HTTP_HEADER_FIELD_SEPARATOR,
+                                ( int ) valueLen, pValue,
+                                HTTP_HEADER_LINE_SEPARATOR );
+
+        /* HTTP_HEADER_LINE_SEPARATOR cannot be written above because snprintf
+         * writes an extra null byte at the end. */
+        memcpy( pBufferCur, HTTP_HEADER_LINE_SEPARATOR, HTTP_HEADER_LINE_SEPARATOR_LEN );
+        pRequestHeaders->headersLen += toAddLen;
```
sarenameas, line 106: With my suggestions above you can add the toAddLen to initHeadersLen and then assign it to pRequestHeaders->headersLen.

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,91 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                                size_t valueLen );
+
+/*-----------------------------------------------------------*/
+
+static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
+                                const char * pField,
+                                size_t fieldLen,
+                                const char * pValue,
+                                size_t valueLen )
+{
```
aggarw13, line 64: All the input parameters should be validated. As this is an internal function, we can add asserts here

libraries/standard/http/src/http_client.c
```diff
@@ -29,6 +34,92 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    /* If we have enough room for the new header line, then write it to the header buffer. */
+    if( ( backtrackHeaderLen + toAddLen ) <= pRequestHeaders->bufferLen )
+    {
+        if( hasTrailingLine == 1u )
+        {
+            pRequestHeaders->headersLen = backtrackHeaderLen;
+        }
+
+        /* Write "Field: Value \r\n\r\n" to headers. */
+        pBufferCur += snprintf( ( char * ) pBufferCur,
```
sarenameas, line 96: pField and pValue are user inputs and there could be encoding errors in snprintf. I believe that we should be checking the output of snprintf.

aggarw13: Sounds like a good opportunity for an assert.

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+        memcpy( header.field,                                         \
+                HTTP_TEST_HEADER_FIELD, HTTP_TEST_HEADER_FIELD_LEN ); \
+        header.fieldLen = HTTP_TEST_HEADER_FIELD_LEN;                 \
+        memcpy( header.value,                                         \
+                HTTP_TEST_HEADER_VALUE, HTTP_TEST_HEADER_VALUE_LEN ); \
+        header.valueLen = HTTP_TEST_HEADER_VALUE_LEN;                 \
+    }                                                                 \
+    while( 0 )
+
+#define reset()                                                      \
```
aggarw13, line 65: These macros can be defined outside `main()`.
The `main()` should only contain test code

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+    test_err = HTTPClient_AddHeader( &reqHeaders,
+                                     header.field, header.fieldLen,
+                                     header.value, header.valueLen );
+    ok( strncmp( ( char * ) reqHeaders.pBuffer,
+                 correctHeader, correctHeaderLen ) == 0 );
+    ok( reqHeaders.headersLen == correctHeaderLen );
+    ok( test_err == HTTP_INSUFFICIENT_MEMORY );
+    /* Add extra header with sufficient memory. */
+    reqHeaders.bufferLen = HTTP_TEST_BUFFER_SIZE;
+    correctHeaderLen = HTTP_TEST_SUFFICIENT_HEADER_LEN +                         \
```
aggarw13, line 112: Don't need the back slashes

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+    while( 0 )
+
+    plan( 16 );
+
+    /* Test the happy path. */
+    reset();
+    fillHeaderStructTemplate();
+    /* We add 1 because snprintf() writes a null byte at the end. */
+    snprintf( correctHeader, HTTP_TEST_SUFFICIENT_HEADER_LEN + 1,
+              "%s%s: %s\r\n\r\n",
```
aggarw13, line 85: Can you add a test where the buffer contains data without the terminating `"\r\n"` to traverse the branch when it shouldn't backtrack?

libraries/standard/http/test/test-HTTPClient_AddHeader.c
yourslab: that is not how it is defined in mock4thewin's example

libraries/standard/http/test/test-HTTPClient_AddHeader.c
yourslab: That's already there. Happy path only starts with "GET / HTTP/1.1 \r\n" in the buffer.

libraries/standard/http/src/http_client.c
yourslab: I should have that test in place already. Is this what you meant?
```
    /* Add extra header with insufficient memory. */
    reqHeaders.pBuffer = buffer;
    reqHeaders.bufferLen = HTTP_TEST_BUFFER_SIZE;
    reqHeaders.headersLen = HTTP_TEST_HEADER_REQUEST_LINE_LEN;
    reqHeaders.bufferLen = reqHeaders.headersLen;
    test_err = HTTPClient_AddHeader( &reqHeaders,
                                     header.field, header.fieldLen,
                                     header.value, header.valueLen );
    ok( strncmp( ( char * ) reqHeaders.pBuffer,
                 correctHeader, correctHeaderLen ) == 0 );
    ok( reqHeaders.headersLen == correctHeaderLen );
    ok( test_err == HTTP_INSUFFICIENT_MEMORY );
```

libraries/standard/http/src/http_client.c
abhidixi11: I agree that assert should be to validate parameters for the internal functions, but not all parameters need validation, especially if the caller has already validated them.

PR #894, opened 2020-04-24
Implement and test HTTPClient_InitializeRequestHeaders method

yourslab:
## Info

This PR contains a complete implementation of HTTPClient_InitializeRequestHeaders() along with a unit test.

## Code Quality
There are no MISRA violations
GNU code complexity found zero violations for -t9.
Building the Tests finds 100% branch coverage:

### DIFF
--- a/libraries/standard/http/src/http_client.c
+++ b/libraries/standard/http/src/http_client.c
@@ -260,7 +260,7 @@ static HTTPStatus_t _writeRequestLine( HTTPRequestHeaders_t * pRequestHeaders,
         pBufferCur += HTTP_PROTOCOL_VERSION_LEN;
         memcpy( pBufferCur,
                 HTTP_HEADER_LINE_SEPARATOR, HTTP_HEADER_LINE_SEPARATOR_LEN );
-        pRequestHeaders->headersLen += toAddLen;
+        pRequestHeaders->headersLen = toAddLen;
     }
 
     return returnStatus;
--- a/libraries/standard/http/src/http_client.c
+++ b/libraries/standard/http/src/http_client.c
@@ -255,7 +255,6 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
                                                   const HTTPRequestInfo_t * pRequestInfo )
 {
     HTTPStatus_t returnStatus = HTTP_SUCCESS;
-    uint8_t * pBufferCur = NULL;
 
     /* Check for NULL parameters. */
     if( pRequestHeaders == NULL )
@@ -303,13 +302,13 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
         /* Empty else MISRA 15.7 */
     }
 
-    pBufferCur = pRequestHeaders->pBuffer;
-    pRequestHeaders->headersLen = 0;
-    /* Clear user-provided buffer. */
-    memset( pRequestHeaders->pBuffer, 0, pRequestHeaders->bufferLen );
-
     if( returnStatus == HTTP_SUCCESS )
     {
+        /* Reset application-provided parameters. */
+        pRequestHeaders->headersLen = 0;
+        /* Clear user-provided buffer. */
+        memset( pRequestHeaders->pBuffer, 0, pRequestHeaders->bufferLen );
+
         /* Write "<METHOD> <PATH> HTTP/1.1\r\n" to start the HTTP header. */
         returnStatus = _writeRequestLine( pRequestHeaders,
                                           pRequestInfo->method,
--- a/libraries/standard/http/src/http_client.c
+++ b/libraries/standard/http/src/http_client.c
@@ -108,6 +108,19 @@ static HTTPStatus_t _getFinalResponseStatus( HTTPParsingState_t parsingState,
 static HTTPStatus_t _receiveAndParseHttpResponse( const HTTPTransportInterface_t * pTransport,
                                                   HTTPResponse_t * pResponse );
 
+/**
+ * @brief This method writes the request line (first line) of the HTTP Header
+ * into #HTTPRequestHeaders_t.pBuffer and updates length accordingly.
+ *
+ * @param pRequestHeaders Request header buffer information.
+ * @param pMethod The HTTP request method e.g. "GET", "POST", "PUT", or "HEAD".
+ * @param methodLen The byte length of the request method.
+ * @param pPath The Request-URI to the objects of interest, e.g. "/path/to/item.txt".
+ * @param pathLen The byte length of the request path.
+ *
+ * @return #HTTP_SUCCESS if successful. If there was insufficient memory in the
+ * application buffer, then #HTTP_INSUFFICIENT_MEMORY is returned.
+ */
 static HTTPStatus_t _writeRequestLine( HTTPRequestHeaders_t * pRequestHeaders,
                                        const char * pMethod,
                                        size_t methodLen,
--- a/libraries/standard/http/src/http_client.c
+++ b/libraries/standard/http/src/http_client.c
@@ -108,6 +108,12 @@ static HTTPStatus_t _getFinalResponseStatus( HTTPParsingState_t parsingState,
 static HTTPStatus_t _receiveAndParseHttpResponse( const HTTPTransportInterface_t * pTransport,
                                                   HTTPResponse_t * pResponse );
 
+static HTTPStatus_t _writeRequestLine( HTTPRequestHeaders_t * pRequestHeaders,
+                                       const char * pMethod,
+                                       size_t methodLen,
+                                       const char * pPath,
+                                       size_t pathLen );
+
 /*-----------------------------------------------------------*/
 
 static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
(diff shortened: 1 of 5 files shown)


libraries/standard/http/src/http_client.c
```diff
@@ -29,23 +34,329 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    size_t initHeadersLen = pRequestHeaders->headersLen;
+    uint8_t hasTrailingLine = 0u;
+
+    /* Backtrack before trailing "\r\n" (HTTP header end) if it's already written.
+     * Note that this method also writes trailing "\r\n" before returning. */
+    if( strncmp( ( char * ) pBufferCur - ( 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ),
+                 "\r\n\r\n", 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ) == 0 )
+    {
+        /* Set this flag to backtrack in case of HTTP_INSUFFICIENT_MEMORY. */
+        hasTrailingLine = 1u;
```
dcgaws, line 96: Should pRequestHeaders->headersLen be updated here as well?

libraries/standard/http/src/http_client.c
```diff
@@ -29,23 +34,329 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    {
+        returnStatus = HTTP_INSUFFICIENT_MEMORY;
+    }
+
+    if( returnStatus == HTTP_SUCCESS )
+    {
+        /* Write "<METHOD> <PATH> HTTP/1.1\r\n" to start the HTTP header. */
+        memcpy( pBufferCur, pMethod, methodLen );
+        pBufferCur += methodLen;
+        memcpy( pBufferCur, SPACE_CHARACTER, 1 );
```
dcgaws, line 167: SPACE_CHARACTER_LEN

libraries/standard/http/src/http_client.c
```diff
@@ -29,23 +34,329 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+
+        pBufferCur += SPACE_CHARACTER_LEN;
+
+        /* Use "/" as default value if <PATH> is NULL. */
+        if( ( pPath == NULL ) || ( pathLen == 0 ) )
+        {
+            /* Revise toAddLen to contain <HTTP_EMPTY_PATH_LEN> instead. */
+            toAddLen = ( toAddLen - pathLen ) + HTTP_EMPTY_PATH_LEN;
```
dcgaws, line 175: Since pathLen is 0 in this case, how do you know that the buffer is big enough for the addition of HTTP_EMPTY_PATH?

libraries/standard/http/src/http_client.c
```diff
@@ -29,23 +34,329 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    {
+        /* Empty else MISRA 15.7 */
+    }
+
+    if( returnStatus == HTTP_SUCCESS )
+    {
+        /* Reset application-provided parameters. */
+        pRequestHeaders->headersLen = 0;
+        /* Clear user-provided buffer. */
+        memset( pRequestHeaders->pBuffer, 0, pRequestHeaders->bufferLen );
```
dcgaws, line 252: A redundant write of the entire buffer is potentially a costly operation. Why is this necessary?

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+    HTTPRequestHeaders_t reqHeaders = HTTP_REQUEST_HEADERS_INITIALIZER;
+    HTTPRequestHeaders_t reqHeadersDflt = HTTP_REQUEST_HEADERS_INITIALIZER;
+    HTTPStatus_t test_err = HTTP_NOT_SUPPORTED;
+    uint8_t buffer[ HTTP_TEST_BUFFER_SIZE ] = { 0 };
+    uint8_t smallBuffer[ HTTP_TEST_SUFFICIENT_HEADER_LEN - 1 ] = { 0 };
+    char correctHeader[ HTTP_TEST_BUFFER_SIZE ] = { 0 };
+    size_t correctHeaderLen = HTTP_TEST_SUFFICIENT_HEADER_LEN;
+    struct Header
+    {
+        char field[ 100 ];
```
dcgaws, line 45: Can you please replace the magic numbers with defines?

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+        memset( buffer, 0, HTTP_TEST_BUFFER_SIZE );                  \
+        memset( correctHeader, 0, HTTP_TEST_SUFFICIENT_HEADER_LEN ); \
+        memset( header.field, 0, 100 );                              \
+        header.fieldLen = 0;                                         \
+        memset( header.value, 0, 100 );                              \
+        header.valueLen = 0;                                         \
+    }                                                                \
+    while( 0 )
+
+    plan( 16 );
```
dcgaws, line 78: What does this do?

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+
+    /* Test the happy path. */
+    reset();
+    fillHeaderStructTemplate();
+    /* We add 1 because snprintf() writes a null byte at the end. */
+    snprintf( correctHeader, HTTP_TEST_SUFFICIENT_HEADER_LEN + 1,
```
dcgaws, line 84: This call doesn't seem to be guarding against the actual length of correctHeader. Also, you're not capturing the return value, so you don't know how much was written.

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+              "%s%s: %s\r\n\r\n",
+              HTTP_TEST_HEADER_REQUEST_LINE,
+              HTTP_TEST_HEADER_FIELD, HTTP_TEST_HEADER_VALUE );
+    correctHeaderLen = HTTP_TEST_SUFFICIENT_HEADER_LEN;
+    /* Set parameters for reqHeaders. */
+    memcpy( buffer, HTTP_TEST_HEADER_REQUEST_LINE, HTTP_TEST_HEADER_REQUEST_LINE_LEN );
```
dcgaws, line 90: How do you know that buffer is big enough? It's declared independently of the defines that determine what's written into it.

libraries/standard/http/test/test-HTTPClient_AddHeader.c
```diff
@@ -1,9 +1,182 @@
...
+    reqHeaders.pBuffer = buffer;
+    reqHeaders.bufferLen = HTTP_TEST_BUFFER_SIZE;
+    reqHeaders.headersLen = HTTP_TEST_HEADER_REQUEST_LINE_LEN;
+    test_err = HTTPClient_AddHeader( &reqHeaders,
+                                     header.field, header.fieldLen,
+                                     header.value, header.valueLen );
+    ok( strncmp( ( char * ) reqHeaders.pBuffer,
```
dcgaws, line 97: These macros are hard to follow.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
```diff
@@ -0,0 +1,189 @@
+#include <string.h>
+
+#include "common.h"
+
+/* Functions are pulled out into their own C files to be tested as a unit. */
+#include "_writeRequestLine.c"
```
dcgaws, line 6: Will all embedded compilers allow this?

libraries/standard/http/src/http_client.c
```diff
@@ -29,23 +34,329 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+{
+    HTTPStatus_t returnStatus = HTTP_SUCCESS;
+    uint8_t * pBufferCur = pRequestHeaders->pBuffer + pRequestHeaders->headersLen;
+    size_t toAddLen = 0;
+    size_t initHeadersLen = pRequestHeaders->headersLen;
+    uint8_t hasTrailingLine = 0u;
+
+    /* Backtrack before trailing "\r\n" (HTTP header end) if it's already written.
+     * Note that this method also writes trailing "\r\n" before returning. */
+    if( strncmp( ( char * ) pBufferCur - ( 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ),
```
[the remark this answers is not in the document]
yourslab: Nice catch :)

libraries/standard/http/src/http_client.c
```diff
@@ -29,23 +34,329 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                 "\r\n\r\n", 2 * HTTP_HEADER_LINE_SEPARATOR_LEN ) == 0 )
+    {
+        /* Set this flag to backtrack in case of HTTP_INSUFFICIENT_MEMORY. */
+        hasTrailingLine = 1u;
+        pBufferCur -= HTTP_HEADER_LINE_SEPARATOR_LEN;
```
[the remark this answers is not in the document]
yourslab: No, a single "\r\n" denotes end of line but two of them denotes end of header. So we just want to go back two spaces to insert a new line then correctly write a new end of header at the end.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
sarenameas: We already #include c source files in the current MQTT tests that run on all of our platforms.

libraries/standard/http/src/http_client.c
```diff
@@ -132,10 +151,171 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+
+    if( ( toAddLen + pRequestHeaders->headersLen ) > pRequestHeaders->bufferLen )
+    {
+        returnStatus = HTTP_INSUFFICIENT_MEMORY;
+    }
+
+    if( returnStatus == HTTP_SUCCESS )
+    {
+        /* Write "<METHOD> <PATH> HTTP/1.1\r\n" to start the HTTP header. */
+        memcpy( pBufferCur, pMethod, methodLen );
```
sarenameas, line 183: I suppose for consistency we will want to use snprintf for this as we had in other places. 

But I am OK with memcpy. There are tradeoffs to snprintf (such as encoding errors), while memcpy will copy as it is.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
```diff
@@ -0,0 +1,199 @@
...
+    test_err = HTTPClient_InitializeRequestHeaders( &reqHeaders, &reqInfo );
+    ok( strncmp( ( char * ) reqHeaders.pBuffer,
+                 expectedHeader, expectedHeaderLen ) == 0 );
+    ok( reqHeaders.headersLen == expectedHeaderLen );
+    ok( test_err == HTTP_SUCCESS );
+    reset();
+
+    /* -----------------------------------------------------------------------*/
+
+    return grade();
```
sarenameas, line 198: We are going to need tests for the other error codes for coverage.

libraries/standard/http/src/http_client.c
```diff
@@ -132,10 +151,171 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+
+        memcpy( pBufferCur, SPACE_CHARACTER, SPACE_CHARACTER_LEN );
+        pBufferCur += SPACE_CHARACTER_LEN;
+
+        memcpy( pBufferCur,
+                HTTP_PROTOCOL_VERSION, HTTP_PROTOCOL_VERSION_LEN );
+        pBufferCur += HTTP_PROTOCOL_VERSION_LEN;
+        memcpy( pBufferCur,
+                HTTP_HEADER_LINE_SEPARATOR, HTTP_HEADER_LINE_SEPARATOR_LEN );
+        pRequestHeaders->headersLen += toAddLen;
```
sarenameas, line 209: I think headersLen should be replaced with the length of this Request-Line. It makes sense only for it to be at the start of the headers.

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -48,8 +48,11 @@
 #define HTTP_HEADER_END_INDICATOR               "\r\n\r\n"
 #define HTTP_HEADER_END_INDICATOR_LEN           ( sizeof( HTTP_HEADER_END_INDICATOR ) - 1 )
 #define HTTP_HEADER_ADD_FORMAT                  "%.*s" HTTP_HEADER_FIELD_SEPARATOR "%.*s" HTTP_HEADER_LINE_SEPARATOR
-#define CARRIAGE_RETURN_CHARACTER               '\r'
-#define NEWLINE_CHARACTER                       '\n'
+#define HTTP_REQUEST_LINE_FORMAT                "%.*s %.*s %.*s"
```
sarenameas, line 51: We are not using this anywhere so delete.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
sarenameas: We are going to convert these tests to the cmock framework and unity after CMAKE is in. For now we are writing tests in what is available. It is an easy task to convert our tests.

dcgaws: The OTA library implementation did this (i.e. include C modules) at some point and we ran into problems, both with subsequent ports and with subsequent code-review visibility. I think we should find a different approach.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
```diff
@@ -0,0 +1,220 @@
...
+    HTTPRequestInfo_t reqInfo = HTTP_REQUEST_INFO_INITIALIZER;
+    HTTPRequestInfo_t reqInfoDflt = HTTP_REQUEST_INFO_INITIALIZER;
+    HTTPStatus_t test_err = HTTP_NOT_SUPPORTED;
+    uint8_t buffer[ HTTP_TEST_BUFFER_SIZE ] = { 0 };
+    char expectedHeader[ HTTP_TEST_BUFFER_SIZE ] = { 0 };
+    size_t expectedHeaderLen = HTTP_TEST_MAX_HEADER_LEN;
+
+/* Write template reqInfo to pass as parameter to
+ * HTTPClient_InitializeRequestHeaders() method. */
+#define fillReqInfoTemplate()                             \
```
dcgaws, line 75: These complex inline macros, defined within a local function scope, are atypical and tough to follow. What's the rationale?

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
```diff
@@ -0,0 +1,220 @@
...
+    do {                                                           \
+        test_err = HTTP_NOT_SUPPORTED;                             \
+        reqHeaders = reqHeadersDflt;                               \
+        reqInfo = reqInfoDflt;                                     \
+        memset( buffer, 0, HTTP_TEST_BUFFER_SIZE );                \
+        memset( expectedHeader, 0, HTTP_TEST_MAX_HEADER_LEN + 1 ); \
+    }                                                              \
+    while( 0 )
+
+    plan( 22 );
```
dcgaws, line 97: What does this do?

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
```diff
@@ -0,0 +1,220 @@
...
+
+    /* Happy Path testing. */
+    reset();
```
dcgaws, line 100: What does this do?

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
```diff
@@ -0,0 +1,220 @@
...
+    expectedHeaderLen = HTTP_TEST_PREFIX_HEADER_LEN + \
+                        HTTP_CONNECTION_CLOSE_VALUE_LEN;
+    ok( snprintf( expectedHeader, expectedHeaderLen + 1,
```
dcgaws, line 103: Are we not using Unity framework? This appears to be a different set of macros.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
sarenameas: We are going to convert these in the future. These WILL be gone. For now we are using https://github.com/dan4thewin/mock4thewin for unit testing.

These includes are only for Dan's testing framework which builds each function separately with their dependent private function for mocking. We do not have Unity in the development branch yet and moving these tests is a separate activity which will be easy to do.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
sarenameas: When we move to Unity this test setup will move to a TEST_SETUP(){} macro area. This is to setup the shared variable for each of the tests separated by `/* -----------------------------------------------------------------------*/`.

This method comes from Dan's testing framework: https://gist.github.com/dan4thewin/6f708bf635f6cb647d9f7bc7f55e4706#the-complete-unit-test

libraries/standard/http/src/private/http_client_internal.h
cobusve: Are you referring to the HTTP_REQUEST_LINE_FORMAT? If you are then yes I agree that unused definitions are code rot that needs to be removed?

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
cobusve: I think we can resolve this for now as it will be corrected as part of the conversion. The issues we ran into in the past was because we included a parameterized file which the name was passed into the compiler on the command line, that is not supported by all, but what is used here is valid C according to the standard on all C compilers (as stated at  section 6.10.2 of the C standard, it is entitled "Source file inclusion" in the spec,  to be 100% clear the description in the standard states "A #include directive shall identify a header or source file that can be processed by the
implementation)" and so forth. 

Let's resolve this one for now.

libraries/standard/http/test/test-HTTPClient_InitializeRequestHeaders.c
cobusve: This is one of the reasons we decided against this method moving forward, CMock does not require you to do these and as such this will be resolved as part of the conversion. For now this is fine as it is in the tets only and not in the library code...

PR #898, opened 2020-04-28
Incremental HTTPClient_Send(): receive functionality only.

sarenameas:
Implemented temporary unit tests. These unit tests will be moved to the unity framework when it is setup. These are used to verify logic in the current PR.

Parsing of the response is not implemented yet. This PR implements only receiving the response from the network and handling when the parser finds no response, an incomplete response, and a complete response.

image

The last PR brought in the **send over transport** functionality only, but there have been fixes to logging and parameter checking.

### DIFF
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -84,8 +84,8 @@ typedef struct HTTPNetworkContext HTTPNetworkContext_t;
 /**
  * @brief Transport interface for sending data over the network.
  *
- * If the number of bytes written returned is not equal to bytesToWrite, then
- * #HTTPClient_Send will return HTTP_NETWORK_ERROR. If a negative value is
+ * If the number of bytes written returned is not equal to @p bytesToWrite, then
+ * #HTTPClient_Send will return #HTTP_NETWORK_ERROR. If a negative value is
  * returned then this #HTTPClient_Send will also return #HTTP_NETWORK_ERROR.
  *
  * @param[in] context User defined context.
@@ -101,14 +101,14 @@ typedef int32_t (* HTTPTransportSend_t )( HTTPNetworkContext_t * pContext,
 /**
  * @brief Transport interface for reading data on the network.
  *
- * This function will read up to bytesToRead amount of data from the network.
+ * This function will read up to @p bytesToRead amount of data from the network.
  *
  * If this function returns a value less than zero, then #HTTPClient_Send will
  * return #HTTP_NETWORK_ERROR.
  *
  * If this function returns less than the bytesToRead and greater than zero,
- * then this function will be invoked again if the data in pBuffer contains a
- * partial HTTP response message and there is room left in the pBuffer.
+ * then this function will be invoked again if the data in @p pBuffer contains a
+ * partial HTTP response message and there is room left in the @p pBuffer.
  * Repeated invocations will stop if this function returns zero.
  *
  * @param[in] context User defined context.
@@ -171,9 +171,6 @@ typedef enum HTTPStatus
     /**
      * @brief Part of the HTTP response was received from the network.
      *
-     * This can occur only if #HTTPTransportRecv_t returns zero before the full
-     * response message was received.
-     *
      * Functions that may return this value:
      * - #HTTPClient_Send
      */
@@ -182,8 +179,8 @@ typedef enum HTTPStatus
     /**
      * @brief No HTTP response was received from the network.
      *
-     * This can occur only if #HTTPTransportRecv_t returns zero when first
-     * invoked.
+     * This can occur only if there was no data received from the transport
+     * interface.
      *
      * Functions that may return this value:
      * - #HTTPClient_Send
@@ -484,7 +481,8 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * more information.
  * @param[in] pRequestHeaders Request configuration containing the buffer of
  * headers to send.
- * @param[in] pRequestBodyBuf Request entity body. This can be NULL.
+ * @param[in] pRequestBodyBuf Optional Request entity body. Set to NULL if there
+ * is no request body.
  * @param[in] reqBodyBufLen The length of the request entity in bytes.
  * @param[in] pResponse The response message and some notable response
  * parameters will be returned here on success.
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -108,7 +108,7 @@ typedef int32_t (* HTTPTransportSend_t )( HTTPNetworkContext_t * pContext,
  *
  * If this function returns less than the bytesToRead and greater than zero,
  * then this function will be invoked again if the data in pBuffer contains a
- * partial HTTP response message and there there is room left in the pBuffer.
+ * partial HTTP response message and there is room left in the pBuffer.
  * Repeated invocations will stop if this function returns zero.
  *
  * @param[in] context User defined context.
@@ -191,7 +191,7 @@ typedef enum HTTPStatus
     HTTP_NO_RESPONSE,
 
     /**
-     * @brief The application buffer was not large enough for HTTP request
+     * @brief The application buffer was not large enough for the HTTP request
      * headers or the HTTP response message.
      *
      * Functions that may return this value:
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -471,7 +471,7 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * parameter pRequestBodyBuf over the transport. The response is received in
  * #HTTPResponse_t.
  *
- * The application should close the connection with the server if any 
+ * The application should close the connection with the server if any
  * HTTP_SECURITY_ALERT_X errors are returned.
  * TODO: List all the security alerts possible after parsing development.
  *
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -471,9 +471,9 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * parameter pRequestBodyBuf over the transport. The response is received in
  * #HTTPResponse_t.
  *
- * The application should close the connection with the server if any
- * HTTP_SECURITY_ALERT_X errors are returned. Any error found in parsing is
- * considered a malformed response and therefore a security alert.
+ * The application should close the connection with the server if any 
+ * HTTP_SECURITY_ALERT_X errors are returned.
+ * TODO: List all the security alerts possible after parsing development.
  *
  * TODO: Expand documentation.
  *
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -488,12 +488,12 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * parameters will be returned here on success.
  *
  * @return One of the following:
- * - #HTTP_SUCCESS
- * - #HTTP_INVALID_PARAMETER
- * - #HTTP_NETWORK_ERROR
- * - #HTTP_PARTIAL_RESPONSE
- * - #HTTP_NO_RESPONSE
- * - #HTTP_INSUFFICIENT_MEMORY
+ * - #HTTP_SUCCESS (If successful.)
+ * - #HTTP_INVALID_PARAMETER (If any provided parameters or their members are invalid.)
+ * - #HTTP_NETWORK_ERROR (Errors in sending or receiving over the transport interface.)
+ * - #HTTP_PARTIAL_RESPONSE (Part of an HTTP response was received in a partially filled response buffer.)
+ * - #HTTP_NO_RESPONSE (No data was received from the transport interface.)
+ * - #HTTP_INSUFFICIENT_MEMORY (The response received could not fit into the response buffer.)
  * TODO: Add more errors for parsing implementation.
  */
 HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -84,7 +84,7 @@ typedef struct HTTPNetworkContext HTTPNetworkContext_t;
 /**
  * @brief Transport interface for sending data over the network.
  *
- * If the number of bytes written returned is less than bytesToWrite, then
+ * If the number of bytes written returned is not equal to bytesToWrite, then
  * #HTTPClient_Send will return HTTP_NETWORK_ERROR. If a negative value is
  * returned then this #HTTPClient_Send will also return #HTTP_NETWORK_ERROR.
  *
@@ -102,10 +102,14 @@ typedef int32_t (* HTTPTransportSend_t )( HTTPNetworkContext_t * pContext,
  * @brief Transport interface for reading data on the network.
  *
  * This function will read up to bytesToRead amount of data from the network.
+ *
  * If this function returns a value less than zero, then #HTTPClient_Send will
- * return #HTTP_NETWORK_ERROR. If this function returns less than the
- * bytesToRead and greater than zero, then this function will be invoked again
- * if the data in pBuffer contains a partial HTTP response message.
+ * return #HTTP_NETWORK_ERROR.
+ *
+ * If this function returns less than the bytesToRead and greater than zero,
+ * then this function will be invoked again if the data in pBuffer contains a
+ * partial HTTP response message and there there is room left in the pBuffer.
+ * Repeated invocations will stop if this function returns zero.
  *
  * @param[in] context User defined context.
  * @param[in] pBuffer Buffer to read network data into.
@@ -122,9 +126,9 @@ typedef int32_t (* HTTPTransportRecv_t )( HTTPNetworkContext_t * pContext,
  */
 typedef struct HTTPTransportInterface
 {
-    HTTPTransportRecv_t recv;
-    HTTPTransportSend_t send;
-    HTTPNetworkContext_t * pContext;
+    HTTPTransportRecv_t recv;        /**< Transport receive interface */
+    HTTPTransportSend_t send;        /**< Transport interface send interface. */
+    HTTPNetworkContext_t * pContext; /**< User defined transport interface context. */
 } HTTPTransportInterface_t;
 
 /**
@@ -142,7 +146,7 @@ typedef enum HTTPStatus
      * - #HTTPClient_Send
      * - #HTTPClient_ReadHeader
      */
-    HTTP_SUCCESS = 0,
+    HTTP_SUCCESS,
 
     /**
      * @brief The HTTP Client library function input an invalid parameter.
@@ -164,23 +168,47 @@ typedef enum HTTPStatus
      */
     HTTP_NETWORK_ERROR,
 
-    HTTP_NOT_SUPPORTED,
+    /**
+     * @brief Part of the HTTP response was received from the network.
+     *
+     * This can occur only if #HTTPTransportRecv_t returns zero before the full
+     * response message was received.
+     *
+     * Functions that may return this value:
+     * - #HTTPClient_Send
+     */
     HTTP_PARTIAL_RESPONSE,
 
     /**
-     * @brief The application buffer was not large enough for HTTP headers or body.
+     * @brief No HTTP response was received from the network.
+     *
+     * This can occur only if #HTTPTransportRecv_t returns zero when first
+     * invoked.
+     *
+     * Functions that may return this value:
+     * - #HTTPClient_Send
+     */
+    HTTP_NO_RESPONSE,
+
+    /**
+     * @brief The application buffer was not large enough for HTTP request
+     * headers or the HTTP response message.
      *
      * Functions that may return this value:
      * - #HTTPClient_InitializeRequestHeaders
      * - #HTTPClient_AddHeader
      * - #HTTPClient_AddRangeHeader
+     * - #HTTPClient_Send
      */
     HTTP_INSUFFICIENT_MEMORY,
+
     HTTP_INTERNAL_ERROR,
     HTTP_SECURITY_ALERT_RESPONSE_HEADERS_SIZE_LIMIT_EXCEEDED,
     HTTP_SECURITY_ALERT_PARSER_INVALID_CHARACTER,
     HTTP_SECURITY_ALERT_INVALID_CONTENT_LENGTH,
     /* TODO: Add return codes as implementation continues. */
+    /* Temporary error code while implementation is in progress. */
+    HTTP_NOT_SUPPORTED,
 } HTTPStatus_t;
 
 /**
@@ -446,19 +474,29 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * parameter pRequestBodyBuf over the transport. The response is received in
  * #HTTPResponse_t.
  *
+ * The application should close the connection with the server if any
+ * HTTP_SECURITY_ALERT_X errors are returned. Any error found in parsing is
+ * considered a malformed response and therefore a security alert.
+ *
  * TODO: Expand documentation.
  *
  * @param[in] pTransport Transport interface, see #HTTPTransportInterface_t for
  * more information.
  * @param[in] pRequestHeaders Request configuration containing the buffer of
  * headers to send.
- * @param[in] pRequestBodyBuf Request entity body.
+ * @param[in] pRequestBodyBuf Request entity body. This can be NULL.
  * @param[in] reqBodyBufLen The length of the request entity in bytes.
  * @param[in] pResponse The response message and some notable response
  * parameters will be returned here on success.
  *
- * @return #HTTP_SUCCESS if successful, an error code otherwise.
- * TODO: Update for exact error codes returned.
+ * @return One of the following:
+ * - #HTTP_SUCCESS
+ * - #HTTP_INVALID_PARAMETER
+ * - #HTTP_NETWORK_ERROR
+ * - #HTTP_PARTIAL_RESPONSE
+ * - #HTTP_NO_RESPONSE
+ * - #HTTP_INSUFFICIENT_MEMORY
+ * TODO: Add more errors for parsing implementation.
  */
 HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
                               const HTTPRequestHeaders_t * pRequestHeaders,
--- a/libraries/standard/http/src/http_client.c
+++ b/libraries/standard/http/src/http_client.c
@@ -1,18 +1,19 @@
 #include <assert.h>
 #include <string.h>
-#include <stdlib.h>
 #include <stdio.h>
 
 #include "http_client.h"
 #include "private/http_client_internal.h"
+#include "private/http_client_parse.h"
 
 /*-----------------------------------------------------------*/
 
 /**
  * @brief Send the HTTP headers over the transport send interface.
  *
- * @param pTransport Transport interface.
- * @param pRequestHeaders Request headers to send, it includes the buffer and length.
+ * @param[in] pTransport Transport interface.
+ * @param[in] pRequestHeaders Request headers to send, it includes the buffer
+ * and length.
  *
  * @return #HTTP_SUCCESS if successful. If there was a network error or less
  * bytes than what was specified were sent, then #HTTP_NETWORK_ERROR is returned.
@@ -23,9 +24,9 @@ static HTTPStatus_t _sendHttpHeaders( const HTTPTransportInterface_t * pTranspor
 /**
  * @brief Send the HTTP body over the transport send interface.
  *
- * @param pTransport Transport interface.
- * @param pRequestBodyBuf Request body buffer.
- * @param reqBodyLen Length of the request body buffer.
+ * @param[in] pTransport Transport interface.
+ * @param[in] pRequestBodyBuf Request body buffer.
+ * @param[in] reqBodyLen Length of the request body buffer.
  *
  * @return #HTTP_SUCCESS if successful. If there was a network error or less
  * bytes than what was specified were sent, then #HTTP_NETWORK_ERROR is returned.
@@ -53,6 +54,58 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
                                 size_t fieldLen,
                                 const char * pValue,
                                 size_t valueLen );
+
+/**
+ * @brief Receive HTTP response from the transport recv interface.
+ *
+ * @param[in] pTransport Transport interface.
+ * @param[in] pResponse Response buffer.
+ * @param[in] bufferLen Length of the response buffer.
+ *
+ * @return Returns #HTTP_SUCCESS if successful. If there was a network error or
+ * more bytes than what was specified were read, then #HTTP_NETWORK_ERROR is
+ * returned.
+ */
+HTTPStatus_t _receiveHttpResponse( const HTTPTransportInterface_t * pTransport,
+                                   uint8_t * pBuffer,
+                                   size_t bufferLen,
+                                   size_t * pBytesReceived );
+
+/**
+ * @brief Get the status of the HTTP response given the parsing state and how
+ * much data is in the response buffer.
+ *
+ * @param[in] parsingState State of the parsing on the HTTP response.
+ * @param[in] totalReceived The amount of network data received in the response
+ * buffer.
+ * @param[in] responseBufferLen The length of the response buffer.
+ *
+ * @return Returns #HTTP_SUCCESS if parsing state denotes it is complete. If the
+ * parsing state denotes it never started, then return #HTTP_NO_RESPONSE. If the
+ * parsing state is incomplete, thenif the response buffer is not full return
+ * #HTTP_PARTIAL_RESPONSE. If the parsing state is incomplete, then if the
+ * response buffer is full return #HTTP_INSUFFICIENT_MEMORY.
+ */
+static HTTPStatus_t _getFinalResponseStatus( HTTPParsingState_t parsingState,
+                                             size_t totalReceived,
+                                             size_t responseBufferLen );
+
+/**
+ * @brief Receive the HTTP response from the network and parse it.
+ *
+ * @param[in] pTransport Transport interface.
+ * @param[in] pResponse Response message to receive data from the network.
+ *
+ * @return Returns #HTTP_SUCCESS if successful. If there was an issue with receiving
+ * the response over the network interface, then #HTTP_NETWORK_ERROR is returned,
+ * please see #_receiveHttpResponse. If there was an issue with parsing, then the
+ * parsing error that occurred will be returned, please see
+ * #_HTTPClient_InitializeParsingContext and #_HTTPClient_ParseResponse. Please
+ * see #_getFinalResponseStatus for the status returned when there were no
+ * network or parsing errors.
+ */
+static HTTPStatus_t _receiveAndParseHttpResponse( const HTTPTransportInterface_t * pTransport,
+                                                  HTTPResponse_t * pResponse );
 
 /*-----------------------------------------------------------*/
 
@@ -210,6 +263,10 @@ static HTTPStatus_t _sendHttpHeaders( const HTTPTransportInterface_t * pTranspor
     HTTPStatus_t returnStatus = HTTP_SUCCESS;
     int32_t transportStatus = 0;
 
+    assert( pTransport != NULL );
+    assert( pTransport->send != NULL );
+    assert( pRequestHeaders != NULL );
+
     /* Send the HTTP headers over the network. */
     transportStatus = pTransport->send( pTransport->pContext,
                                         pRequestHeaders->pBuffer,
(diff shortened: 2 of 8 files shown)


libraries/standard/http/include/http_client.h
```diff
@@ -102,10 +102,14 @@ typedef int32_t (* HTTPTransportSend_t )( HTTPNetworkContext_t * pContext,
...
+ *
  * If this function returns a value less than zero, then #HTTPClient_Send will
- * return #HTTP_NETWORK_ERROR. If this function returns less than the
- * bytesToRead and greater than zero, then this function will be invoked again
- * if the data in pBuffer contains a partial HTTP response message.
+ * return #HTTP_NETWORK_ERROR.
+ *
+ * If this function returns less than the bytesToRead and greater than zero,
+ * then this function will be invoked again if the data in pBuffer contains a
+ * partial HTTP response message and there there is room left in the pBuffer.
```
yourslab, line 111: Double there

sarenameas: Thanks!

libraries/standard/http/test/Makefile
```diff
@@ -29,7 +40,14 @@ include Mock4thewin.mk
 $(TESTS): $(CSDK_ROOT)/platform/posix/iot_clock_posix.o $(CSDK_ROOT)/platform/posix/iot_logging.o
 
 # additional dependency for a specific test
-HTTPClient_Send.c: _sendHttpHeaders.c _sendHttpBody.c _receiveHttpResponse.c
+HTTPClient_Send.c: _sendHttpHeaders.c \
+				   _sendHttpBody.c \
+				   _receiveHttpResponse.c \
+				   _receiveAndParseHttpResponse.c \
+				   _receiveHttpResponse.c \
```
yourslab, line 47: Double dependency of _receiveHttpResponse.c

libraries/standard/http/test/Makefile
```diff
@@ -20,7 +24,14 @@ INCLUDE = . ../include $(CSDK_ROOT) ../../utilities/include ../src
 
 # functions to dump into separate source files
 # Values based on test source file names will be dynamically added.
-FUNCTIONS = _sendHttpHeaders _sendHttpBody _receiveHttpResponse
+FUNCTIONS = _sendHttpHeaders \
+			_sendHttpBody \
+			_receiveHttpResponse \
+			_receiveAndParseHttpResponse \
+			_receiveHttpResponse \
```
yourslab, line 31: Double dependency of _receiveHttpResponse

libraries/standard/http/include/http_client.h
```diff
@@ -84,7 +84,7 @@ typedef struct HTTPNetworkContext HTTPNetworkContext_t;
 /**
  * @brief Transport interface for sending data over the network.
  *
- * If the number of bytes written returned is less than bytesToWrite, then
+ * If the number of bytes written returned is not equal to bytesToWrite, then
  * #HTTPClient_Send will return HTTP_NETWORK_ERROR. If a negative value is
```
aggarw13, line 88: (Nit): `#HTTP_NETWORK_ERROR` for doxygen to highlight the error code.

libraries/standard/http/include/http_client.h
```diff
@@ -84,7 +84,7 @@ typedef struct HTTPNetworkContext HTTPNetworkContext_t;
 /**
  * @brief Transport interface for sending data over the network.
  *
- * If the number of bytes written returned is less than bytesToWrite, then
+ * If the number of bytes written returned is not equal to bytesToWrite, then
```
aggarw13, line 87: (Nit) `@p bytesToWrite`

libraries/standard/http/include/http_client.h
```diff
@@ -122,9 +126,9 @@ typedef int32_t (* HTTPTransportRecv_t )( HTTPNetworkContext_t * pContext,
  */
 typedef struct HTTPTransportInterface
 {
-    HTTPTransportRecv_t recv;
-    HTTPTransportSend_t send;
-    HTTPNetworkContext_t * pContext;
+    HTTPTransportRecv_t recv;        /**< Transport receive interface */
+    HTTPTransportSend_t send;        /**< Transport interface send interface. */
+    HTTPNetworkContext_t * pContext; /**< User defined transport interface context. */
```
aggarw13, line 131: Can this just be `void *`? We don't need to specify a type when we are not responsible for interpreting the time

libraries/standard/http/include/http_client.h
sarenameas: Thanks

libraries/standard/http/include/http_client.h
sarenameas: I didn't know about that doxygen I will add it everywhere thanks

libraries/standard/http/include/http_client.h
```diff
@@ -423,19 +460,29 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ * considered a malformed response and therefore a security alert.
+ *
  * TODO: Expand documentation.
  *
  * @param[in] pTransport Transport interface, see #HTTPTransportInterface_t for
  * more information.
  * @param[in] pRequestHeaders Request configuration containing the buffer of
  * headers to send.
- * @param[in] pRequestBodyBuf Request entity body.
+ * @param[in] pRequestBodyBuf Request entity body. This can be NULL.
```
aggarw13, line 473: (Nit) "@param[in] **(Optional)**" makes it explicit about being _optional_

libraries/standard/http/include/http_client.h
sarenameas: Richard does not want any void pointers

libraries/standard/http/include/http_client.h
```diff
@@ -423,19 +460,29 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ * @return One of the following:
+ * - #HTTP_SUCCESS
+ * - #HTTP_INVALID_PARAMETER
+ * - #HTTP_NETWORK_ERROR
+ * - #HTTP_PARTIAL_RESPONSE
+ * - #HTTP_NO_RESPONSE
+ * - #HTTP_INSUFFICIENT_MEMORY
+ * TODO: Add more errors for parsing implementation.
  */
 HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
```
aggarw13, line 487: Looking at the function again in the API, a `xxx_SendAndReceived` naming would be clearer about the behavior of this function

libraries/standard/http/include/http_client.h
```diff
@@ -164,14 +168,47 @@ typedef enum HTTPStatus
...
+     *
+     * Functions that may return this value:
+     * - #HTTPClient_Send
+     */
     HTTP_PARTIAL_RESPONSE,
+
+    /**
+     * @brief No HTTP response was received from the network.
+     *
+     * This can occur only if #HTTPTransportRecv_t returns zero when first
```
aggarw13, line 185: This seems to be leaking some implementation detail. It could just mention generally that when no data is received from the transport interface.

libraries/standard/http/include/http_client.h
sarenameas: The name of this function has been approved in sign off. It is clear that we are receiving the response since it is in the parameters.

libraries/standard/http/include/http_client.h
sarenameas: Makes sense I will change it

libraries/standard/http/src/http_client.c
```diff
@@ -102,40 +165,204 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                             reqBodyBufLen,
+                             transportStatus );
+        returnStatus = HTTP_NETWORK_ERROR;
+    }
+    else
     {
-        transportStatus = pTransport->send( pTransport->pContext,
-                                            pRequestBodyBuf,
-                                            reqBodyBufLen );
+        IotLogDebugWithArgs( "Sent HTTP body over the transport: Bytes sent = %d.",
```
aggarw13, line 194: In the data section, strings can be compact as "BytesSent=%d" to avoid using extra characters

sarenameas: sure

libraries/standard/http/include/http_client.h
```diff
@@ -423,19 +457,30 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * parameter pRequestBodyBuf over the transport. The response is received in
  * #HTTPResponse_t.
  *
+ * The application should close the connection with the server if any
+ * HTTP_SECURITY_ALERT_X errors are returned. Any error found in parsing is
```
aggarw13, line 461: Can you state all the error codes here to be clear? Some may not understand the presence of the generic `_X`

sarenameas: I do not have them all defined yet. Next PR.

libraries/standard/http/include/http_client.h
```diff
@@ -423,19 +457,30 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ * considered a malformed response and therefore a security alert.
+ *
  * TODO: Expand documentation.
```
aggarw13, line 464: Unresolved TODO?

sarenameas: The function is NOT finished.

libraries/standard/http/include/http_client.h
```diff
@@ -423,19 +457,30 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * parameter pRequestBodyBuf over the transport. The response is received in
  * #HTTPResponse_t.
  *
+ * The application should close the connection with the server if any
+ * HTTP_SECURITY_ALERT_X errors are returned. Any error found in parsing is
+ * considered a malformed response and therefore a security alert.
```
aggarw13, line 462: The "Any error found in parsing is considered a malformed response and therefore a security alert" seems incomplete and or misleading about the quality of the library. 
What about errors arising from internal failures during parsing? Would that be also elevated as a "security alert"? 
These questions are the type of misleading interpretations of the sentence

libraries/standard/http/include/http_client.h
```diff
@@ -423,19 +457,30 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ * @param[in] pRequestBodyBuf Optional Request entity body. Set to NULL if there
+ * is no request body.
  * @param[in] reqBodyBufLen The length of the request entity in bytes.
  * @param[in] pResponse The response message and some notable response
  * parameters will be returned here on success.
  *
- * @return #HTTP_SUCCESS if successful, an error code otherwise.
- * TODO: Update for exact error codes returned.
+ * @return One of the following:
+ * - #HTTP_SUCCESS
```
aggarw13, line 477: Can you add scenarios when these error codes will be returned just like other function docs?

libraries/standard/http/include/http_client.h
sarenameas: Good point deleting.

libraries/standard/http/include/http_client.h
sarenameas: This is not in the v4 docs because people are expected to go to the error code and read about. You guys went above and extra.

libraries/standard/http/src/http_client.c
```diff
@@ -188,15 +415,31 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
-                                      reqBodyBufLen );
+        if( pRequestBodyBuf != NULL )
+        {
+            returnStatus = _sendHttpBody( pTransport,
+                                          pRequestBodyBuf,
+                                          reqBodyBufLen );
+        }
+        else
+        {
+            IotLogDebug( "A request body was not sent: pRequestBodyBuf is NULL." );
```
aggarw13, line 426: Suggest changing it to "No HTTP body in request sent: pRequestBodyBuf is NULL"

sarenameas: I am not changing this. This means the same thing.

libraries/standard/http/include/http_client.h
aggarw13: We raised the bar!

libraries/standard/http/include/http_client.h
aggarw13: Please add a TODO for it then so that is doesn't go unnoticed

libraries/standard/http/src/http_client.c
```diff
@@ -333,15 +557,31 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
+        /* If the application chooses to receive a response, then pResponse
+         * will not be NULL. */
+        if( pResponse != NULL )
+        {
+            returnStatus = _receiveAndParseHttpResponse( pTransport,
+                                                         pResponse );
+        }
+        else
+        {
+            IotLogWarn( "A response was not received: pResponse is NULL. " );
```
aggarw13, line 583: If `pResponse` is optional, then the message should not be logged as a warning, because the API considers it _fine_ to not provide the response buffer.

aggarw13: Also, the message is misleading as the library did not attempt to read from the transport layer

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,204 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
-/*-----------------------------------------------------------*/
-
-static HTTPStatus_t _receiveHttpResponse( const HTTPTransportInterface_t * pTransport,
-                                          HTTPResponse_t * pResponse )
+static HTTPStatus_t _receiveAndParseHttpResponse( const HTTPTransportInterface_t * pTransport,
+                                                  HTTPResponse_t * pResponse )
 {
-    /* TODO: Receive the HTTP response with parsing. */
-    return HTTP_SUCCESS;
+    HTTPStatus_t returnStatus = HTTP_SUCCESS;
```
aggarw13, line 447: assert checks

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,204 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    size_t totalReceived = 0;
+    size_t currentReceived = 0;
+    HTTPParsingContext_t parsingContext = { 0 };
+
+    if( pResponse->pBuffer == NULL )
```
aggarw13, line 452: Seems like this parameter checking should be done at the immediate function of the user-input rather than a level deeper

libraries/standard/http/src/http_client.c
```diff
@@ -54,6 +55,58 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
                                 const char * pValue,
                                 size_t valueLen );
 
+/**
+ * @brief Receive HTTP response from the transport recv interface.
+ *
+ * @param[in] pTransport Transport interface.
+ * @param[in] pResponse Response buffer.
+ * @param[in] bufferLen Length of the response buffer.
```
aggarw13, line 63: Missing `@param[in] BytesReceived`

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    {
+        /* Initialize the parsing context. */
+        returnStatus = _HTTPClient_InitializeParsingContext( &parsingContext,
+                                                             pResponse->pHeaderParsingCallback );
+    }
+
+    /* While there are no errors in the transport recv or parsing, the response
+     * message is not finished, and there is room in the response buffer. */
+    while( ( returnStatus == HTTP_SUCCESS ) &&
+           ( parsingContext.state != HTTP_PARSING_COMPLETE ) &&
```
aggarw13, line 468: Is `parsingContext.state != HTTP_PARSING_COMPLETE` the main indicator of unfinished response?

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                                             size_t responseBufferLen )
+{
+    HTTPStatus_t returnStatus = HTTP_SUCCESS;
+
+    assert( parsingState >= HTTP_PARSING_NONE &&
+            parsingState <= HTTP_PARSING_COMPLETE );
+    assert( totalReceived <= responseBufferLen );
+
+    /* If no parsing occurred, that means network data was never received. */
+    if( parsingState == HTTP_PARSING_NONE )
```
aggarw13, line 409: When `parsingState` is `HTTP_PARSING_NONE`, should `totalReceived` be `0`?  If yes, is it possible to avoid the dual indicators of the same state?

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    }
+    else if( parsingState == HTTP_PARSING_INCOMPLETE )
+    {
+        if( totalReceived == responseBufferLen )
         {
-            IotLogErrorWithArgs( "Error in sending the HTTP body over the "
-                                 "transport interface. Transport status %d.",
-                                 transportStatus );
-            returnStatus = HTTP_NETWORK_ERROR;
+            IotLogErrorWithArgs( "Response is too large for the response buffer"
```
aggarw13, line 420: "Unable to read complete response from transport interface: Response buffer has insufficent space: ResponseBufferSize=%d"

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+    if( parsingState == HTTP_PARSING_NONE )
+    {
+        IotLogErrorWithArgs( "Response not received: Zero returned from "
+                             "transport recv: totalReceived=%d",
+                             totalReceived );
+        returnStatus = HTTP_NO_RESPONSE;
+    }
+    else if( parsingState == HTTP_PARSING_INCOMPLETE )
+    {
+        if( totalReceived == responseBufferLen )
```
aggarw13, line 418: How about the (corner) case when the server sends incomplete response but the partial response fills the buffer? In that case, the error reported is unsufficient memory though the error could also be partial response.

I suppose that it makes sense to prioritize insufficient memory over partial response. But just stating this case to check whether it was considered

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
-        else if( transportStatus != reqBodyBufLen )
+        else
         {
-            IotLogErrorWithArgs( "Failure in sending HTTP headers: Transport layer "
-                                 "did not send the required bytes: Required Bytes=%d"
-                                 ", Sent Bytes=%d.",
-                                 transportStatus );
-            returnStatus = HTTP_NETWORK_ERROR;
+            IotLogErrorWithArgs( "Partial response received: Transport recv "
+                                 "returned zero before the complete response: "
```
aggarw13, line 428: "Received partial response from transport recv(): ResponseSize=%d, TotalBufferSize=%d"

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+        IotLogErrorWithArgs( "Failed to receive HTTP response: Transport recv() "
+                             "returned error: TransportStatus=%d.",
+                             transportStatus );
+        returnStatus = HTTP_NETWORK_ERROR;
+    }
+    else if( (size_t)transportStatus > bufferLen )
+    {
+        /* There is a bug in the transport recv if more bytes are reported
+         * to have been read than the bytes asked for. */
+        IotLogErrorWithArgs( "Failed to receive HTTP response: Transport recv() "
```
aggarw13, line 373: Is this a network error? Can it not be an insufficient buffer space error?

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                             " read more bytes than expected: BytesRead=%d",
```
aggarw13, line 374: Also, log `expectedBytes=%d`

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                             transportStatus );
+    }
+    else
     {
-        transportStatus = pTransport->send( pTransport->pContext,
-                                            pRequestBodyBuf,
-                                            reqBodyBufLen );
+        /* When a zero is returned from the transport recv it will not be
+         * invoked again. */
+        IotLogDebug( "Transport recv() returned 0. Receiving transport data"
```
aggarw13, line 389: "Received zero bytes from trasnport recv(). Response reading is complete."

libraries/standard/http/src/http_client.c
```diff
@@ -54,6 +55,58 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ *
+ * @param[in] pTransport Transport interface.
+ * @param[in] pResponse Response buffer.
+ * @param[in] bufferLen Length of the response buffer.
+ *
+ * @return Returns #HTTP_SUCCESS if successful. If there was a network error or
+ * more bytes than what was specified were read, then #HTTP_NETWORK_ERROR is
+ * returned.
+ */
+HTTPStatus_t _receiveHttpResponse( const HTTPTransportInterface_t * pTransport,
```
aggarw13, line 69: (Nit) I think `_readResponse` is a more appropriate name

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +307,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                                                             pResponse->pHeaderParsingCallback );
+    }
+
+    /* While there are no errors in the transport recv or parsing, the response
+     * message is not finished, and there is room in the response buffer. */
+    while( ( returnStatus == HTTP_SUCCESS ) &&
+           ( parsingContext.state != HTTP_PARSING_COMPLETE ) &&
+           ( totalReceived < pResponse->bufferLen ) )
+    {
+        /* Receive the HTTP response data into the pResponse->pBuffer. */
```
aggarw13, line 471: I think the while condition can be simplified (for better readability) and `break` statement be avoided in the following form:
```
while( shouldRead == 1u ) 
{
    /* Receive the HTTP response data into the pResponse->pBuffer. */
    returnStatus = _receiveHttpResponse( pTransport,
                                         pResponse->pBuffer + totalReceived,
                                         pResponse->bufferLen - totalReceived,
                                         &currentReceived );

    if( returnStatus == HTTP_SUCCESS )
    {
        if( currentReceived > 0 )
        {
            totalReceived += currentReceived;
            /* Data is received into the buffer and must be parsed. */
            returnStatus = _HTTPClient_ParseResponse( &parsingContext,
                                                      pResponse->pBuffer + totalReceived,
                                                      currentReceived );

            shouldRead = ( parsingContext.state == HTTP_PARSING_COMPLETE ) &
                         ( totalReceived < pResponse->bufferLen );
        }
        else
        {
            /* If there was no data received, then end receiving and parsing
             * the response. */
            shouldRead = 0u;
        }
    }
}
```

sarenameas: Looks good thanks

libraries/standard/http/include/http_client.h
sarenameas: I think it's duplicating information.

libraries/standard/http/src/http_client.c
sarenameas: move it up

libraries/standard/http/src/http_client.c
sarenameas: That case is moot, we will never be able to deduce it with the current buffer space. A well behaving server is supposed to send the full response and that is an appropriate assumption to make.

libraries/standard/http/src/http_client.c
sarenameas: sure

libraries/standard/http/src/http_client.c
```diff
@@ -541,6 +542,11 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
         IotLogError( "Parameter check failed: pRequestHeaders->pBuffer is NULL." );
         returnStatus = HTTP_INVALID_PARAMETER;
     }
+    else if( pResponse && ( pResponse->pBuffer == NULL ) )
```
aggarw13, line 545: (Nit): If one operand is being _explicitly_ checked against `NULL`, then the other operand in the expression should too.
`if ( ( pResponse != NULL ) && ( pResponse->pBuffer == NULL ) )`

libraries/standard/http/src/http_client.c
sarenameas: it should be but i find value in printing that it is so

libraries/standard/http/src/http_client.c
sarenameas: sure

libraries/standard/http/src/http_client.c
```diff
@@ -210,29 +264,36 @@ static HTTPStatus_t _sendHttpHeaders( const HTTPTransportInterface_t * pTranspor
...
+                             "did not send the required bytes: RequiredBytes=%d"
+                             ", SentBytes=%d.",
+                             pRequestHeaders->headersLen,
                              transportStatus );
         returnStatus = HTTP_NETWORK_ERROR;
     }
     else
     {
-        /* Empty else MISRA 15.7 */
+        IotLogDebugWithArgs( "Sent HTTP headers over the transport: BytesSent "
```
aggarw13, line 294: _BytesSent=%d_ without spaces

libraries/standard/http/src/http_client.c
sarenameas: The other functions say _send I think this is more consistent. We are receiving from the transport interface anyways.

libraries/standard/http/src/http_client.c
sarenameas: missed one thanks

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +308,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                returnStatus = _HTTPClient_ParseResponse( &parsingContext,
+                                                          pResponse->pBuffer + totalReceived,
+                                                          currentReceived );
+            }
+        }
+
+        /* While there are no errors in the transport recv or parsing, we received
+         * data over the transport, the response message is not finished, and
+         * there is room in the response buffer. */
+        shouldRecv = ( uint8_t ) ( ( returnStatus == HTTP_SUCCESS ) &&
```
aggarw13, line 491: Why not keep this expression in the above happy-path if-case? You wouldn't need the `( returnStatus == HTTP_SUCCESS )` condition

sarenameas: I would still need to check ( returnStatus == HTTP_SUCCESS ) since the loop needs to stop on an error returned from _HTTPClient_ParseResponse().

sarenameas: By putting this at the end we setup the next looping conditions clearly.

libraries/standard/http/src/http_client.c
```diff
@@ -54,6 +55,59 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ */
+static HTTPStatus_t _getFinalResponseStatus( HTTPParsingState_t parsingState,
+                                             size_t totalReceived,
+                                             size_t responseBufferLen );
+
+/**
+ * @brief Receive the HTTP response from the network and parse it.
+ *
+ * @param[in] pTransport Transport interface.
+ * @param[in] pResponse Response message to receive data from the network.
```
yourslab, line 98: Shouldn't this be param[out]?

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
+#ifndef HTTP_CLIENT_PARSE_H_
+#define HTTP_CLIENT_PARSE_H_
+
+#include "http_client.h"
+/* #include "http_parser.h" */
```
aggarw13, line 5: Commented out code

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @brief The state of the response message parsed after
```
aggarw13, line 8: It's the state of the parser, not the response message...
I suggest "States associated with parsing an HTTP response."

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+ * #HTTPClient_ParseResponse returns.
+ */
+typedef enum HTTPParsingState_t
+{
+    HTTP_PARSING_NONE = 0,   /**< The parser has not started reading any response. */
+    HTTP_PARSING_INCOMPLETE, /**< The parse found a partial reponse. */
```
aggarw13, line 14: parse -> parse**r**

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+    HTTP_PARSING_COMPLETE    /**< The parser found the entire response. */
+} HTTPParsingState_t;
+
+/**
+ * The HTTP response parsing context.
+ */
+typedef struct HTTPParsingContext
+{
+    /* http-parser dependencies will be added in the next incremental change. */
+    /* Below will be un-commented when parsing is implemented. */
```
aggarw13, line 24: A TODOD would be good here

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+    #if 0
+        http_parser httpParser; /**< Third-party http-parser context. */
+    #endif
+    HTTPParsingState_t state;   /**< The current state of the HTTP response parsed. */
+    HTTPClient_HeaderParsingCallback_t * pHeaderParsingCallback; /**< Callback to invoke with each header found. */
```
aggarw13, line 29: Should this be the non-pointer type?

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+} HTTPParsingContext_t;
+
+/**
+ * Initialize the HTTP Client response parsing context.
```
aggarw13, line 33: `@brief`

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+ */
+typedef enum HTTPParsingState_t
+{
+    HTTP_PARSING_NONE = 0,   /**< The parser has not started reading any response. */
+    HTTP_PARSING_INCOMPLETE, /**< The parse found a partial reponse. */
+    HTTP_PARSING_COMPLETE    /**< The parser found the entire response. */
+} HTTPParsingState_t;
+
+/**
+ * The HTTP response parsing context.
```
aggarw13, line 19: `@brief` 
Same for the enum

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+        http_parser httpParser; /**< Third-party http-parser context. */
+    #endif
+    HTTPParsingState_t state;   /**< The current state of the HTTP response parsed. */
+    HTTPClient_HeaderParsingCallback_t * pHeaderParsingCallback; /**< Callback to invoke with each header found. */
+} HTTPParsingContext_t;
+
+/**
+ * Initialize the HTTP Client response parsing context.
+ *
+ * @param[in] pParsingContext The state of response parsing to initialize.
```
aggarw13, line 35: "The context to initialize" would be short and sweet.

aggarw13, line 35: `@param[in/out]`

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+ * @param[in] pHeaderParsingCallback Callback to invoked for each complete
```
aggarw13, line 36: (Nit) Suggest "**The** callback **that will be** invoked for each header and value **pair** found **in parsing**."

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+                                                   HTTPClient_HeaderParsingCallback_t * pHeaderParsingCallback );
+
+/**
+ * Parse the input HTTP response buffer.
+ *
+ * This function may be invoked multiple times for different parts of the the
+ * HTTP response. The state of what was last parsed in the response is kept in
+ * #HTTPParsingContext_t.
+ *
+ * The application should close the connection with the server if any
```
aggarw13, line 54: This is not a user-facing file, so this comment is not required here

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+ * HTTP_SECURITY_ALERT_X errors are returned.
+ * TODO: List all the security alerts possible after parsing development.
+ *
+ * @param[in] pParsingState The state of the response parsing.
```
aggarw13, line 58: `@param[in/**out**]`

libraries/standard/http/src/private/http_client_parse.h
```diff
@@ -0,0 +1,72 @@
...
+ * @param[in] bufferLen The length in the buffer to parse.
+ *
+ * @return One of the following:
+ * - #HTTP_SUCCESS
+ * - #HTTP_INVALID_PARAMETER
+ * - #HTTP_SECURITY_ALERT_PARSER_INVALID_CHARACTER
+ * TODO: Other return values are to be added during implementation of parsing.
+ */
+HTTPStatus_t _HTTPClient_ParseResponse( HTTPParsingContext_t * pParsingContext,
+                                        uint8_t * pBuffer,
```
aggarw13, line 69: Parameter should be of `const uint8_t *` type

libraries/standard/http/src/private/http_client_parse.h
sarenameas: This is going to be implemented in the next PR.

libraries/standard/http/src/private/http_client_parse.h
sarenameas: It is only param[in] because things are not being returned.

libraries/standard/http/src/private/http_client_parse.h
sarenameas: It is the state of what has been parsed, so it is the state of the response message.

libraries/standard/http/src/private/http_client_parse.h
sarenameas: Thanks

libraries/standard/http/src/private/http_client_parse.h
sarenameas: I don;t need a todo here because if this is no uncommented I cannot use it.

libraries/standard/http/src/private/http_client_parse.h
sarenameas: Thanks

libraries/standard/http/src/private/http_client_parse.h
sarenameas: Then it doesn't need to be deleted.

libraries/standard/http/src/private/http_client_parse.h
sarenameas: We do not put out for contexts.

libraries/standard/http/src/private/http_client_parse.h
sarenameas: right thanks

libraries/standard/http/src/private/http_client_parse.h
sarenameas: hmm well I guess I will put it

libraries/standard/http/src/http_client_parse.c
```diff
@@ -0,0 +1,17 @@
...
+
+HTTPStatus_t _HTTPClient_InitializeParsingContext( HTTPParsingContext_t * pParsingContext,
+                                                   HTTPClient_HeaderParsingCallback_t * pHeaderParsingCallback )
+{
+    /* This function is to be implenmented. */
+    return HTTP_SUCCESS;
+}
+
+HTTPStatus_t _HTTPClient_ParseResponse( HTTPParsingContext_t * pParsingContext,
+                                        uint8_t * pBuffer,
```
aggarw13, line 11: Add _constness_ here `const uint8_t *`

sarenameas: oh i did and it didnt get pushed

libraries/standard/http/src/private/http_client_parse.h
aggarw13: But it is irrelevant here. The API doc needs to have information meant only for _its_ users. The application is not this file's user.

libraries/standard/http/src/http_client.c
```diff
@@ -247,40 +308,203 @@ static HTTPStatus_t _sendHttpBody( const HTTPTransportInterface_t * pTransport,
...
+                                        pRequestBodyBuf,
+                                        reqBodyBufLen );
+
+    if( transportStatus < 0 )
     {
-        transportStatus = pTransport->send( pTransport->pContext,
-                                            pRequestBodyBuf,
-                                            reqBodyBufLen );
+        IotLogErrorWithArgs( "Failed to send HTTP body: Transport send() "
+                             " returned error: TransportStatus=%d",
```
gkwicker, line 322: We'll want to find a way to convert `transportStatus` to a string for logging here, otherwise we're going to have 1000 issues that say _"I received error -119, what does that mean?"_  Ideally, the transport layer should have a function for converting error codes to strings since every transport layer will do things differently.

sarenameas: That will be part of the demo since the transportStatus is unique to each stack.

PR #899, opened 2020-04-28
HTTPClient_AddRangeHeader() Implementation

aggarw13:
*Issue #, if available:*

*Description of changes:*
* Implement `HTTPClient_AddRangeHeader` API function
* Add unit tests

image


libraries/standard/http/src/http_client.c
```diff
@@ -40,21 +139,149 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
 {
-    return HTTP_NOT_SUPPORTED;
+    HTTPStatus_t returnStatus = HTTP_SUCCESS;
+    char rangeValueBuffer[ MAX_RANGE_REQUEST_VALUE_LEN ] = { 0 };
+    size_t rangeValueLength = 0;
+    int stdRetVal = 0;
+
+    if( pRequestHeaders == NULL )
+    {
+        IotLogError( "Parameter check failed: pRequestHeaders is NULL." );
```
yourslab, line 207: Maybe change to "Parameter check failed: pRequestHeaders interface is NULL." for consistency.

libraries/standard/http/src/http_client.c
```diff
@@ -40,21 +139,149 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
+        else if( rangeStartOrlastNbytes >= 0 )
+        {
+            /* Add the "-" character.*/
+            stdRetVal = sprintf( rangeValueBuffer + rangeValueLength,
+                                 "%c",
+                                 DASH_CHARACTER );
+            /* Check that only a single character was written. */
+            assert( stdRetVal == 1 );
+            rangeValueLength += ( size_t ) stdRetVal;
+        }
```
yourslab, line 274: Is empty else needed here?

libraries/standard/http/src/http_client.c
sarenameas: It should not say "interface"

libraries/standard/http/src/http_client.c
```diff
@@ -40,21 +139,149 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
+        IotLogErrorWithArgs( "Parameter check failed: Invalid range values: "
+                             "rangeStart should be < rangeEnd when both are >= 0: "
+                             "RangeStart=%d, RangeEnd=%d",
+                             rangeStartOrlastNbytes, rangeEnd );
+        returnStatus = HTTP_INVALID_PARAMETER;
+    }
+    else
+    {
+        /* Populate the buffer with the value data for the Range Request.*/
+        stdRetVal = sprintf( rangeValueBuffer,
```
sarenameas, line 242: You must use "snprintf" not sprintf.

libraries/standard/http/src/http_client.c
```diff
@@ -40,21 +139,149 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
+                             "%s%d",
+                             RANGE_REQUEST_HEADER_VALUE_PREFIX,
+                             rangeStartOrlastNbytes );
+        assert( ( stdRetVal >= 0 ) &&
```
sarenameas, line 246: I do not think we should assert here. If there are encoding errors that should be bubbled up to the user.

libraries/standard/http/src/http_client.c
```diff
@@ -150,27 +377,27 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
 
     if( pTransport == NULL )
     {
-        IotLogError( "Parameter check failed: pTransport interface is NULL." );
+        IotLogError( "Parameter check failed:   pTransport interface is NULL." );
```
sarenameas, line 380: What are these weird extra spaces lol?

libraries/standard/http/include/http_client.h
```diff
@@ -403,19 +439,44 @@ HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ *
+ * @note There are 3 different forms of range specification, determined by the
+ * combination of @a rangeStartOrLastNBytes and @a rangeEnd parameter values:
+ * 1. Request containing both parameters for the byte range [rangeStart, rangeEnd]
+ * where @a rangeStartOrLastNBytes <= @a rangeEnd.
+ * Example request: "Range: bytes=0-1024\r\n" for requesting bytes in the range [0, 1024].
  *
- * TODO: Add documentation about rangeStart and rangeEnd configuration.
+ * 2. Request for the last N bytes, represented by @p rangeStartOrLastNBytes.
+ * @p rangeStartOrLastNBytes should be negative and @p rangeEnd should #FILE_END.
```
sarenameas, line 454: #HTTP_RANGE_REQUEST_END_OF_FILE

libraries/standard/http/include/http_client.h
```diff
@@ -403,19 +439,44 @@ HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ * Example request: "Range: bytes=-512\r\n" for requesting the last 512 bytes
+ * (or bytes in the range [0, 512]).
```
sarenameas, line 456: The comment in the parentheses is not correct if the file is greater than 513 bytes.

libraries/standard/http/include/http_client.h
```diff
@@ -403,19 +439,44 @@ HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders,
  * #HTTPRequestHeaders_t.pBuffer.
  *
  * For example, if requesting for the first 1kB of a file the following would be
- * written  "Range: bytes=0-1024\r\n".
+ * written  "Range: bytes=0-1024\r\n\r\n".
```
sarenameas, line 442: The range is inclusive on both ends. So it would be `Range: bytes=0-1023\r\n\r\n`

libraries/standard/http/include/http_client.h
```diff
@@ -403,19 +439,44 @@ HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ *
+ * The trailing "\r\n" that denotes the end of the header lines is overwritten, if it
+ * already exists in the buffer.
+ *
+ * @note There are 3 different forms of range specification, determined by the
+ * combination of @a rangeStartOrLastNBytes and @a rangeEnd parameter values:
+ * 1. Request containing both parameters for the byte range [rangeStart, rangeEnd]
+ * where @a rangeStartOrLastNBytes <= @a rangeEnd.
+ * Example request: "Range: bytes=0-1024\r\n" for requesting bytes in the range [0, 1024].
```
sarenameas, line 451: We should probably change this to 1023 to be consistent with the example change I commented.

libraries/standard/http/src/http_client.c
aggarw13: My thought is that `snprintf` as a standard library functionality should be caught during development rather than leaving the door open for production failure logging. A failure in `snprintf` is not vulnerable to runtime user input. It is using string constants defined by use and user integer inputs whose bit-width is fixed

libraries/standard/http/src/http_client.c
aggarw13: It is not possible to estimate the size of the ASCII version of the range value, due to which I have used `sprintf` here

libraries/standard/http/src/http_client.c
aggarw13: Accidental. Fixed

libraries/standard/http/src/http_client.c
sarenameas: snprintf simply takes the maximum size that can be written, we can write less.

libraries/standard/http/src/http_client.c
aggarw13: Good catch. Added

libraries/standard/http/src/http_client.c
aggarw13: It is not an interface but an object.

libraries/standard/http/src/http_client.c
```diff
@@ -40,21 +139,149 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
...
 /*-----------------------------------------------------------*/
 
 HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
-                                        int32_t rangeStart,
+                                        int32_t rangeStartOrlastNbytes,
                                         int32_t rangeEnd )
 {
-    return HTTP_NOT_SUPPORTED;
+    HTTPStatus_t returnStatus = HTTP_SUCCESS;
+    char rangeValueBuffer[ MAX_RANGE_REQUEST_VALUE_LEN ] = { 0 };
```
sarenameas, line 201: +1 for the null terminator that snprintf adds?

sarenameas: Can we add a test that writes the maximum integers for rangeStart and rangeEnd?

libraries/standard/http/include/http_client.h
aggarw13: Fixed

libraries/standard/http/include/http_client.h
aggarw13: Good catch. Fixed

libraries/standard/http/include/http_client.h
aggarw13: Fixed

libraries/standard/http/src/http_client.c
aggarw13: Good point. I do have a [stress test](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/f67e9c46ed2f300b36b702d3186f6f161d392a92/libraries/standard/http/test/test-HTTPClient_AddRangeHeader.c#L211-L229) with the `INT32_MAX` values for `rangeStart` and `rangeEnd` which runs fine without the NULL (surprisingly). Have added the extra byte space.

libraries/standard/http/include/http_client.h
```diff
@@ -73,6 +73,19 @@
  */
 #define HTTP_RESPONSE_CONNECTION_KEEP_ALIVE_FLAG    0x2U
 
+/**
+ * @brief Flag that represents End of File byte in the range specification of
+ * a Range Request.
+ * This flag should be used ONLY for 2 kinds of range specifications when
+ * creating the Range Request header through the #HTTPClient_AddRangeHeader
+ * function:
+ *  - When the requested range is all bytes > starting range byte to end of file.
```
sarenameas, line 82: I think we should replace ">" with "from the"

libraries/standard/http/include/http_client.h
```diff
@@ -426,19 +439,46 @@ HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+ *
+ * The trailing "\r\n" that denotes the end of the header lines is overwritten, if it
+ * already exists in the buffer.
  *
- * TODO: Add documentation about rangeStart and rangeEnd configuration.
+ * @note There are 3 different forms of range specification, determined by the
+ * combination of @a rangeStartOrLastNBytes and @a rangeEnd parameter values:
+ * 1. Request containing both parameters for the byte range [rangeStart, rangeEnd]
+ * where @a rangeStartOrLastNBytes <= @a rangeEnd.
+ * Example request: "Range: bytes=0-1023\r\n" for requesting bytes in the range [0, 1024].
```
sarenameas, line 451: range [0, 1023]???

libraries/standard/http/src/http_client.c
sarenameas: We really need to snprintf appsec will flag this and cut a ticket. And then we will have to waste time explaining away the flag.
```
snprintf( rangeValueBuffer,
              MAX_RANGE_REQUEST_VALUE_LEN + 1,
              "%s%d"
              RANGE_REQUEST_HEADER_VALUE_PREFIX,
              rangeStartOrlastNbytes );
```

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -85,4 +85,24 @@
...
+ * @brief Constants relating to Range Requests.
+ */
+#define RANGE_REQUEST_HEADER_FIELD               "Range"
+#define RANGE_REQUEST_HEADER_FIELD_LEN           ( sizeof( RANGE_REQUEST_HEADER_FIELD ) - 1 )
+#define RANGE_REQUEST_HEADER_VALUE_PREFIX        "bytes="
+#define RANGE_REQUEST_HEADER_VALUE_PREFIX_LEN    ( sizeof( RANGE_REQUEST_HEADER_VALUE_PREFIX ) - 1 )
+
+/* Maximum value of a 32 bit signed integer is 2,147,483,647. Used for calculating buffer space for
+ * ASCII representation of range values. */
+#define MAX_INT32_NO_OF_DIGITS                   10
```
gkwicker, line 99: This should be named `MAX_INT32_NO_OF_DECIMAL_DIGITS` since you need to know the base to know how long the string representation could be.  Also, what if the number is negative and has a leading `-`?  Not that this would happen in normal use, however one could see it happening in the presence of some other bug.

libraries/standard/http/src/http_client.c
gkwicker: Inclined to agree with @aggarw13 here, although in general we should bubble most errors up.  `snprintf()` is a special case since there is no I/O involved.

libraries/standard/http/src/http_client.c
```diff
@@ -196,10 +196,102 @@ HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
+            /* Add the "-" character.*/
+            stdRetVal = snprintf( rangeValueBuffer + rangeValueLength,
+                                  sizeof( rangeValueBuffer ) - rangeValueLength,
+                                  "%s",
+                                  DASH_CHARACTER );
+            /* Check that only a single character was written. */
+            assert( stdRetVal == DASH_CHARACTER_LEN );
+            rangeValueLength += ( size_t ) stdRetVal;
+        }
+        else
```
gkwicker, line 281: Check the complexity on this function, I believe it is probably > 8.

libraries/standard/http/include/http_client.h
```diff
@@ -73,6 +73,20 @@
...
+ * a Range Request.
+ * This flag should be used ONLY for 2 kinds of range specifications when
+ * creating the Range Request header through the #HTTPClient_AddRangeHeader
+ * function:
+ *  - When the requested range is all bytes from the starting range byte to
+ * the end of file.
+ *  - When the requested range is for the last N bytes of the file.
+ * In both cases, this value should be used for the "rangeEnd" parameter.
+ */
+#define HTTP_RANGE_REQUEST_END_OF_FILE              -1
```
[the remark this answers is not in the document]
aggarw13: `0xFF` is a valid byte range end value for a customer to provide (for example, file bytes in the range [0-255], [255, EoF]).
Thus, a negative value is reserved for the macro as a customer should not pass a negative value otherwise to represent EoF

libraries/standard/http/src/private/http_client_internal.h
aggarw13: Good suggestion about macro renaming. 
As you mentioned, the correct use case of negative value should would only contain a single value in the range as `"Range: bytes=<negative-value>"`. For a misuse of introducing negative value in the `[rangeStart, rangeEnd]` use case, I think the unit tests for `AddRangeHeader` would catch it...

libraries/standard/http/src/http_client.c
aggarw13: I ran complexity locally, and it yielded a threshold score for the function
image

PR #901, opened 2020-04-29
Implement MQTTProcessLoop

muneebahmed10:
*Description of changes:*
Implementation for MQTTProcessLoop. Depends on functions from #874 and #911

### DIFF
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -241,6 +241,23 @@ MQTTStatus_t MQTT_Unsubscribe( MQTTContext_t * const pContext,
  */
 MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext );
 
+/**
+ * @brief Loop to receive packets from the transport interface.
+ *
+ * @param[in] pContext Initialized and connected MQTT context.
+ * @param[in] timeoutMs Minimum time in milliseconds that the receive loop will
+ * run, unless an error occurs.
+ *
+ * @return #MQTTBadParameter if context is NULL;
+ * #MQTTRecvFailed if a network error occurs during reception;
+ * #MQTTSendFailed if a network error occurs while sending an ACK;
+ * #MQTTBadResponse if an invalid packet is received;
+ * #MQTTKeepAliveTimeout if the server has not sent a PINGRESP before
+ * pContext->pingRespTimeoutMs milliseconds;
+ * #MQTTIllegalState if an incoming QoS 1/2 publish or ack causes an
+ * invalid transition for the internal state machine;
+ * #MQTTSuccess on success.
+ */
 MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
                                uint32_t timeoutMs );
 
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -233,8 +233,8 @@ MQTTStatus_t MQTT_Unsubscribe( MQTTContext_t * const pContext,
  */
 MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext );
 
-MQTTStatus_t MQTT_Process( MQTTContext_t * const pContext,
-                           uint32_t timeoutMs );
+MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
+                               uint32_t timeoutMs );
 
 uint16_t MQTT_GetPacketId( MQTTContext_t * const pContext );
 
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -109,6 +109,13 @@ struct MQTTContext
     MQTTConnectionStatus_t connectStatus;
     MQTTApplicationCallbacks_t callbacks;
     uint32_t lastPacketTime;
+    bool controlPacketSent;
+
+    /* Keep alive members. */
+    uint16_t keepAliveIntervalSec;
+    uint32_t pingReqSendTimeMs;
+    uint32_t pingRespTimeoutMs;
+    bool waitingForPingResp;
 };
 
 /**
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -251,7 +251,7 @@ MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext );
  *
  * @return #MQTTBadParameter if context is NULL;
  * #MQTTRecvFailed if a network error occurs during reception;
- * #MQTTSendFailed if a network error occurs while sending an ACK;
+ * #MQTTSendFailed if a network error occurs while sending an ACK or PINGREQ;
  * #MQTTBadResponse if an invalid packet is received;
  * #MQTTKeepAliveTimeout if the server has not sent a PINGRESP before
  * pContext->pingRespTimeoutMs milliseconds;
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -45,7 +45,8 @@ typedef uint32_t (* MQTTGetCurrentTimeFunc_t )( void );
 
 typedef void (* MQTTEventCallback_t )( MQTTContext_t * pContext,
                                        MQTTPacketInfo_t * pPacketInfo,
-                                       uint16_t packetIdentifier );
+                                       uint16_t packetIdentifier,
+                                       MQTTPublishInfo_t * pPublishInfo );
 
 typedef enum MQTTConnectionStatus
 {
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -44,7 +44,8 @@ typedef int32_t (* MQTTTransportSendFunc_t )( MQTTNetworkContext_t context,
 typedef uint32_t (* MQTTGetCurrentTimeFunc_t )( void );
 
 typedef void (* MQTTEventCallback_t )( MQTTContext_t * pContext,
-                                       MQTTPacketInfo_t * pPacketInfo );
+                                       MQTTPacketInfo_t * pPacketInfo,
+                                       uint16_t packetIdentifier );
 
 typedef enum MQTTConnectionStatus
 {
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -73,7 +73,7 @@ static MQTTPubAckType_t getAckFromPacketType( uint8_t packetType );
  * @param[in] bytesToRecv Number of bytes to receive.
  * @param[in] timeoutMs Time remaining to receive the packet.
  *
- * @return Number of bytes received, or -1 on network error.
+ * @return Number of bytes received, or negative number on network error.
  */
 static int32_t recvExact( const MQTTContext_t * const pContext,
                           size_t bytesToRecv,
@@ -198,18 +198,6 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
                                     uint32_t timeoutMs,
                                     MQTTPacketInfo_t * const pIncomingPacket,
                                     bool * const pSessionPresent );
-
-/**
- * @brief Calculate the interval between two timestamps, including when the
- * later value has overflowed.
- *
- * @param[in] later The later time stamp.
- * @param[in] start The earlier time stamp.
- *
- * @return later - start.
- */
-static uint32_t calculateElapsedTime( uint32_t later,
-                                      uint32_t start );
 
 /*-----------------------------------------------------------*/
 
@@ -347,7 +335,7 @@ static int32_t recvExact( const MQTTContext_t * const pContext,
         {
             LogErrorWithArgs( "Network error while receiving packet: ReturnCode=%d",
                               bytesRecvd );
-            totalBytesRecvd = -1;
+            totalBytesRecvd = bytesRecvd;
             receiveError = true;
         }
 
@@ -811,23 +799,6 @@ static MQTTStatus_t sendPublish( MQTTContext_t * const pContext,
     }
 
     return status;
-}
-
-/*-----------------------------------------------------------*/
-
-static uint32_t calculateElapsedTime( uint32_t later,
-                                      uint32_t start )
-{
-    return later - start;
-}
-
-/*-----------------------------------------------------------*/
-
-static MQTTStatus_t receivePacket( MQTTContext_t * const pContext,
-                                   MQTTPacketInfo_t incomingPacket,
-                                   uint32_t remainingTimeMs )
-{
-    return MQTTSuccess;
 }
 
 /*-----------------------------------------------------------*/
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -67,7 +67,7 @@ static MQTTStatus_t validateSubscribeUnsubscribeParams( const MQTTContext_t * co
  * @return #MQTTSendFailed if transport write failed;
  * #MQTTSuccess otherwise.
  */
-static MQTTStatus_t sendPublish( const MQTTContext_t * const pContext,
+static MQTTStatus_t sendPublish( MQTTContext_t * const pContext,
                                  const MQTTPublishInfo_t * const pPublishInfo,
                                  size_t headerSize );
 
@@ -233,6 +233,7 @@ static int32_t recvExact( const MQTTContext_t * const pContext,
     bool shouldBreak = false;
 
     assert( pContext != NULL );
+    assert( bytesToRecv <= pContext->networkBuffer.size );
     pIndex = pContext->networkBuffer.pBuffer;
     recvFunc = pContext->transportInterface.recv;
     getTimeStamp = pContext->callbacks.getTime;
@@ -244,7 +245,7 @@ static int32_t recvExact( const MQTTContext_t * const pContext,
                                pIndex,
                                bytesRemaining );
 
-        if( bytesRecvd > 0 )
+        if( bytesRecvd >= 0 )
         {
             bytesRemaining -= ( size_t ) bytesRecvd;
             totalBytesRecvd += ( int32_t ) bytesRecvd;
@@ -252,14 +253,15 @@ static int32_t recvExact( const MQTTContext_t * const pContext,
         }
         else
         {
-            IotLogError( "Network error while receiving packet." );
+            LogError( "Network error while receiving packet." );
             totalBytesRecvd = -1;
             shouldBreak = true;
         }
 
-        if( calculateElapsedTime( getTimeStamp(), entryTime ) > timeoutMs )
+        if( ( bytesRemaining > 0U) && 
+            ( calculateElapsedTime( getTimeStamp(), entryTime ) > timeoutMs ) )
         {
-            IotLogError( "Time expired while receiving packet." );
+            LogError( "Time expired while receiving packet." );
             shouldBreak = true;
         }
 
@@ -311,10 +313,10 @@ static MQTTStatus_t discardPacket( MQTTContext_t * const pContext,
 
         if( bytesReceived != ( int32_t ) bytesToReceive )
         {
-            IotLogErrorWithArgs( "Receive error while discarding packet."
-                                 "ReceivedBytes=%d, ExpectedBytes=%u.",
-                                 bytesReceived,
-                                 bytesToReceive );
+            LogErrorWithArgs( "Receive error while discarding packet."
+                              "ReceivedBytes=%d, ExpectedBytes=%u.",
+                              bytesReceived,
+                              bytesToReceive );
             shouldBreak = true;
         }
         else
@@ -328,7 +330,7 @@ static MQTTStatus_t discardPacket( MQTTContext_t * const pContext,
             }
             else
             {
-                IotLogError( "Time expired while discarding packet." );
+                LogError( "Time expired while discarding packet." );
                 shouldBreak = true;
             }
         }
@@ -341,8 +343,8 @@ static MQTTStatus_t discardPacket( MQTTContext_t * const pContext,
 
     if( totalBytesReceived == remainingLength )
     {
-        IotLogErrorWithArgs( "Dumped packet. DumpedBytes=%d.",
-                             totalBytesReceived );
+        LogErrorWithArgs( "Dumped packet. DumpedBytes=%d.",
+                          totalBytesReceived );
         /* Packet dumped, so no data is available. */
         status = MQTTNoDataAvailable;
     }
@@ -373,10 +375,10 @@ static MQTTStatus_t receivePacket( MQTTContext_t * const pContext,
 
     if( incomingPacket.remainingLength > pContext->networkBuffer.size )
     {
-        IotLogErrorWithArgs( "Incoming packet length %u exceeds network buffer size %u."
-                             "Incoming packet will be dumped.",
-                             incomingPacket.remainingLength
-                             pContext->networkBuffer );
+        LogErrorWithArgs( "Incoming packet length %u exceeds network buffer size %u."
+                          "Incoming packet will be dumped.",
+                          incomingPacket.remainingLength,
+                          pContext->networkBuffer );
         status = discardPacket( pContext,
                                 incomingPacket.remainingLength,
                                 remainingTimeMs );
@@ -388,15 +390,15 @@ static MQTTStatus_t receivePacket( MQTTContext_t * const pContext,
         if( bytesReceived == ( int32_t ) bytesToReceive )
         {
             /* Receive successful, bytesReceived == bytesToReceive. */
-            IotLogInfoWithArgs( "Packet received. ReceivedBytes=%d.",
-                                bytesReceived );
+            LogInfoWithArgs( "Packet received. ReceivedBytes=%d.",
+                             bytesReceived );
         }
         else
         {
-            IotLogErrorWithArgs( "Packet reception failed. ReceivedBytes=%d, "
-                                 "ExpectedBytes=%u.",
-                                 bytesReceived,
-                                 bytesToReceive );
+            LogErrorWithArgs( "Packet reception failed. ReceivedBytes=%d, "
+                              "ExpectedBytes=%u.",
+                              bytesReceived,
+                              bytesToReceive );
             status = MQTTRecvFailed;
         }
         
@@ -474,15 +476,19 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * const pContext,
 
             if( newState == MQTTStateNull )
             {
+                LogErrorWithArgs( "Failed to update state of publish %u.",
+                                  packetId );
                 status = MQTTIllegalState;
             }
         }
         else
         {
-            IotLogErrorWithArgs( "Failed to send ACK packet: SentBytes=%d, "
-                                 "PacketSize=%u",
-                                 bytesSent,
-                                 MQTT_PUBLISH_ACK_PACKET_SIZE );
+            LogErrorWithArgs( "Failed to send ACK packet: PacketType=%02x, "
+                              "SentBytes=%d, "
+                              "PacketSize=%u",
+                              packetTypeByte,
+                              bytesSent,
+                              MQTT_PUBLISH_ACK_PACKET_SIZE );
             status = MQTTSendFailed;
         }
     }
@@ -529,9 +535,6 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * const pContext )
         else
         {
             status = MQTT_Ping( pContext );
-            /* Sending the packet may have taken some time. Recalculate. */
-            pContext->pingReqSendTimeMs = pContext->callbacks.getTime();
-            pContext->waitingForPingResp = true;
         }
     }
 
@@ -560,7 +563,7 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
     assert( pIncomingPacket != NULL );
 
     status = MQTT_DeserializePublish( pIncomingPacket, &packetIdentifier, &publishInfo );
-    IotLogInfoWithArgs( "De-serialized incoming PUBLISH packet: DeserializerResult=%d", status );
+    LogInfoWithArgs( "De-serialized incoming PUBLISH packet: DeserializerResult=%d", status );
 
     if( status == MQTTSuccess )
     {
@@ -568,8 +571,8 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
                                                       packetIdentifier,
                                                       MQTT_RECEIVE,
                                                       publishInfo.qos );
-        IotLogInfoWithArgs( "State record updated. New state=%d.",
-                            publishRecordState );
+        LogInfoWithArgs( "State record updated. New state=%d.",
+                         publishRecordState );
 
         /* Send PUBACK or PUBREC if necessary. */
         status = sendPublishAcks( pContext,
@@ -617,16 +620,15 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * const pContext,
         case MQTT_PACKET_TYPE_PUBCOMP:
             ackType = getAckFromPacketType( pIncomingPacket->type );
             status = MQTT_DeserializeAck( pIncomingPacket, &packetIdentifier,  &sessionPresent );
-            IotLogInfoWithArgs( "Ack packet deserialized with result: %d.",
-                                status );
+            LogInfoWithArgs( "Ack packet deserialized with result: %d.", status );
             if( status == MQTTSuccess )
             {
                 publishRecordState = MQTT_UpdateStateAck( pContext,
                                                           packetIdentifier,
                                                           ackType,
                                                           MQTT_RECEIVE );
-                IotLogInfoWithArgs( "State record updated. New state=%d.",
-                                    publishRecordState );
+                LogInfoWithArgs( "State record updated. New state=%d.",
+                                 publishRecordState );
 
                 /* Send PUBREL or PUBCOMP if necessary. */
                 status = sendPublishAcks( pContext,
@@ -700,7 +702,7 @@ static MQTTStatus_t validateSubscribeUnsubscribeParams( const MQTTContext_t * co
 
 /*-----------------------------------------------------------*/
 
-static MQTTStatus_t sendPublish( const MQTTContext_t * const pContext,
+static MQTTStatus_t sendPublish( MQTTContext_t * const pContext,
                                  const MQTTPublishInfo_t * const pPublishInfo,
                                  size_t headerSize )
 {
@@ -1158,6 +1160,8 @@ MQTTStatus_t MQTT_Ping( MQTTContext_t * const pContext )
         }
         else
         {
+            pContext->pingReqSendTimeMs = pContext->lastPacketTime;
+            pContext->waitingForPingResp = true;
             LogDebugWithArgs( "Sent %d bytes of PINGREQ packet.",
                               bytesSent );
         }
@@ -1289,7 +1293,7 @@ MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
 {
     MQTTStatus_t status = MQTTBadParameter;
     MQTTGetCurrentTimeFunc_t getTimeStamp = NULL;
-    uint32_t entryTime = 0, remainingTimeMs = timeoutMs;
+    uint32_t entryTime = 0U, remainingTimeMs = timeoutMs, elapsedTime = 0U;
     MQTTPacketInfo_t incomingPacket;
 
     if( pContext != NULL )
@@ -1300,11 +1304,10 @@ MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
     }
     else
     {
-        IotLogError( "MQTT Context cannot be NULL." );
+        LogError( "MQTT Context cannot be NULL." );
     }
     
-    while( ( pContext != NULL ) && 
-           ( calculateElapsedTime( getTimeStamp(), entryTime ) < timeoutMs ) )
+    while( status == MQTTSuccess )
     {
         status = MQTT_GetIncomingPacketTypeAndLength( pContext->transportInterface.recv,
                                                       pContext->transportInterface.networkContext,
@@ -1324,8 +1327,8 @@ MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
         }
         else if( status != MQTTSuccess )
         {
-            IotLogErrorWithArgs( "Receiving incoming packet length failed. Status=%d",
-                                 status );
+            LogErrorWithArgs( "Receiving incoming packet length failed. Status=%d",
+                              status );
         }
         else
         {
@@ -1333,7 +1336,7 @@ MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
             status = receivePacket( pContext, incomingPacket, remainingTimeMs );
         }
 
-        /* Handle received packet. */
+        /* Handle received packet. If no data was read then this will not execute. */
         if( status == MQTTSuccess )
         {
             incomingPacket.pRemainingData = pContext->networkBuffer.pBuffer;
@@ -1351,20 +1354,23 @@ MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
 
         if( status == MQTTNoDataAvailable )
         {
-            /* No data available is still a successful iteration. */
+            /* No data available is not an error. Reset to MQTTSuccess so the
+             * return code will indicate success. */
             status = MQTTSuccess;
         }
 
         if( status != MQTTSuccess )
         {
-            /* Error, break. */
-            IotLogErrorWithArgs( "Exiting receive loop. Error status=%d",
-                                 status );
-            break;
+            LogErrorWithArgs( "Exiting receive loop. Error status=%d", status );
         }
 
-        /* timeoutMs > elapsedTime is checked in the loop condition. */
-        remainingTimeMs = timeoutMs - calculateElapsedTime( getTimeStamp(), entryTime );
+        /* Recalculate remaining time and check if loop should exit. */
+        elapsedTime = calculateElapsedTime( getTimeStamp(), entryTime );
+        remainingTimeMs = timeoutMs - elapsedTime;
+        if( elapsedTime > timeoutMs )
+        {
+            break;
+        }
     }
 
     return status;
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -290,7 +290,7 @@ static MQTTStatus_t discardPacket( MQTTContext_t * const pContext,
     MQTTStatus_t status = MQTTRecvFailed;
     int32_t bytesReceived = 0;
     size_t bytesToReceive = 0U;
-    uint32_t totalBytesReceived, entryTime, elapsedTime;
+    uint32_t totalBytesReceived = 0U, entryTime, elapsedTime;
     uint32_t remainingTimeMs = timeoutMs;
     MQTTGetCurrentTimeFunc_t getTimeStamp = NULL;
     bool shouldBreak = false;
@@ -309,7 +309,7 @@ static MQTTStatus_t discardPacket( MQTTContext_t * const pContext,
 
         bytesReceived = recvExact( pContext, bytesToReceive, remainingTimeMs );
 
-        if( bytesReceived != bytesToReceive )
+        if( bytesReceived != ( int32_t ) bytesToReceive )
         {
             IotLogErrorWithArgs( "Receive error while discarding packet."
                                  "ReceivedBytes=%d, ExpectedBytes=%u.",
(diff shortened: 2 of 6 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -191,10 +506,49 @@ MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext )
...
+            incomingPacket.pRemainingData = pContext->networkBuffer.pBuffer;
+            status = _handleMessage( pContext, &incomingPacket );
+        }
+
+        if( status != MQTTSuccess && status != MQTTNoDataAvailable )
+        {
+            /* Error, break. */
+            IotLogErrorWithArgs( "Exiting receive loop. Error status=%d",
+                                 status );
+            /* TODO signal connection to close? */
```
abhidixi11, line 546: Here, MQTT_GetIncomingPacketTypeAndLength() would have returned one of the network error type.  For now, application can deal with various errors and decide to close and reconnect. When the library starts bubbling up error code, library should return only one error code and application can then look at bubbled up error code from the transport layer.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -431,3 +431,9 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * const pIncomingPacket
 
     return MQTTSuccess;
 }
+
+MQTTStatus_t MQTT_GetIncomingPacketTypeAndLength( MQTTTransportRecvFunc_t readFunc,
```
abhidixi11, line 435: This needs to implemented. The length computation logic can be directly borrowed for V4. Only change needed is instead of getNextByte(), use transport_read with one byte.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -61,6 +64,318 @@ static int32_t sendPacket( MQTTContext_t * pContext, size_t bytesToSend )
...
+{
+    MQTTStatus_t status = MQTTSuccess;
+    MQTTPublishState_t publishRecordState = MQTTStateNull;
+    MQTTPubAckType_t ackType;
+    MQTTPublishInfo_t publishInfo;
+
+    assert( pContext != NULL );
+    assert( pIncomingPacket != NULL );
+
+    /* Determine ack type here so next switch doesn't duplicate code. */
```
abhidixi11, line 293: You can still combine two switch statements like this:
     switch( pIncomingPacket->type )
    {
              case MQTT_PACKET_TYPE_PUBLISH:
              /* Process Publish */
              break;
             case MQTT_PACKET_TYPE_PUBACK:
            ackType = MQTTPuback;
            /* Note no break */
            case MQTT_PACKET_TYPE_PUBREC:
            ackType = MQTTPubrec;
           
        case MQTT_PACKET_TYPE_PUBREL:
            ackType = MQTTPubrel;
            
        case MQTT_PACKET_TYPE_PUBCOMP:
            ackType = MQTTPubcomp;
      /* common ack processing code */
      break;
      default:
      /* error */

}

muneebahmed10: I'm not sure this would work. Wouldn't all cases get overwritten by `MQTTPubcomp` after the fallthrough? I think I'll make it a separate helper function, something like `_getAckFromPacketType`, since the second switch increases complexity a bit.

abhidixi11: I was suggesting to add common code after  the switch actually. ( You can create  bool ackTypeSet ).  Helper function is probably better option though, will reduce complexity as well.

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: Implemented as part of #911

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Removed comment

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -75,6 +77,387 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        /* Update number of bytes to receive. */
+        totalBytesReceived += bytesReceived;
+    }
+
+    if( totalBytesReceived == ( int32_t ) remainingLength )
+    {
+        IotLogInfoWithArgs( "Dumped packet. Dumped bytes=%d",
+                            totalBytesReceived );
+        /* Packet dumped, so no data is available. */
+        status = MQTTNoDataAvailable;
```
abhidixi11, line 163: This status code indicate that packet could not fit the provided buffer, so that application can take some action like increase buffer size.  MQTTNoDataAvailable does not really convey that message.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -75,6 +77,387 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                                          &publishRecordState );
+                if( status == MQTTSuccess )
+                {
+                    /* TODO: There doesn't seem to be a separate callback for acks. */
+                    pContext->callbacks.appCallback( pContext, pIncomingPacket );
+                }
+            }
+            break;
+        case MQTT_PACKET_TYPE_PINGRESP:
+            /* TODO: should this be handled here? pContext->waitingForPingResp = false; */
```
abhidixi11, line 448: We need to call app callback for ping response. Since this loop does not process keep-alive ( does not send ping requests) the responses will be handled by the application.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: The intention here was to log a warning when the packet exceeds the buffer, dump the packet, and then continue the process loop. That's why the status is reset to `MQTTNoDataAvailable`, to indicate that there is no data left to be read and to continue the process loop should continue. However, if the intention is to stop the loop after dumping a packet, then I agree in changing this status, as it would give an error code back to the application.

muneebahmed10: Changed to `MQTTBufferExceeded` so error will bubble back to application.

abhidixi11: OK I looked up V3 as well as v1.4x. and the code silently drops the packet if it exceeds the buffer ( no bubble up) . (https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/f50ae301378ad2546b966521d433ebfbc185c4bf/src/aws_iot_mqtt_client_common_internal.c#L450) After giving some thought, I like the earlier approach. We should change it back to the original behavior and log error, so that  we don't terminate the connection on spurious large packets.

muneebahmed10: Reverted

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+
+/**
+ * @brief Receive a packet from the transport interface.
+ *
+ * @param[in] pContext MQTT Connection context.
+ * @param[in] incomingPacket packet struct with remaining length.
+ *
+ * @return MQTTSuccess or MQTTRecvFailed.
+ */
+static MQTTStatus_t receivePacket( MQTTContext_t * const pContext,
```
abhidixi11, line 203: This function should be changed into a loop with timeout (timeout would be remaining time of the process loop. )
Here is an example: (https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/d039f075e1cc2a2a7fc20edc6868f328d8d36b2f/platform/linux/mbedtls/network_mbedtls_wrapper.c#L318)
recvExact was similar ( which is deleted in another PR)

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -504,10 +927,86 @@ MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext )
...
+        getCurrentTime = pContext->callbacks.getTime;
+        entryTime = getCurrentTime();
+        status = MQTTSuccess;
+    }
+    else
+    {
+        IotLogError( "MQTT Context cannot be NULL." );
+    }
+    
+    while( ( pContext != NULL ) && ( ( getCurrentTime() - entryTime ) < timeoutMs ) )
```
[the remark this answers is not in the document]
abhidixi11: getCurrentTime() can be implemented for MS resolution by using system start time as epoch ( or a choose an epoch so that we can fit in MS resolution). We don't need wallclock time. For now we will leave the usage like this since we will implement is when demo is being developed.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        }
+
+        bytesReceived = pContext->transportInterface.recv( pContext->transportInterface.networkContext,
+                                                           pContext->networkBuffer.pBuffer,
+                                                           bytesToReceive );
+        if( bytesReceived != ( int32_t ) bytesToReceive )
+        {
+            /* Partial receive occurred while trying to dump packet. */
+            IotLogErrorWithArgs( "Partial receive while dumping packet."
+                                 "Received bytes=%d, Expected bytes=%u",
```
aggarw13, line 171: I suggest avoiding whitespaces in the data section of the log (as shown in the examples of the logging standard) to avoid unnecessary characters. (We have been following the same in HTTP). Thus, I suggest changing it to: `ReceivedBytes=%s, ExpectedBytes=%d`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+
+    while( totalBytesReceived < ( int32_t ) remainingLength )
+    {
+        /* Update number of bytes to receive. */
+        if( ( remainingLength - ( size_t ) totalBytesReceived ) < bytesToReceive )
+        {
+            bytesToReceive = remainingLength - ( size_t ) totalBytesReceived;
+        }
+
+        bytesReceived = pContext->transportInterface.recv( pContext->transportInterface.networkContext,
```
aggarw13, line 164: Even though this function calls `recv`, it is named as `dump` which to me gives a different meaning of _sending_ packets over the network. Can you update the name to `receivePacket` for better clarity?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                                                           pContext->networkBuffer.pBuffer,
+                                                           bytesToReceive );
+        if( bytesReceived != ( int32_t ) bytesToReceive )
+        {
+            /* Partial receive occurred while trying to dump packet. */
+            IotLogErrorWithArgs( "Partial receive while dumping packet."
```
aggarw13, line 170: I'm afraid that the "What?" part of the context (or the failure outcome) is missing in the log message here. Can you change to something like: `Unable to receive MQTT packet: Partial receive from transport recv() : ExpectedBytes...` to clearly mention the <Failure-Outcome> and <Cause> sections in the message?

libraries/standard/mqtt/src/mqtt.c
abhidixi11: It is actually dumping the packet, if it does not fit the buffer. (dump == throw away)

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+ * @return #MQTTRecvFailed or #MQTTNoDataAvailable.
+ */
+static MQTTStatus_t dumpPacket( MQTTContext_t * const pContext,
+                                size_t remainingLength,
+                                int32_t bytesAlreadyReceived )
+{
+    MQTTStatus_t status = MQTTRecvFailed;
+    int32_t totalBytesReceived = bytesAlreadyReceived, bytesReceived = 0;
+    assert( pContext != NULL );
+    size_t bytesToReceive = pContext->networkBuffer.size;
```
aggarw13, line 154: The coding standard requires definition of all local variables before any logic in the function. Thus, we have kept `assert()` calls after the local variable definitions in HTTP functions.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+            break;
+        }
+
+        /* Update number of bytes to receive. */
+        totalBytesReceived += bytesReceived;
+    }
+
+    if( totalBytesReceived == ( int32_t ) remainingLength )
+    {
+        IotLogErrorWithArgs( "Dumped packet. Dumped bytes=%d",
```
aggarw13, line 184: Should this be a `Warn` or `Debug` level message instead of an error? It seems like as the intention of this function is to discard packets, then this branch should be the happy-path for this function

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+ * @param[in] bytesAlreadyReceived Bytes previously read from the transport.
+ *
+ * @return #MQTTRecvFailed or #MQTTNoDataAvailable.
+ */
+static MQTTStatus_t dumpPacket( MQTTContext_t * const pContext,
+                                size_t remainingLength,
+                                int32_t bytesAlreadyReceived )
+{
+    MQTTStatus_t status = MQTTRecvFailed;
+    int32_t totalBytesReceived = bytesAlreadyReceived, bytesReceived = 0;
```
aggarw13, line 152: `remainingLength` a

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+
+    bytesReceived = pContext->transportInterface.recv( pContext->transportInterface.networkContext,
+                                                       pContext->networkBuffer.pBuffer,
+                                                       bytesToReceive );
+    totalBytesReceived = bytesReceived;
+
+    /* Partial receive. */
+    if( bytesReceived < ( int32_t ) bytesToReceive )
+    {
+        IotLogWarnWithArgs( "Packet partially received: Bytes received=%d,"
```
aggarw13, line 234: Suggest changing to `Received partial packet from transport recv(): BytesReceived=..."

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                            "Bytes expected=%u",
+                            bytesReceived,
+                            bytesToReceive );
+        /* Handle partial receive. */
```
aggarw13, line 238: What is the design reason for the single attempt to resolve the partial receive problem? Can we depend on the single re-attempt to resolve partial network reads? If we can't, I don't see the value of the single re-attempt to resolve the partial read...

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                                 "received=%d, Bytes expected=%u",
+                                 totalBytesReceived,
+                                 bytesToReceive );
+            status = MQTTRecvFailed;
+        }
+    }
+    /* Too many bytes received. */
+    else if( bytesReceived > ( int32_t ) bytesToReceive )
+    {
+        IotLogErrorWithArgs( "Too many bytes received. Bytes received=%d, ",
```
aggarw13, line 255: Log message missing the "Failure-Outcome" section. Suggest using: `Failure in receving MQTT packet: Received more too many bytes from transport recv(): BytesReceived=%d, BytesExpected=%u)`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        {
+            pContext->controlPacketSent = true;
+            newState = MQTT_UpdateStateAck( pContext,
+                                            packetId,
+                                            packetType,
+                                            MQTT_SEND );
+
+            if( newState != expectedNextState )
+            {
+                status = MQTTIllegalState;
```
aggarw13, line 352: Please add error logs for error branches

libraries/standard/mqtt/src/mqtt.c
aggarw13: My suggestion applies to all logs in the file for their data format

libraries/standard/mqtt/src/mqtt.c
abhidixi11: Yes, correct, please see my comment above about the loop. This code will change

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                            bytesReceived );
+    }
+
+    /* Check if packet exceeds buffer. */
+    if( ( status == MQTTSuccess ) && ( incomingPacket.remainingLength > bytesToReceive ) )
+    {
+        /* Packet exceeds buffer, dump it. */
+        status = dumpPacket( pContext, incomingPacket.remainingLength, totalBytesReceived );
+    }
+
```
aggarw13, line 274: An empty-else branch is needed for MISRA 15.7 compliance.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Send acks for received QoS 1/2 publishes.
+ *
+ * @param[in] pContext MQTT Connection context.
+ * @param[in] packetId, packet ID of original PUBLISH.
```
aggarw13, line 284: Typo with comma present in the param name

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+ * @param[in,out] pPublishState Current/updated publish state in record.
+ *
+ * @return MQTTSuccess or MQTTIllegalState.
+ */
+static MQTTStatus_t sendPublishAcks( MQTTContext_t * const pContext,
```
aggarw13, line 289: `MQTTContext_t * const` represents top-level constness which does carry any value (as the pointer-value is copied into the parameter). I suggest removing the top-level constness to just use `MQTTContext_t * `

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        {
+            bytesSent = sendPacket( pContext,
+                                    pContext->networkBuffer.pBuffer,
+                                    MQTT_PUBLISH_ACK_PACKET_SIZE );
+        }
+
+        if( bytesSent == MQTT_PUBLISH_ACK_PACKET_SIZE )
+        {
+            pContext->controlPacketSent = true;
+            newState = MQTT_UpdateStateAck( pContext,
```
aggarw13, line 345: If this _expects_ a next state, then it means that it is aware of the the state transition information. I would expect the `MQTT_UpdateStateAck` to encapsulate that information, and this function using whatever next state information is provided by it. In this case, the state-transition information is duplicated.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+    MQTTStatus_t status = MQTTSuccess;
+    MQTTPublishState_t publishRecordState = MQTTStateNull;
+    uint16_t packetIdentifier;
+    MQTTPublishInfo_t publishInfo;
+
+    assert( pContext != NULL );
+    assert( pIncomingPacket != NULL );
+
+    status = MQTT_DeserializePublish( pIncomingPacket, &packetIdentifier, &publishInfo );
+    IotLogInfoWithArgs( "Publish packet deserialized with result %d.", status );
```
aggarw13, line 431: `De-serialized incoming PUBLISH packet: DeserializerResult=%d`

aggarw13: Also, please log the stringified version of the result. The MQTT library needs a `strerror` utility

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        {
+            IotLogErrorWithArgs( "Handling partial receive failed. Bytes "
+                                 "received=%d, Bytes expected=%u",
+                                 totalBytesReceived,
+                                 bytesToReceive );
+            status = MQTTRecvFailed;
+        }
+    }
+    /* Too many bytes received. */
+    else if( bytesReceived > ( int32_t ) bytesToReceive )
```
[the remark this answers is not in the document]
muneebahmed10: Removed

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Fixed

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Renamed

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Removed log of partial receive when refactoring

libraries/standard/mqtt/src/mqtt.c

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: No, although the intention of the function is to discard packets, this function should not be called in ideal execution of the process loop. It's mean to handle the error case when the remaining length exceeds the buffer length, and then continue execution

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Implemented timeout loop

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,427 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                             "Incoming packet will be dumped",
+                             incomingPacket.remainingLength
+                             pContext->networkBuffer );
+    }
+    else
+    {
+        bytesToReceive = incomingPacket.remainingLength;
+    }
+
+    bytesReceived = pContext->transportInterface.recv( pContext->transportInterface.networkContext,
```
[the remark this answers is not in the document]
muneebahmed10: Refactored

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Removed partial receive log when refactoring

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Branch removed as it will never execute

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Why? There aren't any `else if`s

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: The `const` is unnecessary, but I believe it was added for MISRA compliance.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
muneebahmed10: Renamed to `getTimeStamp()` to avoid some possible confusion.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: This was refactored, but there has to be a cast either way when adding `bytesReceived`

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Right, `MQTT_UpdateStateAck` validates the state transition, the only important thing is to make sure `MQTTStateNull` is not returned. Changed the check  to remove unnecessary code

aggarw13: Thanks

libraries/standard/mqtt/src/mqtt.c
aggarw13: Oh, nevermind then

libraries/standard/mqtt/src/mqtt.c
aggarw13: Yes, a single cast in calling the function. My suggestion was for the function signature to better represent its input parameters

libraries/standard/mqtt/src/mqtt.c
aggarw13: Can you share the MISRA violation? I have usually seen MISRA violations that suggest constness only for the underlying type when the pointed to memory is not modified in the function (i.e. in this case as `const MQTTContext_t *` type)

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,509 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+    }
+    else
+    {
+        bytesToReceive = incomingPacket.remainingLength;
+        bytesReceived = recvExact( pContext, bytesToReceive, remainingTimeMs );
+        if( bytesReceived == ( int32_t ) bytesToReceive )
+        {
+            /* Receive successful, bytesReceived == bytesToReceive. */
+            IotLogInfoWithArgs( "Packet received. ReceivedBytes=%d.",
+                                bytesReceived );
```
leegeth, line 332: I think we would need to point the buffer read to MQTTPacketInfo.pRemainingData

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,509 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                                            MQTT_SEND );
+
+            if( newState == MQTTStateNull )
+            {
+                status = MQTTIllegalState;
+            }
+        }
+        else
+        {
+            IotLogErrorWithArgs( "Failed to send ACK packet: SentBytes=%d, "
```
leegeth, line 422: Logging which type of Publish packet has failed here would help with debugging problems.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/901/files#diff-34cd787ed56bd43628017def1ba1bae0R1064

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,509 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                pContext->pingRespTimeoutMs )
+            {
+                status = MQTTKeepAliveTimeout;
+            }
+        }
+        else
+        {
+            status = MQTT_Ping( pContext );
+            /* Sending the packet may have taken some time. Recalculate. */
+            pContext->pingReqSendTimeMs = pContext->callbacks.getTime();
```
leegeth, line 473: Don't we need to check if the packet is sent successfully here before updating pingReqSendTimeMs and waitingForPingResp ?

leegeth, line 473: Minor: Can't we use lastPacketTime here instead of calling getTime() again?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -99,6 +101,509 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+
+        /* Send PUBACK or PUBREC if necessary. */
+        status = sendPublishAcks( pContext,
+                                  packetIdentifier,
+                                  &publishRecordState );
+    }
+
+    if( status == MQTTSuccess )
+    {
+        /* TODO: Should the publish info be passed instead? */
```
leegeth, line 522: What is the benefit with that?

libraries/standard/mqtt/src/mqtt.c
abhidixi11: I think this is correct, because in order to determine the timeout scenario, we need to know when the ping request was sent. See line 463 above.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -504,10 +1009,90 @@ MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext )
...
+        entryTime = getTimeStamp();
+        status = MQTTSuccess;
+    }
+    else
+    {
+        IotLogError( "MQTT Context cannot be NULL." );
+    }
+    
+    while( ( pContext != NULL ) && 
+           ( calculateElapsedTime( getTimeStamp(), entryTime ) < timeoutMs ) )
```
leegeth, line 1032: What is the expected behavior when 0 is passed as timeoutMs?

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: All of the information related to the publish is stored in the MQTTPublishInfo_t, including QoS, topic name, and message payload. The only thing the packet info includes is the packet type (PUBLISH + flags), and a pointer to the serialized data.

libraries/standard/mqtt/src/mqtt.c
leegeth: Noticed that this is already handled in MQTT_ProcessLoop.

libraries/standard/mqtt/src/mqtt.c
leegeth: We would need to know when the ping request was sent successfully right? That can be checked by adding if (status == MQTTSuccess) and setting both pingReqSendTimeMs and waitingForPingResp

muneebahmed10: Yes, since we know the last packet sent was a PINGREQ. Also, would it make more sense to move these lines to `MQTT_Ping()`?

leegeth: Yes. That will be a good change.

abhidixi11: Yeah you are correct. I read your " successfully == wait for ping response" . Yes we need ( status == MQTTSuccess) check

libraries/standard/mqtt/src/mqtt.c
leegeth: I also think then should definitely pass the publish info rather than the packet info so as the application can make use of the already de-serialized data.

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -216,7 +227,9 @@ MQTTStatus_t MQTT_Unsubscribe( MQTTContext_t * const pContext,
  */
 MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext );
 
-MQTTStatus_t MQTT_Process( MQTTContext_t * const pContext,
-                           uint32_t timeoutMs );
+MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
```
leegeth, line 230: Please add doxygen comments in here.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Changed the log. Agree with strerror but think it's outside the scope of this PR.

libraries/standard/mqtt/src/mqtt.c
abhidixi11: What do we do for the ack callbacks ? We are only using one callback for reporting all incoming events.

muneebahmed10: The issue is that MQTT_DeserializePublish() does not change `pIncomingPacket` in any way and stores the information in the publish info. If the application wants the deserialized publish information then it will need to call MQTT_DeserializePublish() again, which is wasteful since the data has already been deserialized and is available.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Moved to `MQTT_Ping()`

muneebahmed10: Moved

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: This should execute at least once even with 0 timeout, similar to a do-while. I'll move this check to inside the loop

libraries/standard/mqtt/include/mqtt.h
muneebahmed10: Added

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Added a log for the packet type byte.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Updated

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Added both app callback and keep alive. Not sure if this was your intention

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,435 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+    assert( pContext != NULL );
+    assert( pIncomingPacket != NULL );
+
+    switch( pIncomingPacket->type )
+    {
+        case MQTT_PACKET_TYPE_PUBACK:
+        case MQTT_PACKET_TYPE_PUBREC:
+        case MQTT_PACKET_TYPE_PUBREL:
+        case MQTT_PACKET_TYPE_PUBCOMP:
+            ackType = getAckFromPacketType( pIncomingPacket->type );
```
leegeth, line 618: minor:can this be moved inside if (status == MQTTSuccess)? Doesn't seems like it is needed if there is an error.

muneebahmed10: Yes it can, the reason I didn't put it there was because I didn't want to increase the complexity. I can move it if you think it should be done.

leegeth: I am ok with either apporach.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,447 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        case MQTTPubRelSend:
+            packetTypeByte = MQTT_PACKET_TYPE_PUBREL;
+            packetType = MQTTPubrel;
+            break;
+        case MQTTPubCompSend:
+            packetTypeByte = MQTT_PACKET_TYPE_PUBCOMP;
+            packetType = MQTTPubcomp;
+            break;
+        default:
+            /* Nothing to send. */
```
gkwicker, line 474: Is this expected to be possible in normal operation?  If not, should we log an error?  If so, I'd add elaborate this comment with more detail, e.g. _"This function is called for all possible states, take no action on these."_

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,447 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
+                                           MQTTPacketInfo_t * pIncomingPacket )
+{
+    MQTTStatus_t status = MQTTSuccess;
```
gkwicker, line 562: In general, I usually prefer to see values initialized with failure values rather than success ones.  In this case, you don't need an initializer at all since initialization is guaranteed at line 570.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,447 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+static MQTTStatus_t handleIncomingAck( MQTTContext_t * const pContext,
+                                       MQTTPacketInfo_t * pIncomingPacket )
+{
+    MQTTStatus_t status = MQTTSuccess;
```
gkwicker, line 605: Would prefer this to be initialized to a failure value, if possible.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,447 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+            if( status == MQTTSuccess )
+            {
+                pContext->callbacks.appCallback( pContext,
+                                                 pIncomingPacket,
+                                                 packetIdentifier,
+                                                 NULL );
+            }
+            break;
+        default:
+            /* Not an ack. */
```
gkwicker, line 670: Same comment as in `sendPublishAcks()`, since we're dealing with input from the network here be very clear what is possible/expected here.  Why would I ever be in a function called `handleIncomingAck()` for something which is `"Not an ack"`?   In network packet handling code, we'd usually expect one of three things in a `default:`

1. `assert()` for programming errors
1. log/return failure values for run-time ones
1. _detailed_ comments explaining why neither option above applies.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -657,10 +1204,92 @@ MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext )
...
+        }
+
+        if( status != MQTTSuccess )
+        {
+            LogErrorWithArgs( "Exiting receive loop. Error status=%d", status );
+        }
+
+        /* Recalculate remaining time and check if loop should exit. */
+        elapsedTime = calculateElapsedTime( getTimeStamp(), entryTime );
+        remainingTimeMs = timeoutMs - elapsedTime;
```
gkwicker, line 1285: This illustrates a couple of issues raised in my earlier comments – `calculateElapsedTime()` probably should not be a function since it just a subtraction; why aren't you using it on line 1285?  Also, `getTimeStamp()` should be named such that the units are obvious, e.g. `getTimeStampMs()`.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,447 @@ static int32_t sendPacket( MQTTContext_t * pContext,
 
 /*-----------------------------------------------------------*/
 
+static uint32_t calculateElapsedTime( uint32_t later, uint32_t start )
+{
+    return later - start;
```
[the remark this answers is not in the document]
muneebahmed10: I was hoping this question would come up.

Short answer: This is to handle overflows, and I felt that it would be easy to forget to cast `later - start` into a `uint32_t` every time we want to calculate an elapsed time, so I made a function to return a `uint32_t`.

Long answer: According to [6.3.1.8](http://www.open-std.org/JTC1/sc22/wg14/www/docs/n1256.pdf) of the C standard, both values should get promoted to signed integers in the subtraction. I ran the following test on my machine:
```c
    uint8_t two = 2u, eight = 8u;

    printf("two - eight: %u\n", two - eight);
    if( (two - eight) < 0 )
        printf("two - eight < 0\n");
    else
        printf("two - eight >= 0\n");

    printf("(uint8_t) (two - eight): %u\n", (uint8_t) (two-eight));
    if((uint8_t) (two-eight) < 0)
        printf("(uint8_t) (two - eight) < 0\n");
    else
        printf("(uint8_t) (two - eight) >= 0\n");
```
and this was the result:
```
two - eight: 4294967290
two - eight < 0
(uint8_t) (two - eight): 250
(uint8_t) (two - eight) >= 0
```

So, in the case where the  `later` value has overflowed, and `later < start`, `later - start` will give us the correct answer (since 6.2.5.9 says that arithmetic with unsigned operands will never overflow, and be reduced modulo UINT_MAX), but only if it's converted back into a `uint32_t`. We could do this typecast every time we calculate the elapsed time, but I felt it might be easy to forget to do this, and so it would be better to make it a function.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -41,6 +41,110 @@ static int32_t sendPacket( MQTTContext_t * pContext,
                            const uint8_t * pBufferToSend,
                            size_t bytesToSend );
 
+/**
+ * @brief Calculate the interval between two timestamps, including when the
+ * later value has overflowed.
+ *
+ * @param[in] later The later time stamp.
```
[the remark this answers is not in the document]
muneebahmed10: Milliseconds, will update

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -227,8 +236,25 @@ MQTTStatus_t MQTT_Unsubscribe( MQTTContext_t * const pContext,
...
+/**
+ * @brief Loop to receive packets from the transport interface.
+ *
+ * @param[in] pContext Initialized and connected MQTT context.
+ * @param[in] timeoutMs Minimum time in milliseconds that the receive loop will
+ * run, unless an error occurs.
+ *
+ * @return #MQTTBadParameter if context is NULL;
+ * #MQTTRecvFailed if a network error occurs during reception;
+ * #MQTTSendFailed if a network error occurs while sending an ACK;
```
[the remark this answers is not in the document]
muneebahmed10: Yes, if sending a PINGREQ. Thanks for the catch.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -41,6 +41,110 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+static MQTTPubAckType_t getAckFromPacketType( uint8_t packetType );
+
+/**
+ * @brief Receive bytes into the network buffer, with a timeout.
+ *
+ * @param[in] pContext Initialized MQTT Context.
+ * @param[in] bytesToRecv Number of bytes to receive.
+ * @param[in] timeoutMs Time remaining to receive the packet.
+ *
+ * @return Number of bytes received, or -1 on network error.
```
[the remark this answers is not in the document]
muneebahmed10: The transport read function returns "a negative number" if it encounters an error. I'm not sure if we will have different error codes but I can update the log to have that return value.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,447 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+            break;
+        case MQTT_PACKET_TYPE_PUBREL:
+            ackType = MQTTPubrel;
+            break;
+        case MQTT_PACKET_TYPE_PUBCOMP:
+            ackType = MQTTPubcomp;
+            break;
+        default:
+            /* Not an ack. */
+            assert( 0 );
```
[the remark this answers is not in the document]
muneebahmed10: This function, `getAckFromPacketType`, is only called from `handleIncoming Ack` after it has checked that the type is one of the four values. I will update the comment

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Yes, it is called for all possible states, but action is only required for states indicating an ack should be sent. I will update the comment.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Good catch. The only packet type the server should send is a PUBLISH (handled in another function, or one of the acks above. Anything else is a bad response. I will add an error log.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -130,6 +234,447 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        }
+
+        if( ( bytesRemaining > 0U) && 
+            ( calculateElapsedTime( getTimeStamp(), entryTime ) > timeoutMs ) )
+        {
+            LogError( "Time expired while receiving packet." );
+            shouldBreak = true;
+        }
+
+        if( shouldBreak )
```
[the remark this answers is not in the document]
muneebahmed10: Updated and renamed

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Renamed to `getTimeStampMs()`. See above for reasoning on making `calculateElapsedTime()` its own function

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Changed. The status is updated in each branch of the switch anyway so the initial value doesn't matter here, but I understand why it is a good practice.

libraries/standard/mqtt/src/mqtt.c

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
abhidixi11: We can potentially say value less than zero, rather than -1 to be aligned with transport layer spec.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
abhidixi11: We need good unit tests.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
muneebahmed10: Actually it was -1, because I based this off the original implementation of recvExact we had that returned -1. I changed it to bubble up the exact error code.

PR #908, opened 2020-04-30
Add HTTP plaintext demo

yourslab:
*Description of changes:*
Plaintext demo using new HTTP Client API.


demos/http/http_demo_plaintext/config.h
```diff
@@ -0,0 +1,43 @@
...
+#include "iot_logging_levels.h"
+#include "iot_logging.h"
+
+#define LOG_NAME         ( "HTTP_DEMO" )
+#define IOT_LOG_LEVEL    ( IOT_LOG_DEBUG )
+
+/* Define the IotLog logging interface to enabling logging.
+ * This demo maps the macro to the reference POSIX implementation for logging.
+ * Note: @ref LOG_NAME produces metadata for each log message. */
+#define IotLog( pFormat, ... )              \
```
yourslab, line 35: I'm just defining this generic IotLog with the preset parameters below. Is there a better way to do this @aggarw13?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,341 @@
...
+
+    returnStatus = HTTPClient_InitializeRequestHeaders( &requestHeaders,
+                                                        &requestInfo );
+
+    if( returnStatus == HTTP_SUCCESS )
+    {
+        /* Define the transport interface. */
+        transport.recv = transportRecv;
+        transport.send = transportSend;
+        transport.pContext = tcpSocket;
```
yourslab, line 246: This is giving me warnings because pContext is a struct while tcpSocket is an int. I think we need to avoid defining typedef HTTPNetworkContext_t in http_client.h and leave it up to application to define it. @sarenameas

sarenameas: We need the context to be struct that holds both the socket and the socket error.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,341 @@
...
+        freeaddrinfo( pListHead );
+    }
+
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief The transport send function provided to the MQTT context.
```
sarenameas, line 172: Need to update these MQTT related comments.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,341 @@
...
+ */
+#define PORT                        80
+
+/**
+ * @brief Paths for different HTTP methods for specified host.
+ *
+ * For httpbin.org, see http://httpbin.org/#/HTTP_Methods for details on
+ * supported REST API.
+ */
+#define GET_PATH                    "/ip"
```
sarenameas, line 57: Since we have `#ifdef` for these paths in main(), I think these will need `#ifndef..#endf` around them here. 
If these are not planned to be configurable we need to take out the `#ifdef` constructs from main().

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,353 @@
...
+ */
+#define PORT      80
+
+/**
+ * @brief Paths for different HTTP methods for specified host.
+ *
+ * For httpbin.org, see http://httpbin.org/#/HTTP_Methods for details on
+ * supported REST API.
+ */
+#ifndef GET_PATH
```
sarenameas, line 57: Since you made these configurable you will need to put back the #ifdef in the main

demos/http/http_demo_plaintext/config.h
aggarw13: This is fine for now

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+                            uint16_t port )
+{
+    int status, tcpSocket = -1;
+    struct addrinfo * pListHead = NULL, * pIndex;
+    struct sockaddr * pServerInfo;
+    uint16_t netPort = htons( port );
+    socklen_t serverInfoLength;
+
+    /* Perform a DNS lookup on the given host name. */
+    status = getaddrinfo( pServer, NULL, NULL, &pListHead );
```
gkwicker, line 126: Can this demo run on a FreeRTOS system?  We would usually use [`FreeRTOS_gethostbyname()`](https://github.com/FreeRTOS/FreeRTOS/blob/master/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/include/FreeRTOS_DNS.h#L117) for this when running on FreeRTOS.

yourslab: We will have separate demos for FreeRTOS. This demo is specific to C SDK.

demos/http/http_demo_plaintext/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+target_link_libraries(
+    ${DEMO_NAME}
+    PRIVATE
+        http
+        logging
+)
+
+target_include_directories(
+    logging
```
aggarw13, line 19: Why specifying the include directories for the `logging` and `http` targets here? Wouldn't these libraries have their include directories specified in `logging,cmake` and `http_parser.cmake` respectively?

yourslab: This is just so that `logging` and `http` can find `config.h`. The alternative is to add it as a source.

aggarw13: Okay, we can remove the dependency on `config.h` from `logging.c/.h` files to avoid.  For `http`, it is fine for now to do it this way.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+    #define POST_PATH    "/post"
+#endif
+
+/**
+ * @brief The length in bytes of the user buffer.
+ */
+#define USER_BUFFER_LENGTH          ( 512 )
+
+/**
+ * @brief Some text to send as the request body for a PUT and POST request in
```
aggarw13, line 79: (Nit) Grammatical typos in the sentence "..for ~a~ PUT and POST request"s" in this ~~examples~~ demo"

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+ * the request to store the response headers and body.
+ *
+ * User can also decide to have two separate buffers for request and response.
+ */
+static uint8_t userBuffer[ USER_BUFFER_LENGTH ] = { 0 };
+
+/**
+ * @brief The request body.
+ */
+static uint8_t requestBodyBuffer[] = REQUEST_BODY_TEXT;
```
aggarw13, line 96: No warnings for storing `char *` data in `uint8_t []` type?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+ * @param[in] pServer Host name of server.
+ * @param[in] port Server port.
+ *
+ * @return A file descriptor representing the TCP socket; -1 on failure.
+ */
+static int connectToServer( const char * pServer,
+                            uint16_t port )
+{
+    int status, tcpSocket = -1;
+    struct addrinfo * pListHead = NULL, * pIndex;
```
aggarw13, line 120: Default initialization of `pIndex` as `NULL`?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+    struct sockaddr * pServerInfo;
+    uint16_t netPort = htons( port );
+    socklen_t serverInfoLength;
+
+    /* Perform a DNS lookup on the given host name. */
+    status = getaddrinfo( pServer, NULL, NULL, &pListHead );
+
+    if( status != -1 )
+    {
```
aggarw13, line 129: Can you add logs in the demo?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+        freeaddrinfo( pListHead );
+    }
+
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief The transport send function that defines the transport interface.
```
aggarw13, line 191: It would be helpful for the customer to mention here that this would be used as the `HTTPTransportInterface.send` function to inject in the HTTP library in the demo,

Similarly, for the `transportRecv` function.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+ *
+ * @param[in] pContext User defined context (TCP socket for this demo).
+ * @param[in] pBuffer Buffer to write to the network stack.
```
aggarw13, line 194: (Nit) Buffer containing the bytes to send over the network stack.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+{
+    return ( int32_t ) recv( pContext->tcpSocket, pBuffer, bytesToRead, 0 );
+}
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Send an HTTP request based on a specified method and path.
+ *
+ * @param[in] tcpSocket TCP socket.
```
aggarw13, line 229: `tcpSocket` -> `pContext`

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+ * @param[in] pMethod The HTTP request method.
+ * @param[in] pPath The Request-URI to the objects of interest.
+ *
+ * @return #HTTP_SUCCESS if successful.
```
aggarw13, line 233: This seems incomplete; it can mention "; otherwise, the error status returned by the HTTP library"

demos/http/http_demo_plaintext/http_demo_plaintext.c
aggarw13: Referring to logs for this function

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+    HTTPRequestInfo_t requestInfo = { 0 };
+    HTTPTransportInterface_t transport = { 0 };
+    HTTPResponse_t response = { 0 };
+
+    SdkLog( "Sending HTTP %s request to %s%s\r\n",
+            pMethod, SERVER, pPath );
+
+    /* Initialize the request object. */
+    requestInfo.method = pMethod;
+    requestInfo.methodLen = sizeof( pMethod ) - 1;
```
aggarw13, line 250: This should use `strlen(pMethod)` instead as `sizeof(<pointer-type>)` would be the same for every underlying type,

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+                                                        &requestInfo );
+
+    if( returnStatus == HTTP_SUCCESS )
+    {
+        /* Define the transport interface. */
+        transport.recv = transportRecv;
+        transport.send = transportSend;
+        transport.pContext = pContext;
+
+        /* Initialize the response object. */
```
aggarw13, line 267: Can add a comment to mention with a Note that the buffer is being _re-used_ here,

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+    {
+        returnStatus = HTTP_NETWORK_ERROR;
+    }
+
+    /*********************** Send HTTPS request. ************************/
+
+    /* The client is now connected to the server. This example will send a
+     * GET, HEAD, PUT, and POST request. */
+
+    #ifdef GET_PATH
```
aggarw13, line 316: These pre-processor guards are not required as these macros will always be defined (There are defaults defined at the beginning of the file)

demos/http/http_demo_plaintext/http_demo_plaintext.c
sarenameas: Yea this is not a string literal, it is a pointer type, you will get the size f the pointer here.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,361 @@
...
+        transport.recv = transportRecv;
+        transport.send = transportSend;
+        transport.pContext = pContext;
+
+        /* Initialize the response object. */
+        response.pBuffer = userBuffer;
+        response.bufferLen = USER_BUFFER_LENGTH;
+
+        /* Send the request and receive the response. */
+        returnStatus = HTTPClient_Send( &transport,
```
sarenameas, line 275: For network errors we may want to retry depending on the error. 
```
if( returnStatus== HTTP_NETWORK_ERROR)
{
     //Check the type of network error specific to openssl and we may need to retry to close the connection
}

if( returnStatus== HTTP_SECURITY_ALERT_XXX_ERROR)
{
     //Log error and close the connection
}
```

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: No warnings

sarenameas:
> I only see the GET request being sent on running the demo locally. The remaining method requests are not sent..
> image

This is because HTTPClient_Send() is not in yet.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,553 @@
...
+
+    /* Send GET Request. */
+    if( returnStatus == EXIT_SUCCESS )
+    {
+        httpStatus = _sendHttpRequest( &transport,
+                                       SERVER_HOST,
+                                       HTTP_METHOD_GET,
+                                       GET_PATH );
+
+        if( httpStatus != HTTP_SUCCESS )
```
sarenameas, line 496: This is a little bit weird maybe we can just return EXIT_SUCCESS from _sendHttpRequest? Also because you are not using httpStatus in this function.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Used `strcpy` in main just in case

libraries/standard/http/CMakeLists.txt
```diff
@@ -7,8 +7,7 @@ add_library( http
 
 # HTTP public include path.
 target_include_directories( http PUBLIC ${HTTP_INCLUDE_PUBLIC_DIRS}
-                                        ${LOGGING_INCLUDE_DIRS}
-                                        ${CMAKE_CURRENT_LIST_DIR}/utest )
```
yourslab, line 11: This was initially a hack to include the config header when building http library but is no longer needed because the demo already has a config header included.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * @brief The length in bytes of the user buffer.
+ */
+#define USER_BUFFER_LENGTH                ( 1024 )
+
+/**
+ * @brief Length of an IPv6 address when converted to hex digits.
+ */
+#define IPV6_LENGTH                       ( 40 )
+
+
```
aggarw13, line 94: (Nit): Extra new line can be removed

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+/**
+ * @brief The length in bytes of the user buffer.
+ */
+#define USER_BUFFER_LENGTH                ( 1024 )
+
+/**
+ * @brief Length of an IPv6 address when converted to hex digits.
+ */
+#define IPV6_LENGTH                       ( 40 )
```
aggarw13, line 92: (Nit) This can be renamed as `IPV6_ADDRESS_STRING_LEN` to be explict

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+#define REQUEST_BODY_TEXT_LENGTH    ( sizeof( REQUEST_BODY_TEXT ) - 1 )
+
+/**
+ * @brief A string to store the resolved IP address from the host name.
+ */
+static char resolvedIpAddr[ IPV6_LENGTH ] = { 0 };
+
+/**
+ * @brief A buffer used to store request headers and reused after sending
+ * the request to store the response headers and body.
```
aggarw13, line 109: I think this message should be clearer as it is an important point being shown by the demo for space optimization. 
It could be something like
```
/**
 * @brief A buffer used in the demo for storing HTTP request headers and HTTP response headers and body.
 * 
 * @note This demo shows how the same buffer can be re-used for storing the HTTP response after the HTTP request 
 * is sent out. However, the user can also decided to use separate buffers for storing request and response.
 */
```

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ */
+struct HTTPNetworkContext
+{
+    int tcpSocket;
+};
+
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Performs a DNS lookup on the given host name then establishes a TCP
```
aggarw13, line 131: (Nit) Missing a comma between "host name" and "then"

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * @param[out] pTcpSocket Pointer to TCP socket file descriptor.
+ *
+ * @return EXIT_FAILURE on failure; EXIT_SUCCESS on success.
+ */
+static int connectToServer( const char * pServer,
+                            uint16_t port,
+                            int * pTcpSocket );
+
+/**
+ * @brief The transport send function that defines the transport interface.
```
aggarw13, line 145: (Nit) "The definition of the `send()` function in the transport interface. It is used for sending packets over the TCP transport layer."

The above description can be updated to mention about TLS in the other HTTP demos.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ *
+ * This is passed to the #HTTPTransportInterface.send function and used to
```
aggarw13, line 147: (Nit) This is passed ~~as~~ the #HTTPSTransportInterface.send...

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ *
+ * @return Number of bytes sent; negative value on error.
+ */
+static int32_t transportSend( HTTPNetworkContext_t * pContext,
+                              const void * pBuffer,
+                              size_t bytesToSend );
+
+/**
+ * @brief The transport receive function that defines the transport interface.
+ *
```
aggarw13, line 162: (Nit) To be consistent with the above function,  "This is passed as the #HTTPTransportInterface.recv function used for reading data received from the network."

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * send data over the network.
+ *
+ * @param[in] pContext User defined context (TCP socket for this demo).
+ * @param[in] pBuffer Buffer containing the bytes to send over the network stack.
+ * @param[in] bytesToSend Number of bytes to write to the network.
+ *
+ * @return Number of bytes sent; negative value on error.
+ */
+static int32_t transportSend( HTTPNetworkContext_t * pContext,
+                              const void * pBuffer,
```
aggarw13, line 157: Can this be of `const uint8_t *` type?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * @param[out] pBuffer Buffer to read network data into.
+ * @param[in] bytesToRead Number of bytes requested from the network.
+ *
+ * This is passed to the #HTTPTransportInterface.recv function and used to
+ * receive data over the network.
+ *
+ * @return Number of bytes received; negative value on error.
+ */
+static int32_t transportRecv( HTTPNetworkContext_t * pContext,
+                              void * pBuffer,
```
aggarw13, line 173: Can this be of `uint8_t * `type?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * @brief The transport send function that defines the transport interface.
+ *
+ * This is passed to the #HTTPTransportInterface.send function and used to
+ * send data over the network.
+ *
+ * @param[in] pContext User defined context (TCP socket for this demo).
+ * @param[in] pBuffer Buffer containing the bytes to send over the network stack.
+ * @param[in] bytesToSend Number of bytes to write to the network.
+ *
+ * @return Number of bytes sent; negative value on error.
```
aggarw13, line 154: (Nit) Number of bytes sent **if successful**; otherwise negative value on error

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * @brief Send an HTTP request based on a specified method and path.
+ *
+ * @param[in] pTransportInterface The transport interface for making network calls.
+ * @param[in] pHost The host name of the server.
+ * @param[in] pMethod The HTTP request method.
+ * @param[in] pPath The Request-URI to the objects of interest.
+ *
+ * @return EXIT_FAILURE on failure; EXIT_SUCCESS on success.
+ */
+static int sendHttpRequest( HTTPTransportInterface_t * pTransportInterface,
```
aggarw13, line 186: Can this be of const type `const HTTPTransportInterface_t *`? The members of this object shouldn't be changed

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+/*-----------------------------------------------------------*/
+
+static int connectToServer( const char * pServer,
+                            uint16_t port,
+                            int * pTcpSocket )
+{
+    int returnStatus = EXIT_SUCCESS;
+    struct addrinfo hints, * pIndex, * pListHead = NULL;
+    struct sockaddr * pServerInfo;
+    uint16_t netPort = htons( port );
```
aggarw13, line 200: Does this raise a warning for C89 compliance? C89 does not allow having declarations and "code" together (unfortunately)

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+    /* TCP Socket. */
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+
+    /* Perform a DNS lookup on the given host name. */
+    returnStatus = getaddrinfo( pServer, NULL, &hints, &pListHead );
+
+    if( returnStatus != -1 )
+    {
+        LogInfo( ( "Performing DNS lookup on %s.",
```
aggarw13, line 217: (nit) "Performing DNS lookup: Host=%s" to conform to logging standard

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                           resolvedIpAddr,
+                           sizeof( resolvedIpAddr ) );
+            }
+            else
+            {
+                /* IPv6 */
+                ( ( struct sockaddr_in6 * ) pServerInfo )->sin6_port = netPort;
+                serverInfoLength = sizeof( struct sockaddr_in6 );
+                inet_ntop( pServerInfo->sa_family,
+                           &( ( struct sockaddr_in * ) pServerInfo )->sin_addr,
```
aggarw13, line 248: Should this be cast to `struct sockaddr_in6`?
Has this been tested against an IPV6 address?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                           resolvedIpAddr,
+                           sizeof( resolvedIpAddr ) );
+            }
+
+            LogInfo( ( "Attempting to connect to resolved IP address: %s.",
```
aggarw13, line 253: (Nit) "Attempting to connect to server: Host=%s, IPAddress=%s

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                       resolvedIpAddr ) );
+
+            returnStatus = connect( *pTcpSocket, pServerInfo, serverInfoLength );
+
+            if( returnStatus == -1 )
+            {
+                close( *pTcpSocket );
```
aggarw13, line 260: LogError

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+            /* Fail if no connection could be established. */
+            returnStatus = EXIT_FAILURE;
+            LogError( ( "Could not connect to any resolved IP address from %.*s.\n",
+                        ( int ) strlen( pServer ),
+                        pServer ) );
+        }
+        else
+        {
+            returnStatus = EXIT_SUCCESS;
+            LogInfo( ( "TCP connection established with %.*s.\n",
```
aggarw13, line 281: (Nit) "Established TCP connection: Server=%s"

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+        transportTimeout.tv_usec = ( TRANSPORT_SEND_RECV_TIMEOUT_MS * 1000 );
+
+        /* Set the receive timeout. */
+        if( setsockopt( *pTcpSocket,
+                        SOL_SOCKET,
+                        SO_RCVTIMEO,
+                        ( char * ) &transportTimeout,
+                        sizeof( transportTimeout ) ) < 0 )
+        {
+            LogError( ( "Setting socket receive timeout failed." ) );
```
aggarw13, line 307: You can log the error code as well

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+
+/*-----------------------------------------------------------*/
+
+static int32_t transportSend( HTTPNetworkContext_t * pContext,
+                              const void * pBuffer,
+                              size_t bytesToSend )
+{
+    int32_t bytesSend = 0;
+
+    bytesSend = send( ( int ) pContext->tcpSocket, pBuffer, bytesToSend, 0 );
```
aggarw13, line 339: Is the `int` cast required? Isn't the `tcpSocket` already of `int` type?

aggarw13, line 339: (Nit) Should be bytesSent~~d~~t for the past tense word

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+
+    if( bytesSend < 0 )
+    {
+        /* Check if it was time out */
+        if( ( errno == EAGAIN ) || ( errno == EWOULDBLOCK ) )
+        {
+            /* Set return value to 0 to indicate that send was timed out. */
```
aggarw13, line 346: (nit) "...that send ~~was~~ had timed out"

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+
+/*-----------------------------------------------------------*/
+
+static int32_t transportRecv( HTTPNetworkContext_t * pContext,
+                              void * pBuffer,
+                              size_t bytesToRecv )
+{
+    int32_t bytesReceived = 0;
+
+    bytesReceived = ( int32_t ) recv( ( int ) pContext->tcpSocket, pBuffer, bytesToRecv, 0 );
```
aggarw13, line 362: Both `int32_t` and `int` casts seem to be redundant

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+
+    if( bytesReceived == 0 )
+    {
+        /* Server closed the connection, treat it as an error. */
+        bytesReceived = -1;
```
aggarw13, line 367: How can we assume that zero bytes sent is always due to connection termination by the server?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                            const char * pHost,
+                            const char * pMethod,
+                            const char * pPath )
+{
+    HTTPStatus_t httpStatus = HTTP_SUCCESS;
+    HTTPRequestHeaders_t requestHeaders = { 0 };
+    HTTPRequestInfo_t requestInfo = { 0 };
+    HTTPResponse_t response = { 0 };
+
+    assert( pTransportInterface != NULL );
```
aggarw13, line 398: As this is demo code, `assert` is not strictly required. If using `assert`, please add it in the other private functions as well

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                   ( int32_t ) REQUEST_BODY_TEXT_LENGTH,
+                   ( char * ) requestBodyBuffer ) );
+        /* Send the request and receive the response. */
+        httpStatus = HTTPClient_Send( pTransportInterface,
+                                      &requestHeaders,
+                                      requestBodyBuffer,
+                                      REQUEST_BODY_TEXT_LENGTH,
+                                      &response,
+                                      0 );
+    }
```
aggarw13, line 442: `else` with error logging?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                                                      &requestInfo );
+
+    if( httpStatus == HTTP_SUCCESS )
+    {
+        /* Initialize the response object. The same buffer used for storing
+         * request headers is reused here. */
+        response.pBuffer = userBuffer;
+        response.bufferLen = USER_BUFFER_LENGTH;
+
+        LogInfo( ( "Sending HTTP %s request to %s%s...",
```
aggarw13, line 427: For the server host and path combination, should it be formatted as "%s/%s"?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                   pMethod, SERVER_HOST, pPath ) );
+        LogInfo( ( "Request Headers:\n%.*s",
```
aggarw13, line 429: Could be logged at debug level instead as this could print a lot of characters

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+                   response.pHeaders ) );
+        LogInfo( ( "Response Status:\n%u",
+                   response.statusCode ) );
+        LogInfo( ( "Response Body:\n%.*s\n",
+                   ( int32_t ) response.bodyLen,
+                   response.pBody ) );
+    }
+    else
+    {
+        LogError( ( "Sending HTTP %s request to %s%s failed with status = %u.",
```
aggarw13, line 459: "Failed to send HTTP request: RequestType=%s, ErrorCode=%s"

Please log the stringified version of `httpStatus`

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * @brief Entry point of demo.
+ */
+int main()
+{
+    int returnStatus = EXIT_SUCCESS;
+    HTTPNetworkContext_t socketContext = { 0 };
+    HTTPTransportInterface_t transportInterface = { 0 };
+
+    /* Set the request body. */
+    strncpy( ( char * ) requestBodyBuffer,
```
aggarw13, line 485: You shouldn't need to store the string in a buffer, do you? Can you just pass the macro as a parameter directly?

libraries/standard/http/CMakeLists.txt
aggarw13: Can the logging include directory dependency be added in the demos as well?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+        LogInfo( ( "Request Headers:\n%.*s",
+                   ( int32_t ) requestHeaders.headersLen,
+                   ( char * ) requestHeaders.pBuffer ) );
+        LogInfo( ( "Request Body:\n%.*s\n",
+                   ( int32_t ) REQUEST_BODY_TEXT_LENGTH,
+                   ( char * ) requestBodyBuffer ) );
+        /* Send the request and receive the response. */
+        httpStatus = HTTPClient_Send( pTransportInterface,
+                                      &requestHeaders,
+                                      requestBodyBuffer,
```
aggarw13, line 438: You could just pass `REQUEST_BODY_TEXT` here instead

demos/http/http_demo_plaintext/http_demo_plaintext.c
abhidixi11: I think it should not. It is just initializing variable, but good point. We should make sure compiler runs with c89 option. Out of scope for this PR though.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * receive data over the network.
+ *
+ * @return Number of bytes received; negative value on error.
+ */
+static int32_t transportRecv( HTTPNetworkContext_t * pContext,
+                              void * pBuffer,
+                              size_t bytesToRecv );
+
+/**
+ * @brief Send an HTTP request based on a specified method and path.
```
abhidixi11, line 177: Here we must specify that this function will also process the response because API does process the response.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+    if( returnStatus == EXIT_SUCCESS )
+    {
+        transportInterface.recv = transportRecv;
+        transportInterface.send = transportSend;
+        transportInterface.pContext = &socketContext;
+    }
+
+    /*********************** Send HTTPS request. ************************/
+
+    /* The client is now connected to the server. This example will send a
```
abhidixi11, line 503: We  need to have detailed comments explaining what each sendHttpRequest() is doing and what is end result. In general please look how comments are written in MQTT demos for the flow of the demo:
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c#L816

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+ * @brief Some text to send as the request body for PUT and POST requests in
+ * this demo.
+ */
+#define REQUEST_BODY_TEXT           "Hello, world!"
+#define REQUEST_BODY_TEXT_LENGTH    ( sizeof( REQUEST_BODY_TEXT ) - 1 )
+
+/**
+ * @brief A string to store the resolved IP address from the host name.
+ */
+static char resolvedIpAddr[ IPV6_LENGTH ] = { 0 };
```
abhidixi11, line 105: I think this declaration ( { 0 } ) may not be allowed in c89.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+
+/**
+ * @brief A buffer used to store request headers and reused after sending
+ * the request to store the response headers and body.
+ *
+ * User can also decide to have two separate buffers for request and response.
+ */
+static uint8_t userBuffer[ USER_BUFFER_LENGTH ] = { 0 };
```
abhidixi11, line 113: I think this declaration ( { 0 } ) may not be allowed in c89.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+
+/**
+ * @brief The request body.
+ */
+static uint8_t requestBodyBuffer[ REQUEST_BODY_TEXT_LENGTH ] = { 0 };
```
abhidixi11, line 118: I think this declaration ( { 0 } ) may not be allowed in c89.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+
+static int sendHttpRequest( HTTPTransportInterface_t * pTransportInterface,
+                            const char * pHost,
+                            const char * pMethod,
+                            const char * pPath )
+{
+    HTTPStatus_t httpStatus = HTTP_SUCCESS;
+    HTTPRequestHeaders_t requestHeaders = { 0 };
+    HTTPRequestInfo_t requestInfo = { 0 };
+    HTTPResponse_t response = { 0 };
```
abhidixi11, line 396: I think this declaration ( { 0 } ) may not be allowed in c89.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+/*-----------------------------------------------------------*/
+
+/**
+ * @brief Entry point of demo.
+ */
+int main()
+{
+    int returnStatus = EXIT_SUCCESS;
+    HTTPNetworkContext_t socketContext = { 0 };
+    HTTPTransportInterface_t transportInterface = { 0 };
```
abhidixi11, line 482: I think this declaration ( { 0 } ) may not be allowed in c89.

libraries/standard/http/CMakeLists.txt
yourslab: Yes, `${LOGGING_INCLUDE_DIRS}` is already part.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Yes, Sarena told me about the change to NetworkContext_t. I can make the change for all the demos in a separate PR once we have NetworkContext_t defined.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Good catch

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Good catch

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Good catch

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Hmm that's a little strange because the only other error code here is -1, and this is consistent with MQTT logs as well.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: hmm not sure why it's casted either

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Took out `int` cast, but `int32_t` cast required by definition of transport receive.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: From sockets documentation, "When a stream socket peer has performed an orderly shutdown, the return value will be 0 (the traditional "end-of-file" return)."

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: No, the path already includes the leading /

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: The logs for printing the request are helpful for understanding what the demo does. Changing to debug level means these aren't printed at all. I also feel there aren't that many character being printed when using httpbin.org for this demo. If the user changes the endpoint and feels that there are too many characters being printed, they can always change the log level.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: To fix the above, I will make them static.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Does not raise a warning. I am using the following C flags:
```
-Wall -Wextra -Wstrict-prototypes -Wold-style-definition -Wmissing-prototypes -Wmissing-declarations -Wdeclaration-after-statement
```

demos/http/http_demo_plaintext/http_demo_plaintext.c
abhidixi11: Sure that's fine. We can create sim task for cleaning that up. We have address that for MQTT as well.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    if( httpStatus == HTTP_SUCCESS )
+    {
+        return EXIT_SUCCESS;
+    }
+    else
+    {
+        return EXIT_FAILURE;
+    }
```
muneebahmed10, line 441: I'm not sure if we're following MISRA for demos as well (I saw an empty else clause above) but if we are then we shouldn't have multiple returns either:

```suggestion
    return ( httpStatus == HTTP_SUCCESS )? EXIT_SUCCESS : EXIT_FAILURE;
```

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,524 @@
...
+                                        HTTP_METHOD_POST,
+                                        POST_PATH );
+    }
+
+    /************************** Disconnect. *****************************/
+
+    if( socketContext.tcpSocket != -1 )
+    {
+        shutdown( socketContext.tcpSocket, SHUT_RDWR );
+        close( socketContext.tcpSocket );
```
muneebahmed10, line 520: Should these be `( void )`'ed since they return `int`s? I don't believe we're running MISRA on demos though so it should be fine

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: I declared an extra returnStatus and added the following:
```c
if( httpStatus != HTTP_SUCCESS ){
        returnStatus = EXIT_FAILURE;
}
```

demos/http/http_demo_plaintext/demo_config.h
```diff
@@ -43,4 +43,65 @@
 
 /************ End of logging configuration ****************/
 
+/**
+ * @brief HTTP server host name.
```
aggarw13, line 47: Great change!

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -160,21 +125,25 @@ static int32_t transportSend( HTTPNetworkContext_t * pContext,
...
  * @brief The transport receive function that defines the transport interface.
  *
+ * This is passed as the #HTTPTransportInterface.recv function used for reading
+ * data received from the network.
+ *
  * @param[in] pContext User defined context (TCP socket for this demo).
  * @param[out] pBuffer Buffer to read network data into.
  * @param[in] bytesToRead Number of bytes requested from the network.
  *
  * This is passed to the #HTTPTransportInterface.recv function and used to
```
aggarw13, line 135: I noticed later that the sentence existed here as well :P.  Can you please remove this one?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -359,7 +330,7 @@ static int32_t transportRecv( HTTPNetworkContext_t * pContext,
 {
     int32_t bytesReceived = 0;
 
-    bytesReceived = ( int32_t ) recv( ( int ) pContext->tcpSocket, pBuffer, bytesToRecv, 0 );
+    bytesReceived = ( int32_t ) recv( pContext->tcpSocket, pBuffer, bytesToRecv, 0 );
```
aggarw13, line 333: The `int32_t` cast is redundant as well

yourslab: `recv` and `send` return value is of type `ssize_t`. Is this automatically 32 bits? Not sure.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -456,34 +427,37 @@ static int sendHttpRequest( HTTPTransportInterface_t * pTransportInterface,
...
+
+    return returnStatus;
 }
 
 /*-----------------------------------------------------------*/
 
 /**
  * @brief Entry point of demo.
+ *
+ * This example resolves a domain, then establishes a plaintext TCP connection to
```
aggarw13, line 447: (Nit) "plaintext TCP" may not be clear to readers. To make it clearer, you can say something "..establishes a TCP connection ~~to~~ with an HTTP server to demonstrate HTTP request/response communication without using an encrypted channel (i.e. without TLS)."

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -456,34 +427,37 @@ static int sendHttpRequest( HTTPTransportInterface_t * pTransportInterface,
...
+ * an HTTP server. After which, HTTP_Client APIs are used to send a GET, HEAD,
```
aggarw13, line 448: (Nit) "HTTP_Client APIs" -> "HTTP Client library API"

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -456,34 +427,37 @@ static int sendHttpRequest( HTTPTransportInterface_t * pTransportInterface,
...
+ * PUT, and POST request in that order. For each request, an HTTP response from
```
aggarw13, line 449: (Nit) "For each request, ~~an~~ the HTTP response received from the server is logged."

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Okay I like that description

demos/http/http_demo_plaintext/http_demo_plaintext.c
aggarw13: The local variable `bytesReceived` is of `int32_t` already. A cast should not be _explicitly_ required on the return value

demos/http/http_demo_plaintext/demo_config.h
yourslab: Yah the demo was looking a little too long from all these macros lol

demos/http/http_demo_plaintext/http_demo_plaintext.c
aggarw13: As the variable being assigned to is `int32_t`, and explicit cast does not make a difference, unless the compiler is issuing a warning about it.

yourslab: Removed

demos/http/http_demo_plaintext/demo_config.h
```diff
@@ -0,0 +1,107 @@
...
+#include "logging_stack.h"
+
+/************ End of logging configuration ****************/
+
+/**
+ * @brief HTTP server host name.
+ *
+ * This demo uses httpbin.org: A simple HTTP Request & Response Service.
+ */
+#define SERVER_HOST    "httpbin.org"
```
gkwicker, line 51: Why does the test depend on this service?  Who is it operated by, what is the SLA, are they okay with you pointing a lot of developers (and automated testers) at it with this code, etc.

Why not use a local server?  It would avoid all of these issues.

demos/http/http_demo_plaintext/demo_config.h
```diff
@@ -0,0 +1,107 @@
...
+#endif
+
+#ifndef POST_PATH
+    #define POST_PATH    "/post"
+#endif
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define TRANSPORT_SEND_RECV_TIMEOUT_MS    ( 1000 )
```
gkwicker, line 85: Why is this the right number?

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,526 @@
...
+ * @return EXIT_FAILURE on failure; EXIT_SUCCESS on success.
+ */
+static int sendHttpRequest( const HTTPTransportInterface_t * pTransportInterface,
+                            const char * pHost,
+                            const char * pMethod,
+                            const char * pPath );
+
+/*-----------------------------------------------------------*/
+
+static int connectToServer( const char * pServer,
```
gkwicker, line 164: This should be in common code since the MQTT library (and any others which use TCP) can use it.

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,526 @@
...
+    {
+        freeaddrinfo( pListHead );
+    }
+
+    return returnStatus;
+}
+
+/*-----------------------------------------------------------*/
+
+static int32_t transportSend( HTTPNetworkContext_t * pContext,
```
gkwicker, line 306: Same comment as for the connect method and at line 329 - these should be a common library which can be used by all of your tests of TCP-based protocols.  As a bonus, app developers can use them as well.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Yes that's the plan: Make a common library for connect and transport interface, then update these demos after the PRs are merged in as discussed with @leegeth.

abhidixi11: Agree, common code will be factored out later, right now all demos in MQTT have this function.  So for this PR we will have it in the demo.

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: Yes that's the plan: Make a common library for connect and transport interface, then update these demos after the PRs are merged in as discussed with @leegeth.

abhidixi11: Agree and same comment applies as above. We will factor out transport implementation later. We also need to change HTTPNetworkContext_t to generic NetworkContext_t so that both MQTT and HTTP can use this. We have a separate task to handle that.

demos/http/http_demo_plaintext/demo_config.h
yourslab: Actually, it may not be the right number depending on the platform. Some boards take much longer to send/recv. How can we come up with the right number based on the platform? In fact, the same problem also applies to test stabilization. However, these problems will be addressed in the common library for transport interface and connecting to server.

demos/http/http_demo_plaintext/demo_config.h
yourslab: This service is operated by [postman](postman.com). I was unable to find any data on the SLA or availability of [httpbin.org](httpbin.org). However, we have used this service in the past for AFR demos. Another option would be [postman-echo.com](postman-echo.com). 

We can use a local server. That would be nice because we would be able to determine how many people are running our demos and even where they come from. However, there is added time and cost to setting this up. What is your opinion on this? Should we setup our own server on EC2?

demos/http/http_demo_plaintext/http_demo_plaintext.c
sarenameas: The declarations and code together only pop up when you put "non-assignment code" like asserts, before the declarations, it does not apply to assignment type function usages.

demos/http/http_demo_plaintext/http_demo_plaintext.c
sarenameas: We probably should log an error as there are supposedly more errors: https://pubs.opengroup.org/onlinepubs/009695399/functions/setsockopt.html

demos/http/http_demo_plaintext/http_demo_plaintext.c
```diff
@@ -0,0 +1,552 @@
...
+    requestInfo.pHost = pHost;
+    requestInfo.hostLen = strlen( pHost );
+    requestInfo.method = pMethod;
+    requestInfo.methodLen = strlen( pMethod );
+    requestInfo.pPath = pPath;
+    requestInfo.pathLen = strlen( pPath );
+
+    /* Set "Connection" HTTP header to "keep-alive" so that multiple requests
+     * can be sent over the same established TCP connection. */
+    requestInfo.flags = HTTP_REQUEST_KEEP_ALIVE_FLAG;
```
sarenameas, line 379: This was fixed right? I suppose it is good to be explicit.

demos/http/http_demo_plaintext/http_demo_plaintext.c
sarenameas: Yea i mean if we are just echoing the headers and stuff, then this is OK.

sarenameas: I was looking at the other review and failed to notice this was for the request. It seems superfluous to log the request information. Maybe just the request logging could be debug?

demos/http/http_demo_plaintext/demo_config.h
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @note In general, port 80 is for plaintext HTTP connections. However,
+ * the default plaintext port from the local http server is used below.
+ */
+#define SERVER_PORT    8080
+
+/**
+ * @brief Paths for different HTTP methods for specified host.
+ */
+#ifndef GET_PATH
```
sarenameas, line 65: Given that you have moved the configurations to the config file these should not have a `#ifndef..#endif` around them.
`#ifndef..#endif` are used to define defaults, in-case the demo_config.h does not defined them, and the defaults typically belong in the .c file.

demos/http/http_demo_plaintext/demo_config.h
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * @brief The length in bytes of the user buffer.
+ */
+#define USER_BUFFER_LENGTH                ( 1024 )
+
+/**
+ * @brief Length of an IPv6 address when converted to hex digits.
+ */
+#define IPV6_ADDRESS_STRING_LEN           ( 40 )
```
sarenameas, line 94: This is never going to change, so it should not be in the demo_config.h

demos/http/http_demo_plaintext/demo_config.h
yourslab: Update: After discussing with @gkwicker, we are having the user run a local http server before running the demos: `python3 tools/http-echo-server/http_server.py`

demos/http/http_demo_plaintext/http_demo_plaintext.c
yourslab: It is fixed but yah, it's good to be explicit.

README.md
```diff
@@ -17,4 +17,7 @@ This SDK builds with [CMake](https://cmake.org/), a cross-platform build tool.
 1. Create build directory: `mkdir build && cd build`
 1. Run *cmake* while inside build directory: `cmake ..`
 1. Run this command to build the demos: `make`
-1. Go to the `build/bin` directory to see executables.
+1. Go to the `build/bin` directory and run any demo executables from there.
+1. To run demos prefixed with `http_`:
+    1. Install `Python 3` if it is not yet installed in your system: `sudo apt-get install python3`
```
gkwicker, line 22: 110% awesome

PR #911, opened 2020-04-30
Implement publish deserializers

muneebahmed10:
*Description of changes:*
Implement MQTT_DeserializePublish, MQTT_DeserializeAck, MQTT_GetIncomingPacketTypeAndLength, and MQTT_SerializeAck

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -1123,9 +1123,8 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * const pIncomingPacket
                 break;
         }
     }
-    
 
-    return MQTTSuccess;
+    return status;
 }
 
 /*-----------------------------------------------------------*/
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -20,8 +20,10 @@
  */
 
 #include <string.h>
+#include <assert.h>
 
 #include "mqtt_lightweight.h"
+#include "private/mqtt_internal.h"
 
 /**
  * @brief MQTT protocol version 3.1.1.
@@ -47,6 +49,15 @@
 #define MQTT_CONNECT_FLAG_PASSWORD       ( 6 )    /**< @brief Password present. */
 #define MQTT_CONNECT_FLAG_USERNAME       ( 7 )    /**< @brief User name present. */
 
+/*
+ * Positions of each flag in the first byte of an MQTT PUBLISH packet's
+ * fixed header.
+ */
+#define MQTT_PUBLISH_FLAG_RETAIN         ( 0 )             /**< @brief Message retain flag. */
+#define MQTT_PUBLISH_FLAG_QOS1           ( 1 )             /**< @brief Publish QoS 1. */
+#define MQTT_PUBLISH_FLAG_QOS2           ( 2 )             /**< @brief Publish QoS 2. */
+#define MQTT_PUBLISH_FLAG_DUP            ( 3 )             /**< @brief Duplicate message. */
+
 /**
  * @brief The size of MQTT DISCONNECT packets, per MQTT spec.
  */
@@ -57,10 +68,31 @@
  */
 #define MQTT_DISCONNECT_REMAINING_LENGTH    ( ( uint8_t ) 0 )
 
+/*
+ * Constants relating to CONNACK packets, defined by MQTT 3.1.1 spec.
+ */
+#define MQTT_PACKET_CONNACK_REMAINING_LENGTH        ( ( uint8_t ) 2U )    /**< @brief A CONNACK packet always has a "Remaining length" of 2. */
+#define MQTT_PACKET_CONNACK_SESSION_PRESENT_MASK    ( ( uint8_t ) 0x01U ) /**< @brief The "Session Present" bit is always the lowest bit. */
+
+/*
+ * UNSUBACK, PUBACK, PUBREC, PUBREL, and PUBCOMP always have a remaining length
+ * of 2.
+ */
+#define MQTT_PACKET_SIMPLE_ACK_REMAINING_LENGTH     ( ( uint8_t ) 2 )   /**< @brief A PUBACK packet always has a "Remaining length" of 2. */
+#define MQTT_PACKET_PINGRESP_REMAINING_LENGTH       ( 0U ) /**< @brief A PINGRESP packet always has a "Remaining length" of 0. */
+
 /**
  * @brief Set a bit in an 8-bit unsigned integer.
  */
 #define UINT8_SET_BIT( x, position )    ( ( x ) = ( uint8_t ) ( ( x ) | ( 0x01U << ( position ) ) ) )
+
+/**
+ * @brief Macro for checking if a bit is set in a 1-byte unsigned int.
+ *
+ * @param[in] x The unsigned int to check.
+ * @param[in] position Which bit to check.
+ */
+#define UINT8_CHECK_BIT( x, position )    ( ( ( x ) & ( 0x01U << ( position ) ) ) == ( 0x01U << ( position ) ) )
 
 /**
  * @brief Get the high byte of a 16-bit unsigned integer.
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -192,6 +247,508 @@ static int32_t recvExact( MQTTTransportRecvFunc_t recvFunc,
...
+{
+    MQTTStatus_t status = MQTTSuccess;
+
+    assert( pConnack != NULL );
+    assert( pSessionPresent != NULL );
+    const uint8_t * pRemainingData = pConnack->pRemainingData;
+
+    /* If logging is enabled, declare the CONNACK response code strings. The
+     * fourth byte of CONNACK indexes into this array for the corresponding response. */
+    #if LIBRARY_LOG_LEVEL > IOT_LOG_NONE
```
abhidixi11, line 437: Should we create a function here instead, which will not be linked if logging is disabled. ?

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -192,6 +252,515 @@ static int32_t recvExact( MQTTTransportRecvFunc_t recvFunc,
...
+            IotLogErrorWithArgs( "CONNACK response %hhu is not valid.",
+                                 pRemainingData[ 1 ] );
+
+            status = MQTTBadResponse;
+        }
+        else
+        {
+            /* Print the appropriate message for the CONNACK response code if logs are
+             * enabled. */
+            #if LIBRARY_LOG_LEVEL > IOT_LOG_NONE
```
abhidixi11, line 516: I think #if is not needed (since already inside the function )

libraries/standard/mqtt/src/mqtt_lightweight.c
abhidixi11: Resolving since it has been moved inside a function.

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: Yep, I had forgotten to remove one of the macro guards

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -72,9 +109,32 @@
...
+ * @brief The minimum remaining length for a QoS 0 PUBLISH.
+ *
+ * Includes two bytes for topic name length and one byte for topic name.
+ */
+#define MQTT_MIN_PUBLISH_REMAINING_LENGTH_QOS0      ( 3U )
+
 /*-----------------------------------------------------------*/
 
-static size_t remainingLengthEncodedSize( size_t length )
+static size_t _remainingLengthEncodedSize( size_t length )
```
cobusve, line 137: These names will cause MISRA violations going forward. I get the idea of making internal functions start with underscore but this keeps coming up. If Public functions are prefixed with MQTT_ and private ones are not I think it is clear enough. We should document this in the style guide btw.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -20,8 +20,10 @@
  */
 
 #include <string.h>
+#include <assert.h>
 
 #include "mqtt_lightweight.h"
+#include "private/mqtt_internal.h"
```
cobusve, line 26: I do not like the name of this file, it is not an internal secret and it is not internal private functions. It is the logging interface, can we not just call it that?

cobusve: And I do not see a good reason for this to be "private" in the first place ...

libraries/standard/mqtt/src/mqtt_lightweight.c
cobusve: Looks like we need a "WHY" comment here. A comment explaining that this is in a function so that it can be removed from the build easily would help cusotmers ecplain the code.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -57,11 +68,37 @@
  */
 #define MQTT_DISCONNECT_REMAINING_LENGTH    ( ( uint8_t ) 0 )
 
+/*
+ * Constants relating to CONNACK packets, defined by MQTT 3.1.1 spec.
+ */
+#define MQTT_PACKET_CONNACK_REMAINING_LENGTH        ( ( uint8_t ) 2U )    /**< @brief A CONNACK packet always has a "Remaining length" of 2. */
```
cobusve, line 74: Why are we doing a #define here? The magic numbers smell applies to numbers that we may want to change. This is mandated by the spec so it can never change. It would be better to just have a comment where we check the length.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -57,11 +68,37 @@
...
+#define MQTT_PACKET_CONNACK_SESSION_PRESENT_MASK    ( ( uint8_t ) 0x01U ) /**< @brief The "Session Present" bit is always the lowest bit. */
+
+/**
+ * @brief The size of MQTT PUBACK, PUBREC, PUBREL, and PUBCOMP packets, per MQTT spec.
+ */
+#define MQTT_PUBLISH_ACK_PACKET_SIZE        ( 4U )
+
```
cobusve, line 81: Ditto. These defines just end up adding a lot of lines and noise to the file which is not adding value.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -57,11 +68,37 @@
...
+/*
+ * UNSUBACK, PUBACK, PUBREC, PUBREL, and PUBCOMP always have a remaining length
+ * of 2.
+ */
+#define MQTT_PACKET_SIMPLE_ACK_REMAINING_LENGTH     ( ( uint8_t ) 2 )   /**< @brief PUBACK, PUBREC, PUBREl, PUBCOMP, UNSUBACK Remaining length. */
+#define MQTT_PACKET_PINGRESP_REMAINING_LENGTH       ( 0U ) /**< @brief A PINGRESP packet always has a "Remaining length" of 0. */
+
 /**
```
cobusve, line 89: Same for these 2

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -192,6 +252,513 @@ static int32_t recvExact( MQTTTransportRecvFunc_t recvFunc,
...
+            status = true;
+            break;
+
+        /* Any other packet type is invalid. */
+        default:
+            break;
+    }
+
+    /* The second bit of a PUBREL must be set, so the above switch won't handle it. */
+    if( packetType == MQTT_PACKET_TYPE_PUBREL )
```
cobusve, line 329: I would prefer to see this moved into the switch. You can add it like this:
case MQTT_PACKET_TYPE_PUBREL & 0xF0 :
    if (packetType & 0x02) {
       status = true;
   }

I think it is easier to understand this way?

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: The inclusion of logging has not yet been finalized; this file was added to unblock ongoing development. If we later decide to change the file name, we can just change the include

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: After some discussion, we will be updating the style guide. I reverted this change and removed the leading underscore from function names

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: Changed

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: After (offline?) discussion, will leave the macros here and add comments in code for readability

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: Added a comment explaining the block

PR #915, opened 2020-05-01
Bugfix/http build issues

aggarw13:
*Description of changes:*
* Fix `-Wsign-compare` warning from test code implementation
* Fix build errors and warnings from assert expression in `AddRangeHeader`.
* Enable `assert` utility in unit tests to avoid missing assert expression build issues.
* Update implementation of `_addHeader` to use ~~assert instead of error handling `snprintf` return value~~ `memcpy` instead of `snprintf` due to concerns with `snprintf` failure for ISO 8859-1 character set. Also, update `HTTPClient_AddRangeHeader` API to accept `uint8_t *` types instead of `char *` to support ISO 8859-1 character set

Evidence of passing unit tests _after_ the above changes:
image

### DIFF
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -2,6 +2,7 @@
 #define HTTP_CLIENT_H_
 
 #include <stdint.h>
+#include <stddef.h>
 #include "config.h"
 
 /**
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -408,7 +408,8 @@ typedef struct HTTPResponse
 /**
  * @brief Initialize the request headers, stored in
  * #HTTPRequestHeaders_t.pBuffer, with initial configurations from
- * #HTTPRequestInfo_t.
+ * #HTTPRequestInfo_t. This method is expected to be called before sending a
+ * new request.
  *
  * Upon return, #HTTPRequestHeaders_t.headersLen will be updated with the number
  * of bytes written.
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -445,8 +445,12 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
  *
  * @param[in] pRequestHeaders Request header buffer information.
  * @param[in] pField The header field name to write.
+ * The data should be ISO 8859-1 (Latin-1) encoded per the HTTP standard,
+ * but the API does not perform the character set validation.
  * @param[in] fieldLen The byte length of the header field name.
  * @param[in] pValue The header value to write.
+ * The data should be ISO 8859-1 (Latin-1) encoded per the HTTP standard,
+ * but the API does not perform the character set validation.
  * @param[in] valueLen The byte length of the header field value.
  *
  * @return One of the following:
@@ -455,9 +459,9 @@ HTTPStatus_t HTTPClient_InitializeRequestHeaders( HTTPRequestHeaders_t * pReques
  * - #HTTP_INSUFFICIENT_MEMORY (If application-provided buffer is not large enough to hold headers.)
  */
 HTTPStatus_t HTTPClient_AddHeader( HTTPRequestHeaders_t * pRequestHeaders,
-                                   const char * pField,
+                                   const uint8_t * pField,
                                    size_t fieldLen,
-                                   const char * pValue,
+                                   const uint8_t * pValue,
                                    size_t valueLen );
 
 /**
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -74,6 +74,20 @@
 #define HTTP_RESPONSE_CONNECTION_KEEP_ALIVE_FLAG    0x2U
 
 /**
+ * @brief Flag that represents End of File byte in the range specification of
+ * a Range Request.
+ * This flag should be used ONLY for 2 kinds of range specifications when
+ * creating the Range Request header through the #HTTPClient_AddRangeHeader
+ * function:
+ *  - When the requested range is all bytes from the starting range byte to
+ * the end of file.
+ *  - When the requested range is for the last N bytes of the file.
+ * In both cases, this value should be used for the "rangeEnd" parameter.
+ */
+#define HTTP_RANGE_REQUEST_END_OF_FILE              -1
+
+
+/**
  * @brief The HTTPNetworkContext is an incomplete type. The application must
  * define HTTPNetworkContext to the type of their network context. This context
  * is passed into the network interface functions.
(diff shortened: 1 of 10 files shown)


libraries/standard/http/test/test-HTTPClient_AddRangeHeader.c
```diff
@@ -48,7 +48,7 @@ typedef struct _headers
                                  dataLen + 1,                             \
                                  "%s",                                    \
                                  preexistingData );                       \
-        ok( numBytes == dataLen );                                        \
+        ok( numBytes == ( int ) dataLen );                                \
```
yourslab, line 51: Doesn't this give MISRA warnings?

aggarw13: `-Wsign-compare` build warning

libraries/standard/http/test/common.h
```diff
@@ -7,4 +7,5 @@
 #include "private/http_client_internal.h"
 #include "private/http_client_parse.h"
 
-#define assert( x )
+/*#define assert( x ) */
+#include <assert.h>
```
yourslab, line 11: I remember @sarenameas telling us to avoid doing this

aggarw13: That was only for having 100% branch coverage statistics. However, the advantage of having asserts enabled is prevention of missing out build issues in assert expressions. 
(Also, ideally, we should have written tests for assert expression in unit test code, but it may not be feasible to do that in CMock framework; anyway, if we don't want to do that, we can provide evidence of 100% without asserts in PR descriptions)

aggarw13: Have added a configuration to enable/disable asserts in unit tests for flexibility

libraries/standard/http/test/common.h
```diff
@@ -7,4 +7,8 @@
 #include "private/http_client_internal.h"
 #include "private/http_client_parse.h"
 
-#define assert( x )
+#if TESTS_DISABLE_ASSERT
```
sarenameas, line 10: let's do this instead and then we can add tests for our asserts later
```
static int assertCount_;
#define assertReset()	do { assertCount_ = 0; } while( 0 )
```

libraries/standard/http/src/http_client.c
```diff
@@ -155,21 +155,14 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
 
-        if( ( bytesWritten + HTTP_HEADER_LINE_SEPARATOR_LEN ) != toAddLen )
-        {
-            IotLogErrorWithArgs( "Internal error in snprintf() in _addHeader(). "
-                                 "BytesWritten: %d.", bytesWritten );
-        }
-        else
-        {
-            pBufferCur += bytesWritten;
+        assert( bytesWritten == (int) ( toAddLen - HTTP_HEADER_LINE_SEPARATOR_LEN ) );
```
sarenameas, line 158: We are not asserting here before the input into snprintf is different than the integers we were writing in AddRangeHeader().
The inputs into snprintf here are from user inputs, so we need to bubble up the encoding error.

sarenameas, line 158: We are not asserting here before the input into snprintf is different than the integers we were writing in AddRangeHeader().
The inputs into snprintf here are from user inputs, so we need to bubble up the encoding error.

aggarw13: The `snprintf` is only called after the buffer space is validated.

libraries/standard/http/test/common.h
aggarw13: Sure, I prefer that too.

libraries/standard/http/src/http_client.c
sarenameas: It is not about the space, it's the encoding error from invalid characters that cannot be encoded

aggarw13: If the worry is about non-printable characters, I have tested with the following on gcc:
```
#include <stdio.h>
#include <assert.h>
#include <stdint.h>

int main()
{
    char buffer[ 100 ];
    uint8_t input[] = { 0xFF, 0xFF };

    assert( snprintf( buffer, 100, "%.*s", 2, ( char * ) input ) == 2 );
}
```

sarenameas: This code is to be portable to non-gcc compilers

aggarw13: If the goal is to propagate the error back to the user, then an `INTERNAL_FAILURE` error code cannot be helpful. (Also, it seems like the original code didn't return any error code for the failure: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/http/src/http_client.c#L158-L162)

Ideally, we should provide `INVALID_PARAMETER` back to the caller, but there isn't a reliable way to do without actually doing O(n) traversal on both the header and value string inputs

sarenameas: That is a miss in the original code, but there was supposed to be an error returned.
(continuing conversation offline)

aggarw13: We discussed offline, and decided to use `uint8_t *` instead of `char *` for user-input as the HTTP standard supports [ISO 8859-1]() character encoding which has a larger range set that `char` type.
Also, to avoid any platform-specific `snprintf` issues with printing characters outside the `char` range, we will use `memcpy` to just copy the bytes into the buffer.

libraries/standard/http/test/common.h
abhidixi11: Please stick to one variable naming either _xyzAbc or xyzAbc_ to avoid confusion.

aggarw13: Good point. Will change to `_assertCount`

aggarw13: Updated

libraries/standard/http/test/common.h
```diff
@@ -7,4 +7,6 @@
 #include "private/http_client_internal.h"
 #include "private/http_client_parse.h"
 
-#define assert( x )
+static int _assertCount;
+#define assertReset()    do { _assertCount = 0; } while( 0 )
+#define assert( x )      do { _assertCount++; } while( 0 )
```
sarenameas, line 12: Oh just realized that we only want to increase the _assertCount if x == false

aggarw13: Fixed

libraries/standard/http/src/http_client.c
```diff
@@ -148,28 +148,33 @@ static HTTPStatus_t _addHeader( HTTPRequestHeaders_t * pRequestHeaders,
...
-        {
-            IotLogErrorWithArgs( "Internal error in snprintf() in _addHeader(). "
-                                 "BytesWritten: %d.", bytesWritten );
-        }
-        else
-        {
-            pBufferCur += bytesWritten;
+        /* Copy the header name into the buffer. */
+        memcpyRetVal = memcpy( pBufferCur, pField, fieldLen );
+        assert( memcpyRetVal == pBufferCur );
```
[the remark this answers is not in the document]
aggarw13: The risk is only protecting against any unexpected behavior of `memcpy` of the vendor toolchain during development. I know that the risk is low, but there is no _cost_ of adding the assert for debug builds, is there?

libraries/standard/http/src/http_client.c
```diff
@@ -302,7 +307,8 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
                               RANGE_REQUEST_HEADER_VALUE_PREFIX,
                               rangeStartOrlastNbytes );
         assert( ( stdRetVal >= 0 ) &&
```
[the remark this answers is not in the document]
aggarw13: As the `snprintf` is expected to generally succeed and that it is not vulnerable to any run-time input from the user, it did not make sense to add error handling logic in the production code. 
If there is an issue with the C standard library, it is expected to be caught at development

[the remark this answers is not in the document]
aggarw13: Another thing to note is that I will be removing the `snprintf` call to replace with `memcpy` calls in a subsequent PR as use of `snprintf` violates a MISRA directive. Thus, this code calling the `snprintf` is only temporary.  I will still have _asserts_ on the return value of `memcpy` though

libraries/standard/http/include/http_client.h
```diff
@@ -2,6 +2,7 @@
 #define HTTP_CLIENT_H_
 
 #include <stdint.h>
+#include <stddef.h>
```
sarenameas, line 5: What is stddef.h for?

aggarw13: `size_t`

libraries/standard/http/src/http_client.c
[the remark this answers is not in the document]
aggarw13: @sarenameas pointed out that `memcpy` does not return any error. `memcpy` API doc:https://pubs.opengroup.org/onlinepubs/009695399/functions/memcpy.html.
So, have removed the assert checks for `memcpy`

libraries/standard/http/include/http_client.h
cobusve: It also includes the definition of NULL

PR #920, opened 2020-05-04
Http misra/fix violations

aggarw13:
*Description of changes:*
Address MISRA violations
Replace use of `snprintf` with `memcpy` for MISRA compliance

### DIFF
--- a/libraries/standard/http/src/http_client.c
+++ b/libraries/standard/http/src/http_client.c
@@ -126,7 +126,7 @@ static uint8_t _convertInt32ToAscii( int32_t value,
 {
     /* As input value may be altered and MISRA C 2012 rule 17.8 prevents modification
      * of parameter, a local copy of the parameter is stored. */
-    uint32_t absoluteValue = 0;
+    uint32_t absoluteValue = 0u;
     uint8_t numOfDigits = 0u;
     uint8_t index = 0u;
     uint8_t isNegative = 0u;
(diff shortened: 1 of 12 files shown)


libraries/standard/http/src/http_client.c
```diff
@@ -108,19 +107,80 @@ static HTTPStatus_t _getFinalResponseStatus( HTTPParsingState_t parsingState,
...
+ *
+ * @return Returns the number of bytes written to @p pBuffer.
+ */
+static uint8_t _convertInt32ToAscii( int32_t value,
+                                     char * pBuffer,
+                                     size_t bufferlength );
+
+/*-----------------------------------------------------------*/
+
+static uint8_t _convertInt32ToAscii( int32_t value,
```
yourslab, line 124: Couldn't we also use itoa and define a separate method for returning the length of an int32_t?

aggarw13: `itoa` is not part of ANSI C as mentioned under the Portability section of its reference doc: http://www.cplusplus.com/reference/cstdlib/itoa/

We only want to add dependencies on ANSI C defined functions

libraries/standard/http/src/http_client.c
```diff
@@ -1,6 +1,4 @@
-#include <assert.h>
```
sarenameas, line 1: Why is this deleted? Where will this header exist?

libraries/standard/http/src/private/http_client_internal.h
```diff
@@ -1,6 +1,16 @@
 #ifndef HTTP_CLIENT_INTERNAL_H_
 #define HTTP_CLIENT_INTERNAL_H_
 
+#include "config.h"
+
+#if defined( HTTP_ENABLE_ASSERT ) && ( HTTP_ENABLE_ASSERT == 1 )
```
sarenameas, line 6: I think we shouldn't have this and instead if customers are going to compile out asserts they will use the ANSI C definition NDEBUG in their compiler level macros.

libraries/standard/http/src/http_client.c
aggarw13: `http_client_internal.h`

libraries/standard/http/src/private/http_client_internal.h
aggarw13: We discussed offline, and decided to defer the decision on supporting configurable asserts per library in consultation with the larger team.

PR #924, opened 2020-05-07
Add third-party http-parser as a Git Submodule.

sarenameas:
This brings in https://github.com/nodejs/http-parser as a submodule under **libraries/standard/http/third_party/http_parser**

This is needed for the HTTP Client library to parse the incoming response and to find a header in a completed response.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libraries/standard/http/third_party/http_parser"]
+	path = libraries/standard/http/third_party/http_parser
+	url = git@github.com:nodejs/http-parser.git


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libraries/standard/http/third_party/http_parser"]
```
aggarw13, line 1: As this library is specific to HTTP, it is fine to keep it under the `http` folder for now. However, for shared 3rd party libraries like `tinycbor` or a JSON utility, there would be a common folder created. Would this library be moved there then?

yourslab: I guess putting `http_parser` under `http` makes it clear that `http` is the only library using it. But I am already using a common folder for `CMock` because it will likely be used by all libraries.

PR #926, opened 2020-05-07
Add CMake build for demos and unit tests (part 2)

yourslab:
See **[CMake build for demos ONLY (part 1)](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/929)**

To understand the process introduced by this PR for writing unit tests in C SDK using CMock/Unity, please see **[docs/testing.md](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/00f037c9c4aa84a22590dca3e626f95d9f715e11/docs/utesting.md)**

The cmake functions from `tools/cmock` directory were introduced in this [PR from AFR](https://github.com/aws/amazon-freertos/pull/1720).


CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
...
+cmake_minimum_required( VERSION 3.5.0 )
+project( AwsIotDeviceSdkEmbeddedC
+         VERSION 4.0.2
+         LANGUAGES C )
+
+# Allow the project to be organized into folders.
+set_property( GLOBAL PROPERTY USE_FOLDERS ON )
+
+# Use C99.
+set( CMAKE_C_STANDARD 99 )
```
sarenameas, line 11: I think we are not using C99 but ANSI C

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "libraries/standard/http/third_party/http_parser"]
 	path = libraries/standard/http/third_party/http_parser
 	url = https://github.com/nodejs/http-parser.git
+[submodule "libraries/3rdparty/CMock"]
+	path = libraries/3rdparty/CMock
+	url = https://github.com/ThrowTheSwitch/CMock
```
leegeth, line 6: do we always pull from master of CMock or is there a release tag that we are planning to use?

sarenameas: If we put the release version in here, it has no effect on how gitsubmodules works and will not get automatically pulled. We need to go into the submodule and checkout the release we want and check that change in.

yourslab: AFR has it this way too but putting a release tag isn't a bad idea.

CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
...
+# Do not allow in-source build.
+if( ${PROJECT_SOURCE_DIR} STREQUAL ${PROJECT_BINARY_DIR} )
+    message( FATAL_ERROR "In-source build is not allowed. Please build in a separate directory, such as ${PROJECT_SOURCE_DIR}/build." )
+endif()
+
+# Import global configurations.
+include("tools/cmake/filePaths.cmake")
+
+# Configure options to always show in CMake GUI.
+option( ATOMIC_USE_PORT
```
sarenameas, line 23: We don't have a custom atomic port so this should be gone?

libraries/standard/http/utest/README.md
```diff
@@ -0,0 +1,118 @@
...
+       This compilation step would create an object library called *project_name_real.a*
+    1. Replace all directories in *test_include_directories* with what the test
+       code needs (project_name_utest.c)
+       ```cmake
+       list(APPEND test_include_directories
+                    /my/test/code/needs
+                )
+        ```
+
+    1. And thats it for this file..
```
sarenameas, line 106: that's

libraries/standard/mqtt/utest/README.md
```diff
@@ -0,0 +1,118 @@
+# Running Unit Tests
+
+1. In your GNU compatible environment like WSL, mingw, Linux OS, MAC OS, etc., open a terminal.
```
sarenameas, line 3: Looks like we will need to delete these examples of environments or else we have to test them all. So just "In your GNU compatible environment, open a terminal."

CMakeLists.txt
```diff
@@ -0,0 +1,111 @@
...
+    target_include_directories(cmock PUBLIC
+            ${ROOT_DIR}/libraries/3rdparty/CMock/src
+            ${ROOT_DIR}/libraries/3rdparty/CMock/vendor/unity/src/
+            ${ROOT_DIR}/libraries/3rdparty/CMock/examples
+            )
+    target_link_libraries(cmock unity)
+
+    add_custom_target(coverage
+            COMMAND ${CMAKE_COMMAND} -P ${CMAKE_SOURCE_DIR}/tools/cmock/coverage.cmake
+            DEPENDS cmock unity http_utest mqtt_utest
```
sarenameas, line 102: In the README, for a new library do you add that this needs to get updated?

libraries/standard/http/utest/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
+include("../filePaths.cmake")
+project ("http unit test")
+cmake_minimum_required (VERSION 3.13)
+
+# ====================  Define your project name (edit) ========================
+set(project_name "http")
+
+# =====================  Create your mock here  (edit)  ========================
+
+# list the files to mock here
```
sarenameas, line 10: This is really helpful thanks

libraries/standard/http/utest/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+    1. Replace all the files in *real_source_files* with the source files you will
+       be testing. This will create a library with the name *project_file_real.a*
+       ```cmake
+       list(APPEND real_source_files
+                    code_to_test.c
+                )
+       ```
+    1. Replace all the dirctories in *real_include_directories*  with the
+       directories the source under test will inluce headers from
```
sarenameas, line 91: include

libraries/standard/http/utest/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+1. In your GNU compatible environment, open a terminal.
+1. Install ruby: `sudo apt install ruby-full`
+1. Install lcov: `sudo apt install lcov`
+1. Go to the root directory of this repository.
+1. Update submodules (if you haven't already): `git submodule update --init --recursive` 
+1. Create build directory: `mkdir build && cd build`
+1. Run cmake while inside build directory: `cmake ..`
+1. Run this command to build and run the tests: `make coverage`
+1. Go to `build` and open `http_utest_out.txt` to view logs
```
alfred2g, line 11: I think having the readme in every folder is a bit of an overkill
if something ever changes (typo, bug, new step, removing a step etc..) we will have to go to all folders it could grow up to a lot , and make the changes in each one.
I would suggest in a folder called Doc , UnitTestDoc etc

yourslab: Yah it has been quite a hassle editing both README's lol

CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+# Do not allow in-source build.
+if( ${PROJECT_SOURCE_DIR} STREQUAL ${PROJECT_BINARY_DIR} )
+    message( FATAL_ERROR "In-source build is not allowed. Please build in a separate directory, such as ${PROJECT_SOURCE_DIR}/build." )
+endif()
+
+# Import global configurations.
+include("tools/cmake/filePaths.cmake")
+
+# Configure options to always show in CMake GUI.
+option( BUILD_TESTS
```
aggarw13, line 23: I think we should avoid the need of this flag. We should be able to build both tests and demos, with each of those targets configured to include their appropriate config.h path, without needing this flag

CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+        "Set this to ON to build both demo and test executables. When OFF, only demo executables are built."
+        ON )
+option( BUILD_CLONE_SUBMODULES
```
aggarw13, line 26: I think this option can also be avoided. (Imo) We should start with the minimal feature set of build options for CMake, and add more if needed in the future.

CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+set( ALLOW_SHARED_LIBRARIES ON )
+    
+include( CMakeDependentOption )
+CMAKE_DEPENDENT_OPTION( BUILD_SHARED_LIBS
+                        "Set this to ON to build all libraries as shared libraries. When OFF, libraries build as static libraries."
+                        ON "${ALLOW_SHARED_LIBRARIES}"
+                        OFF )
+
+# Set the platform named based on the host OS if not defined.
+if( NOT DEFINED PLATFORM_NAME )
```
aggarw13, line 40: I would suggest keeping platform specific configuration in the test/demos' CMakeLists instead of being present in the master CMakeLists.txt. We want the test/demos to be self-contained with information of platform-specific stack being defined by them.

CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+endif()
+
+# Set output directories.
+set( CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin )
+set( CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib )
+set( CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib )
+
+
+if( ${BUILD_TESTS} )
+    include("${ROOT_DIR}/tools/cmock/create_test.cmake")
```
aggarw13, line 60: It would be better to keep all CMock/Unity specific configuration in a separate CMakeLists.txt file, and just include that file in this master CMakeLists.txt. The master CMakeLists.txt should only contain common configuration and utilities across libraries, along with includes for sub-CMakeLists.txt

CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+    link_directories("${CMAKE_BINARY_DIR}/lib"
+            )
+
+    add_library(cmock STATIC
+            "${ROOT_DIR}/libraries/3rdparty/CMock/src/cmock.c"
+            )
+
+    set_target_properties(cmock PROPERTIES
+            ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+            POSITION_INDEPENDENT_CODE ON
```
aggarw13, line 76: Why is this flag always enabled for the `cmock` target? Can it not allow static library builds?

CMakeLists.txt
yourslab: Do you mean removing this flag and just building BOTH tests and demos all the time? We had it setup in v4 beta_1 to only build demos when this flag is off. That is what I'm trying to do here as well. As an end user, I probably don't care about the tests anyway.

CMakeLists.txt
yourslab: Mocks are shared libraries to prevent code coverage from displaying cmock generated files. In that case, we need -fPIC when linking a static library to a shared object.

.gitmodules
cobusve: Git stores the actual commit-id in the git database, you can retrieve it as explained here https://stackoverflow.com/questions/3983829/how-can-i-get-a-git-submodules-associated-commit-id-from-a-past-commit-in-the-p

CMakeLists.txt
cobusve: I think if you build both demos and tests you will have to have CMock installed and ruby in order to create the mocks as part of the compilation of the demo, even if you are not interested in the tests at all. Is that a good idea?

I think a flag to disable building of the tests is probably a reasonable way to avoid this problem.

CMakeLists.txt
cobusve: We should think about this, my best guess this was added becuase building including the cloning of submodules requires a round-trip TLS connection to github for every submodule which is something which can easily add minutes or even hours to the nightly CI. 

For Context the ESP-IDF submodule is 180MB, the Microchip Harmony one is also huge, I was looking at the Cypress one which is 150MB last week ...

Is there a different way to give the option to not re-fetch the submodules every time?

.gitmodules
leegeth: Ok. Thanks.

libraries/standard/http/filePaths.cmake
```diff
@@ -0,0 +1,23 @@
...
+# HTTP library source files.
+set( HTTP_SOURCES
+     "${MODULES_DIR}/standard/http/src/http_client.c"
+     "${MODULES_DIR}/standard/http/src/http_client_parse.c" )
+
+# HTTP library Include directories.
+set( HTTP_INCLUDE_PUBLIC_DIRS
+     "${MODULES_DIR}/standard/http/include"
+     "${MODULES_DIR}/standard/utilities/include"
+     "${ROOT_DIR}/platform/include"
```
leegeth, line 17: Can you please remove the platform include directory from this variable and handle separately? This will not be the same path when reused in different repositories.

CMakeLists.txt
yourslab: I am having cmake update submodules from now on whenever this option is activated.

libraries/standard/http/utest/config.h
```diff
@@ -0,0 +1,27 @@
...
+#define IOT_LOG_LEVEL_HTTP          IOT_LOG_DEBUG
+
+#define USE_AWS_IOT_CSDK_LOGGING    1
+
+#ifdef USE_AWS_IOT_CSDK_LOGGING
+
+/* Include file for POSIX reference implementation. */
+    #include "iot_logging.h"
+
+/* Define the IotLog logging interface to enabling logging.
```
leegeth, line 13: to enabling -> to enable

libraries/standard/mqtt/filepaths.cmake
```diff
@@ -0,0 +1,21 @@
...
+# MQTT library source files.
+set( MQTT_SOURCES
+     "${MODULES_DIR}/standard/mqtt/src/mqtt.c"
+     "${MODULES_DIR}/standard/mqtt/src/mqtt_lightweight.c" )
+
+# MQTT library Include directories.
+set( MQTT_INCLUDE_PUBLIC_DIRS
+     "${MODULES_DIR}/standard/mqtt/include"
+     "${MODULES_DIR}/standard/utilities/include"
+     "${ROOT_DIR}/platform/include" )
```
leegeth, line 17: Can you please remove the platform include directory from this variable and handle separately? This will not be the same path when reused in different repositories.

libraries/standard/mqtt/utest/config.h
```diff
@@ -0,0 +1,63 @@
...
+#define IOT_LOG_LEVEL_HTTP          IOT_LOG_DEBUG
+
+#define USE_AWS_IOT_CSDK_LOGGING    1
+
+#ifdef USE_AWS_IOT_CSDK_LOGGING
+
+/* Include file for POSIX reference implementation. */
+    #include "iot_logging.h"
+
+/* Define the IotLog logging interface to enabling logging.
```
leegeth, line 13: minor:to enabling -> to enable

CMakeLists.txt
aggarw13: Hmm, good point about CMock and Ruby dependencies

PR #927, opened 2020-05-08
Fix MISRA violations

leegeth:
*Description of changes:*
Fix MISRA violations in MQTT library in development branch.
Coverity config file used to run MISRA check can be found here https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/scripts/coverity_misra.config.

This fix doesn't cover:
Violations in uncompleted demo.
Violation in functions which are still not complete in development branch.
Violations related to logs and asserts.
Rule 8.7 for internal linkage public APIs. This will be resolved when demos/another API invokes those.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -222,11 +222,11 @@ static MQTTStatus_t calculateSubscriptionPacketSize( const MQTTSubscribeInfo_t *
  * #MQTTBadParameter if invalid parameters are passed;
  * #MQTTSuccess otherwise.
  */
-static MQTTStatus_t validateSubscribeUnsubscribeParams( const MQTTSubscribeInfo_t * const pSubscriptionList,
-                                                        size_t subscriptionCount,
-                                                        uint16_t packetId,
-                                                        size_t remainingLength,
-                                                        const MQTTFixedBuffer_t * const pBuffer );
+static MQTTStatus_t validateSubscriptionSerializeParams( const MQTTSubscribeInfo_t * const pSubscriptionList,
+                                                         size_t subscriptionCount,
+                                                         uint16_t packetId,
+                                                         size_t remainingLength,
+                                                         const MQTTFixedBuffer_t * const pBuffer );
 
 /**
  * @brief Serialize an MQTT CONNECT packet in the given buffer.
@@ -320,6 +320,10 @@ static uint8_t * encodeString( uint8_t * pDestination,
 {
     uint8_t * pBuffer = NULL;
 
+    /* Typecast const char * typed source buffer to const uint8_t *.
+     * This is to use same type buffers in memcpy. */
+    const uint8_t * pSourceBuffer = ( const uint8_t * ) source;
+
     assert( pDestination != NULL );
 
     pBuffer = pDestination;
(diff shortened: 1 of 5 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -343,37 +346,6 @@ static uint8_t * encodeString( uint8_t * pDestination,
 
 /*-----------------------------------------------------------*/
 
-static int32_t recvExact( MQTTTransportRecvFunc_t recvFunc,
```
muneebahmed10, line 346: Is this function not needed?

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -619,7 +590,7 @@ static bool incomingPacketValid( uint8_t packetType )
         case ( MQTT_PACKET_TYPE_PUBREL & 0xF0U ):
 
             /* The second bit of a PUBREL must be set. */
-            if( packetType & 0x02U )
+            if( ( packetType & 0x02U ) > 0U )
```
muneebahmed10, line 593: Is this really a MISRA violation? I thought that `if( <value> )` was a common idiom in C.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -332,7 +332,10 @@ static uint8_t * encodeString( uint8_t * pDestination,
     *pBuffer = UINT16_LOW_BYTE( sourceLength );
     pBuffer++;
 
-    /* Copy the string into pBuffer. */
+    /* Copy the string into pBuffer.
+     * This memcpy intentionally copies bytes from a void * buffer into
+     * a uint8_t * buffer. */
+    /* coverity[misra_c_2012_rule_21_15_violation] */
```
muneebahmed10, line 338: I know we had this message before, but I think we should try to have as few suppressions as possible going forward, and it looks like it might be simple to fix. Do either of the following work, or do they lead to another violation?
```
char *pDest = ( char * ) pBuffer;
( void ) memcpy( pDest, source, sourceLength);
```
or
```
const uint8_t *pSourceStr = ( const uint8_t * ) source;
( void ) memcpy( pBuffer, pSourceStr, sourceLength );
```

libraries/standard/mqtt/src/mqtt_lightweight.c
leegeth: No. It is not used anywhere and I see that PR #901 brings a read function

libraries/standard/mqtt/src/mqtt_lightweight.c
leegeth: Yes. This was a violation. 

Event misra_c_2012_rule_14_4_violation: | The condition expression "packetType & 2U" does not have an essentially boolean type
-- | --

libraries/standard/mqtt/src/mqtt_lightweight.c
leegeth: Addressed it. Thanks for the suggestions

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -533,10 +506,11 @@ static void serializePublishCommon( const MQTTPublishInfo_t * pPublishInfo,
         LogDebugWithArgs( "Copying PUBLISH payload of length =%lu to buffer",
                           pPublishInfo->payloadLength );
 
-        /* This memcpy intentionally copies bytes from a void * buffer into
-         * a uint8_t * buffer. */
-        /* coverity[misra_c_2012_rule_21_15_violation] */
-        ( void ) memcpy( pIndex, pPublishInfo->pPayload, pPublishInfo->payloadLength );
+        /* Typecast const void * typed payload buffer to const uint8_t *.
+         * This is to use same type buffers in memcpy. */
+        pPayloadBuffer = ( const uint8_t * ) pPublishInfo->pPayload;
```
muneebahmed10, line 511: Isn't casting from void * to anything another MISRA violation? So we'll have to suppress a violation anyway

leegeth: No. This cast didn't throw any violation when using our config

muneebahmed10: That's likely due to this: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/scripts/coverity_misra.config#L35

Should we consider re-enabling it for this branch?

leegeth: That brings up a good point, I think we should evaluate all the deviations that we have added in the config for the development branch.
@abhidixi11 What do you think?

PR #928, opened 2020-05-08
Implement HTTPClient_ReadHeader() Part 2

aggarw13:
*Description of changes:*
* Complete functionality for `HTTPClient_ReadHeader` API function by implementing the internal `findHeaderInResponse` function that makes calls to the 3rd party `http-parser` 
* Update unit tests to mock out `http-parser` function dependencies

- [x] Complexity score to be <=8


libraries/standard/http/src/http_client.c
```diff
@@ -929,16 +953,230 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
+        IotLogErrorWithArgs( "Header found in response but http-parser returned error: ParserError=%s",
+                             http_errno_description( HTTP_PARSER_ERRNO( &( parser ) ) ) );
+        returnStatus = HTTP_INTERNAL_ERROR;
+    }
+    else if( ( returnStatus == HTTP_HEADER_NOT_FOUND ) && ( ( parser.http_errno != HPE_OK ) ) )
+    {
+        IotLogErrorWithArgs( "Header not found search in response: http-parser returned error: ParserError=%s",
+                             http_errno_description( HTTP_PARSER_ERRNO( &( parser ) ) ) );
+        returnStatus = HTTP_INVALID_RESPONSE;
+    }
```
sarenameas, line 1224: empty else needed here?

libraries/standard/http/src/http_client.c
```diff
@@ -929,16 +953,230 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
+                             "HeaderName=%.*s, HeaderValue=%.*s",
+                             fieldLen, pField,
+                             *pValueLen, *pValueLoc );
+    }
+
+    if( ( returnStatus == HTTP_SUCCESS ) && ( ( parser.http_errno != HPE_CB_header_value ) ) )
+    {
+        IotLogErrorWithArgs( "Header found in response but http-parser returned error: ParserError=%s",
+                             http_errno_description( HTTP_PARSER_ERRNO( &( parser ) ) ) );
+        returnStatus = HTTP_INTERNAL_ERROR;
```
sarenameas, line 1217: Could you please document this error in the HTTPStatus_t enum since you are using it?

libraries/standard/http/src/http_client.c
```diff
@@ -929,16 +953,230 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
+    }
+    else if( ( returnStatus == HTTP_HEADER_NOT_FOUND ) && ( ( parser.http_errno != HPE_OK ) ) )
+    {
+        IotLogErrorWithArgs( "Header not found search in response: http-parser returned error: ParserError=%s",
```
sarenameas, line 1221: "Header not found search in response"

"Header not found in response search"??

libraries/standard/http/src/http_client.c
```diff
@@ -929,16 +1031,210 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
+                         ( ( findHeaderContext_t * ) pHttpParser->data )->fieldLen,
+                         ( ( findHeaderContext_t * ) pHttpParser->data )->pField );
+
+    /* No further parsing is required; thus, indicate the parser to stop parsing. */
+    return HTTP_PARSER_STOP_PARSING;
+}
+
+/*-----------------------------------------------------------*/
+
+static HTTPStatus_t findHeaderInResponse( const uint8_t * pBuffer,
```
sarenameas, line 1137: Does findHeaderInResponse.c.gcov have 100% coverage?
```
function findHeaderInResponse called N returned 100% blocks executed 100%
```

libraries/standard/http/test/test-HTTPClient_ReadHeader.c
```diff
@@ -13,80 +32,109 @@ static const char * pTestResponse = "HTTP/1.1 200 OK"
                                     "test-header1: test-value1\r\n"
                                     "test-header2: test-value2\r\n"
                                     "\r\n";
-static const size_t headerFieldInRespLoc = 15;
-static const size_t headerFieldInRespLen = strlen( "test-header0" );
-static const size_t headerValInRespLoc = 29;
-static const size_t headerValInRespLen = strlen( "test-value0" );
+static const size_t headerFieldInRespLoc = 42;
+static const size_t headerFieldInRespLen = strlen( "test-header1" );
```
sarenameas, line 36: Please use these instead:
```
static const size_t headerFieldInRespLen = sizeof( "test-header1" ) - 1u;
...
static const size_t headerValInRespLen = sizeof( "test-value1" ) - 1u;
```

libraries/standard/http/src/http_client.c
```diff
@@ -929,16 +953,230 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
 
 /*-----------------------------------------------------------*/
 
+static const int HTTP_PARSER_STOP_PARSING = 1;
```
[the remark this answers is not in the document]
aggarw13: Are these expected to be used in any other source?

libraries/standard/http/test/test-HTTPClient_ReadHeader.c
aggarw13: What's the reason of preferring `sizeof` over `strlen` in a test file? We have control over the test response used in the test file to be confident that it is a constant string...

libraries/standard/http/src/http_client.c
```diff
@@ -929,16 +953,230 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
+static const int HTTP_PARSER_CONTINUE_PARSING = 0;
+
+/**
+ * @brief An aggregator that represents the user-provided parameters to the
+ * #HTTPClient_ReadHeader API function, to be used as context parameter
```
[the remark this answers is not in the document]
aggarw13: Corrected, thanks

libraries/standard/http/src/http_client.c
aggarw13: Fixed

libraries/standard/http/src/http_client.c
aggarw13: Added

libraries/standard/http/src/http_client.c
```diff
@@ -929,16 +953,230 @@ HTTPStatus_t HTTPClient_Send( const HTTPTransportInterface_t * pTransport,
...
+    else if( context.valueFound == 0u )
+    {
+        /* The response buffer is invalid as only the header field was found
+         * in the "<field>: <value>\r\n" format of an HTTP header. */
+        /* IotLogErrorWithArgs( "Unable to find header value in response: " */
+        /*                      "Unexpected state of http-parser: RequestedHeader=%.*s", */
+        /*                      "ParserError=%s", */
+        /*                      fieldLen, pField, */
+        /*                      http_errno_description( HTTP_PARSER_ERRNO( &( parser ) ) ) ); */
+        returnStatus = HTTP_INTERNAL_ERROR;
```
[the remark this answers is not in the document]
aggarw13: Based on my testing of the `http-parser`, if a header field is parsed by the http-parser, then anything after the `: ` would be treated as the value. This represents the error condition when only the field is found, but the value isn't. Thus, it represents a parser error, which is internal to the library

libraries/standard/http/src/http_client.c
aggarw13: Added

libraries/standard/http/src/http_client.c
aggarw13: Nope, this is what I get 
`function findHeaderInResponse called 10 returned 100% blocks executed 90%`

libraries/standard/http/test/test-HTTPClient_ReadHeader.c
sarenameas: Using strlen() here is a bit weird, I thought global static variables needed to be initialized with expressions known at compile time. This also gets rid of the warning spit out.

libraries/standard/http/src/http_client.c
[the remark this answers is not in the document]
sarenameas: Most of the macros in that header are not used by other source. At this point it's for consistency with having only functions and their prototypes in this file.

libraries/standard/http/test/test-HTTPClient_ReadHeader.c
aggarw13: Good point about using compile-time constants. (In C++, they use a separate type specifier `constexpr` to represent compile-time computable expressions). Changed to `sizeof()`

libraries/standard/http/src/http_client.c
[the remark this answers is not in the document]
aggarw13: Okay, moved to the internal header.

libraries/standard/http/src/http_client.c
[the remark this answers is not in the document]
sarenameas: If the header field is found, and if anything after ": " is considered value, then valueFound would be set to 1 for the characters after. 
If the header field is found, and the buffer ends at ": " or the buffer ends without ": ". then valueFound would be set to 0 and that would be an invalid response, not an internal error to the library.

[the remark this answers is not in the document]
aggarw13: I had tested the case when the buffer ends without `: ` and http-parser recognizes everything after the `: ` as the value. 
Test Response:
```
 const char * pInvalidResponse = "HTTP/1.1 200 OK"
                                    "test-header0: test-value0\r\n"
                                    "test-header1: "
                                    "test-header2test-value2";
```
Header searched was `test-header1` and the value it parsed out was `test-header2test-value2`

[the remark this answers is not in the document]
aggarw13: We discussed offline and @sarenameas clarified that she referred to the following case when the header field is found but the value is not found by the `http-parser`.
```
"HTTP/1.1 200 OK"
"test-header0: test-value0\r\n"
 "test-header1: "
```
Thus, I have updated the error handling for this case to return `INVALID_RESPONSE`  to the user.

PR #930, opened 2020-05-12
Convert HTTPClient_InitializeRequestHeaders unit test to use Unity

yourslab:
No mocks are being used here. Same set of tests as the C Preprocessor variant.

### DIFF
--- a/libraries/standard/http/utest/CMakeLists.txt
+++ b/libraries/standard/http/utest/CMakeLists.txt
@@ -31,8 +31,9 @@ list(APPEND real_source_files
 list(APPEND real_include_directories
             .
             ${HTTP_INCLUDE_PUBLIC_DIRS}
-            ${HTTP_INCLUDE_PRIVATE_DIRS}
-            ${LOGGING_INCLUDE_DIRS}
+            ${HTTP_TEST_INCLUDE_PRIVATE_DIRS}
+            "${ROOT_DIR}/platform/include"
+            "${MODULES_DIR}/standard/http/third_party"
         )
 
 # =====================  Create UnitTest Code here (edit)  =====================
--- a/libraries/standard/http/utest/CMakeLists.txt
+++ b/libraries/standard/http/utest/CMakeLists.txt
@@ -25,7 +25,7 @@ list(APPEND mock_define_list
 # list the files you would like to test here
 list(APPEND real_source_files
             ${HTTP_SOURCES}
-            ${LOGGING_SOURCES}
+            "${MODULES_DIR}/standard/http/third_party/http_parser/http_parser.c"
         )
 # list the directories the module under test includes
 list(APPEND real_include_directories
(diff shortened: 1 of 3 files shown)


libraries/standard/http/utest/http_utest.c
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/* ============== Testing HTTPClient_InitializeRequestHeaders =============== */
+
+/**
+ * @brief Initialize pRequestInfo with test-defined macros.
+ *
+ * @param[in] pRequestInfo Initial request header configurations.
+ */
+void setupRequestInfo( HTTPRequestInfo_t * pRequestInfo )
```
sarenameas, line 88: These should be static

libraries/standard/http/utest/CMakeLists.txt
```diff
@@ -25,13 +25,15 @@ list(APPEND mock_define_list
 # list the files you would like to test here
 list(APPEND real_source_files
             ${HTTP_SOURCES}
+            "${ROOT_DIR}/libraries/standard/http/third_party/http_parser/http_parser.c"
```
aggarw13, line 28: What about using the `MODULES_DIR` variable instead?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -195,6 +195,8 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * const pContext,
                                             pWillInfo,
                                             &remainingLength,
                                             &packetSize );
+        printf( "%d\n", status );
```
aggarw13, line 198: What is this change for?

libraries/standard/http/utest/http_utest.c
```diff
@@ -9,6 +9,55 @@
 #include "private/http_client_internal.h"
 #include "private/http_client_parse.h"
 
+#define HTTP_TEST_REQUEST_METHOD        "GET"
```
aggarw13, line 12: I suggest using strongly typed variables instead of macros (to obtain type checking from the compiler). Will not hold the PR though if you prefer using macros for constants 😃

libraries/standard/http/utest/CMakeLists.txt
```diff
@@ -25,13 +25,15 @@ list(APPEND mock_define_list
 # list the files you would like to test here
 list(APPEND real_source_files
             ${HTTP_SOURCES}
+            "${MODULES_DIR}/standard/http/third_party/http_parser/http_parser.c"
```
aggarw13, line 28: We would need these to be mocked instead of building the source of the 3rd party. 
(However, that could be handled in a separate PR)

libraries/standard/http/utest/http_utest.c
sarenameas: This is already defined in the http_client.h, can you not reuse it?

PR #934, opened 2020-05-13
MQTT CONNACK receive in MQTT_Connect

leegeth:
*Description of changes:*
MQTT CONNACK receive in MQTT_Connect.

MQTT CONNACK receive is updated by using the functions MQTT_GetIncomingPacketTypeAndLength and receivePacket. 

receivePacket will be implemented when merging PR #901. Currently only a stub is used.

### DIFF
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -72,7 +72,7 @@
 /**
  * @brief Timeout for receiving CONNACK packet in milli seconds.
  */
-#define CONNACK_RECV_TIMEOUT    (1000)
+#define CONNACK_RECV_TIMEOUT_MS    (1000)
 
 /*-----------------------------------------------------------*/
 
@@ -353,7 +353,7 @@ static int establishMqttSession( MQTTContext_t * pContext, SSL * pSslContext )
     connectInfo.pPassword = NULL;
     connectInfo.passwordLength = 0;
 
-    mqttStatus = MQTT_Connect( pContext, &connectInfo, NULL, CONNACK_RECV_TIMEOUT, NULL );
+    mqttStatus = MQTT_Connect( pContext, &connectInfo, NULL, CONNACK_RECV_TIMEOUT_MS, NULL );
 
     if( mqttStatus != MQTTSuccess )
     {
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -69,6 +69,11 @@
  */
 #define CLIENT_IDENTIFIER_LENGTH    ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) )
 
+/**
+ * @brief Timeout for receiving CONNACK packet in milli seconds.
+ */
+#define CONNACK_RECV_TIMEOUT    (1000)
+
 /*-----------------------------------------------------------*/
 
 /**
@@ -348,7 +353,7 @@ static int establishMqttSession( MQTTContext_t * pContext, SSL * pSslContext )
     connectInfo.pPassword = NULL;
     connectInfo.passwordLength = 0;
 
-    mqttStatus = MQTT_Connect( pContext, &connectInfo, NULL, NULL );
+    mqttStatus = MQTT_Connect( pContext, &connectInfo, NULL, CONNACK_RECV_TIMEOUT, NULL );
 
     if( mqttStatus != MQTTSuccess )
     {
--- a/demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
+++ b/demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
@@ -62,6 +62,11 @@
  */
 #define CLIENT_IDENTIFIER_LENGTH    ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) )
 
+/**
+ * @brief Timeout for receiving CONNACK packet in milli seconds.
+ */
+#define CONNACK_RECV_TIMEOUT    (1000)
+
 /*-----------------------------------------------------------*/
 
 /**
@@ -242,7 +247,7 @@ static int establishMqttSession( MQTTContext_t * pContext, int tcpSocket )
     connectInfo.pPassword = NULL;
     connectInfo.passwordLength = 0;
 
-    mqttStatus = MQTT_Connect( pContext, &connectInfo, NULL, NULL );
+    mqttStatus = MQTT_Connect( pContext, &connectInfo, NULL, CONNACK_RECV_TIMEOUT, NULL );
 
     if( mqttStatus != MQTTSuccess )
     {
--- a/demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
+++ b/demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
@@ -65,7 +65,7 @@
 /**
  * @brief Timeout for receiving CONNACK packet in milli seconds.
  */
-#define CONNACK_RECV_TIMEOUT    (1000)
+#define CONNACK_RECV_TIMEOUT_MS    (1000)
 
 /*-----------------------------------------------------------*/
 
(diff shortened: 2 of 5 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -151,6 +169,61 @@ static MQTTStatus_t validateSubscribeUnsubscribeParams( const MQTTContext_t * co
...
+             * read for CONNACK. */
+            status = receivePacket( pContext,
+                                    *incomingPacket,
+                                    0U );
+        }
+
+        /* Loop until there is data to read or the retry count has not
+         * exceeded the limit. */
+    } while( ( status == MQTTNoDataAvailable ) &&
+             ( ++retryCount < CONNACK_RECV_RETRY_LIMIT ) );
```
muneebahmed10, line 210: Are we sure we want to use a loop counter and not a timer? Based on how MQTT_Connect() is written, this will execute almost immediately after sending the CONNECT packet. What if the CONNACK is not available on the transport layer by that time?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -25,6 +25,10 @@
 #include "mqtt.h"
 #include "private/mqtt_internal.h"
 
+
+/* Transport recv retry limit for CONNACK packet. */
+#define CONNACK_RECV_RETRY_LIMIT    3U
```
muneebahmed10, line 30: Whenever I read "retry limit" I have to think for a bit, because we count the initial attempt as a "retry". Can the comment reflect that the initial attempt is included in this number?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -151,6 +169,61 @@ static MQTTStatus_t validateSubscribeUnsubscribeParams( const MQTTContext_t * co
...
+                                                      incomingPacket );
+
+        if( status == MQTTSuccess )
+        {
+            /* Reading the remainder of the packet by transport recv.
+             * The timeout is passed as 0 as it is only 2 more bytes to
+             * read for CONNACK. */
+            status = receivePacket( pContext,
+                                    *incomingPacket,
+                                    0U );
```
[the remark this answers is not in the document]
leegeth: yes. That's correct. Since we have successfully read packet type and remaining length from the network. It will be 2 more bytes to read from the network buffer.

libraries/standard/mqtt/src/mqtt.c
leegeth: It may not be immediately executed because there is a timeout associated with a transport recv. So the sequence will be 
1. Send Connect packet using transport send.
2. Call transport recv. Read network buffer if data is available, otherwise timeout after the implementation configured transport recv timeout.
3. Keep calling the recv until the configured retry limit or a success or a failure to read data.
4. Read packet type and length.
5. Read the remainder of packet. for CONNACK it will be 2 bytes.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -151,6 +169,61 @@ static MQTTStatus_t validateSubscribeUnsubscribeParams( const MQTTContext_t * co
...
+    MQTTStatus_t status = MQTTSuccess;
+    uint8_t retryCount = 0U;
+
+    assert( pContext != NULL );
+    assert( incomingPacket != NULL );
+
+    do
+    {
+        /* Transport read for incoming CONNACK packet type and length. */
+        status = MQTT_GetIncomingPacketTypeAndLength( pContext->transportInterface.recv,
```
abhidixi11, line 193: We must verify here that incoming packet is indeed a conn ack, before we try to read remaining packet.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
muneebahmed10: It makes sense to use recvExact. However, the check pContext->networkBuffer.size >= 2 should be performed first as recvExact assumes the buffer is large enough for the packet

libraries/standard/mqtt/src/mqtt.c
leegeth: Agree.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
leegeth: I think it still makes sense to use recvPacket as it has a wrapper logic to verify the length of the buffer and logic to dump the packet.

libraries/standard/mqtt/src/mqtt.c
leegeth: Changed the logic to use timeouts. Please check

libraries/standard/mqtt/src/mqtt.c
leegeth: Removed the retry logic.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -214,6 +233,122 @@ static MQTTStatus_t sendPublish( const MQTTContext_t * const pContext,
...
+                                                      pContext->transportInterface.networkContext,
+                                                      incomingPacket );
+
+        /* Loop until there is data to read or if the timeout has not expired. */
+    } while( ( status == MQTTNoDataAvailable ) &&
+             ( calculateElapsedTime( getTimeStamp(), entryTime ) < timeoutMs ) );
+
+    if( status == MQTTSuccess )
+    {
+        /* Check if the timeout has expired. */
```
abhidixi11, line 291: I think we should make at least one read attempt if we have successfully received type and length. Basically time check block can be deleted.

leegeth: Yes Done

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -69,6 +69,11 @@
  */
 #define CLIENT_IDENTIFIER_LENGTH    ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) )
 
+/**
+ * @brief Timeout for receiving CONNACK packet in milli seconds.
+ */
+#define CONNACK_RECV_TIMEOUT    (1000)
```
gkwicker, line 75: Units need to be in the name, e.g. `CONNACK_RECV_TIMEOUT_MS`.

demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
```diff
@@ -62,6 +62,11 @@
  */
 #define CLIENT_IDENTIFIER_LENGTH    ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) )
 
+/**
+ * @brief Timeout for receiving CONNACK packet in milli seconds.
+ */
+#define CONNACK_RECV_TIMEOUT    (1000)
```
gkwicker, line 68: See previous comment, `CONNACK_RECV_TIMEOUT_MS`.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -214,6 +245,107 @@ static MQTTStatus_t sendPublish( const MQTTContext_t * const pContext,
...
+/*-----------------------------------------------------------*/
+
+static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
+                                    uint32_t timeoutMs,
+                                    MQTTPacketInfo_t * const incomingPacket,
+                                    bool * const pSessionPresent )
+{
+    MQTTStatus_t status = MQTTSuccess;
+    MQTTGetCurrentTimeFunc_t getTimeStamp = NULL;
+    uint32_t entryTime = 0U, remainingTimeMs = 0U, timeTaken = 0U;
```
gkwicker, line 272: All time values should have units in the name, e.g. `entryTimeMs`, `timeTakenMs`.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -214,6 +245,107 @@ static MQTTStatus_t sendPublish( const MQTTContext_t * const pContext,
...
+    assert( pContext->callbacks.getTime != NULL );
+
+    getTimeStamp = pContext->callbacks.getTime;
+    /* Get the entry time for the function. */
+    entryTime = getTimeStamp();
+
+    do
+    {
+        /* Transport read for incoming CONNACK packet type and length. */
+        status = MQTT_GetIncomingPacketTypeAndLength( pContext->transportInterface.recv,
```
gkwicker, line 285: I'd note that this is a blocking call so that the reader knows this isn't a busy waiting loop.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: Done

demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
leegeth: Fixed.

libraries/standard/mqtt/src/mqtt.c
leegeth: Added it

libraries/standard/mqtt/src/mqtt.c
leegeth: Done.

PR #935, opened 2020-05-13
Update publish with state machine updates

leegeth:
*Description of changes:*
Update MQTT PUBLISH with state machine updates.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -393,6 +393,7 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
     size_t remainingLength = 0UL, packetSize = 0UL, headerSize = 0UL;
     int32_t bytesSent = 0;
     MQTTStatus_t status = MQTTSuccess;
+    MQTTPublishState_t publishStatus = MQTTStateNull;
 
     /* Validate arguments. */
     if( ( pContext == NULL ) || ( pPublishInfo == NULL ) )
@@ -453,6 +454,10 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
         status = sendPublish( pContext,
                               pPublishInfo,
                               headerSize );
+
+        /* TODO. When a publish fails, the reserved state has to be cleaned
+         * up. This will have to be done once an API in state machine is
+         * available. */
     }
 
     if( status == MQTTSuccess )
@@ -461,10 +466,26 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
          * Only to be done for QoS1 or QoS2. */
         if( pPublishInfo->qos > MQTTQoS0 )
         {
-            status = MQTT_UpdateStatePublish( pContext,
-                                              packetId,
-                                              MQTT_SEND,
-                                              pPublishInfo->qos );
+            /* TODO MQTT_UpdateStatePublish will be updated to return
+             * MQTTStatus_t instead of MQTTPublishState_t. Update the
+             * code when that change is made. */
+            publishStatus = MQTT_UpdateStatePublish( pContext,
+                                                     packetId,
+                                                     MQTT_SEND,
+                                                     pPublishInfo->qos );
+
+            if( publishStatus == MQTTStateNull )
+            {
+                LogErrorWithArgs( "Update state for publish failed with status =%u."
+                                  " However PUBLISH packet is sent to the broker."
+                                  " Any further handling of ACKs for the packet Id"
+                                  " will fail.",
+                                  publishStatus );
+
+                /* TODO. Need to remove this update once MQTT_UpdateStatePublish is
+                 * refactored with return type of MQTTStatus_t. */
+                status = MQTTBadParameter;
+            }
         }
     }
 
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -23,6 +23,7 @@
 #include <assert.h>
 
 #include "mqtt.h"
+#include "mqtt_state.h"
 #include "private/mqtt_internal.h"
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -375,41 +438,41 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
...
     if( status == MQTTSuccess )
     {
-        /* Send header first. */
-        bytesSent = sendPacket( pContext,
-                                pContext->networkBuffer.pBuffer,
-                                headerSize );
-
-        if( bytesSent < 0 )
+        /* Reserve state for publish message. Only to be done for QoS1 or QoS2. */
+        if( pPublishInfo->qos > MQTTQoS0 )
```
muneebahmed10, line 442: This isn't strictly necessary since `MQTT_ReserveState()` was written with the possibility of calling it with QoS 0 in mind. However, since that function is meant to be internal use only, it presents us with the opportunity to convert its checks into asserts and reduce the amount of code. It isn't necessary for this PR, but we need to remember to go back to it and remove its unnecessary code.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -375,41 +438,41 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
...
-            {
-                LogError( "Transport send failed for PUBLISH payload." );
-                status = MQTTSendFailed;
-            }
-            else
-            {
-                LogDebugWithArgs( "Sent %d bytes of PUBLISH payload.",
-                                  bytesSent );
-            }
+    if( status == MQTTSuccess )
```
muneebahmed10, line 458: What happens if the publish was not successful? Should we add an `MQTT_PurgeState()` function to cleanup and remove the packet id from the state record?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -375,41 +438,41 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
...
+    {
+        /* Update state machine after PUBLISH is sent.
+         * Only to be done for QoS1 or QoS2. */
+        if( pPublishInfo->qos > MQTTQoS0 )
+        {
+            status = MQTT_UpdateStatePublish( pContext,
```
muneebahmed10, line 464: `MQTT_UpdateStatePublish` returns an `MQTTPublishState_t`, not an `MQTTStatus_t`. This is done so that the library knows the next state of incoming publishes, and whether it should send a PUBACK or PUBREC. However, we might want to refactor it to return a status and have the publish state as an output parameter. Or, since these functions are internal, we can split them into separate functions for incoming and outgoing publishes (I think this option would be better since it will result in the simplest functions). If we want to do that, then it should be done before this PR is merged.

libraries/standard/mqtt/src/mqtt.c
abhidixi11: Should we add state only after publish is successful ?

libraries/standard/mqtt/src/mqtt.c
leegeth: I don't totally agree with this. Our current comments (https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/mqtt/include/mqtt_state.h#L48) indicates that it is either QoS1 or 2. Also I think if it has to be called only for QoS1 or 2, it will definitely help the readability of the function that the state machine has to be invoked only specific QoSs instead for all and then asserting it. And what happens for QoS0 publishes with asserts enabled.

muneebahmed10: There might be some confusion, but my suggestion was to update `MQTT_ReserveState()` to change its `if(qos == MQTTQoS0)` into just an `assert(qos != MQTTQoS0)`, since this should be the only place that function is called from. The code in this file should stay as it is.

libraries/standard/mqtt/src/mqtt.c
leegeth: I agree a refactor would be good. I think definitely the return type has to be addressed in this PR itself. Thanks for that catch.  This PR doesn't need to be blocked for that change.  Depending on which PR gets in first, It will just have to be handled in the rebase.

libraries/standard/mqtt/src/mqtt.c
leegeth: Ok. Yea I was thinking to update the call sites for the function. I have again a question related to the proposed optimization regarding the function, Isn't MQTT_ReserveState() a public API function? Asserts are not used in a public API. Given the composable tenet that we are following, I think it makes sense for us to have the state machine available for the customers in its full form rather than making MQTT_ReserveState() an internal function. What do you think?
Anyway, I think this can be handled in a separate PR based on what we decide.

libraries/standard/mqtt/src/mqtt.c
leegeth: > What happens if the publish was not successful? Should we add an `MQTT_PurgeState()` function to cleanup and remove the packet id from the state record?

Yes I think, this function would be required. I will add  a TODO to address that in the comments once it is available.

leegeth: > Should we add state only after publish is successful ?

no. We should try to send a publish only after MQTT_ReserveState() is successful.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -375,41 +439,61 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
...
+
+            if( publishStatus == MQTTStateNull )
             {
-                LogDebugWithArgs( "Sent %d bytes of PUBLISH payload.",
-                                  bytesSent );
+                LogErrorWithArgs( "Update state for publish failed with status =%u."
+                                  " However PUBLISH packet is sent to the broker."
+                                  " Any further handling of ACKs for the packet Id"
+                                  " will fail.",
+                                  publishStatus );
```
muneebahmed10, line 483: Don't really need this since we know it's MQTTStateNull, but we're going to change this on refactor anyway.

PR #937, opened 2020-05-13
Update logConnackResponse function to log CONNACK error response with…

aggarw13:
*Description of changes:*
Remove dependency of `LOG_NONE` from MQTT file.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -718,30 +718,31 @@ static void logConnackResponse( uint8_t responseCode )
     switch( responseCode )
     {
         case 0u:
-            LogError( "Connection accepted." );
+            LogInfo( "Connection accepted." );
             break;
 
         case 1u:
-            LogError( "Connection refused: unacceptable protocol version." );
+            LogInfo( "Connection refused: unacceptable protocol version." );
             break;
 
         case 2u:
-            LogError( "Connection refused: identifier rejected." );
+            LogInfo( "Connection refused: identifier rejected." );
             break;
 
         case 3u:
-            LogError( "Connection refused: server unavailable" );
+            LogInfo( "Connection refused: server unavailable" );
             break;
 
         case 4u:
-            LogError( "Connection refused: bad user name or password." );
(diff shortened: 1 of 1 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -714,21 +714,36 @@ static void logConnackResponse( uint8_t responseCode )
...
-            "Connection refused: bad user name or password.",     /* 4 */
-            "Connection refused: not authorized."                 /* 5 */
-        };
-        LogErrorWithArgs( "%s", pConnackResponses[ responseCode ] );
-    #endif /* if LIBRARY_LOG_LEVEL > LOG_NONE */
+    /* Log an error based on the CONNACK response code. */
+    switch( responseCode )
+    {
+        case 0u:
+            LogError( "Connection accepted." );
```
muneebahmed10, line 721: If each response code gets its own log statement then let's take the opportunity to change this to `LogInfo`, since it's indicating success.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -714,21 +714,36 @@ static void logConnackResponse( uint8_t responseCode )
...
+        case 4u:
+            LogError( "Connection refused: bad user name or password." );
+            break;
+
+        case 5u:
+            LogError( "Connection refused: not authorized." );
+            break;
+
+        default:
+            break;
```
muneebahmed10, line 745: There should be a comment in the default branch for MISRA

libraries/standard/mqtt/src/mqtt_lightweight.c
aggarw13: Okay, strange that it was logged as error earlier.

libraries/standard/mqtt/src/mqtt_lightweight.c
aggarw13: Okay, added

libraries/standard/mqtt/src/mqtt_lightweight.c
aggarw13: Changed to `LogInfo`

PR #942, opened 2020-05-15
test: MQTT_SerializeConnect unit tests using Unity Framework

yourslab:
Create two methods that achieve 100% branch coverage on MQTT_SerializeConnect:
- `test_MQTT_SerializeConnect_invalid_params` calls Mqtt_SerializeConnect using NULL parameters and insufficient buffer size until we receive all possible MQTTBadParameter and MQTTNoMemory errors.
- `test_MQTT_SerializeConnect_happy_paths` calls MQTT_SerializeConnect successfully using different parameters until we have full coverage on the private method, serializeConnectPacket(...).

### DIFF
--- a/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
+++ b/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
@@ -14,24 +14,30 @@
  * @brief Test-defined macro for MQTT username.
  */
 #define MQTT_TEST_USERNAME               "username"
-#define MQTT_TEST_USERNAME_LEN           sizeof( MQTT_TEST_USERNAME ) - 1
+#define MQTT_TEST_USERNAME_LEN           ( sizeof( MQTT_TEST_USERNAME ) - 1 )
 
 /**
  * @brief Test-defined macro for MQTT password.
  */
 #define MQTT_TEST_PASSWORD               "password"
-#define MQTT_TEST_PASSWORD_LEN           sizeof( MQTT_TEST_PASSWORD ) - 1
+#define MQTT_TEST_PASSWORD_LEN           ( sizeof( MQTT_TEST_PASSWORD ) - 1 )
 
 /**
  * @brief Test-defined macro for MQTT topic.
  */
 #define MQTT_TEST_TOPIC                  "topic"
-#define MQTT_TEST_TOPIC_LEN              sizeof( MQTT_TEST_TOPIC ) - 1
+#define MQTT_TEST_TOPIC_LEN              ( sizeof( MQTT_TEST_TOPIC ) - 1 )
 
 /**
  * @brief Length of the client identifier.
  */
-#define CLIENT_IDENTIFIER_LEN            sizeof( CLIENT_IDENTIFIER ) - 1
+#define MQTT_CLIENT_IDENTIFIER_LEN       ( sizeof( MQTT_CLIENT_IDENTIFIER ) - 1 )
+
+/**
+ * @brief Payload for will info.
+ */
+#define MQTT_SAMPLE_PAYLOAD              "payload"
+#define MQTT_SAMPLE_PAYLOAD_LEN          ( sizeof( MQTT_SAMPLE_PAYLOAD ) - 1 )
 
 /* MQTT CONNECT flags. */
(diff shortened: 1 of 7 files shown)


libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,177 @@
+#include <stdbool.h>
```
leegeth, line 1: Please don't include stdbool.h directly as it is a C99+ header. I think you can just delete it as it is handled here
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/mqtt/include/mqtt_lightweight.h#L26

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,177 @@
...
+    /* Repeat with QoS2. */
+    willInfo.qos = MQTTQoS2;
+    mqttStatus = MQTT_GetConnectPacketSize( &connectInfo,
+                                            &willInfo,
+                                            &remainingLength,
+                                            &packetSize );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    mqttStatus = MQTT_SerializeConnect( &connectInfo, &willInfo,
+                                        remainingLength, &networkBuffer );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
```
leegeth, line 174: I think it would make a good addition to validate the serialized buffer values as well here.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,425 @@
...
+
+/* Called at the beginning of the whole suite. */
+void suiteSetUp()
+{
+}
+
+/* Called at the end of the whole suite. */
+int suiteTearDown( int numFailures )
+{
+    ( void ) numFailures;
```
yourslab, line 105: Need to return this

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,449 @@
...
+                                            &remainingLength,
+                                            &packetSize );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    /* Make sure buffer has enough space. */
+    TEST_ASSERT_GREATER_OR_EQUAL( packetSize, networkBuffer.size );
+    mqttStatus = MQTT_SerializeConnect( &connectInfo, &willInfo,
+                                        remainingLength, &networkBuffer );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    verifySerializedConnectPacket( &connectInfo, &willInfo,
+                                   remainingLength, &networkBuffer );
```
leegeth, line 446: I think a few more cases can be added here to check different branches( Will Info being null, username and  password being null, clean session as false, retain flag in willinfo as false. etc. )

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,487 @@
...
+
+
+    /* Re-initialize objects for branch coverage. */
```
leegeth, line 451: Please describe what is the combination getting tested here.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,496 @@
...
+                            uint16_t sourceLength )
+{
+    uint8_t * pBuffer = NULL;
+
+    /* Typecast const char * typed source buffer to const uint8_t *.
+     * This is to use same type buffers in memcpy. */
+    const uint8_t * pSourceBuffer = ( const uint8_t * ) source;
+
+    TEST_ASSERT_NOT_NULL( pDestination );
+
```
gkwicker, line 210: Why no `TEST_ASSERT_NOT_NULL( source );` here?

yourslab: Good point. I think it would be more appropriate to just call this `pBuffer` instead.

PR #955, opened 2020-05-19
Fix issues in MQTT library.

leegeth:
Below are the issues fixed in this commit
1. Remove the incorrect and duplicate pointer initialization in
sendPacket function.
2. Remove NULL check for the packet Id pointer passed to MQTT_DeserializeAck
when packet type is either CONNACK or PINGRESP.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -1943,9 +1943,8 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * const pIncomingPacket
     if( ( pIncomingPacket == NULL ) || ( pSessionPresent == NULL ) )
     {
         LogErrorWithArgs( "Argument cannot be NULL: pIncomingPacket=%p, "
-                          "pPacketId=%p, pSessionPresent=%p",
+                          "pSessionPresent=%p",
                           pIncomingPacket,
-                          pPacketId,
                           pSessionPresent );
         status = MQTTBadParameter;
     }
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -1949,6 +1949,16 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * const pIncomingPacket
                           pSessionPresent );
         status = MQTTBadParameter;
     }
+    /* Pointer for packet identifier cannot be NULL for packets other than
+     * CONNACK and PINGRESP. */
+    else if( ( pPacketId == NULL ) &&
```
muneebahmed10, line 1953: Do we want to allow args to be `NULL`? Also, if we do this, then we might as well allow `pSessionPresent` to be `NULL` for everything except CONNACK

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -1940,15 +1940,24 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * const pIncomingPacket
...
+             ( ( pIncomingPacket->type != MQTT_PACKET_TYPE_CONNACK ) &&
+               ( pIncomingPacket->type != MQTT_PACKET_TYPE_PINGRESP ) ) )
+    {
+        LogErrorWithArgs( "pPacketId cannot be NULL for packet type %u.",
```
muneebahmed10, line 1957: Nit: either `%02x` or some strerror utility will make this more readable

libraries/standard/mqtt/src/mqtt_lightweight.c
leegeth: I think yes. It can definitely save some memory from the calling function if they don't need those params. You are right about pSessionPresent. I will add that to the PR.

PR #957, opened 2020-05-20
Import lightweight unit tests

muneebahmed10:
*Description of changes:*

Import unit tests for lightweight API from [existing tests](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c#L2000) (Line 2000 onwards). Does not include tests for functions that are not in v4_beta (`MQTT_SerializeAck()` and `MQTT_SerializePublishHeader()`)

Total coverage:
* Lines: 91% coverage
* Functions: 91% coverage
* Branches: 90% coverage

Tested functions have 100% coverage. Besides altering names and parameters to fit the new API, some changes were made to existing tests to achieve coverage, which are called out in the comments below.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -847,7 +847,7 @@ static MQTTStatus_t calculateSubscriptionPacketSize( const MQTTSubscribeInfo_t *
     packetSize += sizeof( uint16_t );
 
     /* Sum the lengths of all subscription topic filters; add 1 byte for each
-     * subscription's QoS if type is IOT_MQTT_SUBSCRIBE. */
+     * subscription's QoS if type is MQTT_SUBSCRIBE. */
     for( i = 0; i < subscriptionCount; i++ )
     {
         /* Add the length of the topic filter. MQTT strings are prepended
@@ -1316,8 +1316,12 @@ MQTTStatus_t MQTT_GetConnectPacketSize( const MQTTConnectInfo_t * const pConnect
                     pPacketSize ) );
         status = MQTTBadParameter;
     }
-
-    if( status == MQTTSuccess )
+    else if( ( pConnectInfo->clientIdentifierLength == 0U ) || ( pConnectInfo->pClientIdentifier == NULL ) )
+    {
+        LogError( ( "Mqtt_GetConnectPacketSize() client identifier must be set." ) );
+        status = MQTTBadParameter;
+    }
+    else
     {
         /* Add the length of the client identifier. */
         connectPacketSize += pConnectInfo->clientIdentifierLength + sizeof( uint16_t );
@@ -1396,6 +1400,15 @@ MQTTStatus_t MQTT_SerializeConnect( const MQTTConnectInfo_t * const pConnectInfo
                     pBuffer->size,
                     connectPacketSize ) );
         status = MQTTNoMemory;
+    }
+    else if ( ( pWillInfo != NULL ) && 
+              ( ( pWillInfo->pPayload == NULL ) || ( pWillInfo->pTopicName == NULL ) ) )
+    {
+        LogError( ( "Argument cannot be NULL: pWillInfo->pPayload=%p, "
+                    "pWillInfo->pTopicName=%p",
+                    pWillInfo->pPayload,
+                    pWillInfo->pTopicName ) );
+        status = MQTTBadParameter;
     }
     else
     {
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -337,7 +337,10 @@ static uint8_t * encodeString( uint8_t * pDestination,
     pBuffer++;
 
     /* Copy the string into pBuffer. */
-    ( void ) memcpy( pBuffer, pSourceBuffer, sourceLength );
+    if( pSourceBuffer != NULL )
+    {
+        ( void ) memcpy( pBuffer, pSourceBuffer, sourceLength );
+    }
 
     /* Return the pointer to the end of the encoded string. */
     pBuffer += sourceLength;
@@ -1267,6 +1270,7 @@ static void serializeConnectPacket( const MQTTConnectInfo_t * const pConnectInfo
         pIndex = encodeString( pIndex,
                                pWillInfo->pTopicName,
                                pWillInfo->topicNameLength );
+
         pIndex = encodeString( pIndex,
                                pWillInfo->pPayload,
                                ( uint16_t ) pWillInfo->payloadLength );
@@ -1401,13 +1405,9 @@ MQTTStatus_t MQTT_SerializeConnect( const MQTTConnectInfo_t * const pConnectInfo
                     connectPacketSize ) );
         status = MQTTNoMemory;
     }
-    else if ( ( pWillInfo != NULL ) && 
-              ( ( pWillInfo->pPayload == NULL ) || ( pWillInfo->pTopicName == NULL ) ) )
+    else if ( ( pWillInfo != NULL ) && ( pWillInfo->pTopicName == NULL ) )
     {
-        LogError( ( "Argument cannot be NULL: pWillInfo->pPayload=%p, "
-                    "pWillInfo->pTopicName=%p",
-                    pWillInfo->pPayload,
-                    pWillInfo->pTopicName ) );
+        LogError( ( "pWillInfo->pTopicName cannot be NULL if Will is present." ) );
         status = MQTTBadParameter;
     }
     else
--- a/libraries/standard/mqtt/utest/mqtt_config.h
+++ b/libraries/standard/mqtt/utest/mqtt_config.h
@@ -22,7 +22,7 @@
 
 /************ End of logging configuration ****************/
 
-/* Set network context to socket (int). */
+/* Set network context to double pointer to buffer (uint8_t**). */
 typedef uint8_t ** MQTTNetworkContext_t;
 
 /**
--- a/libraries/standard/mqtt/utest/mqtt_config.h
+++ b/libraries/standard/mqtt/utest/mqtt_config.h
@@ -16,14 +16,14 @@
 
 /* Configure name and log level for the MQTT library. */
 #define LIBRARY_LOG_NAME     "MQTT"
-#define LIBRARY_LOG_LEVEL    LOG_INFO
+#define LIBRARY_LOG_LEVEL    LOG_NONE
 
 #include "logging_stack.h"
 
 /************ End of logging configuration ****************/
 
 /* Set network context to socket (int). */
-typedef int MQTTNetworkContext_t;
+typedef uint8_t ** MQTTNetworkContext_t;
 
 /**
  * @brief The maximum number of MQTT PUBLISH messages that may be pending
--- a/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
+++ b/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
@@ -513,16 +513,6 @@ void test_MQTT_SerializeConnect( void )
     TEST_ASSERT_GREATER_OR_EQUAL( packetSize, bufferSize );
     status = MQTT_SerializeConnect( &connectInfo, &willInfo, remainingLength, &fixedBuffer );
     TEST_ASSERT_EQUAL( MQTTSuccess, status );
-
-    /* For this example, MQTT_GetConnectPacketSize() will return
-     * packetSize = remainingLength +2 (two byte fixed header).
-     * Make sure MQTT_SerializeConnect()
-     * fails  when remaining length is more than packet size. */
-    /* TODO: Since MQTT_SerializeConnect trusts the remaining length to be
-     * accurate, either change the implementation or remove this test, since it
-     * should fail. */
-    status = MQTT_SerializeConnect( &connectInfo, &willInfo, remainingLength + 4, &fixedBuffer );
-    TEST_ASSERT_EQUAL_INT( MQTTBadParameter, status );
 }
 
 /* ========================================================================== */
@@ -733,20 +723,6 @@ void test_MQTT_SerializeSubscribe( void )
                                       remainingLength,
                                       &fixedBuffer );
     TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
-
-    /* For this example, MQTT_GetSubscriptionPacketSize() will return
-     * packetSize = remainingLength +2 (two byte fixed header).
-     * Make sure MQTT_SerializeSubscribe()
-     * fails  when remaining length is more than packet size. */
-    /* TODO: Since MQTT_SerializeSubscribe trusts the remaining length to be
-     * accurate, either change the implementation or remove this test, since it
-     * should fail. */
-    status = MQTT_SerializeSubscribe( &subscriptionList,
-                                      subscriptionCount,
-                                      PACKET_ID,
-                                      remainingLength + 4,
-                                      &fixedBuffer );
-    TEST_ASSERT_EQUAL_INT( MQTTNoMemory, status );
 }
 
 /**
@@ -826,19 +802,6 @@ void test_MQTT_SerializeUnsubscribe( void )
                                         remainingLength,
                                         &fixedBuffer );
     TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
-
-    /* For this example, MQTT_GetSubscriptionPacketSize() will return
-     * packetSize = remainingLength +2, make sure MQTT_SerializeUnsubscribe()
-     * fails  when remaining length is more than packet size. */
-    /* TODO: Since MQTT_SerializeUnsubscribe trusts the remaining length to be
-     * accurate, either change the implementation or remove this test, since it
-     * should fail. */
-    status = MQTT_SerializeUnsubscribe( &subscriptionList,
-                                        subscriptionCount,
-                                        PACKET_ID,
-                                        remainingLength + 4,
-                                        &fixedBuffer );
-    TEST_ASSERT_EQUAL_INT( MQTTNoMemory, status );
 }
 
 /* ========================================================================== */
--- a/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
+++ b/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
@@ -5,6 +5,25 @@
 /* Include paths for public enums, structures, and macros. */
 #include "mqtt_lightweight.h"
 
+#define MQTT_MAX_REMAINING_LENGTH                   ( 268435455UL )
+#define MQTT_PACKET_CONNACK_REMAINING_LENGTH        ( ( uint8_t ) 2U )    /**< @brief A CONNACK packet always has a "Remaining length" of 2. */
+#define MQTT_PACKET_CONNACK_SESSION_PRESENT_MASK    ( ( uint8_t ) 0x01U ) /**< @brief The "Session Present" bit is always the lowest bit. */
+#define MQTT_PACKET_SIMPLE_ACK_REMAINING_LENGTH     ( ( uint8_t ) 2 ) /**< @brief PUBACK, PUBREC, PUBREl, PUBCOMP, UNSUBACK Remaining length. */
+#define MQTT_PACKET_PINGRESP_REMAINING_LENGTH       ( 0U )            /**< @brief A PINGRESP packet always has a "Remaining length" of 0. */
+#define MQTT_PACKET_PUBACK_REMAINING_LENGTH         ( 2U )
+#define MQTT_PACKET_UNSUBACK_REMAINING_LENGTH       ( 2U )
+/*
+ * Client identifier and length to use for the MQTT API tests.
+ */
+#define CLIENT_IDENTIFIER           ( "test" )                                           /**< @brief Client identifier. */
+#define CLIENT_IDENTIFIER_LENGTH    ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) ) /**< @brief Length of client identifier. */
+/*
+ * Will topic name and length to use for the MQTT API tests.
+ */
+#define TEST_TOPIC_NAME             ( "/test/topic" )                                  /**< @brief An arbitrary topic name. */
+#define TEST_TOPIC_NAME_LENGTH      ( ( uint16_t ) ( sizeof( TEST_TOPIC_NAME ) - 1 ) ) /**< @brief Length of topic name. */
+
+
 /**
  * @brief MQTT protocol version 3.1.1.
  */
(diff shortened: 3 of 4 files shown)


libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,1198 @@
...
+    status = MQTT_SerializePublish( &publishInfo,
+                                    PACKET_ID,
+                                    remainingLength,
+                                    &fixedBuffer );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    mqttPacketInfo.type = buffer[ 0 ];
+    mqttPacketInfo.remainingLength = ( size_t ) buffer[ 1 ];
+    mqttPacketInfo.pRemainingData = &buffer[ 2 ];
+    status = MQTT_DeserializePublish( &mqttPacketInfo, &packetIdentifier, &publishInfo );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
```
muneebahmed10, line 1115: Added test of QoS publish

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,1198 @@
...
+
+    /* Remaining length of 128 needs 2 bytes. */
+    bufPtr = buffer;
+    buffer[ 0 ] = MQTT_PACKET_TYPE_PUBLISH;
+    buffer[ 1 ] = 0x80;
+    buffer[ 2 ] = 0x01;
+    status = MQTT_GetIncomingPacketTypeAndLength( mockReceive, networkContext, &mqttPacket );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    TEST_ASSERT_EQUAL_INT( MQTT_PACKET_TYPE_PUBLISH, mqttPacket.type );
+    TEST_ASSERT_EQUAL_INT( 128, mqttPacket.remainingLength );
```
muneebahmed10, line 1146: Added this test for branch coverage

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,1198 @@
...
+    bufPtr = buffer;
+    buffer[ 0 ] = MQTT_PACKET_TYPE_PUBREL & 0xF0U;
+    status = MQTT_GetIncomingPacketTypeAndLength( mockReceive, networkContext, &mqttPacket );
+    TEST_ASSERT_EQUAL( MQTTBadResponse, status );
+
+    /* Receive type then fail. */
+    bufPtr = buffer;
+    buffer[ 0 ] = MQTT_PACKET_TYPE_PUBREL;
+    status = MQTT_GetIncomingPacketTypeAndLength( mockReceiveSucceedThenFail, networkContext, &mqttPacket );
+    TEST_ASSERT_EQUAL( MQTTBadResponse, status );
```
muneebahmed10, line 1197: Added these tests due to differences between the old network interface and the new transport receive function, as well as branch coverage.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -0,0 +1,1198 @@
...
+
+/* Called at the end of the whole suite. */
+int suiteTearDown( int numFailures )
+{
+    return numFailures;
+}
+
+/* ========================================================================== */
+
+static int32_t mockReceive( MQTTNetworkContext_t context,
```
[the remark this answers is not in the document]
muneebahmed10: I don't believe we can, since the receive function is only a function pointer, and not defined in an external header file.

libraries/standard/mqtt/utest/mqtt_config.h
```diff
@@ -16,14 +16,14 @@
...
-#define LIBRARY_LOG_LEVEL    LOG_INFO
+#define LIBRARY_LOG_LEVEL    LOG_NONE
 
 #include "logging_stack.h"
 
 /************ End of logging configuration ****************/
 
 /* Set network context to socket (int). */
-typedef int MQTTNetworkContext_t;
+typedef uint8_t ** MQTTNetworkContext_t;
```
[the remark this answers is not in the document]
muneebahmed10: Updated

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    willInfo.topicNameLength = 4;
+    willInfo.pPayload = "testload";
+    willInfo.payloadLength = 8;
+    status = MQTT_GetConnectPacketSize( &connectInfo, &willInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    /* Make sure remaining size returned is 32 = 16 + 16. */
+    TEST_ASSERT_EQUAL_INT( 32, remainingLength );
+    /* Make sure packet size is 34 = 18 + 16. */
+    TEST_ASSERT_EQUAL_INT( 34, packetSize );
+}
```
leegeth, line 391: I see that the paths for pUserName != NULL and pPassword != NULL and connectPacketSize > MQTT_PACKET_CONNECT_MAX_SIZE are covered in the unit tests for MQTT_SerializeConnect. Considering MQTT_GetConnectPacketSize as a unit, I think those cases belong here.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    connectInfo.pClientIdentifier = CLIENT_IDENTIFIER;
+    connectInfo.clientIdentifierLength = UINT16_MAX;
+    connectInfo.pPassword = "";
+    connectInfo.passwordLength = UINT16_MAX;
+    connectInfo.pUserName = "";
+    connectInfo.userNameLength = UINT16_MAX;
+    willInfo.pTopicName = TEST_TOPIC_NAME;
+    willInfo.topicNameLength = UINT16_MAX;
+    willInfo.payloadLength = UINT16_MAX + 2;
+    status = MQTT_GetConnectPacketSize( &connectInfo, &willInfo, &remainingLength, &packetSize );
```
leegeth, line 428: This is unit tests for MQTT_GetConnectPacketSize as mentioned in the earlier comment.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    memset( ( void * ) &connectInfo, 0x0, sizeof( connectInfo ) );
+    connectInfo.pClientIdentifier = "TEST";
+    connectInfo.clientIdentifierLength = 4;
+    status = MQTT_GetConnectPacketSize( &connectInfo, NULL, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    /* Make sure buffer has enough space */
+    TEST_ASSERT_GREATER_OR_EQUAL( packetSize, bufferSize );
+    /* Make sure test succeeds. */
+    status = MQTT_SerializeConnect( &connectInfo, NULL, remainingLength, &fixedBuffer );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
```
leegeth, line 442: It will be a good addition to verify the serialized buffer as well here. That would be adding value to tests as we will be verifying the if the serialization(The main output of the function MQTT_SerializeConnect) is as expected. The same comment applies to all the success cases of serialization APIs in this file.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    connectInfo.userNameLength = 4;
+    connectInfo.pPassword = "PASS";
+    connectInfo.passwordLength = 4;
+    status = MQTT_GetConnectPacketSize( &connectInfo, NULL, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL( MQTTSuccess, status );
+    TEST_ASSERT_GREATER_OR_EQUAL( packetSize, bufferSize );
+    status = MQTT_SerializeConnect( &connectInfo, NULL, remainingLength, &fixedBuffer );
+    TEST_ASSERT_EQUAL( MQTTSuccess, status );
+
+    /* Serialize connect with LWT. */
```
leegeth, line 456: please add a description here about which parameter is bad so that it will help readability of the code.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    willInfo.qos = MQTTQoS1;
+    willInfo.pTopicName = "test";
+    willInfo.topicNameLength = ( uint16_t ) 4;
+    status = MQTT_GetConnectPacketSize( &connectInfo, &willInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL( MQTTSuccess, status );
+    TEST_ASSERT_GREATER_OR_EQUAL( packetSize, bufferSize );
+    status = MQTT_SerializeConnect( &connectInfo, &willInfo, remainingLength, &fixedBuffer );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, status );
+
+    ( void ) memset( &willInfo, 0x00, sizeof( MQTTPublishInfo_t ) );
```
leegeth, line 468: please add a small description in comments for this test.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    willInfo.qos = MQTTQoS1;
+    willInfo.pPayload = "test";
+    willInfo.payloadLength = ( uint16_t ) strlen( willInfo.pPayload );
+    status = MQTT_GetConnectPacketSize( &connectInfo, &willInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL( MQTTSuccess, status );
+    TEST_ASSERT_GREATER_OR_EQUAL( packetSize, bufferSize );
+    status = MQTT_SerializeConnect( &connectInfo, &willInfo, remainingLength, &fixedBuffer );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, status );
+
+    ( void ) memset( &willInfo, 0x00, sizeof( MQTTPublishInfo_t ) );
```
leegeth, line 479: same comment about a small description in comments.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    status = MQTT_SerializeConnect( &connectInfo, &willInfo, remainingLength, &fixedBuffer );
+    TEST_ASSERT_EQUAL( MQTTSuccess, status );
+
+    /* For this example, MQTT_GetConnectPacketSize() will return
+     * packetSize = remainingLength +2 (two byte fixed header).
+     * Make sure MQTT_SerializeConnect()
+     * fails  when remaining length is more than packet size. */
+    /* TODO: Since MQTT_SerializeConnect trusts the remaining length to be
+     * accurate, either change the implementation or remove this test, since it
+     * should fail. */
```
leegeth, line 515: What is the plan to address this? I think fixing it in the library is a good solution.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+ * @brief Tests that MQTT_GetSubscribePacketSize works as intended.
+ */
+void test_MQTT_GetSubscribePacketSize( void )
+{
+    MQTTSubscribeInfo_t subscriptionList;
+    size_t subscriptionCount = 0;
+    size_t remainingLength = 0;
+    size_t packetSize = 0;
+    MQTTStatus_t status = MQTTSuccess;
+    MQTTSubscribeInfo_t fourThousandSubscriptions[ 4096 ] = { 0 };
```
leegeth, line 532: This test will increase the required stack size by at least 4KB * sizeof(MQTTSubscribeInfo_t ).  As long as unit tests are not targeted to run on targets that are memory constrained, this should be ok.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Agree, although I think that needs to be part of a separate PR (since this PR deals mainly with importing existing tests and verifying the buffer would require adding some new code)

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+                                      remainingLength,
+                                      &fixedBuffer );
+    TEST_ASSERT_EQUAL_INT( MQTTBadParameter, status );
+
+    status = MQTT_SerializeSubscribe( &subscriptionList,
+                                      subscriptionCount,
+                                      PACKET_ID,
+                                      remainingLength,
+                                      NULL );
+    TEST_ASSERT_EQUAL_INT( MQTTBadParameter, status );
```
leegeth, line 687: Don't we have 2 more paths to cover for invalid params? subscriptionCount == 0U and packetSize > pBuffer->size

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    status = MQTT_GetPublishPacketSize( &publishInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, status );
+
+    /* Good case succeeds. */
+    publishInfo.pTopicName = "/test/topic";
+    publishInfo.topicNameLength = sizeof( "/test/topic" );
+    publishInfo.pPayload = "";
+    publishInfo.payloadLength = 0;
+    status = MQTT_GetPublishPacketSize( &publishInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL( MQTTSuccess, status );
```
leegeth, line 868: Don't we have another branch to cover here "if( pPublishInfo->qos > MQTTQoS0 )" for calculating the packet size?

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: subscriptionCount == 0 is handled in line 704. packetSize > pBuffer->size is handled in line 726, though I think that's only because the buffer is too small. We can make that a separate test where we more explicitly set the buffer size to a low value

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: That's already handled when testing MQTT_SerializePublish, though I can copy that code over here too if that's desirable

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    status = MQTT_GetPublishPacketSize( &publishInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    /* Make sure buffer has enough space */
+    TEST_ASSERT_GREATER_OR_EQUAL( packetSize, bufferSize );
+    status = MQTT_SerializePublish( &publishInfo,
+                                    PACKET_ID,
+                                    remainingLength,
+                                    &fixedBuffer );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+}
```
leegeth, line 974: How about success cases for QoS1 and QoS0?

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1154 @@ static size_t encodeString( uint8_t * pDestination,
...
+    publishInfo.pPayload = "";
+    publishInfo.payloadLength = 0;
+    status = MQTT_GetPublishPacketSize( &publishInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL( MQTTSuccess, status );
+}
+
+/**
+ * @brief Tests that MQTT_SerializePublish works as intended.
+ */
+void test_MQTT_SerializePublish( void )
```
leegeth, line 874: There is a new function to serialize only header of publish packet MQTT_SerializePublishHeader. That function also need to be tested. Since this is import of the existing tests from v4_beta, I think it can be considered in a separate PR.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Good point, I'll rearrange this so both QoS 0 and 2 are used. I don't think success case for QoS 1 is necessary since it's treated identically to QoS 2

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
leegeth: QoS2

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
leegeth: Sounds good to me.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
leegeth: Yes. I think that test belong here.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Good catch, it seems this was misplaced in v4_beta and copied over to here. I'll move the test and delete it here

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Removed test since NULL payload should be allowed

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Added

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Added

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: I think removing these will be best, I left these here so other reviewers could chime in. Will remove these after another round of review

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -1267,9 +1267,12 @@ static void serializeConnectPacket( const MQTTConnectInfo_t * const pConnectInfo
         pIndex = encodeString( pIndex,
                                pWillInfo->pTopicName,
                                pWillInfo->topicNameLength );
-        pIndex = encodeString( pIndex,
-                               pWillInfo->pPayload,
-                               ( uint16_t ) pWillInfo->payloadLength );
+        if( pWillInfo->pPayload != NULL )
```
leegeth, line 1270: please add a comment here why we have to add this check instead of a parameter  validation.

muneebahmed10: Now that I think about it, this is incorrect since the length of the payload needs to be encoded in the next two bytes. Moved the check to inside `encodeString`, before the `memcpy` is called.

leegeth: Yes. That is true. Length of 0 has to be added. We just need to skip the memcpy.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Handled in #968

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Verification of serialized buffers added in #968

leegeth: Looks good to me.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -235,6 +323,1151 @@ static size_t encodeString( uint8_t * pDestination,
...
+
+/**
+ * @brief Tests that MQTT_SerializeConnect works as intended.
+ */
+void test_MQTT_SerializeConnect( void )
+{
+    MQTTConnectInfo_t connectInfo;
+    MQTTPublishInfo_t willInfo;
+    size_t remainingLength = 0;
+    uint8_t buffer[ 70 ];
```
gkwicker, line 429: One thing I like to do in unit tests like this is to add extra space at the beginning and end of any buffer which will be written by the function under test, and fill those spaces with special values.  Then you execute the function, giving it an address just after your special values and verify afterwards that your special values are still in place.  This is particularly important when the function writing the buffer depend on some passed in length value which could be bogus (e.g. -1).

I usually create some utility functions to do this and use them throughout whenever a buffer is getting formatted.

muneebahmed10: Added

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -323,6 +333,45 @@ static size_t encodeString( uint8_t * pDestination,
...
+ * @param[in] bufferLength Total length of buffer.
+ */
+static void padAndResetBuffer( uint8_t * pBuffer, size_t bufferLength )
+{
+    int i = 0;
+
+    for( i = 0; i < BUFFER_PADDING_LENGTH; i++ )
+    {
+        pBuffer[ i ] = BUFFER_PADDING_BYTE;
+        pBuffer[ bufferLength - 1 - i ] = BUFFER_PADDING_BYTE;
```
gkwicker, line 350: nice

PR #967, opened 2020-05-27
Ci/coverity integration

nateglims:
*Description of changes:* Changes to enable static analysis in CI:

1) Add MISRA config for coverity from `v4_beta` branch. This provides a set of MISRA rules that are unconditionally ignored.
2) Explicitly link `pthreads` and `CMAKE_DL_LIBS`. These are used by SSL and on some environments (such as our build nodes) CMake has trouble reasoning about the existence of these libraries or if they are required. The `CMAKE_THREAD_PREFER_PTHREAD` setting and `${CMAKE_DL_LIBS}` appear to be best practices.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,9 @@ set_property( GLOBAL PROPERTY USE_FOLDERS ON )
 # Use C90.
 set( CMAKE_C_STANDARD 90 )
 set( CMAKE_C_STANDARD_REQUIRED ON )
+set( CMAKE_THREAD_PREFER_PTHREAD ON )
+
+find_package(Threads REQUIRED)
 
 # Do not allow in-source build.
 if( ${PROJECT_SOURCE_DIR} STREQUAL ${PROJECT_BINARY_DIR} )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,9 +10,6 @@ set_property( GLOBAL PROPERTY USE_FOLDERS ON )
 # Use C90.
 set( CMAKE_C_STANDARD 90 )
 set( CMAKE_C_STANDARD_REQUIRED ON )
-set( CMAKE_THREAD_PREFER_PTHREAD ON )
-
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ set_property( GLOBAL PROPERTY USE_FOLDERS ON )
 # Use C90.
 set( CMAKE_C_STANDARD 90 )
 set( CMAKE_C_STANDARD_REQUIRED ON )
+set( CMAKE_THREAD_PREFER_PTHREAD ON )
+
+find_package(Threads REQUIRED)
```
yourslab, line 15: Does it still work if we move this to `demos/mqtt/mqtt_demo_basic_tls/CMakeLists.txt` or `demos/CMakeLists.txt` (preferably the first).

demos/mqtt/mqtt_demo_basic_tls/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ target_link_libraries(
         mqtt
         OpenSSL::Crypto
         OpenSSL::SSL
+        Threads::Threads
```
leegeth, line 20: can you please add some comments here to indicate why Threads need to be linked in here even though we are not directly using Threads in the TLS demo?

nateglims: Sure, but so you're aware CMake will pull in pthreads when linking OpenSSL either way. I just had to make it explicit for the build environment.

CMakeLists.txt
nateglims: Appears so.

demos/mqtt/mqtt_demo_basic_tls/CMakeLists.txt
nateglims: Added for Thread and the Dynamic Loading library.

PR #968, opened 2020-05-27
Increase coverage of lightweight unit tests

muneebahmed10:
*Description of changes:*
Adds unit tests for `MQTT_SerializePublishHeader`, `MQTT_SerializeAck`, `MQTT_GetDisconnectPacketSize`, and `MQTT_GetPingreqPacketSize`. Branch coverage of mqtt_lightweight.c is 100%.

This should not be merged until #957 is.

### DIFF
--- a/libraries/standard/mqtt/include/mqtt_lightweight.h
+++ b/libraries/standard/mqtt/include/mqtt_lightweight.h
@@ -478,6 +478,10 @@ MQTTStatus_t MQTT_GetPingreqPacketSize( size_t * pPacketSize );
  */
 MQTTStatus_t MQTT_SerializePingreq( const MQTTFixedBuffer_t * const pBuffer );
 
+MQTTStatus_t MQTT_GetIncomingPacket( MQTTTransportRecvFunc_t recvFunc,
+                                     MQTTNetworkContext_t networkContext,
+                                     MQTTPacketInfo_t * const pIncomingPacket );
+
 /**
  * @brief Deserialize an MQTT PUBLISH packet.
  *
--- a/libraries/standard/mqtt/include/mqtt_lightweight.h
+++ b/libraries/standard/mqtt/include/mqtt_lightweight.h
@@ -478,10 +478,6 @@ MQTTStatus_t MQTT_GetPingreqPacketSize( size_t * pPacketSize );
  */
 MQTTStatus_t MQTT_SerializePingreq( const MQTTFixedBuffer_t * const pBuffer );
 
-MQTTStatus_t MQTT_GetIncomingPacket( MQTTTransportRecvFunc_t recvFunc,
-                                     MQTTNetworkContext_t networkContext,
-                                     MQTTPacketInfo_t * const pIncomingPacket );
-
 /**
  * @brief Deserialize an MQTT PUBLISH packet.
  *
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -712,42 +712,24 @@ static MQTTStatus_t processPublishFlags( uint8_t publishFlags,
 
 static void logConnackResponse( uint8_t responseCode )
 {
+    const char * const pConnackResponses[ 6 ] =
+    {
+        "Connection accepted.",                               /* 0 */
+        "Connection refused: unacceptable protocol version.", /* 1 */
+        "Connection refused: identifier rejected.",           /* 2 */
+        "Connection refused: server unavailable",             /* 3 */
+        "Connection refused: bad user name or password.",     /* 4 */
+        "Connection refused: not authorized."                 /* 5 */
+    };
+
     /* Avoid unused parameter warning when assert and logs are disabled. */
     ( void ) responseCode;
+    ( void ) pConnackResponses;
 
     assert( responseCode <= 5 );
 
     /* Log an error based on the CONNACK response code. */
-    switch( responseCode )
-    {
-        case 0u:
-            LogInfo( ( "Connection accepted." ) );
-            break;
-
-        case 1u:
-            LogInfo( ( "Connection refused: unacceptable protocol version." ) );
-            break;
-
-        case 2u:
-            LogInfo( ( "Connection refused: identifier rejected." ) );
-            break;
-
-        case 3u:
-            LogInfo( ( "Connection refused: server unavailable" ) );
-            break;
-
-        case 4u:
-            LogInfo( ( "Connection refused: bad user name or password." ) );
-            break;
-
-        case 5u:
-            LogInfo( ( "Connection refused: not authorized." ) );
-            break;
-
-        default:
-            /* Empty default MISRA 16.4. */
-            break;
-    }
+    LogError( ( "%s", pConnackResponses[ responseCode ] ) );
 }
 
 /*-----------------------------------------------------------*/
@@ -1803,6 +1785,11 @@ MQTTStatus_t MQTT_SerializeAck( const MQTTFixedBuffer_t * const pBuffer,
         LogError( ( "Insufficient memory for packet." ) );
         status = MQTTNoMemory;
     }
+    else if( packetId == 0U )
+    {
+        LogError( ( "Packet ID cannot be 0." ) );
+        status = MQTTBadParameter;
+    }
     else
     {
         switch( packetType )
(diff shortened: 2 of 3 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -712,42 +712,24 @@ static MQTTStatus_t processPublishFlags( uint8_t publishFlags,
...
 {
+    const char * const pConnackResponses[ 6 ] =
+    {
+        "Connection accepted.",                               /* 0 */
+        "Connection refused: unacceptable protocol version.", /* 1 */
+        "Connection refused: identifier rejected.",           /* 2 */
+        "Connection refused: server unavailable",             /* 3 */
+        "Connection refused: bad user name or password.",     /* 4 */
+        "Connection refused: not authorized."                 /* 5 */
+    };
```
muneebahmed10, line 723: Originally this function was like this, but in a macro guard so that is would only be allocated if logging was enabled. #937 converted this into a switch statement, since that macro is no longer defined anywhere. This reverts it back, since part of that switch was unreachable, and I'm going to trust the compiler to not allocate memory for unused variables.

libraries/standard/mqtt/include/mqtt_lightweight.h
```diff
@@ -478,10 +478,6 @@ MQTTStatus_t MQTT_GetPingreqPacketSize( size_t * pPacketSize );
  */
 MQTTStatus_t MQTT_SerializePingreq( const MQTTFixedBuffer_t * const pBuffer );
 
-MQTTStatus_t MQTT_GetIncomingPacket( MQTTTransportRecvFunc_t recvFunc,
```
leegeth, line 481: Why is this removed. I think we are planning to implement this function.

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -1929,15 +1926,6 @@ MQTTStatus_t MQTT_SerializePingreq( const MQTTFixedBuffer_t * const pBuffer )
...
 
-MQTTStatus_t MQTT_GetIncomingPacket( MQTTTransportRecvFunc_t recvFunc,
-                                     MQTTNetworkContext_t networkContext,
-                                     MQTTPacketInfo_t * const pIncomingPacket )
-{
-    return MQTTSuccess;
-}
-
-/*-----------------------------------------------------------*/
-
```
leegeth, line 1927: same comment as above.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -1570,6 +1606,67 @@ void test_MQTT_SerializePublishHeader( void )
...
+    expectedPacket[ 0 ] = packetType;
+    expectedPacket[ 1 ] = 2U;
+    expectedPacket[ 2 ] = PACKET_ID >> 8;
+    expectedPacket[ 3 ] = PACKET_ID & UINT8_MAX;
+
+    /* Verify parameters. */
+    status = MQTT_SerializeAck( NULL, packetType, PACKET_ID );
+    TEST_ASSERT_EQUAL_INT( MQTTBadParameter, status );
+
+    status = MQTT_SerializeAck( &fixedBuffer, packetType, 0 );
```
leegeth, line 1684: missed validation?

libraries/standard/mqtt/include/mqtt_lightweight.h
muneebahmed10: No, it's not in our API doc, and we wrote MQTT_GetIncomingPacketTypeAndLength instead. I believe we intended to remove it before, but never got around to it.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Are you referring to the reason for the expected failure? I can add in a comment that packet ID 0 is not allowed.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -968,6 +1009,15 @@ void test_MQTT_SerializePublish( void )
...
     TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    memset( ( void * ) expectedPacket, 0x00, sizeof( expectedPacket ) );
+    pIterator = expectedPacket;
+    /* Dup = 0x8, QoS2 = 0x4, Retain = 0x1. 8 + 4 + 1 = 0xD. */
+    *pIterator++ = MQTT_PACKET_TYPE_PUBLISH | 0xD;
+    pIterator += encodeRemainingLength( pIterator, remainingLength );
+    pIterator += encodeString( pIterator, publishInfo.pTopicName, publishInfo.topicNameLength );
+    *pIterator++ = UINT16_HIGH_BYTE( PACKET_ID );
+    *pIterator++ = UINT16_LOW_BYTE( PACKET_ID );
+    TEST_ASSERT_EQUAL_MEMORY( expectedPacket, buffer, packetSize );
```
leegeth, line 1020: I think it would add value if we can add the buffer validation as part of all the Success cases. Line 976. What are your thoughts?

libraries/standard/mqtt/include/mqtt_lightweight.h
leegeth: I think the plan is to still keep it and implement it. It is an additional API to provide the customer the choice of avoiding multiple transport reads. It came up in the design proposal.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
leegeth: Did you miss yo do this step here? TEST_ASSERT_EQUAL_INT( MQTTBadParameter, status );

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: I'm not sure I understand. Nothing is written into the buffer by line 976, that's just making sure the calculated packet size does not exceed the buffer size

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Oh, I didn't see that. Yes

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
leegeth: Sorry, wrong line number, I meant 982

muneebahmed10: That buffer is verified in line 993. Everything between 982 and 993 just generates the expected packet. Should it be rearranged so the buffer verification is closer to when the status is tested?

leegeth: Yea. Ok. I see the problem. I missed to notice the assert in 993 and was thinking that it is setting the expected buffer for the next test. I think it is ok as it is.

libraries/standard/mqtt/include/mqtt_lightweight.h
muneebahmed10: After some offline discussion it does seem like that is the plan, so I will re add this function.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -1542,6 +1632,178 @@ void test_MQTT_GetIncomingPacketTypeAndLength( void )
...
+
+    /* Success case. */
+    status = MQTT_GetPublishPacketSize( &publishInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    status = MQTT_SerializePublishHeader( &publishInfo,
+                                          PACKET_ID,
+                                          remainingLength,
+                                          &fixedBuffer,
+                                          &headerSize );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
```
leegeth, line 1727: missed to verify the serialized buffer?

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -1542,6 +1632,178 @@ void test_MQTT_GetIncomingPacketTypeAndLength( void )
...
+    publishInfo.pPayload = "test";
+    publishInfo.payloadLength = 4;
+    status = MQTT_GetPublishPacketSize( &publishInfo, &remainingLength, &packetSize );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    status = MQTT_SerializePublishHeader( &publishInfo,
+                                          0,
+                                          remainingLength,
+                                          &fixedBuffer,
+                                          &headerSize );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
```
leegeth, line 1739: missed to verify the serialized buffer?

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Yes, added

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: Added

PR #969, opened 2020-05-27
CBMC infrastructure for HTTPv2

markrtuttle:
This pull request adds CBMC infrastructure to the HTTPv2 project.


libraries/standard/http/cbmc/proofs/demo/Makefile
```diff
@@ -0,0 +1,82 @@
+METHOD=IotMqtt_Disconnect
+ENTRY=harness
+
+################################################################
+# Proof assumptions
+
+# Bound the number of subscriptions in a list (BOUND = MAX-1)
+SUBSCRIPTION_COUNT_MAX=2
+DEF += -DSUBSCRIPTION_COUNT_MAX=$(SUBSCRIPTION_COUNT_MAX)
```
danielsn, line 9: I wonder if there is a nice macro to do this?

libraries/standard/http/cbmc/proofs/demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+# allow mqtt mallocs to fail for coverage
+DEF += -include mqtt_state.h
+DEF += -DIotMqtt_MallocMessage=malloc_can_fail
+DEF += -DIotMqtt_MallocOperation=malloc_can_fail
+
+################################################################
+
+SRCDIR ?= $(abspath $(PROOFDIR)/../../..)
+
+INC += \
```
danielsn, line 20: Is this different for each proof? Or can it go in the base makefile?

libraries/standard/http/cbmc/proofs/demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+PROOF_SOURCES += $(SRCDIR)/cbmc/stubs/IotListDouble_RemoveAllMatches.c
+PROOF_SOURCES += $(SRCDIR)/cbmc/stubs/IotSemaphore.c
+
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/iot_mqtt_api.c
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/iot_mqtt_helper.c
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/iot_mqtt_operation.c
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/iot_mqtt_network.c
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/iot_mqtt_serialize.c
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/iot_mqtt_subscription.c
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/iot_mqtt_validate.c
```
danielsn, line 78: Not critical, but I like to sort the lines alphabetically e.g. unwindset goes after proof_sources.

libraries/standard/http/cbmc/proofs/demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+
+DEFINES = $(DEF)
```
danielsn, line 80: why is this needed?

libraries/standard/http/cbmc/proofs/demo/Makefile
markrtuttle: Do you mean macro that would take SUB_CNT and 2 and produce the 2 lines?  I'm inclined to leave this as it is, I think macros can obfuscate, and I prefer to leave the macros for things that must be right (eg, how to invoke cbmc) and not for stuff like this that most programmers see every day.

libraries/standard/http/cbmc/proofs/demo/Makefile
markrtuttle: In proof hierarchies that have proofs at differing depths in the proof hierarchy, in this case (and in FreeRTOS), each proof is going to have to define SRCDIR rising up differing amounts to the source root.

libraries/standard/http/cbmc/proofs/demo/Makefile
markrtuttle: I will construct a real demo and not just a sketch.

libraries/standard/http/cbmc/proofs/demo/Makefile
markrtuttle: I like INC and DEF and you like INC and DEFINES.  I'd like to agree on INC and DEF or INCLUDES and DEFINES.  I had already used DEF in the MQTT makefile, and just renamed it.  But I will construct a real demo.

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+DO_GOTO_CC =  $(call DO_AND_LOG_COMMAND,$(GOTO_CC),$(CBMC_VERBOSITY) $(1) $(2) -o $(3), $(call LOG_FROM_ENTRY,$(3)))
+
+#1: flags, 2: source, 3: target
+DO_GOTO_INSTRUMENT = $(call DO_AND_LOG_COMMAND,$(GOTO_INSTRUMENT),$(CBMC_VERBOSITY) $(1) $(2) $(3), $(call LOG_FROM_ENTRY,$(3)))
+
+#1: flags, 2: source, 3: logfile
+DO_CBMC =  $(call DO_AND_LOG_IGNORING_ERROR_10,$(CBMC),$(CBMC_VERBOSITY) --flush $(1) $(2), $(3))
+
+#1: message 2: source 3: dest
+DO_NOOP_COPY = cp $(2) $(3); echo $(1) | tee $(call LOG_FROM_ENTRY,$(3))
```
karkhaz, line 238: If cp fails (for example because the source file or destination directory doesn't exist), this command will still succeed. Up to you whether you think this is worth fixing. Something like this ought to work (haven't tested):
```sh
cp $(2) $(3); export SUCCESS=$$?; echo $(1) | tee $(call LOG_FROM_ENTRY,$(3)); exit $${SUCCESS}
```

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+PROOF_SOURCES ?=
+
+################################################################
+################################################################
+## Section II: This section is for project-specific definitions
+
+
+################################################################
+################################################################
+## Section II: This section defines the process of running a proof
```
karkhaz, line 183: This should be Section III. Better yet, there's no need to number the sections.

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+CBMC ?= cbmc
+GOTO_ANALYZER ?= goto-analyzer
+GOTO_CC ?= goto-cc
+GOTO_INSTRUMENT ?= goto-instrument
+VIEWER ?= cbmc-viewer
+MAKE_SOURCES ?= make-sources
+VIEWER2 ?= viewer
+
+GOTODIR ?= $(PROOFDIR)/gotos
+LOGDIR ?= $(PROOFDIR)/logs
```
karkhaz, line 199: This seems to be putting all output files into subdirectories of where the proof is written, thus polluting the source tree.

You could build entirely out of tree with a small modification. Something like:
```sh
ROOTDIR ?= /tmp/mqtt-proofs/$(notdir $(PROOFDIR))
GOTODIR ?= $(ROOTDIR)/gotos
LOGDIR ?= $(ROOTDIR)/logs
```
You could then make this even fancier by generating a new temporary directory for each run. If you know that this will only be run on Unix, something like
```sh
ROOTDIR ?= $(shell mktemp -d -t mqtt-proof-XXXXXX)/$(notdir $(PROOFDIR))
```
This would obviate the need to `make clean` every time. You would then want to `echo Files will be written to $(PROOFDIR)` in whichever recipe runs first.

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+
+report2: logs/sources.json logs/property.xml logs/property-list.xml logs/coverage.xml
+	$(VIEWER2) \
+	--make-sources logs/sources.json \
+	--result logs/property.xml \
+	--coverage logs/coverage.xml \
+	--property logs/property-list.xml \
+	--srcdir $(SRCDIR) \
+	--goto $(ENTRY_GOTO).goto \
+	--htmldir html \
```
karkhaz, line 357: `--htmldir $(PROOFDIR)/html`

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+
+property-list: logs/property-list.xml
+
+coverage: logs/coverage.xml
+
+report: logs/property.txt logs/property-list.xml logs/coverage.xml
+	$(VIEWER) \
+	--block logs/coverage.xml \
+	--goto $(ENTRY_GOTO).goto \
+	--htmldir html \
```
karkhaz, line 339: `--htmldir $(PROOFDIR)/html`

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+	--property logs/property-list.xml \
```
karkhaz, line 340: `--property $(LOGDIR)/property-list.xml`

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+	--result logs/property.txt \
```
karkhaz, line 341: `--result $(LOGDIR)/property.txt`

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+
+property: logs/property.txt
+
+property-list: logs/property-list.xml
+
+coverage: logs/coverage.xml
+
+report: logs/property.txt logs/property-list.xml logs/coverage.xml
+	$(VIEWER) \
+	--block logs/coverage.xml \
```
karkhaz, line 337: `--block $(LOGDIR)/coverage.xml`

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+	$(call DO_CBMC,$(CBMCFLAGS) $(CHECKFLAGS) --unwinding-assertions --show-properties --xml-ui,$<,$@)
+
+logs/coverage.xml: $(ENTRY_GOTO).goto
+	$(call DO_CBMC,$(CBMCFLAGS) $(COVERFLAGS) --cover location --xml-ui,$<,$@)
+
+property: logs/property.txt
+
+property-list: logs/property-list.xml
+
+coverage: logs/coverage.xml
```
karkhaz, line 333: s/`logs`/`$(LOGDIR)`/

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+logs/property.txt: $(ENTRY_GOTO).goto
+	$(call DO_CBMC,$(CBMCFLAGS) $(CHECKFLAGS) --trace,$<,$@)
+
+logs/property.xml: $(ENTRY_GOTO).goto
+	$(call DO_CBMC,$(CBMCFLAGS) $(CHECKFLAGS) --unwinding-assertions --trace --xml-ui,$<,$@)
+
+logs/property-list.xml: $(ENTRY_GOTO).goto
+	$(call DO_CBMC,$(CBMCFLAGS) $(CHECKFLAGS) --unwinding-assertions --show-properties --xml-ui,$<,$@)
+
+logs/coverage.xml: $(ENTRY_GOTO).goto
```
karkhaz, line 326: s/`logs`/`$(LOGDIR)`

libraries/standard/http/cbmc/proofs/Makefile.common
```diff
@@ -0,0 +1,375 @@
...
+	$(RM) *.goto
+	$(RM) $(DEPENDENT_GOTOS)
+	$(RM) *.log
+	$(RM) TAGS*
+	$(RM) *~ \#*
+
+veryclean: clean
+	$(RM) -r html
+	$(RM) -r logs
+	$(RM) -r gotos
```
karkhaz, line 371: `$(PROOFDIR)/html`, `$(LOGDIR)`, `$(GOTODIR)`. Speaking of which, you may wish to define `$(REPORTDIR)`.

libraries/standard/http/cbmc/proofs/demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+
+################################################################
+# Proof assumptions
+
+# Bound the number of subscriptions in a list (BOUND = MAX-1)
+SUBSCRIPTION_COUNT_MAX=2
+DEF += -DSUBSCRIPTION_COUNT_MAX=$(SUBSCRIPTION_COUNT_MAX)
+
+# allow mqtt mallocs to fail for coverage
+DEF += -include mqtt_state.h
```
karkhaz, line 12: Why is this not under `INC`?

libraries/standard/http/cbmc/proofs/demo/Makefile
karkhaz, line 20: Is there a reason that you're using the `+=` on each line for `DEF`, but using a single `+=` with line continuations for `INC`?

If there's not a good reason, I personally prefer the `DEF` style, line continuations are annoying and easy to mess up.

libraries/standard/http/cbmc/proofs/demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+	-I$(SRCDIR)/libraries/standard/mqtt/include \
+	-I$(SRCDIR)/libraries/standard/mqtt/src \
+	-I$(SRCDIR)/libraries/standard/mqtt/src/private \
+	-I$(SRCDIR)/ports/common/include \
+
+DEF += \
+        -DIOT_SYSTEM_TYPES_FILE=\"$(SRCDIR)/ports/posix/include/iot_platform_types_posix.h\" \
+        -Diotmqtt_EXPORTS \
+
+COMPILE_FLAGS = -std=gnu99
```
karkhaz, line 35: In other projects, we use `+=` for everything, even for things that are defined for the first time. This makes it easy to add additional compile flags when invoking make.

libraries/standard/http/cbmc/proofs/demo/Makefile
```diff
@@ -0,0 +1,82 @@
...
+LINK_FLAGS = -std=gnu99
+
+NONDET_STATIC=--nondet-static
```
karkhaz, line 38: If this is a best practices demo, I would prefer a comment to indicate that this is optional. You've already written a comment about NONDET_STATIC in Makefile.common, but I would actually prefer to see that comment here, since people are more likely to read it. Also make it clear that it should be set to empty, rather than unset, to switch it off.

sarenameas:
I think review can be closed?

markrtuttle:
Closing this pull request against aws/aws-iot-device-sdk-embedded-C because the changes debugged and discussed here have been merged directly in to awslabs/aws-templates-for-cbmc-proofs.

PR #971, opened 2020-05-28
Server auth demo

leegeth:
*Description of changes:*
Demo for showing the use of MQTT APIs to establish an MQTT session,
subscribe to a topic, publish to a topic, receive incoming publishes,
unsubscribe from a topic and disconnect the MQTT session.
This demo authorizes the MQTT broker it is connecting to.


demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -184,18 +431,40 @@ static SSL * tlsSetup( int tcpSocket )
...
         if( pRootCaFile != NULL )
         {
             pRootCa = PEM_read_X509( pRootCaFile, NULL, NULL, NULL );
         }
+        else
+        {
+            LogError( ( "Unable to find the certificate file in the path"
+                        " provided by SERVER_CERT_PATH(%.*s).",
+                        SERVER_CERT_PATH_LEN,
+                        SERVER_CERT_PATH ) );
```
yourslab, line 445: Is it preferable to use %.*s over %s?

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -70,34 +97,248 @@
...
+static int32_t transportSend( MQTTNetworkContext_t pSslContext,
+                              const void * pMessage,
+                              size_t bytesToSend );
+
+/**
+ * @brief The transport receive function provided to the MQTT context.
+ *
+ * @param[in] pSslContext Pointer to SSL context.
+ * @param[out] pBuffer Buffer for receiving data.
+ * @param[in] bytesToSend Size of pBuffer.
```
yourslab, line 223: Should be `bytesToRecv`. I think it's also like this in plaintext demo.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -41,26 +58,36 @@
...
-#define PORT                   8883
+#define BROKER_PORT               8883
 
 /**
  * @brief Path of the file containing the server's root CA certificate.
  *
  * This certificate should be PEM-encoded.
  */
-#define SERVER_CERT            "mosquitto.org.crt"
+#define SERVER_CERT_PATH          "mosquitto.org.crt"
```
yourslab, line 80: Is it okay if we set `SERVER_CERT_PATH` to `certificates/mosquitto.org.crt`? Because the certificate needs to have the same path relative to the executable when compiling using `cmake`. Otherwise, we have to put `mosquitto.org.crt` together with other executables in the same directory.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -70,34 +97,248 @@
...
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define TRANSPORT_SEND_RECV_TIMEOUT_MS      ( 100 )
+
+/*-----------------------------------------------------------*/
+
+/* @brief errno to check transport error. */
+extern int errno;
```
yourslab, line 155: When is this used?

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -184,18 +431,40 @@ static SSL * tlsSetup( int tcpSocket )
...
+        else
+        {
+            LogError( ( "Failed to parse the server certificate from"
+                        " file %.*s. Please validate the certificate.",
+                        SERVER_CERT_PATH_LEN,
+                        SERVER_CERT_PATH ) );
+
+            /* Assert here in the step so as to help configure the path
+             * to certificate file. */
+            assert( false );
```
yourslab, line 466: I don't really have a good understanding of how asserts work under the hood. Is the file stream from the certificate automatically closed with a failed assert?

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: %s relies on internally calculating strlen, %.*s comes with a benefit that we can pass the already known string length. Also %s relies on the presence of a null character to the end. In this above instance it shouldn't make a difference.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: Yea. Thanks. Fixed.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: Yea, I thought about that. Yea, that is a good idea.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: Not used. removed. Thanks

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: Yea. That brings a good point. asert doesn't clean the resources properly. I think I should avoid assert in this case.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
yourslab: Cool. If you add this to the end of the `CMakeLists.txt` file for this demo, it will automatically be copied over:
```cmake
add_custom_command(
        TARGET ${DEMO_NAME} POST_BUILD
        COMMAND ${CMAKE_COMMAND} -E copy
                ${CMAKE_CURRENT_SOURCE_DIR}/certificates/mosquitto.org.crt
                ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/certificates/mosquitto.org.crt)
```

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
yourslab: Good to know. Thank you.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: I was also thinking the same. Just did that in the latest patch.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -289,108 +610,252 @@ static int32_t transportRecv( MQTTNetworkContext_t pSslContext,
...
                            uint16_t packetIdentifier,
                            MQTTPublishInfo_t * pPublishInfo )
 {
+    assert( pContext != NULL );
+    assert( pPacketInfo != NULL );
+
+    /* Handle incoming publish. The lower 4 bits of the publish packet
+     * type is used for the dup, QoS, and retain flags. Hence masking
+     * out the lower bits to check if the packet is publish. */
+    if( ( pPacketInfo->type & 0xF0U ) == MQTT_PACKET_TYPE_PUBLISH )
```
muneebahmed10, line 717: This could just be moved to the switch since SUBACK, UNSUBACK, and PINGRESP all have their lower four bits set to 0

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -184,59 +432,86 @@ static SSL * tlsSetup( int tcpSocket )
...
 
         if( pRootCaFile != NULL )
         {
             pRootCa = PEM_read_X509( pRootCaFile, NULL, NULL, NULL );
         }
+        else
+        {
+            LogError( ( "Unable to find the certificate file in the path"
+                        " provided by SERVER_CERT_PATH(%.*s).",
+                        SERVER_CERT_PATH_LEN,
```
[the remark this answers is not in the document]
leegeth: Line 400 was of type size_t. That was the reason we needed a cast to int. Here the conversion is to a wider type from uint16_t.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -184,59 +432,86 @@ static SSL * tlsSetup( int tcpSocket )
...
-            sslStatus = X509_STORE_add_cert( SSL_CTX_get_cert_store( pSslSetup ),
-                                             pRootCa );
+            status = X509_STORE_add_cert( SSL_CTX_get_cert_store( pSslSetup ),
+                                          pRootCa );
+        }
+        else
+        {
+            LogError( ( "Failed to parse the server certificate from"
+                        " file %.*s. Please validate the certificate.",
+                        SERVER_CERT_PATH_LEN,
```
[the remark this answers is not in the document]
leegeth: Line 400 was of type size_t. That was the reason we needed a cast to int. Here the conversion is to a wider type from uint16_t.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: I thought about that and intentionally made the change this way to make it clear to the reader that only PUBLISH packet type will have to mask the lower bits.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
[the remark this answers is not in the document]
muneebahmed10: I haven't tested this code, was just commenting based on what I've seen from previous build warnings:

/media/psf/Home/Documents/CSDK/libraries/standard/http/src/http_client.c:1200:20: warning: field precision specifier ‘.*’ expects argument of type ‘int’, but argument 2 has type ‘size_t {aka long unsigned int}’ [-Wformat=]

If there's no issue with uint16_t then it's fine

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -41,26 +57,36 @@
...
-#define PORT                   8883
+#define BROKER_PORT               ( 8883 )
 
 /**
  * @brief Path of the file containing the server's root CA certificate.
  *
  * This certificate should be PEM-encoded.
  */
-#define SERVER_CERT            "mosquitto.org.crt"
+#define SERVER_CERT_PATH          "certificates/mosquitto.org.crt"
```
muneebahmed10, line 79: Does this demo only run with mosquitto? Could this be changed to another cert at compile time with a cmake flag?

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -253,7 +528,23 @@ static SSL * tlsSetup( int tcpSocket )
...
+                    BROKER_ENDPOINT ) );
+        status = EXIT_FAILURE;
+    }
+    else
+    {
+        LogInfo( ( "Established a TLS connection to %.*s.\n\n",
+                   BROKER_ENDPOINT_LENGTH,
+                   BROKER_ENDPOINT ) );
+        status = EXIT_SUCCESS;
+    }
```
muneebahmed10, line 545: These lines seem odd to me since it's effectively `status = !status`. I think the return value should be a separate variable

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
yourslab: Maybe we can copy all *.crt or *.pem into the build/bin/certificates directory. The cmake flag works too though, but we would need to document that for the user.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -41,26 +57,36 @@
...
  *
  * This certificate should be PEM-encoded.
  */
-#define SERVER_CERT            "mosquitto.org.crt"
+#define SERVER_CERT_PATH          "certificates/mosquitto.org.crt"
+
+/**
+ * @brief Length of path to server certificate.
+ */
+#define SERVER_CERT_PATH_LEN      ( ( uint16_t ) ( sizeof( SERVER_CERT_PATH ) - 1 ) )
```
yourslab, line 84: This one ends with _LEN while the others end with _LENGTH

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
yourslab: Actually, I'm not sure if we should open the certificate using `fopen`. The relative directory is based on the location of the current working directory rather than the location of the executable. We could `chdir()` to the path of the executable, but the code we write to find that path might not always work depending on the OS. Maybe we should just place the certificate in the header file like in AFR.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -184,59 +432,86 @@ static SSL * tlsSetup( int tcpSocket )
...
+            SSL_set_verify( *pSslContext, SSL_VERIFY_PEER, NULL );
 
-            sslStatus = SSL_set_fd( pSslContext, tcpSocket );
+            status = SSL_set_fd( *pSslContext, tcpSocket );
         }
         else
         {
-            sslStatus = 0;
+            LogError( ( "Failed to create a new SSL context." ) );
+            status = 0;
```
yourslab, line 479: Might be better to use `SSL_ERROR_SSL` macro here

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: Yea agree. fixed.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: yea. Thanks. fixed.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: I like that proposal. That way we will be consistent in server auth and mutual auth demo. Will update the PR to do that. Thanks

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -253,83 +630,421 @@ static SSL * tlsSetup( int tcpSocket )
...
+
+static void handleIncomingPublish( MQTTPublishInfo_t * pPublishInfo,
+                                   uint16_t packetIdentifier )
+{
+    assert( pPublishInfo != NULL );
+
+    /* Process incoming Publish. */
+    LogInfo( ( "Incoming QOS : %d.", pPublishInfo->qos ) );
+
+    /* Verify the received publish is for the we have subscribed to. */
```
abhidixi11, line 941: Since the demo is  receiving echo of Publish message sent, may be we should verify packetId as well.  Store lastPacketIdSent and compare it with received packet id.

leegeth: Those packet Ids won't be the same. That will be part of 2 independent publishes. Spec only mentions about maintaining the packet Ids for Publishes and the acks for it.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -253,83 +630,421 @@ static SSL * tlsSetup( int tcpSocket )
...
+    /* Check all the QoS2 publish packets that are still waiting for
+     * a PUBREC. This can be identified by querying the state machine
+     * for all the outgoing publish packets which are in state
+     * #MQTTPubRecPending. */
+    do
+    {
+        /* Find the packet ids which are in #MQTTPubRecPending state.
+         * Using the same cursor will help restart the search from the
+         * last searched position. */
+        packetId = MQTT_StateSelect( pContext, MQTTPubRecPending, &cursor );
```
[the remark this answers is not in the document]
leegeth: Yea. Makes sense. Will fix it

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -70,43 +99,356 @@
...
  * @brief Length of client identifier.
  */
-#define CLIENT_IDENTIFIER_LENGTH    ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) )
+#define CLIENT_IDENTIFIER_LENGTH            ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) )
 
 /**
  * @brief Timeout for receiving CONNACK packet in milli seconds.
  */
-#define CONNACK_RECV_TIMEOUT_MS     ( 1000 )
+#define CONNACK_RECV_TIMEOUT_MS             ( 1000 )
```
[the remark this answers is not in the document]
leegeth: yes. Thanks

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -70,43 +99,356 @@
...
+
+/**
+ * @brief Delay between MQTT publishes in seconds.
+ */
+#define DELAY_BETWEEN_PUBLISHES_SECONDS     ( 1U )
+
+/**
+ * @brief Transport timeout in milliseconds for transport send and receive.
+ */
+#define TRANSPORT_SEND_RECV_TIMEOUT_MS      ( 20 )
```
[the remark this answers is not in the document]
leegeth: No. this is used in a signed type.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
abhidixi11: If you have verified it then I'm OK. Thanks.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
leegeth: Thinking about it again, I am keeping it this way because the mosquitto server certificate is downloadable as a file and a small transformation of that is needed to keep it as macro. File will be easier for an end customer.

abhidixi11: We can specify path in demo_config.h ( like we do in iot_config.h) in v4_beta.

demos/mqtt/mqtt_demo_basic_tls/demo_config.h
```diff
@@ -43,6 +43,38 @@
 
 /************ End of logging configuration ****************/
 
+/**
+ * @brief MQTT server host name.
+ *
+ * This demo uses the Mosquitto test server. This is a public MQTT server; do not
+ * publish anything sensitive to this server.
```
abhidixi11, line 50: We should add a comment here stating that mosquitto server can downloaded and used in local setup from <webstite> , in case public server in inaccessible or not desirable.

leegeth: Sounds good. Will add that. Thanks.

PR #974, opened 2020-05-28
Fix warnings when logs are enabled in MQTT

leegeth:
*Description of changes:*
Fix build warnings in logs. These warnings are reported when logs are enabled.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -1387,7 +1387,7 @@ MQTTStatus_t MQTT_SerializeConnect( const MQTTConnectInfo_t * const pConnectInfo
                     connectPacketSize ) );
         status = MQTTNoMemory;
     }
-    else if ( ( pWillInfo != NULL ) && ( pWillInfo->pTopicName == NULL ) )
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -2027,7 +2027,7 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * const pIncomingPacket
 
             /* Any other packet type is invalid. */
             default:
-                LogError( ( "IotMqtt_DeserializeResponse() called with unknown packet type:(%lu).", pIncomingPacket->type ) );
+                LogError( ( "IotMqtt_DeserializeResponse() called with unknown packet type:(%u).", pIncomingPacket->type ) );
```
muneebahmed10, line 2030: I'd prefer this to be changed to hex while we're at it.
```suggestion
                LogError( ( "IotMqtt_DeserializeResponse() called with unknown packet type:(%02x).", pIncomingPacket->type ) );
```

leegeth: Agree. Fixed. Thanks

PR #978, opened 2020-06-02
Add HTTP mutual auth demo

yourslab:
Demo creates a mutual TLS connection to AWS IoT Core then publishes a message to a topic named `topic` with `QoS=1`. Note that this does not work out of the box. User needs to modify the `demo_config.h` file to use their endpoint for IoT Core and also copy their client certificates in the `certificates` directory.


demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,137 @@
...
+#define LIBRARY_LOG_LEVEL    LOG_INFO
+
+#include "logging_stack.h"
+
+/************ End of logging configuration ****************/
+
+/**
+ * @brief HTTP server host name.
+ */
+#define SERVER_HOST                "server.cryptomix.com"
```
sarenameas, line 49: Why are we using this instead of AWS IoT?

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,137 @@
...
+ */
+#define CLIENT_CERT_PATH_LENGTH    ( ( uint16_t ) sizeof( CLIENT_CERT_PATH ) - 1 )
+
+/**
+ * @brief Path of the file containing the client's key for TLS client authentication.
+ *
+ *
+ * @note This key should be PEM-encoded.
+ */
+#define CLIENT_KEY_PATH            "certificates/client.key"
```
sarenameas, line 97: Assuming the same changes as the other PR are going to made for configurable macros here too.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
```diff
@@ -0,0 +1,834 @@
...
+
+/**
+ * @brief The HTTP Client library transport layer interface.
+ */
+static HTTPTransportInterface_t transportInterface;
+
+/**
+ * @brief Represents header data that will be sent in an HTTP request.
+ */
+static HTTPRequestHeaders_t requestHeaders;
```
sarenameas, line 85: Assuming the same changes for the other PR are going to made here as well.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
```diff
@@ -0,0 +1,834 @@
...
+        transportInterface.recv = transportRecv;
+        transportInterface.send = transportSend;
+        transportInterface.pContext = &networkContext;
+    }
+
+    /*********************** Send HTTPS request. ************************/
+
+    if( returnStatus == EXIT_SUCCESS )
+    {
+        returnStatus = sendHttpRequest( &transportInterface,
```
sarenameas, line 806: If we use AWS IoT we will only be able to post.

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * @brief HTTP server host name.
+ */
+#define SERVER_HOST                       "placeholder.iot.us-west-2.amazonaws.com"
+
+/**
+ * @brief HTTP server port number.
+ *
+ * @note Port 443 is normally used for HTTPS connections. However, because 443
```
sarenameas, line 54: Where did you find this information? We use port 443 in FreeRTOS.org demos.

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,108 @@
...
+ * is reserved for MQTT in AWS IoT Core, port 8443 is used instead.
+ */
+#define SERVER_PORT                       8443
+
+/**
+ * @brief Path of the file containing the server's root CA certificate for TLS authentication.
+ *
+ * @note This certificate should be PEM-encoded.
+ */
+#define SERVER_CERT_PATH                  "certificates/AmazonRootCA1.crt"
```
sarenameas, line 64: Noticed in the MQTT PR these are blank strings

demos/http/http_demo_mutual_tls/demo_config.h
sarenameas: Saw the AwS IoT documentation has use this port so i take back my comment,

demos/http/http_demo_mutual_tls/demo_config.h
yourslab: hmm idk, would be nice to let the user know that relative paths work here

demos/http/http_demo_mutual_tls/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+    PUBLIC
+        ${LOGGING_INCLUDE_DIRS}
+)
+
+#Copy the certificates and client key to the binary directory.
+add_custom_command(
+    TARGET
+        ${DEMO_NAME}
+    POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_directory
```
sarenameas, line 46: oh it's right here lol thanks

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
```diff
@@ -0,0 +1,923 @@
...
+        LogInfo( ( "Poll timed out and there is no data to read from the buffer." ) );
+    }
+    else
+    {
+        LogError( ( "Poll returned with status = %d.", pollStatus ) );
+        bytesReceived = -1;
+    }
+
+    return bytesReceived;
+}
```
gkwicker, line 730: Everything above this line to do with the network & TLS should be in a separate file since it has nothing to with HTTP.  They have opened a file called `http_demo_mutual_tls.c` to learn how to create and use an HTTP/TLS connection, but more than 75% of the file has nothing to do with HTTP.

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,118 @@
...
+#include "logging_stack.h"
+
+/************ End of logging configuration ****************/
+
+/**
+ * @brief HTTP server endpoint for your AWS IoT Core.
+ *
+ * @note This can be found in the "Settings" section of AWS IoT Core
+ */
+#define IOT_CORE_ENDPOINT                 "placeholder.iot.us-west-2.amazonaws.com"
```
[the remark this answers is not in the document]
sarenameas: Yea I mentioned in Lee's demo. We should probably delete it from the config.

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * @brief AWS IoT Core server port number for HTTPS connections.
+ *
+ * @note Port 443 is normally used for HTTPS connections. However, because 443
+ * is reserved for MQTT in AWS IoT Core, port 8443 is used instead.
```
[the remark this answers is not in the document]
sarenameas: Sigv4 is at the application level, so I don't think I would relate the two.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
yourslab: I completely agree. We already have a story for this that will put all TCP/TLS-related functions in the `platform/posix` directory. We will then refactor all demos for both MQTT and HTTP.

demos/http/http_demo_mutual_tls/certificates/AmazonRootCA1.crt
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
yourslab: I will include a link and have CMake automatically download this certificate in the correct directory.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
```diff
@@ -0,0 +1,957 @@
...
+                                                      &requestInfo );
+
+    if( httpStatus == HTTP_SUCCESS )
+    {
+        /* Initialize the response object. The same buffer used for storing
+         * request headers is reused here. */
+        response.pBuffer = userBuffer;
+        response.bufferLen = USER_BUFFER_LENGTH;
+
+        LogInfo( ( "Sending HTTP %.*s request to %.*s%.*s...",
```
sarenameas, line 820: IotLogDebug to match the other demos?

yourslab: We have this as LogInfo for all the other demos. The LogDebug only applies when printing the request headers and body.

demos/http/http_demo_mutual_tls/demo_config.h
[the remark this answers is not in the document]
yourslab: Added support so that ALPN is used depending on the port defined.

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -50,24 +50,41 @@
...
+ */
+#define IOT_CORE_ALPN_PROTOCOL_NAME       "\x0ex-amzn-http-ca"
 
 /**
  * @brief AWS IoT Core server port number for HTTPS connections.
  *
- * @note Port 443 is normally used for HTTPS connections. However, because 443
- * is reserved for MQTT in AWS IoT Core, port 8443 is used instead.
+ * For this demo, an X.509 certificate is used to verify the client.
+ * However, it is also possible to use Signature Version 4 as detailed here:
```
gkwicker, line 72: I wouldn't include lines 72&73 yet since (AFAIK) this library does not support SigV4 authentication.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
```diff
@@ -590,6 +601,32 @@ static int tlsSetup( int tcpSocket,
...
+            {
+                LogError( ( "SSL_set_alpn_protos failed to set ALPN protos. %s",
+                            pAlpnProtos ) );
+            }
+            else
+            {
+                sslStatus = 1;
+            }
+        }
+
```
gkwicker, line 629: Let's move the networking/TLS code to a separate file.  The reader has opened a file called `http_demo_mutual_tls.c`to learn about how to use the HTTP library, and the first 75% of the file has nothing to do with HTTP.  Also, moving the networking/TLS code to a separate file makes reuse possible, since that code is not HTTP-specific.

sarenameas: I believe we want to push this PR in with the networking code first, then refactor later.

demos/http/http_demo_mutual_tls/demo_config.h
sarenameas: We will never add Sigv4 comments into this demo. Sigv4 is needed for AWS IoT only when doing a request to the non-account-specific AWS IoT endpoint for an AWS API command like: `GET /endpoint?endpointType=endpointType HTTP/1.1`. When making requests to general AWS service endpoints we would be initially connecting over TLS with just server authentication.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
sarenameas: oh right I looked at this wrong thanks

demos/http/http_demo_mutual_tls/demo_config.h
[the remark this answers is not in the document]
sarenameas: Please see offline conversation.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
yourslab: Yah, the plain is to make the changes after the PR for the common networking library is finished.

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * #define IOT_CORE_ENDPOINT                 "your-aws-iot-core-endpoint"
+ */
+
+/**
+ * @brief ALPN protocol name to be sent as part of the ClientHello message.
+ *
+ * @note When using ALPN, port 443 must be used to connect to AWS IoT Core.
+ */
+#define IOT_CORE_ALPN_PROTOCOL_NAME       "\x0ex-amzn-http-ca"
```
sarenameas, line 66: There is no default for this, but we do not expect customers to change it so it might need to be in the .c file.

demos/http/http_demo_mutual_tls/demo_config.h
```diff
@@ -0,0 +1,135 @@
...
+#define ROOT_CA_CERT_PATH                 "certificates/AmazonRootCA1.crt"
+
+/**
+ * @brief Path of the file containing the client's certificate for TLS
+ * authentication to AWS IoT Core.
+ *
+ * @note This certificate should be PEM-encoded and must have an associated
+ * policy from AWS IoT core for the demo to function correctly.
+ */
+#define CLIENT_CERT_PATH                  "certificates/client.crt"
```
sarenameas, line 97: What is the reason for not putting the #error default macros for the client cert and private key?

sarenameas: As discussed, customers are also going to need some instructions somewhere that tells them where to paste their certificates.

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
```diff
@@ -0,0 +1,351 @@
...
+#ifndef USER_BUFFER_LENGTH
+    #define USER_BUFFER_LENGTH    ( 2048 )
+#endif
+
+/**
+ * @brief ALPN protocol name to be sent as part of the ClientHello message.
+ *
+ * @note When using ALPN, port 443 must be used to connect to AWS IoT Core.
+ */
+#ifndef IOT_CORE_ALPN_PROTOCOL_NAME
```
sarenameas, line 86: this is not configurable so let's take out the surrounding #ifdndef..#endif

demos/http/http_demo_mutual_tls/http_demo_mutual_tls.c
```diff
@@ -0,0 +1,351 @@
...
+    ServerInfo_t serverInfo;
+
+    ( void ) argc;
+    ( void ) argv;
+
+    /* Initialize TLS credentials. */
+    ( void ) memset( &opensslCredentials, 0, sizeof( opensslCredentials ) );
+    opensslCredentials.pClientCertPath = CLIENT_CERT_PATH;
+    opensslCredentials.pPrivateKeyPath = CLIENT_PRIVATE_KEY_PATH;
+    opensslCredentials.pRootCaPath = ROOT_CA_CERT_PATH;
```
sarenameas, line 292: We are expecting these certificates to always be NULL terminated?

yourslab: As discussed offline, this is done because the underlying API call requires NULL terminated strings.

PR #980, opened 2020-06-04
MQTT - handling re-sends when a session is reestablished.

leegeth:
*Description of changes:*
Handling the resend and receive of duplicate packets when an MQTT session is reestablished.

These are the cases that is handled in this PR,

When a session is reestablished,
Outgoing publishes:
1. Resend PUBLISHes for which and ack is not received(PUBACK and PUBREC).
2. Resend PUBRELs for PUBLISHes for which PUBCOMP was not received.

Incoming publishes:  
1. Handle duplicate QoS1 and QoS2 PUBLISHes
2. Handle duplicate PUBRELs

### DIFF
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -27,6 +27,13 @@
 #define MQTT_STATE_CURSOR_INITIALIZER    ( size_t ) 0
 
 /**
+ * @brief Invalid packet identifier.
+ *
+ * Zero is an invalid packet identifier as per MQTT v3.1.1 spec.
+ */
+#define MQTT_PACKET_ID_INVALID           ( ( uint16_t ) 0U )
+
+/**
  * @brief Value indicating either send or receive.
  */
 typedef enum MQTTStateOperation
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -27,13 +27,6 @@
 #define MQTT_STATE_CURSOR_INITIALIZER    ( size_t ) 0
 
 /**
- * @brief Invalid packet identifier.
- *
- * Zero is an invalid packet identifier as per MQTT v3.1.1 spec.
- */
-#define MQTT_PACKET_ID_INVALID           ( ( uint16_t ) 0U )
-
-/**
  * @brief Value indicating either send or receive.
  */
 typedef enum MQTTStateOperation
@@ -124,5 +117,21 @@ MQTTPublishState_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
 uint16_t MQTT_StateSelect( const MQTTContext_t * pMqttContext,
                            MQTTPublishState_t searchState,
                            MQTTStateCursor_t * pCursor );
+
+/**
+ * @brief Get the publish state for the publish packet with the given packet id.
+ *
+ * @param[in] pMqttContext Initialized MQTT context.
+ * @param[in] packetId The packet id of the packet to search for.
+ * @param[in] incomingPublish Whether to search for the publish packet in
+ * incoming publish records or the outgoing publish records. True for searching
+ * in incoming publish record; false for searching outgoing publish record.
+ *
+ * @return #MQTTStateNull if no publish packet found or if invalid parameters
+ * are passed; state of the publish packet with given packet id otherwise.
+ */
+MQTTPublishState_t MQTT_GetPacketState( const MQTTContext_t * pMqttContext,
+                                        uint16_t packetId,
+                                        bool incomingPublish );
 
 #endif /* ifndef MQTT_STATE_H */
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -26,6 +26,7 @@
 #include "mqtt_state.h"
 #include "private/mqtt_internal.h"
 
+
 /*-----------------------------------------------------------*/
 
 /**
@@ -111,12 +112,15 @@ static MQTTStatus_t receivePacket( MQTTContext_t * const pContext,
  * @param[in] pContext MQTT Connection context.
  * @param[in] packetId packet ID of original PUBLISH.
  * @param[in,out] pPublishState Current/updated publish state in record.
+ * @param[in] stateNeedUpdate Whether the state need to be updated after
+ * sending an ack or not.
  *
  * @return #MQTTSuccess or #MQTTIllegalState.
  */
 static MQTTStatus_t sendPublishAcks( MQTTContext_t * const pContext,
                                      uint16_t packetId,
-                                     MQTTPublishState_t * pPublishState );
+                                     MQTTPublishState_t * pPublishState,
+                                     bool stateNeedUpdate );
 
 /**
  * @brief Send a keep alive PINGREQ if the keep alive interval has elapsed.
@@ -198,6 +202,79 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
                                     uint32_t timeoutMs,
                                     MQTTPacketInfo_t * const pIncomingPacket,
                                     bool * const pSessionPresent );
+
+/**
+ * @brief Resends pending acks for a re-established MQTT session.
+ *
+ * @param[in] pContext Initialized MQTT context.
+ *
+ * @return #MQTTSendFailed if transport send failed;
+ * #MQTTSuccess otherwise.
+ */
+static MQTTStatus_t resendPendingAcks( MQTTContext_t * const pContext );
+
+/**
+ * @brief Resends PUBREL for the outgoing QoS2 publish and updates
+ * the state machine based on the state of the packet.
+ *
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] state Current state of the packet.
+ *
+ * @return #MQTTSendFailed if transport send failed;
+ * #MQTTSuccess otherwise.
+ */
+static MQTTStatus_t resendPubrelForPacketInState( MQTTContext_t * const pContext,
+                                                  MQTTPublishState_t state );
+
+/**
+ * @brief Checks if an incoming publish was received before.
+ *
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] packetId Packet identifier of the incoming publish packet.
+ * @param[in] qos Qos level of the incoming publish packet.
+ * @param[out] pIsStateUpdateNeededAfterAck Whether a state update is required
+ * after sending an ack for the incoming publish.
+ * @param[out] pPublishRecordState Publish record state for the ack to be sent.
+ *
+ * @return true if publish was already received and present in the state records;
+ * false otherwise.
+ */
+static bool checkIfPublishWasReceivedBefore( MQTTContext_t * const pContext,
+                                             uint16_t packetId,
+                                             MQTTQoS_t qos,
+                                             bool * pIsStateUpdateNeededAfterAck,
+                                             MQTTPublishState_t * pPublishRecordState );
+
+/**
+ * @brief Checks if packet with a given state is present in state records.
+ *
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] packetId Packet identifier to query in the state records.
+ * @param[in] state Packet state to query for in the state records.
+ *
+ * @return true if packet id with a given state is present in the state records;
+ * false otherwise.
+ */
+static bool isPacketStatePresentInStateRecords( MQTTContext_t * const pContext,
+                                                uint16_t packetId,
+                                                MQTTPublishState_t state );
+
+/**
+ * @brief Checks if a PUBREL packet was received before.
+ *
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] packetId Packet identifier of the incoming publish packet.
+ * @param[out] pIsStateUpdateNeededAfterAck Whether a state update is required
+ * after sending an ack for the incoming PUBREL.
+ * @param[out] pPublishRecordState Publish record state for the ack to be sent.
+ *
+ * @return true if PUBREL was already received and present in the state records;
+ * false otherwise.
+ */
+static bool checkIfPubrelWasReceivedBefore( MQTTContext_t * const pContext,
+                                            uint16_t packetId,
+                                            bool * pIsStateUpdateNeededAfterAck,
+                                            MQTTPublishState_t * pPublishRecordState );
 
 /*-----------------------------------------------------------*/
 
@@ -465,7 +542,8 @@ static MQTTStatus_t receivePacket( MQTTContext_t * const pContext,
 
 static MQTTStatus_t sendPublishAcks( MQTTContext_t * const pContext,
                                      uint16_t packetId,
-                                     MQTTPublishState_t * pPublishState )
+                                     MQTTPublishState_t * pPublishState,
+                                     bool stateNeedUpdate )
 {
     MQTTStatus_t status = MQTTSuccess;
     MQTTPublishState_t newState = MQTTStateNull;
@@ -519,16 +597,21 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * const pContext,
         if( bytesSent == ( int32_t ) MQTT_PUBLISH_ACK_PACKET_SIZE )
         {
             pContext->controlPacketSent = true;
-            newState = MQTT_UpdateStateAck( pContext,
-                                            packetId,
-                                            packetType,
-                                            MQTT_SEND );
 
-            if( newState == MQTTStateNull )
+            /* Only update the state machine if the state update is required. */
+            if( stateNeedUpdate )
             {
-                LogError( ( "Failed to update state of publish %u.",
-                            packetId ) );
-                status = MQTTIllegalState;
+                newState = MQTT_UpdateStateAck( pContext,
+                                                packetId,
+                                                packetType,
+                                                MQTT_SEND );
+
+                if( newState == MQTTStateNull )
+                {
+                    LogError( ( "Failed to update state of publish %u.",
+                                packetId ) );
+                    status = MQTTIllegalState;
+                }
             }
         }
         else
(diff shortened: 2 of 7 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -601,30 +818,124 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
     LogInfo( ( "De-serialized incoming PUBLISH packet: DeserializerResult=%d", status ) );
 
     if( status == MQTTSuccess )
+    {
+        /* Check if dup bit is set by the broker. Duplicates are not possible
+         * for QoS0 publishes. */
+        if( ( publishInfo.dup ) && ( ( publishInfo.qos == MQTTQoS1 ) ||
+                                     ( publishInfo.qos == MQTTQoS2 ) ) )
```
muneebahmed10, line 825:
```suggestion
        if( ( publishInfo.dup ) && ( publishInfo.qos > MQTTQoS0 ) )
```

has less conditions and would be easier for unit test coverage

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -199,6 +203,79 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+
+/**
+ * @brief Checks if an incoming publish was received before.
+ *
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] packetId Packet identifier of the incoming publish packet.
+ * @param[in] qos Qos level of the incoming publish packet.
+ * @param[out] pIsStateUpdateNeededAfterAck Whether a state update is required
+ * after sending an ack for the incoming publish.
+ * @param[out] pPublishRecordState Publish record state for the ack to be sent.
```
muneebahmed10, line 237: Don't think this parameter is ever used

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -199,6 +203,79 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+ *
+ * @return true if publish was already received and present in the state records;
+ * false otherwise.
+ */
+static bool checkIfPublishWasReceivedBefore( MQTTContext_t * const pContext,
+                                             uint16_t packetId,
+                                             MQTTQoS_t qos,
+                                             bool * pIsStateUpdateNeededAfterAck,
+                                             MQTTPublishState_t * pPublishRecordState );
```
muneebahmed10, line 246: Not really a fan of long names, since ultimately they lead to increased complexity due to splitting parameters across lines:
```suggestion
static bool isDupPublish( MQTTContext_t * const pContext,
                          uint16_t packetId,
                          MQTTQoS_t qos,
                          bool * pStateUpdateNeeded );
```
or something like `publishRecordExists`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -199,6 +203,79 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] packetId Packet identifier to query in the state records.
+ * @param[in] state Packet state to query for in the state records.
+ *
+ * @return true if packet id with a given state is present in the state records;
+ * false otherwise.
+ */
+static bool isPacketStatePresentInStateRecords( MQTTContext_t * const pContext,
+                                                uint16_t packetId,
+                                                MQTTPublishState_t state );
```
muneebahmed10, line 260: Shorter name:
```suggestion
static bool isPacketStatePresent( MQTTContext_t * const pContext,
                                  uint16_t packetId,
                                  MQTTPublishState_t state );
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -601,30 +818,124 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
...
                                                       publishInfo.qos );
         LogInfo( ( "State record updated. New state=%d.",
                    publishRecordState ) );
+    }
 
+    if( status == MQTTSuccess )
+    {
+        /* Pass the incoming publish to application, except the QoS2 duplicate
+         * incoming publishes. */
+        if( ( !hasPublishReceivedBefore ) || ( publishInfo.qos != MQTTQoS2 ) )
```
muneebahmed10, line 852: Why not QoS1?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -890,6 +1218,88 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+static MQTTStatus_t resendPendingAcks( MQTTContext_t * const pContext )
+{
+    MQTTStatus_t status = MQTTSuccess;
+
+    assert( pContext != NULL );
+
+    /* Check if there is any pending acks that need to be resend.
+     * Library owns resending PUBREL for a QoS2 outgoing publish.
+     * PUBREL packet will be resend if there are any packets at state
+     * #MQTTPubCompPending or #MQTTPubRelSend. */
```
muneebahmed10, line 1230: Why only outgoing publishes? Do we not need to resend PUBACK/PUBREC/PUBCOMP for incoming publishes?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -764,7 +1093,6 @@ static MQTTStatus_t sendPublish( MQTTContext_t * const pContext,
 
     assert( pContext != NULL );
     assert( pPublishInfo != NULL );
-    assert( headerSize > 0 );
```
muneebahmed10, line 767: This is fine to keep here. Fix is in #979

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -199,6 +203,79 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+ * after sending an ack for the incoming PUBREL.
+ * @param[out] pPublishRecordState Publish record state for the ack to be sent.
+ *
+ * @return true if PUBREL was already received and present in the state records;
+ * false otherwise.
+ */
+static bool checkIfPubrelWasReceivedBefore( MQTTContext_t * const pContext,
+                                            uint16_t packetId,
+                                            bool * pIsStateUpdateNeededAfterAck,
+                                            MQTTPublishState_t * pPublishRecordState );
```
muneebahmed10, line 277: This is never used

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -199,6 +203,79 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] packetId Packet identifier of the incoming publish packet.
+ * @param[out] pIsStateUpdateNeededAfterAck Whether a state update is required
+ * after sending an ack for the incoming PUBREL.
+ * @param[out] pPublishRecordState Publish record state for the ack to be sent.
+ *
+ * @return true if PUBREL was already received and present in the state records;
+ * false otherwise.
+ */
+static bool checkIfPubrelWasReceivedBefore( MQTTContext_t * const pContext,
```
muneebahmed10, line 274: This looks like a very specific function that could very easily be generalized to handle PUBREC as well. Do we not care about it?

libraries/standard/mqtt/src/mqtt.c
leegeth: yea. Thanks. Will remove it.

libraries/standard/mqtt/src/mqtt.c
leegeth: I like the name publishRecordExists. Thanks.

libraries/standard/mqtt/src/mqtt.c
leegeth: Yea. Thanks

libraries/standard/mqtt/src/mqtt.c
leegeth: QoS1 will always be passed to application as it allows duplicates(at least once delivery). On the other hand, library will not pass any duplicate QoS2 publishes to application(at most once).

libraries/standard/mqtt/src/mqtt.c
leegeth: A rebase will handle it. For me to unblock my unit tests, I would need to have this change.

libraries/standard/mqtt/src/mqtt.c
leegeth: Only PUBLISH and PUBREL packets will be resent as per MQTT V3.1.1. PUBLISH will be sent from application.

libraries/standard/mqtt/src/mqtt.c
leegeth: This is used when a MQTT_UpdateStatePublish call is not required to update the incoming publish state.

libraries/standard/mqtt/src/mqtt.c
leegeth: Only PUBLISH and PUBREL packets will be resend as per the spec. http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718103

libraries/standard/mqtt/src/mqtt.c
leegeth: This is used when MQTT_UpdateStateAck is not required to move to the next state.

libraries/standard/mqtt/src/mqtt.c
leegeth: yea. Agree. Thanks

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -539,46 +628,88 @@ static MQTTStatus_t receivePacket( MQTTContext_t * const pContext,
...
     }
+}
+
+/*-----------------------------------------------------------*/
+
+static MQTTStatus_t updateAckState( MQTTContext_t * const pContext,
+                                    uint16_t packetId,
+                                    bool stateNeedUpdate,
+                                    MQTTPubAckType_t packetType,
+                                    MQTTPublishState_t * const pNewState )
```
muneebahmed10, line 669: This function does not seem necessary to me either, since the complexity is 8 if you put this code back in `sendPublishAcks` (at least due to the changes in #979). Also, the `pNewState` param will never be used by `sendPublishAcks` due to the fact that this functions returns `MQTTIllegalState` on error.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -143,6 +143,17 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * const pContext );
...
+/**
+ * @brief Handle received MQTT ack for publish messages.
+ *
+ * @param[in] pContext MQTT Connection context.
+ * @param[in] pIncomingPacket Incoming packet.
+ *
+ * @return MQTTSuccess, MQTTIllegalState, or deserialization error.
+ */
+static MQTTStatus_t handleIncomingPubAck( MQTTContext_t * const pContext,
+                                          MQTTPacketInfo_t * pIncomingPacket );
```
muneebahmed10, line 155: Also not necessary due to complexity reduction in #979.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: We should specify that in a comment

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: If we rely on that, it's very likely that we'd forget about this, since deleting this line won't cause any merge conflicts in the rebase. If it's blocking you from running unit tests, build them with `cmake .. -DBUILD_TESTS="ON" -DCMAKE_C_FLAGS="-DNDEBUG"`

libraries/standard/mqtt/include/mqtt_state.h
```diff
@@ -26,6 +26,13 @@
 
 #define MQTT_STATE_CURSOR_INITIALIZER    ( size_t ) 0
 
+/**
+ * @brief Invalid packet identifier.
+ *
+ * Zero is an invalid packet identifier as per MQTT v3.1.1 spec.
+ */
+#define MQTT_PACKET_ID_INVALID           ( ( uint16_t ) 0U )
```
muneebahmed10, line 34: I know this was in `mqtt_state.c` before, but it seems like it would make more sense to be in `mqtt.h` instead of `mqtt_state.h`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -586,13 +789,148 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * const pContext )
 
 /*-----------------------------------------------------------*/
 
+static bool packetStatePresent( MQTTContext_t * const pContext,
+                                uint16_t packetId,
+                                MQTTPublishState_t state )
+{
+    bool hasPublishReceived = false;
+    uint16_t packetIdFromStoredPublish = MQTT_PACKET_ID_INVALID;
+    MQTTStateCursor_t cursor = 0U;
```
muneebahmed10, line 798: Why not have this be
```suggestion
    MQTTStateCursor_t cursor = MQTT_STATE_CURSOR_INITIALIZER;
```
That's what the macro's there for, after all.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -586,13 +789,148 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * const pContext )
...
+        packetIdFromStoredPublish = MQTT_StateSelect( pContext,
+                                                      state,
+                                                      &cursor );
+
+        /* If there was a publish received already with the same
+         * packet id.*/
+        if( packetIdFromStoredPublish == packetId )
+        {
+            /* The incoming packet was received earlier and in state
+             * MQTTPubRelPending. */
```
muneebahmed10, line 817: This function doesn't say anything about only searching for PubRelPending

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -586,13 +789,148 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * const pContext )
...
+    assert( packetId != MQTT_PACKET_ID_INVALID );
+
+    /* Find out if the publish packet has already received and state
+     * is maintained in the state machine. */
+    do
+    {
+        /* Query state machine for packet ids in given state. */
+        packetIdFromStoredPublish = MQTT_StateSelect( pContext,
+                                                      state,
+                                                      &cursor );
```
muneebahmed10, line 810: We don't really *need* to use state select here. Given how `packetStatePresent` is used in the rest of the code, it seems like a function that finds if a packet ID exists regardless of state might be more useful:
```c
records = ( incomingPublish )? pContext->incomingPublishRecords : pContext->outgoingPublishRecords;
for( int i = 0; i < MQTT_STATE_ARRAY_MAX_COUNT; i++)
{
    if( records[ i ].packetId == packetId )
    {
        hasPublishReceived = true;
        *pRetrievedState = records[i].publishState;
        break;
    }
}
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -890,6 +1211,196 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+                            packetId ) );
+            }
+            else
+            {
+                LogDebug( ( "Resent PUBREL for packet id %u.",
+                            packetId ) );
+            }
+        }
+    } while( packetId != MQTT_PACKET_ID_INVALID &&
+             status == MQTTSuccess );
```
muneebahmed10, line 1289:
```suggestion
    } while( ( packetId != MQTT_PACKET_ID_INVALID ) &&
             ( status == MQTTSuccess ) );
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -890,6 +1211,196 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+    MQTTPublishState_t publishRecordState = MQTTPubRelSend;
+
+    /* State machine needs update only if packet is not already in state
+     * #MQTTPubCompPending. */
+    if( state == MQTTPubRelSend )
+    {
+        stateNeedUpdate = true;
+    }
+
+    do
```
muneebahmed10, line 1261: You could reduce nesting in this loop:
```c
uint16_t packetId = MQTT_StateSelect( pContext, state, &cursor );

/* other code */

while( ( status == MQTTSuccess ) && ( packetId != MQTT_PACKET_ID_INVALID ) )
{
    status = sendPublishAcks( pContext, packetId, &publishRecordState, stateNeedUpdate );
    /* Log status */
    packetId = MQTT_StateSelect( pContext, state, &cursor );
}
```
but I don't think it's necessarily better or worse than what you already have here

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -890,6 +1211,196 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+         * last searched position. */
+        packetId = MQTT_StateSelect( pContext, state, &cursor );
+
+        /* Check if a valid packet id is obtained. */
+        if( packetId != MQTT_PACKET_ID_INVALID )
+        {
+            /* Resend pending PUBREL. */
+            status = sendPublishAcks( pContext,
+                                      packetId,
+                                      &publishRecordState,
```
muneebahmed10, line 1274: Do we expect `sendPublishAcks` to update `publishRecordState`? In that case, shouldn't it be reset to `MQTTPubRelSend` in every iteration of the loop?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -890,6 +1211,196 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+    MQTTStatus_t status = MQTTSuccess;
+    MQTTStateCursor_t cursor = MQTT_STATE_CURSOR_INITIALIZER;
+    uint16_t packetId = MQTT_PACKET_ID_INVALID;
+    bool stateNeedUpdate = false;
+
+    assert( pContext != NULL );
+    assert( state == MQTTPubCompPending || state == MQTTPubRelSend );
+
+    /* Resending PUBREL packet. */
+    MQTTPublishState_t publishRecordState = MQTTPubRelSend;
```
muneebahmed10, line 1252: I think variables need to be declared before asserts to conform to our style

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -890,6 +1211,196 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+        {
+            LogError( ( "Update state for publish failed with status =%u."
+                        " However PUBLISH packet is sent to the broker."
+                        " Any further handling of ACKs for the packet Id"
+                        " will fail.",
+                        publishStatus ) );
+
+            /* TODO. Need to remove this update once MQTT_UpdateStatePublish is
+             * refactored with return type of MQTTStatus_t. */
+            status = MQTTBadParameter;
```
muneebahmed10, line 1395: I think `MQTTIllegalState` is more appropriate here than `MQTTBadParameter`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1077,57 +1596,25 @@ MQTTStatus_t MQTT_Publish( MQTTContext_t * const pContext,
...
+                                   &headerSize );
     }
 
     if( status == MQTTSuccess )
     {
-        /* Reserve state for publish message. Only to be done for QoS1 or QoS2. */
-        if( pPublishInfo->qos > MQTTQoS0 )
+        /* Reserve state for publish message. Only to be done for QoS1 or QoS2
+         * and a non duplicate publish. */
+        if( ( pPublishInfo->qos > MQTTQoS0 ) && ( !pPublishInfo->dup ) )
```
muneebahmed10, line 1617: MISRA:
```suggestion
        if( ( pPublishInfo->qos > MQTTQoS0 ) && ( pPublishInfo->dup == false ) )
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1410,6 +1875,7 @@ MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * const pContext,
         {
             /* Receive packet. Remaining time is recalculated at the end of the loop. */
             status = receivePacket( pContext, incomingPacket, remainingTimeMs );
+            LogDebug( ( "After receiving packet with status %u.", status ) );
```
muneebahmed10, line 1878: Reword to something like "Packet reception concluded with result ..."

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -586,13 +700,108 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * const pContext )
...
+    assert( pContext != NULL );
+    assert( packetId != MQTT_PACKET_ID_INVALID );
+    assert( pIsStateUpdateNeededAfterAck != NULL );
+    assert( pPublishRecordState != NULL );
+    assert( qos == MQTTQoS1 || qos == MQTTQoS2 );
+
+    /* Get the packet state from the state engine. */
+    publishState = MQTT_GetPacketState( pContext, packetId, true );
+
+    if( qos == MQTTQoS1 )
```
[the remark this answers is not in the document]
leegeth: I like the suggestion, I will keep the if - else, but take the pointers to optimize it. Thanks

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -601,30 +810,125 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
...
+     * not duplicate. No action to be taken here. */
+    else if( publishState == MQTTPubRelPending )
+    {
+        /* This is the first time PUBREL is received for this packet. */
+    }
+
+    /* No valid state exists for PUBREL. Just respond with a PUBCOMP
+     * without updating state. Marking this as duplicate as this would
+     * indicate a state in which application would have lost the states of
+     * messages received before reestablishing a session. */
```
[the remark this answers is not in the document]
leegeth: Not in the duplicate handling code, I think that would be caught as a invalid state transition by our state engine.

libraries/standard/mqtt/src/mqtt.c
leegeth: Yes.good catch

leegeth:
Closing this PR as this change is refactored and handled in a different approach in PR #1007.

PR #985, opened 2020-06-05
Fix: HTTP header context initialization and usage of transport send

yourslab:
This PR makes the following fixes for the HTTP Client Library:
- `findHeaderContext_t * pContext` is set AFTER the assert checks in `findHeaderFieldParserCallback` and `findHeaderValueParserCallback`.
- `sendHttpData` assumes that all data can be sent after a single invocation of transport send. However, if the socket write buffer is too small for all data being sent, this will not be the case. 
    - `test_HTTPClient_Send_less_bytes_request_headers` and `test_HTTPClient_Send_less_bytes_request_body` are updated to return successfully because of the above change

### DIFF
--- a/libraries/standard/http/src/http_client.c
+++ b/libraries/standard/http/src/http_client.c
@@ -1519,7 +1519,7 @@ static HTTPStatus_t sendHttpData( const HTTPTransportInterface_t * pTransport,
 {
     HTTPStatus_t returnStatus = HTTP_SUCCESS;
     const uint8_t * pIndex = pData;
-    int32_t bytesSent = 0;
+    int32_t transportStatus = 0;
     size_t bytesRemaining = dataLen;
 
     assert( pTransport != NULL );
@@ -1529,18 +1529,18 @@ static HTTPStatus_t sendHttpData( const HTTPTransportInterface_t * pTransport,
     /* Loop until all data is sent. */
     while( bytesRemaining > 0UL )
     {
-        bytesSent = pTransport->send( pTransport->pContext,
-                                      pData,
-                                      dataLen );
+        transportStatus = pTransport->send( pTransport->pContext,
+                                            pIndex,
+                                            bytesRemaining );
 
-        if( bytesSent > 0 )
+        if( transportStatus > 0 )
         {
-            bytesRemaining -= ( size_t ) bytesSent;
-            pIndex += bytesSent;
+            bytesRemaining -= ( size_t ) transportStatus;
+            pIndex += transportStatus;
             LogDebug( ( "Sent HTTP data over the transport: "
                         "BytesSent=%d, BytesRemaining=%ul, "
                         "TotalBytesSent=%d.",
-                        bytesSent,
+                        transportStatus,
                         bytesRemaining,
                         dataLen - bytesRemaining ) );
         }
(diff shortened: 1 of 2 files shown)


libraries/standard/http/src/http_client.c
```diff
@@ -1518,33 +1518,45 @@ static HTTPStatus_t sendHttpData( const HTTPTransportInterface_t * pTransport,
...
-                    transportStatus ) );
-        returnStatus = HTTP_NETWORK_ERROR;
+        bytesSent = pTransport->send( pTransport->pContext,
+                                      pData,
+                                      dataLen );
+
+        if( bytesSent > 0 )
+        {
+            bytesRemaining -= ( size_t ) bytesSent;
+            totalBytesSent += bytesSent;
```
muneebahmed10, line 1539: Actually, the only use of `totalBytesSent` seems to be in the debug log. I think this could just be deleted and line 1546 changed to `dataLen - bytesRemaining`

libraries/standard/http/src/http_client.c
```diff
@@ -1980,6 +1990,8 @@ static int findHeaderValueParserCallback( http_parser * pHttpParser,
     assert( pVaLueLoc != NULL );
     assert( valueLen > 0u );
 
+    pContext = ( findHeaderContext_t * ) pHttpParser->data;
```
muneebahmed10, line 1993: Delete line 2000 as it's redundant now

yourslab: oops thanks

yourslab: you catch things so fast

libraries/standard/http/utest/http_send_utest.c
```diff
@@ -1083,8 +1083,8 @@ void test_HTTPClient_Send_null_response( void )
 {
     HTTPStatus_t returnStatus = HTTP_SUCCESS;
 
-    memcpy( requestHeaders.pBuffer, 
-            HTTP_TEST_REQUEST_PUT_HEADERS, 
+    memcpy( requestHeaders.pBuffer,
+            HTTP_TEST_REQUEST_PUT_HEADERS,
```
yourslab, line 1087: Any changes from above are from uncrustify (using the VS Code plugin)

sarenameas: Thanks

libraries/standard/http/src/http_client.c
```diff
@@ -1518,36 +1518,46 @@ static HTTPStatus_t sendHttpData( const HTTPTransportInterface_t * pTransport,
...
-                    ( unsigned long ) dataLen,
-                    transportStatus ) );
-        returnStatus = HTTP_NETWORK_ERROR;
+        transportStatus = pTransport->send( pTransport->pContext,
+                                            pIndex,
+                                            bytesRemaining );
+
+        if( transportStatus > 0 )
+        {
+            bytesRemaining -= ( size_t ) transportStatus;
```
gkwicker, line 1538: defensive programming opportunity: if there is a bug in the `send()` routine on line 1532 which causes it to return a value greater than `bytesRemaining`, we won't exit the loop as intended with the test on line 1530 since `bytesRemaining` is unsigned.  Consider sanity checking the value of `transport_status` prior to the assignment at line 1538, e.g.:

```c
if( transportStatus > bytesRemaining )
{
    transportStatus = bytesRemaining;
    LogError( ( "Unexpected value returned from pTransport->send()...
}
```

PR #992, opened 2020-06-11
Update HTTPNetworkContext_t to NetworkContext_t (struct ptr type)

yourslab:
*Summary of changes*
- Rename `HTTPNetworkContext_t` to `NetworkContext_t` and use struct ptr type:
```c
/**
 * @brief The NetworkContext is an incomplete type. The application must
 * define NetworkContext to the type of their network context. This context
 * is passed into the network interface functions.
 */
struct NetworkContext;
typedef struct NetworkContext * NetworkContext_t;
```

- Update HTTP tests and demos to reflect this change

### DIFF
--- a/libraries/standard/http/include/http_client.h
+++ b/libraries/standard/http/include/http_client.h
@@ -131,12 +131,12 @@
 #define HTTP_RANGE_REQUEST_END_OF_FILE              -1
 
 /**
- * @brief The HTTPNetworkContext is an incomplete type. The application must
- * define HTTPNetworkContext to the type of their network context. This context
+ * @brief The NetworkContext is an incomplete type. The application must
+ * define NetworkContext to the type of their network context. This context
  * is passed into the network interface functions.
  */
-struct HTTPNetworkContext;
-typedef struct HTTPNetworkContext HTTPNetworkContext_t;
+struct NetworkContext;
+typedef struct NetworkContext * NetworkContext_t;
 
 /**
  * @brief Transport interface for sending data over the network.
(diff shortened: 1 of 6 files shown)


libraries/standard/http/include/http_client.h
```diff
@@ -130,12 +130,12 @@
...
- * @brief The HTTPNetworkContext is an incomplete type. The application must
- * define HTTPNetworkContext to the type of their network context. This context
+ * @brief The NetworkContext is an incomplete type. The application must
+ * define NetworkContext to the type of their network context. This context
  * is passed into the network interface functions.
  */
-struct HTTPNetworkContext;
-typedef struct HTTPNetworkContext HTTPNetworkContext_t;
+struct NetworkContext;
+typedef struct NetworkContext * NetworkContext_t;
```
yourslab, line 138: This is defined here for now but will be transferred once we make the common transport library. Also, I still think it's a little weird that this is the only struct ptr type in the entire library, but that's just me.

muneebahmed10: No, I think it's pretty weird too

abhidixi11: It follows what FreeRTOS kernel does with opaque types. This is to void pointers.

abhidixi11: Looking at the entire PR now, I think, HTTP library itself should not reinterpret NetworkContext_t , it should be defined in network_interface.h (eventually). Make sure the implementation is same as MQTT, since we are expected to share the transport implementation.

PR #993, opened 2020-06-11
Fix MQTT MISRA violations

muneebahmed10:
*Description of changes:*
Fixes most MISRA violations in MQTT. The remaining violations with our current configuration are all 8.7 (External linkage for functions used in one translation unit), which is required for API functions. Note that the changes in commit "Remove useless const" aren't due to MISRA violations, we had just been adding const because we had them in v4_beta (due to other MISRA violations relating to function pointers). However, since C is pass by value I didn't find value in keeping them.

### DIFF
--- a/demos/logging-stack/logging_stack.h
+++ b/demos/logging-stack/logging_stack.h
@@ -40,7 +40,6 @@
 #define LOG_METADATA_ARGS      __FILE__, __LINE__
 
 /* Common macro for all logging interface macros. */
-/* TODO - Replace printf with an implementation function. */
 #if !defined( DISABLE_LOGGING )
     #define SdkLog( string )    printf string
 #else
--- a/demos/logging-stack/logging_stack.h
+++ b/demos/logging-stack/logging_stack.h
@@ -41,7 +41,9 @@
 
 /* Common macro for all logging interface macros. */
 /* TODO - Replace printf with an implementation function. */
-#define SdkLog( string )    printf string
+#ifndef SdkLog
+    #define SdkLog( string )    printf string
+#endif
 
 /* Check that LIBRARY_LOG_LEVEL is defined and has a valid value. */
 #if !defined( LIBRARY_LOG_LEVEL ) ||       \
--- a/demos/logging-stack/logging_stack.h
+++ b/demos/logging-stack/logging_stack.h
@@ -41,8 +41,10 @@
 
 /* Common macro for all logging interface macros. */
 /* TODO - Replace printf with an implementation function. */
-#ifndef SdkLog
+#if !defined( DISABLE_LOGGING )
     #define SdkLog( string )    printf string
+#else
+    #define SdkLog( string )
 #endif
 
 /* Check that LIBRARY_LOG_LEVEL is defined and has a valid value. */
--- a/demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
+++ b/demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
@@ -588,7 +588,7 @@ static int establishMqttSession( MQTTContext_t * pContext,
     int status = EXIT_SUCCESS;
     MQTTStatus_t mqttStatus;
     MQTTConnectInfo_t connectInfo;
-    char sessionPresent;
+    bool sessionPresent;
     MQTTTransportInterface_t transport;
     MQTTFixedBuffer_t networkBuffer;
     MQTTApplicationCallbacks_t callbacks;
--- a/libraries/standard/mqtt/include/mqtt_lightweight.h
+++ b/libraries/standard/mqtt/include/mqtt_lightweight.h
@@ -22,17 +22,17 @@
 #ifndef MQTT_LIGHTWEIGHT_H
 #define MQTT_LIGHTWEIGHT_H
 
-/* bools are only defined in C99+ */
-#if defined( __cplusplus ) || __STDC_VERSION__ >= 199901L
-    #include <stdbool.h>
-#elif !defined( bool )
-    #define bool     signed char
-    #define false    0
-    #define true     1
-#endif
-
 #include <stddef.h>
 #include <stdint.h>
+
+/* bools are only defined in C99+ */
+#if defined( __cplusplus ) || ( defined( __STDC_VERSION__ ) && ( __STDC_VERSION__ >= 199901L ) )
+    #include <stdbool.h>
+#elif !defined( bool )
+    #define bool     int8_t
+    #define false    ( int8_t ) 0
+    #define true     ( int8_t ) 1
+#endif
 
 /* Include config file before other headers. */
 #include "mqtt_config.h"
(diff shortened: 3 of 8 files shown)


libraries/standard/mqtt/include/mqtt_lightweight.h
```diff
@@ -22,18 +22,18 @@
...
-#if defined( __cplusplus ) || __STDC_VERSION__ >= 199901L
+#if defined( __cplusplus ) || ( defined( __STDC_VERSION__ ) && ( __STDC_VERSION__ >= 199901L ) )
     #include <stdbool.h>
 #elif !defined( bool )
-    #define bool     signed char
-    #define false    0
-    #define true     1
+    #define bool     int8_t
+    #define false    ( int8_t ) 0
+    #define true     ( int8_t ) 1
```
muneebahmed10, line 34: I think some wanted to get rid of our `#define` of `bool`, and others wanted to keep it. I don't believe there was a clear decision on that matter, and I don't think removing it is in scope for this PR. Note that due to the changes in this PR, removal (if desired) should be straightforward, just a search and replace of `bool` with the desired type, `false` with `0`, and `true` with `1`

demos/logging-stack/logging_stack.h
```diff
@@ -41,7 +41,9 @@
 
 /* Common macro for all logging interface macros. */
 /* TODO - Replace printf with an implementation function. */
-#define SdkLog( string )    printf string
+#ifndef SdkLog
+    #define SdkLog( string )    printf string
+#endif
```
muneebahmed10, line 46: This was changed so that we can easily disable logging from the command line with `-D'SdkLog(string)='`

demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
```diff
@@ -588,7 +588,7 @@ static int establishMqttSession( MQTTContext_t * pContext,
     int status = EXIT_SUCCESS;
     MQTTStatus_t mqttStatus;
     MQTTConnectInfo_t connectInfo;
-    char sessionPresent;
+    bool sessionPresent;
```
sarenameas, line 591: I thought bools were not allowed

demos/logging-stack/logging_stack.h
sarenameas: This is nice, but I think it may be better to have a "DISABLE_LOGGING" master macro or something?

muneebahmed10: Changed

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -140,11 +145,21 @@ static bool validateTransitionPublish( MQTTPublishState_t currentState,
             switch( qos )
             {
                 case MQTTQoS1:
-                    isValid = ( newState == MQTTPubAckPending );
+                    /* Removing the below cast results in a MISRA 10.3 violation (required, implicit cast
+                     * from a boolean), and keeping it results in a 10.5 violation (advisory, explicit cast).
+                     * The violation can only be resolved if the variable is of a boolean type, which is
+                     * not present in C89. */
+                    /* coverity[misra_c_2012_rule_10_5_violation] */
```
abhidixi11, line 152: We should not have annotation and comment in a separate block.  If use of bool is causing this us to use annotation, I think , we should remove bool. Instead may be use BOOL / Bool something other than c99 bool.

muneebahmed10: 1. I'm unsure what you mean by "C99 mode" but according to our cmake, we compile with `std=c89`. If we compile with C99 instead, then this isn't a violation, since the types of these values are actually booleans and not whatever we define them as for C89. This is only a violation when using C89.
1. The same violation occurs whether we define bool as int32_t, uint32_t, uint8_t, etc. The issue is that MISRA considers the outputs of relational operators as "essentially boolean types", which is different from "essentially signed" or "essentially unsigned", and won't allow even explicit casts between them. Since there's no data type in C89 that's "essentially boolean", MISRA will complain if we store these values to a variable. So removing bool and replacing it with anything else will not help either.
1. Yes, that is an option, however I was reluctant to do that earlier since it would greatly increase complexity of our functions due to increased nesting and lines of code, and they would have had to be split up in ways that didn't seem the most logical. However, I realized that instead of `if`s, we can just use `? :` instead, as that both solved the 10.5 issue and did not noticeably increase the complexity.

I reverted the annotation change and used the ternary operator instead

demos/logging-stack/logging_stack.h
```diff
@@ -41,7 +41,11 @@
 
 /* Common macro for all logging interface macros. */
 /* TODO - Replace printf with an implementation function. */
```
gkwicker, line 43: Who is this `TODO` targeted to?  A developer _of_ this library, or a developer _who uses_ this library?

demos/logging-stack/logging_stack.h
```diff
@@ -41,7 +41,11 @@
...
-#define SdkLog( string )    printf string
+#if !defined( DISABLE_LOGGING )
+    #define SdkLog( string )    printf string
```
gkwicker, line 45: Where are the `( )`?  This won't compile, is that the intent?

libraries/standard/mqtt/src/mqtt_state.c
abhidixi11: Regarding #1 and #2 I thought in C89 / ANSI mode value of boolean should be int or "essentially signed". That's why I asked if Coverity is  throwing that error because it sees bool which is c99 and considers it a data type.  #3 was basically the same as you change, which avoid any casts. Your change looks good, approved. ( Just make sure ternary operator is not a problem MISRA )

abhidixi11: Regarding #1 and #2 I thought in C89 / ANSI mode value of boolean should be int or "essentially signed". That's why I asked if Coverity is  throwing that error because it sees bool which is c99 and considers it a data type.  #3 was basically the same as you change, which avoid any casts. Your change looks good, approved. ( Just make sure ternary operator is not a problem MISRA )

abhidixi11: Regarding #1 and #2 I thought in C89 / ANSI mode value of boolean should be int or "essentially signed". That's why I asked if Coverity is  throwing that error because it sees bool, which is c99 and considers it a "essentially boolean".  #3 was basically the same as you change, which avoid any casts. Your change looks good, approved. ( Just make sure ternary operator is not a problem with MISRA )

demos/logging-stack/logging_stack.h
muneebahmed10: I believe this is the reason why all of our logs use double parenthesis `( ( ) )` instead of single `( )`. I think it was made this way since variable arguments in macros are not supported in C89, so all of it is passed at once to printf

demos/logging-stack/logging_stack.h
muneebahmed10: For us, since printf may not be thread-safe on all POSIX compliant MCUs. Archit has a PoC for using a logging thread for log message inputs

gkwicker: Then I'd make sure that it is clear to anyone using this library that it's not their responsibility to implement that `TODO`.  I think we need some more explanation here.

aggarw13: The TODO is just temporary until the PR for the logging servicing thread is raised. If there is concern about it being misinterpreted, we can remove it completely for now.

PR #994, opened 2020-06-11
MQTT_Connect with retries

leegeth:
*Description of changes:*
Update the logic in MQTT_Connect to work with retries if a timeout of 0ms is passed to the MQTT library.

### DIFF
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -164,19 +164,31 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
  *    A network read would be retried upto the number of times configured by
  *    #MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT for the CONNACK packet.
  *
- * @brief param[in] pContext Initialized MQTT context.
- * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
- * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
- * @brief param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
- * packet. In certain conditions, depending on the network conditions or the
- * transport interface receive timeout configured, this function can spend more
- * time in receiving CONNACK packet than the provided timeout duration. However,
- * the probability of exceeding the timeout is very low assuming the timeouts
- * for CONNACK receive and transport interface receive are configured correctly.
- * Refer to the transport interface documentation for more details about the
- * timeouts. @note A zero timeout will make use of the retries for receiving
+ * @param[in] pContext Initialized MQTT context.
+ * @param[in] pConnectInfo MQTT CONNECT packet parameters.
+ * @param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
+ * @param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
+ * packet.
+ * @note A zero timeout will make use of the retries for receiving
  * CONNACK.
- * @brief param[out] pSessionPresent Whether a previous session was present.
+ * @note The API may take more time than the given timeout in certain
+ * conditions as listed below.
+ * 1. Timeouts are incorrectly configured - If the timeout passed to this API
+ * is less than the transport receive timeout and if a CONNACK packet is not
+ * received within the transport receive timeout, the API will spend more time
+ * than the timeoutMs parameter passed. The timeoutMs parameter passed to this
+ * API should be greater than the transport receive timeout. Please refer to the
+ * transport interface documentation for more details about timeout
+ * configurations.
+ * 2. CONNACK packet is received right after expiry of the timeout -
+ * It is possible the CONNACK packet is received towards the expiry of the
+ * timeout passed to the API. In the worst case, it can take upto
+ * timeoutMs + transport receive timeout to receive the packet type and
+ * remaining length of CONNACK packet from network. The API will attempt to
+ * read the remaining 2 bytes after reading the packet type and remaining
+ * length. In this case, API will spend more time than timeoutMs for receiving
+ * CONNACK.
+ * @param[out] pSessionPresent Whether a previous session was present.
  * Only relevant if not establishing a clean session.
  *
  * @return #MQTTNoMemory if the #MQTTContext_t.networkBuffer is too small to
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -152,7 +152,7 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
                         const MQTTFixedBuffer_t * pNetworkBuffer );
 
 /**
- * @brief Establish a MQTT session.
+ * @brief Establish an MQTT session.
  *
  * This function will send MQTT CONNECT packet and receive a CONNACK packet.
  * The send and receive from the network is through the transport interface.
@@ -168,8 +168,8 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
  * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
  * @brief param[in] timeoutMs Timeout in milliseconds for receiving
- * CONNACK packet. A zero timeout value will attempt to receive through
- * transport receive once.
+ * CONNACK packet. A zero timeout value will attempt a single reception
+ * from the transport interface.
  * @brief param[out] pSessionPresent Whether a previous session was present.
  * Only relevant if not establishing a clean session.
  *
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -154,11 +154,22 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
 /**
  * @brief Establish a MQTT session.
  *
+ * This function will send MQTT CONNECT packet and receive a CONNACK packet.
+ * The send and receive from the network is through the transport interface.
+ * The maximum wait for receiving a CONNACK is managed in 2 ways.
+ * 1. If a #getTime interface function is passed to the MQTT library:
+ *    Parameter #timeoutMs will be used for a #getTime based maximum timeout
+ *    for the network read.
+ * 2. If a #getTime interface function is not passed to the MQTT library:
+ *    A network read would be retried upto the number of times configured by
+ *    #MQTT_CONNACK_RECEIVE_RETRY_COUNT.
+ *
  * @brief param[in] pContext Initialized MQTT context.
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
  * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
  * @brief param[in] timeoutMs Timeout in milliseconds for receiving
- * CONNACK packet.
+ * CONNACK packet. A zero timeout value will attempt to receive through
+ * transport receive once.
  * @brief param[out] pSessionPresent Whether a previous session was present.
  * Only relevant if not establishing a clean session.
  *
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -170,11 +170,12 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
  * @brief param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
  * packet. In certain conditions, depending on the network conditions or the
  * transport interface receive timeout configured, this function can spend more
- * time in receiving CONNACK packet as compared to the timeout provided. However
- * the probability of exceeding the timeout is very low assuming the the timeouts
+ * time in receiving CONNACK packet than the provided timeout duration. However,
+ * the probability of exceeding the timeout is very low assuming the timeouts
  * for CONNACK receive and transport interface receive are configured correctly.
  * Refer to the transport interface documentation for more details about the
- * timeouts. A zero timeout will make use of the retries for receiving CONNACK.
+ * timeouts. @note A zero timeout will make use of the retries for receiving
+ * CONNACK.
  * @brief param[out] pSessionPresent Whether a previous session was present.
  * Only relevant if not establishing a clean session.
  *
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -167,9 +167,14 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
  * @brief param[in] pContext Initialized MQTT context.
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
  * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
- * @brief param[in] timeoutMs Timeout in milliseconds for receiving
- * CONNACK packet. A zero timeout value will attempt a single reception
- * from the transport interface.
+ * @brief param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
+ * packet. In certain conditions, depending on the network conditions or the
+ * transport interface receive timeout configured, this function can spend more
+ * time in receiving CONNACK packet as compared to the timeout provided. However
+ * the probability of exceeding the timeout is very low assuming the the timeouts
+ * for CONNACK receive and transport interface receive are configured correctly.
+ * Refer to the transport interface documentation for more details about the
+ * timeouts. A zero timeout will make use of the retries for receiving CONNACK.
  * @brief param[out] pSessionPresent Whether a previous session was present.
  * Only relevant if not establishing a clean session.
  *
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -157,12 +157,12 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
  * This function will send MQTT CONNECT packet and receive a CONNACK packet.
  * The send and receive from the network is through the transport interface.
  * The maximum wait for receiving a CONNACK is managed in 2 ways.
- * 1. If a #getTime interface function is passed to the MQTT library:
+ * 1. If #timeoutMs is greater than 0:
  *    Parameter #timeoutMs will be used for a #getTime based maximum timeout
- *    for the network read.
- * 2. If a #getTime interface function is not passed to the MQTT library:
+ *    for the network read for CONNACK packet.
+ * 2. If #timeoutMs is 0:
  *    A network read would be retried upto the number of times configured by
- *    #MQTT_CONNACK_RECEIVE_RETRY_COUNT.
+ *    #MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT for the CONNACK packet.
  *
  * @brief param[in] pContext Initialized MQTT context.
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -154,40 +154,25 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
 /**
  * @brief Establish an MQTT session.
  *
- * This function will send MQTT CONNECT packet and receive a CONNACK packet.
- * The send and receive from the network is through the transport interface.
- * The maximum wait for receiving a CONNACK is managed in 2 ways.
+ * This function will send MQTT CONNECT packet and receive a CONNACK packet. The
+ * send and receive from the network is done through the transport interface.
+ *
+ * The maximum time this function waits for a CONNACK is decided in one of the
+ * following ways:
  * 1. If #timeoutMs is greater than 0:
- *    Parameter #timeoutMs will be used for a #getTime based maximum timeout
- *    for the network read for CONNACK packet.
+ *    #getTime is used to ensure that the function does not wait more than #timeoutMs
+ *    for CONNACK.
  * 2. If #timeoutMs is 0:
- *    A network read would be retried upto the number of times configured by
- *    #MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT for the CONNACK packet.
+ *    The network receive for CONNACK is retried up to the number of times configured
+ *    by #MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT.
  *
  * @param[in] pContext Initialized MQTT context.
- * @param[in] pConnectInfo MQTT CONNECT packet parameters.
- * @param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
- * @param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
- * packet.
- * @note A zero timeout will make use of the retries for receiving
- * CONNACK.
- * @note The API may take more time than the given timeout in certain
- * conditions as listed below.
- * 1. Timeouts are incorrectly configured - If the timeout passed to this API
- * is less than the transport receive timeout and if a CONNACK packet is not
- * received within the transport receive timeout, the API will spend more time
- * than the timeoutMs parameter passed. The timeoutMs parameter passed to this
- * API should be greater than the transport receive timeout. Please refer to the
- * transport interface documentation for more details about timeout
- * configurations.
- * 2. CONNACK packet is received right after expiry of the timeout -
- * It is possible the CONNACK packet is received towards the expiry of the
- * timeout passed to the API. In the worst case, it can take upto
- * timeoutMs + transport receive timeout to receive the packet type and
- * remaining length of CONNACK packet from network. The API will attempt to
- * read the remaining 2 bytes after reading the packet type and remaining
- * length. In this case, API will spend more time than timeoutMs for receiving
- * CONNACK.
+ * @param[in] pConnectInfo MQTT CONNECT packet information.
+ * @param[in] pWillInfo Last Will and Testament. Pass NULL if Last Will and
+ * Testament is not used.
+ * @param[in] timeoutMs Maximum time in milliseconds to wait for a CONNACK packet.
+ * A zero timeout makes use of the retries for receiving CONNACK as configured with
+ * #MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT .
  * @param[out] pSessionPresent Whether a previous session was present.
  * Only relevant if not establishing a clean session.
  *
@@ -199,6 +184,24 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
  * #MQTTNoDataAvailable if no data available to receive in transport until
  * the #timeoutMs for CONNACK;
  * #MQTTSuccess otherwise.
+ *
+ * @note This API may spend more time than provided in the timeoutMS parameters in
+ * certain conditions as listed below:
+ *
+ * 1. Timeouts are incorrectly configured - If the timeoutMS is less than the
+ *    transport receive timeout and if a CONNACK packet is not received within
+ *    the transport receive timeout, the API will spend the transport receive
+ *    timeout (which is more time than the timeoutMs). It is the case of incorrect
+ *    timeout configuration as the timeoutMs parameter passed to this API must be
+ *    greater than the transport receive timeout. Please refer to the transport
+ *    interface documentation for more details about timeout configurations.
+ *
+ * 2. Partial CONNACK packet is received right before the expiry of the timeout - It
+ *    is possible that first two bytes of CONNACK packet (packet type and remaining
+ *    length) are received right before the expiry of the timeoutMS. In that case,
+ *    the API makes one more network receive call in an attempt to receive the remaining
+ *    2 bytes. In the worst case, it can happen that the remaining 2 bytes are never
+ *    received and this API will end up spending timeoutMs + transport receive timeout.
  */
 MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
                            const MQTTConnectInfo_t * pConnectInfo,
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -1128,7 +1128,8 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
         }
         else
         {
-            breakFromLoop = ( loopCount++ >= MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT ) ? true : false;
+            breakFromLoop = ( loopCount >= MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT ) ? true : false;
+            loopCount++;
         }
 
         /* Loop until there is data to read or if we have exceeded the timeout/retries. */
@@ -1193,23 +1194,6 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
     if( status == MQTTSuccess )
     {
         LogInfo( ( "Received MQTT CONNACK successfully from broker." ) );
-    }
-    else if( status == MQTTNoDataAvailable )
-    {
-        if( useRetry == false )
-        {
-            LogError( ( "Failed to receive CONNACK within the timeout of %ums."
-                        " Increasing the timeout may help resolve this issue.",
-                        timeoutMs ) );
-        }
-        else
-        {
-            LogError( ( "Failed to receive CONNACK within the max number"
-                        " of retries of %u to read from network. Increasing the"
-                        " number of retries with config MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT"
-                        " may help resolve the issue.",
-                        MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT ) );
-        }
     }
     else
     {
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -1121,7 +1121,6 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
          *    A value of 0 for the config will try once to read CONNACK. */
         if( timeoutMs > 0U )
         {
-            /* Loop runs only once for 0 timeoutMs. */
             breakFromLoop = ( calculateElapsedTime( getTimeStamp(), entryTimeMs ) >= timeoutMs ) ? true : false;
         }
         else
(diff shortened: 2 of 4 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -948,14 +954,33 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+        }
+
+        /* Loop until there is data to read or if we have exceeded the timeout/retries. */
+    } while( ( status == MQTTNoDataAvailable ) && ( !breakFromLoop ) );
 
     if( status == MQTTSuccess )
     {
         /* Time taken in this function so far. */
-        timeTakenMs = calculateElapsedTime( getTimeStamp(), entryTimeMs );
+        if( getTimeStamp != NULL )
```
muneebahmed10, line 980: In the event of `getTimeStamp` being `NULL`, then `timeTakenMs` won't be updated and be 0, the condition `timeTakenMs < timeoutMs` will hold, and `remainingTimeMs` will be set to `timeoutMs`. Then `receivePacket` will be called with a nonzero remaining time, and will have the potential to be an infinite loop if no data can be received in `recvExact`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -948,14 +954,33 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+         *    Loop times out based on the maximum number of retries config
+         *    MQTT_CONNACK_RECEIVE_RETRY_COUNT. */
+        if( getTimeStamp != NULL )
+        {
+            /* Loop runs only once for 0 timeoutMs. */
+            breakFromLoop = ( bool ) ( calculateElapsedTime( getTimeStamp(), entryTimeMs ) >= timeoutMs );
+        }
+        else
+        {
+            breakFromLoop = ( bool ) ( loopCount++ >= MQTT_CONNACK_RECEIVE_RETRY_COUNT );
```
muneebahmed10, line 971: Maybe add a comment clarifying that the maximum number of iterations this loop can run is `MQTT_CONNACK_RECEIVE_RETRY_COUNT + 1` (1 try + number of retries)

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -948,14 +954,33 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
                                                       pIncomingPacket );
 
-        /* Loop until there is data to read or if the timeout has not expired. */
-    } while( ( status == MQTTNoDataAvailable ) &&
-             ( calculateElapsedTime( getTimeStamp(), entryTimeMs ) < timeoutMs ) );
+        /* The loop times out based on 2 conditions.
+         * 1. If a non-NULL getTime function is provided by the application:
+         *    Loop times out based on the timeout calculated by getTime()
+         *    function.
+         * 2. If a non-NULL getTime function is not provided by the application:
```
muneebahmed10, line 961:
```suggestion
         * 2. If a getTime function is not provided by the application:
```

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -626,6 +626,42 @@ void test_MQTT_Connect_receiveConnack( void )
     MQTT_DeserializeAck_ExpectAnyArgsAndReturn( MQTTBadResponse );
     status = MQTT_Connect( &mqttContext, &connectInfo, NULL, timeout, &sessionPresent );
     TEST_ASSERT_EQUAL_INT( MQTTBadResponse, status );
+
+    /* Same set ot tests with retries. MQTT_CONNACK_RECEIVE_RETRY_COUNT is 2*/
```
muneebahmed10, line 630:
```suggestion
    /* Same set of tests with retries. MQTT_CONNACK_RECEIVE_RETRY_COUNT is 2. */
```

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -141,11 +141,22 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * const pContext,
...
+ *    A network read would be retried upto the number of times configured by
+ *    #MQTT_CONNACK_RECEIVE_RETRY_COUNT.
+ *
  * @brief param[in] pContext Initialized MQTT context.
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
  * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
  * @brief param[in] timeoutMs Timeout in milliseconds for receiving
- * CONNACK packet.
+ * CONNACK packet. A zero timeout value will attempt to receive through
+ * transport receive once.
```
muneebahmed10, line 159:
```suggestion
 * CONNACK packet. A zero timeout value will attempt a single reception
 * from the transport interface.
```

libraries/standard/mqtt/src/mqtt.c
leegeth: I think the word retry speaks for itself. Anyway if you had to rethink, I think it makes sense to add comment regarding that.

libraries/standard/mqtt/src/mqtt.c
leegeth: I think you are mentioning about a case in which timeoutMs is non zero and getTime is NULL. Ideally the timeoutMs doesn't matter in that case, but as you have pointed out our logic may not be handling it. I will zero out timeoutMs if getTime is NULL

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -141,11 +141,22 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * const pContext,
 /**
  * @brief Establish a MQTT session.
```
aggarw13, line 142: (Nit) Established **_an_** MQTT session

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -246,7 +257,12 @@ MQTTStatus_t MQTT_Unsubscribe( MQTTContext_t * const pContext,
 MQTTStatus_t MQTT_Disconnect( MQTTContext_t * const pContext );
 
 /**
- * @brief Loop to receive packets from the transport interface.
+ * @brief Loop to receive packets from the transport interface. Handles keep
+ * alive.
+ *
+ * @note This function assumes that pContext->callbacks.getTime is defined.
```
aggarw13, line 263: (Minor) Suggest re-wording the text to mention that `getTime` is a requirement for this function. So, suggest something like:
"_This function requires a valid definition of pContext->callbacks.getTime function. It is required for supporting the keep-alive mechanism that sends ping requests at configured keep-alive intervals of idle connection time._"

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -26,6 +26,12 @@
 #include "mqtt_state.h"
 #include "private/mqtt_internal.h"
 
+
+/* Default value for the CONNACK receive retries. */
+#ifndef MQTT_CONNACK_RECEIVE_RETRY_COUNT
+    #define MQTT_CONNACK_RECEIVE_RETRY_COUNT    ( 5U )
```
aggarw13, line 32: Should this raise an **error** instead of using the default value? The main reason being that default values can go unnoticed by the user..

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -144,11 +158,34 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
  *
  * @param[in] pContext MQTT Connection context.
  * @param[in] pIncomingPacket Incoming packet.
+ * @param[in] manageKeepAlive Flag indicating if PINGRESPs should not be given
```
aggarw13, line 161: (Minor) To make the parameter more meaningful, I suggest renaming it to something like `notifyAppOfPingResponse`..

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -144,11 +158,34 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
...
+ * @return #MQTTRecvFailed if a network error occurs during reception;
+ * #MQTTSendFailed if a network error occurs while sending an ACK or PINGREQ;
+ * #MQTTBadResponse if an invalid packet is received;
+ * #MQTTKeepAliveTimeout if the server has not sent a PINGRESP before
+ * pContext->pingRespTimeoutMs milliseconds;
+ * #MQTTIllegalState if an incoming QoS 1/2 publish or ack causes an
+ * invalid transition for the internal state machine;
+ * #MQTTSuccess on success.
+ */
+static MQTTStatus_t receiveSingleIteration( MQTTContext_t * pContext,
```
aggarw13, line 186: (Minor) How about re-naming it as receive**Loop**SingleIteration?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -264,6 +305,13 @@ static uint32_t calculateElapsedTime( uint32_t later,
 
 /*-----------------------------------------------------------*/
 
+static uint32_t getTimeStampNoSystemTime()
```
aggarw13, line 308: Can this dummy function be avoided? It seems like a technical debt introduced for supporting existing private functions that require the `getTime` function...

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -665,28 +729,17 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * const pContext,
                 status = sendPublishAcks( pContext,
                                           packetIdentifier,
                                           &publishRecordState );
-
-                if( status == MQTTSuccess )
-                {
```
aggarw13, line 670: Was invocation of `pContext->callbacks.appCallback` removed for the buffer corruption issue from `sendPublishAcks`?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -630,18 +690,22 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * const pContext,
...
-                                       MQTTPacketInfo_t * pIncomingPacket )
+                                       MQTTPacketInfo_t * pIncomingPacket,
+                                       bool manageKeepAlive )
 {
     MQTTStatus_t status = MQTTBadResponse;
     MQTTPublishState_t publishRecordState = MQTTStateNull;
     uint16_t packetIdentifier;
     /* Need a dummy variable for MQTT_DeserializeAck(). */
-    bool sessionPresent = false;
+    bool sessionPresent = false, callAppCallback = true;
```
aggarw13, line 700: (Minor) The declaration of `callAppCallback` seems to be misplaced in relation to the comment about `MQTT_DeserializeAck`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -714,6 +758,81 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * const pContext,
...
+        if( status == MQTTSuccess )
+        {
+            /* Reset the status to indicate that we should not try to read
+             * a packet from the transport interface. */
+            status = MQTTNoDataAvailable;
+        }
+    }
+    else if( status != MQTTSuccess )
+    {
+        LogError( ( "Receiving incoming packet length failed. Status=%d",
```
aggarw13, line 802: Can you log error codes as strings if stringification utility is available for MQTT?

aggarw13, line 802: Should the error log message be something like "Failed to receive incoming packet type and length"? 
Are we sure that the network read was successful but the failure occurred in determining packet length and type?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -714,6 +758,81 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * const pContext,
...
+                    status ) );
+    }
+    else
+    {
+        /* Receive packet. Remaining time is recalculated before calling this
+         * function. */
+        status = receivePacket( pContext, incomingPacket, remainingTimeMs );
```
aggarw13, line 809: (It may be out of scope of this PR) Is there a reason that `incomingPacket` is passed by value instead of being passed as a pointer? 
Looking at the type [`MQTTPacketInfo_t`](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/mqtt/include/mqtt_lightweight.h#L245-L261), it seems like passing by value will involve a larger copy of bytes than just passing as a pointer..

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -714,6 +758,81 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * const pContext,
...
+        {
+            status = handleIncomingPublish( pContext, &incomingPacket );
+        }
+        else
+        {
+            status = handleIncomingAck( pContext, &incomingPacket, manageKeepAlive );
+        }
+    }
+
+    if( status == MQTTNoDataAvailable )
```
aggarw13, line 829: If `MQTTNoDataAvailable` is not passed to the user, then should it be a public-facing `MQTTStatus_t` type?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -830,14 +954,33 @@ static MQTTStatus_t receiveConnack( MQTTContext_t * const pContext,
...
+         * 1. If a non-NULL getTime function is provided by the application:
+         *    Loop times out based on the timeout calculated by getTime()
+         *    function.
+         * 2. If a non-NULL getTime function is not provided by the application:
+         *    Loop times out based on the maximum number of retries config
+         *    MQTT_CONNACK_RECEIVE_RETRY_COUNT. */
+        if( getTimeStamp != NULL )
+        {
+            /* Loop runs only once for 0 timeoutMs. */
+            breakFromLoop = ( bool ) ( calculateElapsedTime( getTimeStamp(), entryTimeMs ) >= timeoutMs );
```
aggarw13, line 967: Is an explicit cast to `bool` type required? If the toolchain is C99 compliant, wouldn't it auto resolve to bool type? For C89 compilers, it would generate a 1 or 0 where `bool` would be aliased for `int8_t`

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: I chose manageKeepAlive since I thought it was more meaningful to indicate whether or not keep alive is being handled by the library, rather than indicating if the app should be notified of pingresps. The latter implies the question "why shouldn't the app be notified?" - and the answer to which is that the library is managing keep alive. I can update the param description to better reflect this

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Latest commit in #990 removes this

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Yes, it's one of the possible return values of `MQTT_GetIncomingPacketTypeAndLength` in the lightweight API

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: The cast is not necessary for any compiler. This was entirely for MISRA, since it considers `1`s and `0`s obtained as a result of comparison/logical operators as "essentially boolean type" even in C89, where there is no boolean type. Note that this would still be an issue even if we didn't use the `bool` alias. However, this cast was another MISRA violation, so we ended up using a ternary `? true : false` at the end instead.

libraries/standard/mqtt/src/mqtt.c
leegeth: This config is not always used. It is only required if the API is called with 0 timeout value. We switch to retries only if we find it out at run time. I don't think raising an error would be a good idea for handling a run time behavior, hence used default value

libraries/standard/mqtt/src/mqtt.c
leegeth: Was handled in PR #990

libraries/standard/mqtt/src/mqtt.c
leegeth: Already handled as part of #990

leegeth: This comment is in the scope of PR#990

libraries/standard/mqtt/src/mqtt.c
leegeth: Yea, it is out of the scope of this PR, but I think it is a good point.

libraries/standard/mqtt/src/mqtt.c
leegeth: related to PR #990.

libraries/standard/mqtt/include/mqtt.h
leegeth: related to PR #990

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -951,10 +965,30 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
...
+        if( useRetry == false )
+        {
+            /* Loop runs only once for 0 timeoutMs. */
+            breakFromLoop = ( calculateElapsedTime( getTimeStamp(), entryTimeMs ) >= timeoutMs ) ? true : false;
+        }
+        else
+        {
+            breakFromLoop = ( loopCount >= MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT ) ? true : false;
+            loopCount++;
+        }
```
aggarw13, line 988: Could the `useRetry` flag be avoided with the following?
```suggestion
        /* If timeout is passed, use timeout value to determine whether we should retry reception of CONNACK. */
        if( timeout == 1u )
        {
            /* Loop runs only once for 0 timeoutMs. */
            breakFromLoop = ( calculateElapsedTime( getTimeStamp(), entryTimeMs ) >= timeoutMs ) ? true : false;
        }
        /* Use the configured retry count to determine if we should retry reception of CONNACK. */
        else
        {
            breakFromLoop = ( loopCount >= MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT ) ? true : false;
            loopCount++;
        }
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -951,10 +965,30 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
...
 
-        /* Loop until there is data to read or if the timeout has not expired. */
-    } while( ( status == MQTTNoDataAvailable ) && ( timeExpired != true ) );
+        /* Loop until there is data to read or if we have exceeded the timeout/retries. */
```
aggarw13, line 990: Should there be some delay or back off logic in between consecutive transport receive calls?

leegeth: There is already a delay as part of the transport interface configured by the application. So the transport read is not going to immediately return, but wait till the timeout is hit if there is no data to read. I think adding a back off and retry may not be required here as library retrying may be enough as CONNACK is going to be the only packet being received.

libraries/standard/mqtt/src/mqtt.c
leegeth: Yea. Thanks. Added it for an additional logic and later removed. Will make that fix.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -971,7 +1003,10 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
...
          * Attempt to read once even if the timeout has expired at this point.
          * Invoking receivePacket with remainingTime as 0 would attempt to
-         * recv from network once.*/
+         * recv from network once. If using retries, the remainder of the
+         * CONNACK packet is tried to be read only once. Reading once would be
+         * good as the packet type and remaining length was already read. Hence,
+         * the probability of the remaining 2 bytes available to read is very high. */
         if( pIncomingPacket->type == MQTT_PACKET_TYPE_CONNACK )
         {
             status = receivePacket( pContext,
```
aggarw13, line 1012: If the passed timeout duration has transpired (i.e. `remainingTimeMs` is 0), calling transport recv could add additional delay depending on the platform specific implementation of the recv function. which yields non-deterministic behavior across platforms.
Has this scenario been considered? The extra delay (beyond the total timeout duration) in execution time of the function wouldn't matter if timeout >> transport receive timeout value, however, in the case of the two timeout values being comparable (in scale), the extra delay of the function could be significant (relatively). 

May be it is not a problem, if we have the assumption that 2 extra bytes should have been received in the network buffer if the CONNACK packet type and remaining length were

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -971,7 +1003,10 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
         /* Reading the remainder of the packet by transport recv.
          * Attempt to read once even if the timeout has expired at this point.
```
aggarw13, line 1004: (Nit): "...even if the timeout **duration** has expired..."

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -814,6 +849,29 @@ void test_MQTT_Connect_happy_path()
     status = MQTT_Connect( &mqttContext, &connectInfo, &willInfo, timeout, &sessionPresent );
     TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
     TEST_ASSERT_EQUAL_INT( MQTTConnected, mqttContext.connectStatus );
+
+    /* CONNACK receive with retries. */
```
aggarw13, line 853: This seems to be the test case where the first call to transport receive is successful. If so, should the comment say "CONNACK receive without retry execution."?

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -690,6 +691,40 @@ void test_MQTT_Connect_receiveConnack( void )
...
+
+    /* Did not receive a CONNACK. */
+    incomingPacket.type = MQTT_PACKET_TYPE_PINGRESP;
+    incomingPacket.remainingLength = 0;
+    MQTT_GetIncomingPacketTypeAndLength_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_GetIncomingPacketTypeAndLength_ReturnThruPtr_pIncomingPacket( &incomingPacket );
+    status = MQTT_Connect( &mqttContext, &connectInfo, NULL, 0U, &sessionPresent );
+    TEST_ASSERT_EQUAL_INT( MQTTBadResponse, status );
+
+    /* Transport receive failure when receiving rest of packet. */
```
aggarw13, line 712: Does rest of packet represent the 2 bytes after remaining length? Is this test case for failure path in this call: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/3b2352962e761bd2edbd48890d4e932e979795fb/libraries/standard/mqtt/src/mqtt.c#L1012

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: The timeout is not meant to be the maximum time that the function will take, but rather the minimum time (unless there's an error). We document this in the parameter descriptions: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/mqtt/include/mqtt.h#L259

libraries/standard/mqtt/utest/mqtt_config.h
```diff
@@ -56,5 +56,14 @@ typedef uint8_t ** NetworkContext_t;
...
 #define MQTT_CLIENT_IDENTIFIER              "testclient"
 
+/**
+ * @brief Retry count for reading CONNACK from network.
+ *
+ * #MQTT_Connect() can be using retries. If timeout passed as 0 to MQTT_Connect(),
+ * retries are used to attempt to read from network. The maximum retry count is
+ * specified by this config.
+ */
+#define MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT    ( 2U )
```
muneebahmed10, line 66: Do we not need to update mqtt_config.h for the demos as well?

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -690,6 +691,40 @@ void test_MQTT_Connect_receiveConnack( void )
     MQTT_DeserializeAck_ExpectAnyArgsAndReturn( MQTTBadResponse );
     status = MQTT_Connect( &mqttContext, &connectInfo, NULL, timeout, &sessionPresent );
     TEST_ASSERT_EQUAL_INT( MQTTBadResponse, status );
+
+    /* Same set of tests with retries. MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT is 2*/
```
muneebahmed10, line 695: If it's the same set of tests, I think we should put them in a new function: `test_MQTT_Connect_receiveConnack_retries`

libraries/standard/mqtt/src/mqtt.c
leegeth: I think Muneeb's comment helped clear the first part of the question. I will take an action on me to update the documentation of MQTT_Connect() as well to match the above mentioned documentation.

For the second part, by the time we find that we could get a valid packet type and remaining length, the possibility is very high that we get the remainder of 2 bytes as well for connack.

libraries/standard/mqtt/utest/mqtt_utest.c
leegeth: What I meant is CONNACK receive using the retry logic. I think it is confusing. Will update the comments. Thanks

libraries/standard/mqtt/utest/mqtt_utest.c
leegeth: yes correct.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -951,10 +965,28 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
...
+         *    Loop times out based on the timeout calculated by getTime()
+         *    function.
+         * 2. If timeoutMs is 0:
+         *    Loop times out based on the maximum number of retries config
+         *    MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT. This config will control
+         *    maximum the number of retry attempts to read the CONNACK packet.
+         *    A value of 0 for the config will try once to read CONNACK. */
+        if( timeoutMs > 0U )
+        {
+            /* Loop runs only once for 0 timeoutMs. */
```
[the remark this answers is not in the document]
leegeth: ?

libraries/standard/mqtt/utest/mqtt_config.h
leegeth: No, we don't. We need this config only for demos which are using the MQTT library without time dependency.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
muneebahmed10: Line 979 does not match lines 972-976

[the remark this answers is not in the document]
leegeth: oh yea. Thanks

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -692,6 +693,66 @@ void test_MQTT_Connect_receiveConnack( void )
...
+
+    memset( ( void * ) &mqttContext, 0x0, sizeof( mqttContext ) );
+    MQTT_Init( &mqttContext, &transport, &callbacks, &networkBuffer );
+
+    /* Everything before receiving the CONNACK should succeed. */
+    MQTT_SerializeConnect_IgnoreAndReturn( MQTTSuccess );
+    MQTT_GetConnectPacketSize_IgnoreAndReturn( MQTTSuccess );
+
+    /* Test with retries. MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT is 2*/
+    /* Nothing received from transport interface. */
```
muneebahmed10, line 723: Minor:
```suggestion
    /* Test with retries. MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT is 2.
     * Nothing received from transport interface. */
```

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -160,9 +160,14 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
...
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
  * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
- * @brief param[in] timeoutMs Timeout in milliseconds for receiving
- * CONNACK packet. A zero timeout value will attempt a single reception
- * from the transport interface.
+ * @brief param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
+ * packet. In certain conditions, depending on the network conditions or the
+ * transport interface receive timeout configured, this function can spend more
+ * time in receiving CONNACK packet as compared to the timeout provided. However
+ * the probability of exceeding the timeout is very low assuming the the timeouts
```
aggarw13, line 167: (Nit): the ~the~

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -160,9 +160,14 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
...
+ * for CONNACK receive and transport interface receive are configured correctly.
+ * Refer to the transport interface documentation for more details about the
+ * timeouts. A zero timeout will make use of the retries for receiving CONNACK.
```
aggarw13, line 170: Maybe a `@note` tag for the last sentence can be easier to notice about use of retries for the zero timeout value.

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -160,9 +160,14 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
  * @brief param[in] pContext Initialized MQTT context.
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
  * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
- * @brief param[in] timeoutMs Timeout in milliseconds for receiving
- * CONNACK packet. A zero timeout value will attempt a single reception
- * from the transport interface.
+ * @brief param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
+ * packet. In certain conditions, depending on the network conditions or the
+ * transport interface receive timeout configured, this function can spend more
+ * time in receiving CONNACK packet as compared to the timeout provided. However
```
aggarw13, line 166: (Nit): "..more time in receiving CONNACK packet ~~as compared to the~~ than the provided timeout duration. However, (comma)...."

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -26,6 +26,21 @@
 #include "mqtt_state.h"
 #include "private/mqtt_internal.h"
 
+
+/**
```
aggarg, line 30:
```
/**
 * @brief The number of retries for receiving CONNACK.
 *
 * The MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT will be used only when the
 * timeoutMs parameter of #MQTT_Connect() is passed as 0 . The transport receive
 * for CONNACK will be retried MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT
 * times before timing out. A value of 0 for this config will cause the transport receive
 * for CONNACK  to be invoked only once.
 */
```

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -814,6 +875,29 @@ void test_MQTT_Connect_happy_path()
     status = MQTT_Connect( &mqttContext, &connectInfo, &willInfo, timeout, &sessionPresent );
     TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
     TEST_ASSERT_EQUAL_INT( MQTTConnected, mqttContext.connectStatus );
+
+    /* CONNACK receive with timeoutMs=0. Retry logic will be used. */
```
aggarg, line 879: Do we need a test case where retry logic is not used?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -969,9 +1000,12 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
         }
 
         /* Reading the remainder of the packet by transport recv.
-         * Attempt to read once even if the timeout has expired at this point.
+         * Attempt to read once even if the timeout has expired.
          * Invoking receivePacket with remainingTime as 0 would attempt to
-         * recv from network once.*/
+         * recv from network once. If using retries, the remainder of the
```
aggarg, line 1005: We do not re-try even if retries are left?

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -145,13 +145,30 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
...
+ * 2. If #timeoutMs is 0:
+ *    A network read would be retried upto the number of times configured by
+ *    #MQTT_MAX_CONNACK_RECEIVE_RETRY_COUNT for the CONNACK packet.
  *
  * @brief param[in] pContext Initialized MQTT context.
  * @brief param[in] pConnectInfo MQTT CONNECT packet parameters.
  * @brief param[in] pWillInfo Last Will and Testament. Pass NULL if not used.
- * @brief param[in] timeoutMs Timeout in milliseconds for receiving
- * CONNACK packet.
+ * @brief param[in] timeoutMs Timeout in milliseconds for receiving CONNACK
```
aggarg, line 163: This explanation is not clear and hand-wavy. In what conditions this can spend more time? What is correct configuration? What is the low probability condition?

libraries/standard/mqtt/utest/mqtt_utest.c
muneebahmed10: The test cases above this do not use retry logic, since it's the value of the timeout that determines this

libraries/standard/mqtt/include/mqtt.h
muneebahmed10: Also, I think it's a little odd this is tagged with a `@brief`, when the description is anything but

libraries/standard/mqtt/src/mqtt.c
leegeth: Thanks

libraries/standard/mqtt/src/mqtt.c
leegeth: No. We don't. After reading the packet type and length, it is the remaining 2 bytes that we would need to read. The delay is required to wait for the reception of CONNACK packet after sending a CONNECT packet. Once CONNACK packet type and length is available, we would need a follow up read just to read the remaining 2 bytes.  Also in this case, we would be expecting only one packet from broker.

libraries/standard/mqtt/include/mqtt.h
leegeth: Ok Thanks. Will reword it.

PR #995, opened 2020-06-11
Reconnect Logic

abhidixi11:
*Description of changes:*

Implementation of Reconnect Logic for POSIX platform.  This PR implements sample code for exponential backoff and reconnect when a connection attempt to a server fails. It also modifies Plain Text demo to use the reconnect logic. 


platform/platformFilePaths.cmake
```diff
@@ -0,0 +1,12 @@
...
+# in their Cmake based build system by including this file.
+
+# PLATFORM library source files.
+set( PLATFORM_SOURCES
+     "${PLATFORM_DIR}/posix/src/reconnect_posix.c" )
+
+# PLATFORM library Public Include directories.
+set( PLATFORM_INCLUDE_PUBLIC_DIRS
+     "${PLATFORM_DIR}/include"
+     "${PLATFORM_DIR}/*/include" )
```
yourslab, line 12: Overall, looks good but not sure if we wanna include all header files from `${PLATFORM_DIR}/*/include`. For example, we probably don't want to add POSIX includes for AFR. Maybe we can have separate `CMakeLists.txt` for each subdirectory and possibly separate target for each that is selectively linked to main platform target.

platform/posix/src/reconnect_posix.c
```diff
@@ -0,0 +1,65 @@
...
+ * @file reconnect_posix.c
+ * @brief Implementation of the backoff logic when connection fails to the server fails.
+ */
+
+/* Standard includes. */
+#include <unistd.h>
+#include "reconnect.h"
+
+static uint32_t reconnectTimeoutSec = RECONNECT_INITIAL_TIMEOUT_SECONDS;
+static uint32_t attemptsDone = 0;
```
yourslab, line 32: How come above is static global? Maybe we can just declare inside each function.

demos/mqtt/mqtt_demo_plaintext/mqtt_demo_plaintext.c
```diff
@@ -423,6 +438,43 @@ static int connectToServer( const char * pServer,
...
+     * a timeout. Timeout value will exponentially increase till maximum
+     * attemps are reached.
+     *
+     *  while( ( EXIT_FAILURE == connectToServer( BROKER_ENDPOINT, BROKER_PORT, pTcpSocket ) ) &&
+     *  ( true == reconnectBackoffAndSleep() ) ) {}
+     */
+
+    do
+    {
+        status = connectToServer( BROKER_ENDPOINT, BROKER_PORT, pTcpSocket );
```
yourslab, line 459: We can possibly have multiple `connectToServer` functions, so we can probably have an interface for this.

abhidixi11: In past we discussed to have an interface using function pointers, but during the design review, it was discouraged. connectToServerWithBackoffRetries() will be implemented by the application using reconnectBackoffReset() and reconnectBackoffAndSleep()

platform/platformFilePaths.cmake
abhidixi11: Yes, this was make shift cmake file. During implementation we can change it as needed.

platform/include/reconnect.h
```diff
@@ -0,0 +1,77 @@
...
+/* @brief Max number of connect attempts, set this value to 0 if the client
+ * must try connecting to the server forever */
+#define MAX_RECONNECT_ATTEMPS                4U
+/* @brief Initial fixed timeout value in seconds between two successive
+ connects. A random jitter value is added to every timeout value  */
+#define RECONNECT_INITIAL_TIMEOUT_SECONDS    1U
+/* @brief Max timout value in seconds */
+#define MAX_RECONNECT_TIMEOUT                32U
+/* @brief Max jitter value in seconds */
+#define MAX_JITTER_VALUE_SECONDS             5U
```
jainamt, line 47: If a customer has a large fleet of devices, we will expect them to modify these values, right? How would a customer know what value to choose for their fleet? 

And how would a customer discover that these values need to be updated for their fleet.

abhidixi11: Good point, we will have these values from the existing SDKs as default.  ( I'll update those ) This example code is making sure that the customers do use some kind of exponential backoff mechanism along with jitter for re-connection, but if required they will have to fine tune it based on their use case.

abhidixi11: I have updated the max value ( from 32 to 128 seconds) as per V3 CSDK. 
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/d039f075e1cc2a2a7fc20edc6868f328d8d36b2f/samples/linux/subscribe_publish_library_sample/aws_iot_config.h#L53

platform/include/reconnect.h
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * @brief Simple platfrom specific exponential backoff function. The application
+ * must use this function between connection failures to add exponential delay.
+ * This function will block the calling task for the current timeout value.
+ *
+ * @param[in, out] reconnectParam structure containing reconnection parameters.
+ *
+ * @return true after successful sleep, false when all attempts are exhausted.
+ */
+bool Transport_reconnectBackoffAndSleep( TransportReconnectParams_t * reconnectParams );
```
yourslab, line 78: Nit: I think it should be Transport_ReconnectBackoffAndSleep where the first letter of the function is capitalized to be consistent with naming convention.

platform/posix/transport/CMakeLists.txt
```diff
@@ -25,3 +25,12 @@ add_library( openssl_posix
...
                            sockets_posix )
+
+# Create target for POSIX implementation of reconnect logic.
+add_library( reconnect_posix 
+               ${RECONNECT_SOURCES} )
+
+target_include_directories( reconnect_posix 
+                              PUBLIC 
+                                ${COMMON_TRANSPORT_INCLUDE_PUBLIC_DIRS}
+                                ${LOGGING_INCLUDE_DIRS} )
```
yourslab, line 36: Please add newline

platform/posix/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# Include filepaths for source and include.
+# include( "../platformFilePaths.cmake" )
+
+# PLATFORM library target.
+# add_library( platform
+#             ${PLATFORM_SOURCES} )
```
yourslab, line 6: Please remove file if not used

platform/include/reconnect.h
yourslab: Or maybe we can just have it as `Reconnect_BackoffAndSleep`

platform/posix/CMakeLists.txt
abhidixi11: Hmm.. This file is most likely not needed.  I'll get rid of it.

PR #998, opened 2020-06-15
Fix buffer corruption for incoming publishes

leegeth:
*Issue #, if available:*
Network buffer gets corrupted when invoking an application callback for incoming publishes.

*Description of changes:*
Root cause - Publish acks uses the same buffer as the incoming publishes. If a publish ack is sent before the incoming publish is processed by the application, new serialized data for acks will be overwritten in the same buffer. Incoming publish will thus be corrupted.

Fix - Fix is to invoke application callback before attempting to send acks.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -293,6 +293,7 @@ static MQTTPubAckType_t getAckFromPacketType( uint8_t packetType )
 
         case MQTT_PACKET_TYPE_PUBCOMP:
         default:
+
             /* This function is only called after checking the type is one of
              * the above four values, so packet type must be PUBCOMP here. */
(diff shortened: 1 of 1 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -293,6 +293,7 @@ static MQTTPubAckType_t getAckFromPacketType( uint8_t packetType )
 
         case MQTT_PACKET_TYPE_PUBCOMP:
         default:
+
             /* This function is only called after checking the type is one of
              * the above four values, so packet type must be PUBCOMP here. */
             assert( packetType == MQTT_PACKET_TYPE_PUBCOMP );
```
aggarw13, line 299: Is this check redundant considering that there exists the case matching for `MQTT_PACKET_TYPE_PUBCOMP`  just before the default?

PR #999, opened 2020-06-15
Rename MQTTNetworkContext_t to NetworkContext_t

muneebahmed10:
*Description of changes:*
Renames instances of `MQTTNetworkContext_t` to `NetworkContext_t`.

### DIFF
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -38,7 +38,7 @@ typedef struct MQTTContext                MQTTContext_t;
 struct MQTTTransportInterface;
 typedef struct MQTTTransportInterface     MQTTTransportInterface_t;
 
-typedef int32_t (* MQTTTransportSendFunc_t )( MQTTNetworkContext_t context,
+typedef int32_t (* MQTTTransportSendFunc_t )( NetworkContext_t context,
                                               const void * pMessage,
                                               size_t bytesToSend );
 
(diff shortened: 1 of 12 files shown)


libraries/standard/mqtt/include/mqtt.h
```diff
@@ -38,7 +38,7 @@ typedef struct MQTTContext                MQTTContext_t;
 struct MQTTTransportInterface;
 typedef struct MQTTTransportInterface     MQTTTransportInterface_t;
 
-typedef int32_t (* MQTTTransportSendFunc_t )( MQTTNetworkContext_t context,
```
aggarw13, line 41: The `http_client.h` file defines `NetworkContext_t` as `typedef struct NetworkContext * NetWorkContext_t` . 
Can you update this to be consistent with HTTP?

muneebahmed10: I don't see why we need to add a `typedef` where one's not already present, and why it needs to be `typedef`ed to a struct within the library itself. Currently we `typedef` the network context in the `mqtt_config.h` file, and if a struct is desired then it can be configured that way.

aggarw13: I am fine with not explicity typedef'ng the `NetworkContext_t`. I was primarily pointing at the inconsistency between the libraries. We can make HTTP not typedef it instead for the network context.
Besides, can you add a comment to mention (for improved documentation) that the type needs to be user-defined?

abhidixi11: It will be addressed in a PR where we are writing common transport utility functions. We will have a common header file declaring transport interface for both HTTP and MQTT.  Documentation should be part of that PR.

PR #1007, opened 2020-06-22
MQTT - Handle resends

leegeth:
*Description of changes:*
Handling the resend and receive of duplicate packets when an MQTT session is reestablished.

These are the cases that is handled in this PR,

When a session is reestablished,
Outgoing publishes:

Resend PUBLISHes for which and ack is not received(PUBACK and PUBREC).
Resend PUBRELs for PUBLISHes for which PUBCOMP was not received.
Incoming publishes:

Handle duplicate QoS1 and QoS2 PUBLISHes
Handle duplicate PUBRELs.

This change also handles the message ordering requirement in MQTT3.1.1 spec while re sending publishes and PUBRELs.

### DIFF
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -132,7 +132,7 @@ struct MQTTContext
  * @brief Initialize an MQTT context.
  *
  * This function must be called on an MQTT context before any other function.
- * 
+ *
  * @note The getTime callback function must be defined. If there is no time
  * implementation, it is the responsibility of the application to provide a
  * dummy function to always return 0, and provide 0 timeouts for functions. This
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -26,6 +26,13 @@
 #include "mqtt_config.h"
 #include "mqtt_lightweight.h"
 
+/**
+ * @brief Invalid packet identifier.
+ *
+ * Zero is an invalid packet identifier as per MQTT v3.1.1 spec.
+ */
+#define MQTT_PACKET_ID_INVALID    ( ( uint16_t ) 0U )
+
 struct MQTTApplicationCallbacks;
 typedef struct MQTTApplicationCallbacks   MQTTApplicationCallbacks_t;
 
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -118,7 +118,7 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
  * This function will need to be called to get the packet for which a PUBREL
  * need to be sent when a session is reestablished. Calling this function
  * repeatedly until packet id is 0 will give all the packets for which
- * a PUBREL need to be resent in the order.
+ * a PUBREL need to be resent in the correct order.
  *
  * @param[in] pMqttContext Initialized MQTT context.
  * @param[in,out] pCursor Index at which to start searching.
@@ -134,7 +134,7 @@ uint16_t MQTT_PubrelToResend( const MQTTContext_t * pMqttContext,
  * This function will need to be called to get the packet for which a publish
  * need to be sent when a session is reestablished. Calling this function
  * repeatedly until packet id is 0 will give all the packets for which
- * a publish need to be resent in the order.
+ * a publish need to be resent in the correct order.
  *
  * @param[in] pMqttContext Initialized MQTT context.
  * @param[in,out] pCursor Index at which to start searching.
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -113,23 +113,12 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
                                   MQTTPublishState_t * pNewState );
 
 /**
- * @brief Get the packet ID and index of a publish in a specified state.
+ * @brief Get the packet ID of next pending PUBREL ack to be resent.
  *
- * @param[in] pMqttContext Initialized MQTT context.
- * @param[in] searchState The state to search for.
- * @param[in,out] pCursor Index at which to start searching.
- */
-uint16_t MQTT_StateSelect( const MQTTContext_t * pMqttContext,
-                           MQTTPublishState_t searchState,
-                           MQTTStateCursor_t * pCursor );
-
-/**
- * @brief Get the packet ID of next pending PUBREL ack to be resend.
- *
- * This function will be called to get the packet for which a PUBREL
- * need to send when a session is reestablished. Calling this function
+ * This function will need to be called to get the packet for which a PUBREL
+ * need to be sent when a session is reestablished. Calling this function
  * repeatedly until packet id is 0 will give all the packets for which
- * a PUBREL need to be resent.
+ * a PUBREL need to be resent in the order.
  *
  * @param[in] pMqttContext Initialized MQTT context.
  * @param[in,out] pCursor Index at which to start searching.
@@ -138,6 +127,20 @@ uint16_t MQTT_StateSelect( const MQTTContext_t * pMqttContext,
 uint16_t MQTT_AckToResend( const MQTTContext_t * pMqttContext,
                            MQTTStateCursor_t * pCursor,
                            MQTTPublishState_t * pState );
+
+/**
+ * @brief Get the packet ID of next pending publish to be resent.
+ *
+ * This function will need to be called to get the packet for which a publish
+ * need to be sent when a session is reestablished. Calling this function
+ * repeatedly until packet id is 0 will give all the packets for which
+ * a publish need to be resent in the order.
+ *
+ * @param[in] pMqttContext Initialized MQTT context.
+ * @param[in,out] pCursor Index at which to start searching.
+ */
+uint16_t MQTT_PublishToResend( const MQTTContext_t * pMqttContext,
+                               MQTTStateCursor_t * pCursor );
 
 /**
  * @brief State to string conversion for state engine.
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -124,6 +124,22 @@ uint16_t MQTT_StateSelect( const MQTTContext_t * pMqttContext,
                            MQTTStateCursor_t * pCursor );
 
 /**
+ * @brief Get the packet ID of next pending PUBREL ack to be resend.
+ *
+ * This function will be called to get the packet for which a PUBREL
+ * need to send when a session is reestablished. Calling this function
+ * repeatedly until packet id is 0 will give all the packets for which
+ * a PUBREL need to be resent.
+ *
+ * @param[in] pMqttContext Initialized MQTT context.
+ * @param[in,out] pCursor Index at which to start searching.
+ * @param[out] pState State indicating that PUBREL packet need to be sent.
+ */
+uint16_t MQTT_AckToResend( const MQTTContext_t * pMqttContext,
+                           MQTTStateCursor_t * pCursor,
+                           MQTTPublishState_t * pState );
+
+/**
  * @brief State to string conversion for state engine.
  *
  * @param[in] state The state to convert to a string.
--- a/libraries/standard/mqtt/include/mqtt_state.h
+++ b/libraries/standard/mqtt/include/mqtt_state.h
@@ -124,9 +124,9 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
  * @param[in,out] pCursor Index at which to start searching.
  * @param[out] pState State indicating that PUBREL packet need to be sent.
  */
-uint16_t MQTT_AckToResend( const MQTTContext_t * pMqttContext,
-                           MQTTStateCursor_t * pCursor,
-                           MQTTPublishState_t * pState );
+uint16_t MQTT_PubrelToResend( const MQTTContext_t * pMqttContext,
+                              MQTTStateCursor_t * pCursor,
+                              MQTTPublishState_t * pState );
 
 /**
  * @brief Get the packet ID of next pending publish to be resent.
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -116,6 +116,19 @@ static MQTTStatus_t receivePacket( const MQTTContext_t * pContext,
 static uint8_t getAckTypeToSend( MQTTPublishState_t state );
 
 /**
+ * @brief Update state records after publish acks are sent.
+ *
+ * @param[in] pContext MQTT Connection context.
+ * @param[in] packetId packet ID of original PUBLISH.
+ * @param[in] packetType Packet type of the ack packet sent.
+ *
+ * @return #MQTTSuccess or #MQTTIllegalState.
+ */
+static MQTTStatus_t updateAckSentState( MQTTContext_t * pContext,
+                                        uint16_t packetId,
+                                        MQTTPubAckType_t packetType );
+
+/**
  * @brief Send acks for received QoS 1/2 publishes.
  *
  * @param[in] pContext MQTT Connection context.
@@ -148,6 +161,17 @@ static MQTTStatus_t handleKeepAlive( MQTTContext_t * pContext );
  */
 static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
                                            MQTTPacketInfo_t * pIncomingPacket );
+
+/**
+ * @brief Handle received MQTT publish acks.
+ *
+ * @param[in] pContext MQTT Connection context.
+ * @param[in] pIncomingPacket Incoming packet.
+ *
+ * @return MQTTSuccess, MQTTIllegalState, or deserialization error.
+ */
+static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
+                                       MQTTPacketInfo_t * pIncomingPacket );
 
 /**
  * @brief Handle received MQTT ack.
@@ -241,6 +265,37 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
  * #MQTTSuccess otherwise.
  */
 static MQTTStatus_t resendPendingAcks( MQTTContext_t * const pContext );
+
+/**
+ * @brief Serializes a PUBLISH message.
+ *
+ * @brief param[in] pContext Initialized MQTT context.
+ * @brief param[in] pPublishInfo MQTT PUBLISH packet parameters.
+ * @brief param[in] packetId Packet Id of the publish packet.
+ * @brief param[out] pHeaderSize Size of the serialized PUBLISH header.
+ *
+ * @return #MQTTNoMemory if pBuffer is too small to hold the MQTT packet;
+ * #MQTTBadParameter if invalid parameters are passed;
+ * #MQTTSuccess otherwise.
+ */
+static MQTTStatus_t serializePublish( MQTTContext_t * const pContext,
+                                      const MQTTPublishInfo_t * const pPublishInfo,
+                                      uint16_t packetId,
+                                      size_t * const pHeaderSize );
+
+/**
+ * @brief Function to validate #MQTT_Publish parameters.
+ *
+ * @brief param[in] pContext Initialized MQTT context.
+ * @brief param[in] pPublishInfo MQTT PUBLISH packet parameters.
+ * @brief param[in] packetId Packet Id for the MQTT PUBLISH packet.
+ *
+ * @return #MQTTBadParameter if invalid parameters are passed;
+ * #MQTTSuccess otherwise.
+ */
+static MQTTStatus_t validatePublishParams( MQTTContext_t * const pContext,
+                                           const MQTTPublishInfo_t * const pPublishInfo,
+                                           uint16_t packetId );
 
 /*-----------------------------------------------------------*/
 
@@ -543,6 +598,44 @@ static uint8_t getAckTypeToSend( MQTTPublishState_t state )
 
 /*-----------------------------------------------------------*/
 
+static MQTTStatus_t updateAckSentState( MQTTContext_t * pContext,
+                                        uint16_t packetId,
+                                        MQTTPubAckType_t packetType )
+{
+    MQTTStatus_t status = MQTTSuccess;
+    MQTTPublishState_t newState = MQTTStateNull;
+
+    assert( pContext != NULL );
+    assert( packetId != MQTT_PACKET_ID_INVALID );
+
+    status = MQTT_UpdateStateAck( pContext,
+                                  packetId,
+                                  packetType,
+                                  MQTT_SEND,
+                                  &newState );
+
+    /* A PUBCOMP packet can be sent without having a state present
+     * in state engine. This happens when an ack is sent for an
+     * incoming PUBREL for a packet without a stored record. */
+    if( ( status == MQTTRecordNotPresent ) &&
+        ( packetType == MQTTPubcomp ) )
+    {
+        status = MQTTSuccess;
+        LogDebug( ( "Sent PUBCOMP for packet id %u which did not have"
+                    " a stored record.",
+                    packetId ) );
+    }
+
+    if( status != MQTTSuccess )
+    {
+        LogError( ( "Failed to update state of publish %u.", packetId ) );
+    }
+
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
 static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
                                      uint16_t packetId,
                                      MQTTPublishState_t publishState )
@@ -575,28 +668,7 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
         if( bytesSent == ( int32_t ) MQTT_PUBLISH_ACK_PACKET_SIZE )
         {
             pContext->controlPacketSent = true;
-            status = MQTT_UpdateStateAck( pContext,
-                                          packetId,
-                                          packetType,
-                                          MQTT_SEND,
-                                          &newState );
-
-            /* A PUBCOMP packet can be sent without having a state present
-             * in state engine. This happens when an ack is sent for an
-             * incoming PUBREL for a packet without a stored record. */
-            if( ( status == MQTTStateNotPresent ) &&
-                ( packetType == MQTTPubcomp ) )
-            {
-                status = MQTTSuccess;
-                LogDebug( ( "Sent PUBCOMP for packet id %u which did not have"
-                            " a stored record.",
-                            packetId ) );
-            }
-
-            if( status != MQTTSuccess )
-            {
-                LogError( ( "Failed to update state of publish %u.", packetId ) );
-            }
+            status = updateAckSentState( pContext, packetId, packetType );
         }
         else
         {
@@ -720,6 +792,79 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
 
 /*-----------------------------------------------------------*/
 
+static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
+                                       MQTTPacketInfo_t * pIncomingPacket )
+{
+    MQTTStatus_t status = MQTTBadResponse;
+    MQTTPublishState_t publishRecordState = MQTTStateNull;
+    uint16_t packetIdentifier;
+    MQTTPubAckType_t ackType;
+    MQTTEventCallback_t appCallback;
+
+    assert( pContext != NULL );
+    assert( pIncomingPacket != NULL );
+    assert( pContext->callbacks.appCallback != NULL );
+
+    appCallback = pContext->callbacks.appCallback;
+
+    ackType = getAckFromPacketType( pIncomingPacket->type );
+    status = MQTT_DeserializeAck( pIncomingPacket, &packetIdentifier, NULL );
+    LogInfo( ( "Ack packet deserialized with result: %s.",
+               MQTT_Status_strerror( status ) ) );
+
+    if( status == MQTTSuccess )
+    {
+        status = MQTT_UpdateStateAck( pContext,
+                                      packetIdentifier,
+                                      ackType,
+                                      MQTT_RECEIVE,
+                                      &publishRecordState );
+
+        if( status == MQTTSuccess )
+        {
+            LogInfo( ( "State record updated. New state=%d.",
+                       publishRecordState ) );
+        }
+
+        /* PUBREL is received and there is no state record exists.
+         * This happens when PUBCOMP was successfully sent from client
+         * and broker failed to receive it due to a connection failure. */
+        else if( ( status == MQTTRecordNotPresent ) &&
+                 ( pIncomingPacket->type == MQTT_PACKET_TYPE_PUBREL ) )
+        {
+            status = MQTTSuccess;
+
+            /* Calculate the state ack for PUBCOMP to be sent. */
+            publishRecordState = MQTT_CalculateStateAck( ackType,
+                                                         MQTT_RECEIVE,
+                                                         MQTTQoS2 );
+        }
+        else
+        {
+            LogError( ( "Updating the state engine for packet id %u"
+                        " failed with error %s.",
+                        packetIdentifier,
+                        MQTT_Status_strerror( status ) ) );
+        }
+    }
+
+    if( status == MQTTSuccess )
+    {
+        /* Invoke application callback to hand the buffer over to application
+         * before sending acks. */
+        appCallback( pContext, pIncomingPacket, packetIdentifier, NULL );
+
+        /* Send PUBREL or PUBCOMP if necessary. */
+        status = sendPublishAcks( pContext,
+                                  packetIdentifier,
+                                  publishRecordState );
+    }
+
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
 static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
                                        MQTTPacketInfo_t * pIncomingPacket,
                                        bool manageKeepAlive )
@@ -736,7 +881,6 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
      * sending any PUBREL or PUBCOMP. However, for other cases, we invoke it
      * at the end to reduce the complexity of this function. */
     bool invokeAppCallback = false;
-    MQTTPubAckType_t ackType;
     MQTTEventCallback_t appCallback = NULL;
     bool pubrelRecordMissing = false;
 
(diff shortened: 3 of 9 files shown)


libraries/standard/mqtt/include/mqtt_state.h
```diff
@@ -113,15 +113,34 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
...
+ * a publish need to be resent in the order.
  *
  * @param[in] pMqttContext Initialized MQTT context.
- * @param[in] searchState The state to search for.
  * @param[in,out] pCursor Index at which to start searching.
  */
-uint16_t MQTT_StateSelect( const MQTTContext_t * pMqttContext,
-                           MQTTPublishState_t searchState,
-                           MQTTStateCursor_t * pCursor );
+uint16_t MQTT_PublishToResend( const MQTTContext_t * pMqttContext,
```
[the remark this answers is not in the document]
leegeth: Yes.

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -26,6 +26,13 @@
 #include "mqtt_config.h"
 #include "mqtt_lightweight.h"
 
+/**
+ * @brief Invalid packet identifier.
+ *
```
sarenameas, line 31: I apologize, I checked in a file with an extra space left in mqtt.h, could you rebase uncrustify this file?

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -323,23 +544,246 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
+    uint16_t packetId;
+
+    assert( records != NULL );
+
     if( shouldDelete == true )
     {
+        /* Mark the record as invalid. */
         records[ recordIndex ].packetId = MQTT_PACKET_ID_INVALID;
-        records[ recordIndex ].qos = MQTTQoS0;
-        records[ recordIndex ].publishState = MQTTStateNull;
```
muneebahmed10, line 335: What's wrong with keeping this? I think it makes the records cleaner

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -323,23 +544,246 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
+    /* Only outgoing publish records need to be searched. */
+    assert( ( outgoingStates & searchStates ) > 0U );
+    records = pMqttContext->outgoingPublishRecords;
+
+    while( *pCursor < MQTT_STATE_ARRAY_MAX_COUNT )
+    {
+        /* Check if any of the search states are present. */
+        stateCheck = UINT16_CHECK_BIT( searchStates, records[ *pCursor ].publishState ) ? true : false;
+
+        if( stateCheck == true )
```
muneebahmed10, line 621: I think this could just be:
```suggestion
        if( UINT16_CHECK_BIT( searchStates, records[ *pCursor ].publishState ) )
```

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -323,23 +544,246 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
+                                packetId,
+                                qos,
+                                newState );
+        }
+        /* Send operation. */
+        else
+        {
+            /* Skip updating record when publish is resend and no state
+             * update is required. */
+            if( currentState != newState )
```
muneebahmed10, line 767: Is this necessary? Even if the function is called it will only update to itself

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -323,23 +544,246 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
-        records[ recordIndex ].publishState = MQTTStateNull;
     }
     else
     {
-        records[ recordIndex ].publishState = newState;
+        /* For QoS2 messages, in order to preserve the message ordering, when
+         * a PUBREC is received for an outgoing publish, the record should be
+         * moved to the last. This move will help preserve the order in which
+         * a PUBREL needs to be resent in case of a session reestablishment. */
+        if( newState == MQTTPubRelSend )
```
muneebahmed10, line 569: What if instead of adding this logic here, add it to updateStateAck? That would keep this function simple and straightforward and move the reordering logic up to a higher level.
```
shouldDeleteRecord = ( ( newState == MQTTPublishDone ) || ( newState == MQTTPubRelSend ) ) ? true : false;
...
if( isTransitionValid == true )
{
    updateRecord( records, recordIndex, newState, shouldDeleteRecord );
    status = MQTTSuccess;
    if( newState == MQTTPubRelSend )
    {
        addRecord( records, MQTT_STATE_ARRAY_MAX_COUNT, packetId, MQTTQoS2, MQTTPubRelSend );
    }
}
```

libraries/standard/mqtt/include/mqtt.h
leegeth: Sure will do.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -565,16 +655,7 @@ static MQTTStatus_t sendPublishAcks( MQTTContext_t * pContext,
...
-                                          packetId,
-                                          packetType,
-                                          MQTT_SEND,
-                                          &newState );
-
-            if( status != MQTTSuccess )
-            {
-                LogError( ( "Failed to update state of publish %u.", packetId ) );
-            }
+            status = updateAckSentState( pContext, packetId, packetType );
```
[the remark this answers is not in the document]
leegeth: Yea. It was needed when I had some other checks in this function and increased the overall complexity. I have taken those out after an optimization, but I think it is still beneficial to have as we are already at complexity 7 and this helps. So I think I will keep it. Any thoughts?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -647,18 +729,44 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
...
         /* Invoke application callback to hand the buffer over to application
-         * before sending acks. */
-        pContext->callbacks.appCallback( pContext,
-                                         pIncomingPacket,
-                                         packetIdentifier,
-                                         &publishInfo );
+         * before sending acks.
+         * Application callback will be invoked for all publishes, except for
+         * duplicate QoS2 incoming publish. QoS2 ensures at most once delivery. */
+        if( ( duplicatePublish == false ) || ( publishInfo.qos != MQTTQoS2 ) )
```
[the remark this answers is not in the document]
leegeth: Yes we would need it, A QoS1 publish can still be received when the state is pubacksend.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -687,8 +853,8 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
      * sending any PUBREL or PUBCOMP. However, for other cases, we invoke it
      * at the end to reduce the complexity of this function. */
     bool invokeAppCallback = false;
-    MQTTPubAckType_t ackType;
     MQTTEventCallback_t appCallback = NULL;
+    bool pubrelRecordMissing = false;
```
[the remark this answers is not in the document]
leegeth: It was part of an earlier change, removed it. Not sure why Coverity didn't catch that.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -323,23 +544,246 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
+    uint16_t outgoingStates = 0U;
+    const MQTTPubAckInfo_t * records = NULL;
+    bool stateCheck = false;
+
+    assert( pMqttContext != NULL );
+    assert( searchStates != 0U );
+    assert( pCursor != NULL );
+
+    /* Create a bit map with all the outgoing publish states. */
+    UINT16_SET_BIT( outgoingStates, MQTTPublishSend );
```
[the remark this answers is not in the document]
leegeth: Yes

libraries/standard/mqtt/src/mqtt_state.c
leegeth: We don't need that. We are just depending on packet ID for determining if a record is present or not. When a new record is added, we are anyway updating all the fields. Hence I don't think this clean up is required. It came up in a discussion with Dan.

libraries/standard/mqtt/src/mqtt_state.c
leegeth: I would prefer it to be assigned to a variable for making the debugging easier.

libraries/standard/mqtt/src/mqtt_state.c
leegeth: Why to do that if we know nothing to be done in this case?

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -354,10 +362,12 @@ static void setupSubscriptionInfo( MQTTSubscribeInfo_t * pSubscribeInfo )
 static void expectProcessLoopCalls( MQTTContext_t * const pContext,
                                     MQTTStatus_t deserializeStatus,
                                     MQTTPublishState_t stateAfterDeserialize,
+                                    MQTTStatus_t updateStateStatus,
                                     MQTTStatus_t serializeStatus,
                                     MQTTPublishState_t stateAfterSerialize,
                                     MQTTStatus_t processLoopStatus,
-                                    bool incomingPublish )
+                                    bool incomingPublish,
+                                    MQTTPublishInfo_t * pPubInfo )
```
muneebahmed10, line 370: The parameter list for this function is getting pretty large, and I don't see any comments describing each parameter either. At this point - since a lot of these parameters are related to each other - would it be more maintainable to create a struct to encapsulate most of these parameters and pass that in instead? I won't block the PR for this though

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -425,21 +440,35 @@ static void expectProcessLoopCalls( MQTTContext_t * const pContext,
...
+        else
+        {
+            switch( updateStateStatus )
+            {
+                case MQTTSuccess:
+                    expectMoreCalls = true;
+                    break;
+
+                case MQTTStateCollision:
+                    expectMoreCalls = ( pPubInfo->dup == true ) ? true : false;
```
muneebahmed10, line 465: We don't need to worry about the extra MISRA code for this file
```suggestion
                    expectMoreCalls = pPubInfo->dup;
```

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -768,6 +797,128 @@ void test_MQTT_Connect_partial_receive()
...
+    incomingPacket.remainingLength = 2;
+    MQTT_GetIncomingPacketTypeAndLength_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_GetIncomingPacketTypeAndLength_ReturnThruPtr_pIncomingPacket( &incomingPacket );
+    /* Return with a session present flag. */
+    sessionPresent = true;
+    MQTT_DeserializeAck_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_DeserializeAck_ReturnThruPtr_pSessionPresent( &sessionPresent );
+    /* No packets to resend. */
+    MQTT_PubrelToResend_ExpectAnyArgsAndReturn( MQTT_PACKET_TYPE_INVALID );
+    status = MQTT_Connect( &mqttContext, &connectInfo, NULL, timeout, &sessionPresent );
```
muneebahmed10, line 839: Shouldn't the two `sessionPresent`s be different variables? I don't think it's good practice to mock setting a variable with itself

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -768,6 +797,128 @@ void test_MQTT_Connect_partial_receive()
...
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+    TEST_ASSERT_EQUAL_INT( MQTTConnected, mqttContext.connectStatus );
+
```
muneebahmed10, line 842: Why not add a test for `sessionPresent == true` here?

libraries/standard/mqtt/utest/mqtt_state_utest.c
```diff
@@ -68,45 +69,193 @@ static void fillRecord( MQTTPubAckInfo_t * records,
...
 /* ========================================================================== */
 
 void test_MQTT_ReserveState( void )
 {
     MQTTContext_t mqttContext = { 0 };
     MQTTStatus_t status;
+    const uint16_t PACKET_ID = 1;
+    const uint16_t PACKET_ID2 = 2;
+    const uint16_t PACKET_ID3 = 3;
+    const size_t index = 0;
```
muneebahmed10, line 92: I don't think it's worth declaring this const as 0. I think it's more readable to just leave 0, since that immediately tells me we're testing the first element

libraries/standard/mqtt/utest/mqtt_state_utest.c
```diff
@@ -68,45 +69,193 @@ static void fillRecord( MQTTPubAckInfo_t * records,
...
+                      MQTTQoS1,
+                      MQTTPublishSend );
+    /* Any new add should result in no memory error. */
+    status = MQTT_ReserveState( &mqttContext, PACKET_ID2, MQTTQoS1 );
+    TEST_ASSERT_EQUAL( MQTTNoMemory, status );
+
+    /* Alternate free spots.
+     * Pre condition - 1 0 1 0 1 0 1 0 1 0.
+     * Add an element will skip to compact the array and the resulting state
+     * should be - 1 0 1 0 1 0 1 0 1 0. */
```
muneebahmed10, line 180:
```suggestion
     * should be - 1 0 1 0 1 0 1 0 1 1. */
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -647,18 +729,44 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
...
+        {
+            LogInfo( ( "State record updated. New state=%s.",
+                       MQTT_State_strerror( publishRecordState ) ) );
+        }
+
+        /* A collision in the state engine and a duplicate flag set in the
+         * incoming packet would be indicating a duplicate incoming publish. */
+        else if( ( status == MQTTStateCollision ) && ( publishInfo.dup == true ) )
+        {
+            status = MQTTSuccess;
```
muneebahmed10, line 743: If there was a state collision for an incoming publish, shouldn't we reset the state to PubRecSend/PubAckSend? Otherwise there will be more state collisions when we're sending the acks. It looks like if there was a collision, then publishRecordState will be MQTTStateNull, meaning sendPublishAcks on line 772 won't send any acks for it. However, according to the spec: 

"Until it has received the corresponding PUBREL packet, the Receiver MUST acknowledge any subsequent PUBLISH packet with the same Packet Identifier by sending a PUBREC. It MUST NOT cause duplicate messages to be delivered to any onward recipients in this case"

For QoS 2, I think we will always end up sending a PUBREC. The only state that occurs after receiving the PUBREL is PubCompSend. If we haven't sent the PUBCOMP, then the server has no way of knowing whether we ever received the PUBREL, so it should be fine the send the PUBREC. If we have sent the PUBCOMP, then we're required to treat this as a new publish and send a PUBREC.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
leegeth: I just did a quick experiment of copying over the code back and complexity jumped to 11. So I think we have a conclusion here.

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -1,5 +1,4 @@
 #include <string.h>
-#include <stdbool.h>
```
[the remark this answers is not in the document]
leegeth: Ok

libraries/standard/mqtt/utest/mqtt_utest.c
leegeth: Hi, I had a different set of opinions about the function we use. I see a few problems with the function already.
1. I think the readability of the tests are not great with the function. I have to spent a good amount of time to understand what the function is doing and spent another good amount time to write any tests for ProcessLoop. I think if we are not the developer of the function, the readability aspect is really poor.
2. Debugging issues - debugging was very difficult for any failure as the line number and the expected and actual values are not really pointing to the test failed in the group, Tests need to be rerun again with logs enabled or enabling tests one by one in the group. I think getting rid of the function itself will be a good change.

Anyway those are not in the scope of the PR, I think we would need to come back to it.

libraries/standard/mqtt/utest/mqtt_utest.c
leegeth: Yea. I think that is a good idea to verify the session present as well.

libraries/standard/mqtt/src/mqtt.c
leegeth: As discussed, the state returned from MQTT_UpdateStatePublish will either be PubRecSend/PubAckSend depending on the QoS level of the publish.  So that will take care of sending the acks.

Also if we have already sent a PUBCOMP, the state of the publish will not be present and this will be considered as a new publish.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
muneebahmed10: Are you sure? These were my results on this branch:
Before copying the code:
```
$ complexity -t 1 mqtt.c | grep sendPublishAcks
    5      42      35   mqtt.c(625): sendPublishAcks
```
After copying the code:
```
$ complexity -t 1 mqtt.c | grep sendPublishAcks
    7      51      43   mqtt.c(625): sendPublishAcks
```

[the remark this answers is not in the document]
leegeth: Yes. After copying over.
Complexity Scores
Score | ln-ct | nc-lns| file-name(line): proc-name
   11      89      67   libraries/standard/mqtt/src/mqtt_state.c(914): MQTT_UpdateStateAck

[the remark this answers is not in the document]
muneebahmed10: That doesn't make sense. Copying the lines:
```
            status = MQTT_UpdateStateAck( pContext,
                                        packetId,
                                        packetType,
                                        MQTT_SEND,
                                        &newState );

            if( status != MQTTSuccess )
            {
                LogError( ( "Failed to update state of publish %u.", packetId ) );
            }
```
to line 658 of `sendPublishAcks` in mqtt.c will not change the complexity of `MQTT_UpdateStateAck` in mqtt_state.c. They're different functions in different files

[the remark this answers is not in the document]
leegeth: I see what happened. I was taking a look at the wrong function MQTT_UpdateStateAck. I see the right function now and Your finding is absolutely right and made the changes. Thanks

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -671,12 +750,70 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
...
+                        packetIdentifier,
+                        MQTT_Status_strerror( status ) ) );
+        }
+    }
+
+    if( status == MQTTSuccess )
+    {
+        /* Invoke application callback to hand the buffer over to application
+         * before sending acks. */
+        appCallback( pContext, pIncomingPacket, packetIdentifier, NULL );
```
muneebahmed10, line 799: Should this be moved to the above if block? Would it make sense to hand the packet to the application regardless of the outcome of updating the state engine?

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: After offline conversation, I think there should be comment explaining each case, to avoid future confusion

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -113,6 +145,65 @@ static void updateRecord( MQTTPubAckInfo_t * records,
                           MQTTPublishState_t newState,
                           bool shouldDelete );
 
+/**
+ * @brief Get the packet ID and index of a publish in a specified state.
+ *
+ * @note The API is limited to search states of only outgoing publish records
+ * or incoming publish records at a time. The usage of API only demands the
+ * search of only either of the records. Hence this is an optimization to
+ * limit the search only to one of the records at a time.
```
dan4thewin, line 154: You lost me.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -154,6 +245,26 @@ static bool validateTransitionPublish( MQTTPublishState_t currentState,
 
             break;
 
+        /* Below cases are for validating the resends of publish when a session is
+         * reestablished. */
+        case MQTTPubAckPending:
+
+            /* When a session is reestablished outgoing QoS1 publishes in state
+             * #MQTTPubAckPending can be resend. This will result in a no
+             * transition. */
```
dan4thewin, line 254: comma after reestablished; resend -> resent
how about "This will..." -> "The state remains the same."

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -154,6 +245,26 @@ static bool validateTransitionPublish( MQTTPublishState_t currentState,
...
+            isValid = ( newState == MQTTPubAckPending ) ? true : false;
+
+            break;
+
+        case MQTTPubRecPending:
+
+            /* When a session is reestablished outgoing QoS2 publishes in state
+             * #MQTTPubRecPending can be resend. This will result in a no
+             * transition. */
```
dan4thewin, line 263: same comment

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -182,13 +295,45 @@ static bool validateTransitionAck( MQTTPublishState_t currentState,
...
-            /* Incoming publish, QoS 2. */
-            isValid = ( newState == MQTTPubCompSend ) ? true : false;
+
+            /* Incoming publish, QoS 2.
+             * There are 2 valid transitions possible.
+             * 1. MQTTPubRelPending -> MQTTPubCompSend : A PUBREL ack is received
+             *    when publish record state is MQTTPubRelPending. This is the
+             *    normal state transition without any connection interuptions.
+             * 2. MQTTPubRelPending -> MQTTPubRelPending : Receiving a duplicate
+             *    QoS2 publish can result in this no transition.
```
dan4thewin, line 305: "this no transition" -> "a transition to the same state"

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -182,13 +295,45 @@ static bool validateTransitionAck( MQTTPublishState_t currentState,
...
-            /* Incoming publish, QoS 2. */
-            isValid = ( newState == MQTTPublishDone ) ? true : false;
+
+            /* Incoming publish, QoS 2.
+             * There are 2 valid transitions possible.
+             * 1. MQTTPubCompSend -> MQTTPublishDone : A PUBCOMP ack is sent
+             *    after receiving a PUBREL from broker. This is the
+             *    normal state transition without any connection interuptions.
+             * 2. MQTTPubCompSend -> MQTTPubCompSend : Receiving a duplicate PUBREL
+             *    can result in this no transition.
```
dan4thewin, line 328: same comment

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -182,13 +295,45 @@ static bool validateTransitionAck( MQTTPublishState_t currentState,
...
+             *    This can happen in the below state transition.
+             *    1. A TCP connection failure happened before sending a PUBCOMP
+             *       for an incoming PUBREL.
+             *    2. Reestablish MQTT session.
+             *    3. MQTT broker resends the un-acked PUBREL.
+             *    4. Receiving the PUBREL again will result in this no transition. */
```
dan4thewin, line 334: same comment

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -202,8 +347,23 @@ static bool validateTransitionAck( MQTTPublishState_t currentState,
             break;
 
         case MQTTPubCompPending:
-            /* Outgoing publish, Qos 2. */
-            isValid = ( newState == MQTTPublishDone ) ? true : false;
+
+            /* Outgoing publish, Qos 2.
```
dan4thewin, line 351: same grammar issue

libraries/standard/mqtt/src/mqtt.c
leegeth: No. I don't think so. If state engine already recognized an issue with an incoming packet, I don't think application should receive the packet.

libraries/standard/mqtt/src/mqtt.c
leegeth: Agree.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -323,23 +546,242 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
+    assert( pCursor != NULL );
+
+    /* Create a bit map with all the outgoing publish states. */
+    UINT16_SET_BIT( outgoingStates, MQTTPublishSend );
+    UINT16_SET_BIT( outgoingStates, MQTTPubAckPending );
+    UINT16_SET_BIT( outgoingStates, MQTTPubRecPending );
+    UINT16_SET_BIT( outgoingStates, MQTTPubRelSend );
+    UINT16_SET_BIT( outgoingStates, MQTTPubCompPending );
+
+    /* Only outgoing publish records need to be searched. */
```
dan4thewin, line 591: Should we also assert that only outgoing states are present in searchStates? (i.e. `(~outgoingStates & searchStates) == 0` )

libraries/standard/mqtt/include/mqtt_state.h
```diff
@@ -113,15 +113,34 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
                                   MQTTPublishState_t * pNewState );
 
 /**
- * @brief Get the packet ID and index of a publish in a specified state.
+ * @brief Get the packet ID of next pending PUBREL ack to be resent.
+ *
+ * This function will need to be called to get the packet for which a PUBREL
+ * need to be sent when a session is reestablished. Calling this function
+ * repeatedly until packet id is 0 will give all the packets for which
+ * a PUBREL need to be resent in the order.
```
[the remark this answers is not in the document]
leegeth: Yes. Thanks

libraries/standard/mqtt/include/mqtt_state.h
```diff
@@ -113,15 +113,34 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
...
+                              MQTTStateCursor_t * pCursor,
+                              MQTTPublishState_t * pState );
+
+/**
+ * @brief Get the packet ID of next pending publish to be resent.
+ *
+ * This function will need to be called to get the packet for which a publish
+ * need to be sent when a session is reestablished. Calling this function
+ * repeatedly until packet id is 0 will give all the packets for which
+ * a publish need to be resent in the order.
```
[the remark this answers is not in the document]
leegeth: Yes. Thanks

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1131,6 +1337,12 @@ MQTTStatus_t MQTT_Connect( MQTTContext_t * pContext,
                                  pSessionPresent );
     }
 
+    /* Resend all the PUBREL when reestablishing a session. */
+    if( ( status == MQTTSuccess ) && ( *pSessionPresent == true ) )
```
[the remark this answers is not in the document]
leegeth: That's a good question. I think If we adhere to MQTT protocol, checking session present flag itself will be good as per the spec http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718033. 
Even though, I think it brings a good question that whether we should consider cleansession and session present is true, a violation of the spec and throw an error back to application. I think yes, What do you think?

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -323,23 +546,242 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
+    MQTTPublishState_t newState = MQTTStateNull;
+    bool isTransitionValid = false;
+
+    assert( pMqttContext != NULL );
+    assert( packetId != MQTT_PACKET_ID_INVALID );
+    assert( qos != MQTTQoS0 );
+    assert( pNewState != NULL );
+
+    newState = *pNewState;
+    isTransitionValid = validateTransitionPublish( currentState, newState, opType, qos );
```
muneebahmed10, line 747: Can you add a comment about how this will always succeed if opType is a receive? This is due to the fact that the passed in currentState must be MQTTStateNull, since MQTT_UpdateStatePublish does not perform a lookup for receives.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -20,9 +20,28 @@
...
+ */
+#define UINT16_SET_BIT( x, position )      ( ( x ) = ( uint16_t ) ( ( x ) | ( ( uint16_t ) 0x01U << ( ( uint16_t ) position ) ) ) )
+
+/**
+ * @brief Macro for checking if a bit is set in a 16-bit unsigned integer.
+ *
+ * @param[in] x The unsigned 16-bit integer to check.
+ * @param[in] position Which bit to check.
+ */
+#define UINT16_CHECK_BIT( x, position )    ( ( ( x ) & ( ( uint16_t ) 0x01U << ( ( uint16_t ) position ) ) ) == ( 0x01U << ( ( uint16_t ) position ) ) )
```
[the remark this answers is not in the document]
leegeth: Sorry, I didn't understand that comment.

libraries/standard/mqtt/src/mqtt_state.c
leegeth: This comment need to be reworked. Thanks

libraries/standard/mqtt/src/mqtt_state.c
leegeth: Thanks

libraries/standard/mqtt/src/mqtt_state.c
leegeth: Thanks

libraries/standard/mqtt/src/mqtt_state.c
leegeth: Thanks

libraries/standard/mqtt/src/mqtt_state.c
leegeth: Thanks

libraries/standard/mqtt/src/mqtt_state.c
leegeth: Yes. That will be a good addition.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -647,18 +700,44 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
...
     {
         /* Invoke application callback to hand the buffer over to application
-         * before sending acks. */
-        pContext->callbacks.appCallback( pContext,
-                                         pIncomingPacket,
-                                         packetIdentifier,
-                                         &publishInfo );
+         * before sending acks.
+         * Application callback will be invoked for all publishes, except for
+         * duplicate QoS2 incoming publish. QoS2 ensures at most once delivery. */
```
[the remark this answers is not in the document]
leegeth: That is a good suggestion.I think it makes sense. I will make that change.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -707,8 +707,34 @@ static MQTTStatus_t handleIncomingPublish( MQTTContext_t * pContext,
...
+         *       #MQTTStateCollision status from the state engine. The publish
+         *       data is passed to the application even if it is a duplicate as
+         *       QoS1 is at least once delivery.
+         *    b. QoS2 - If a PUBREC is not successfully sent for the incoming
+         *       publish or the PUBREC sent is not successfully received by the
+         *       broker due to a connection issue, it can result in broker
+         *       sending out a duplicate publish with dup flag set, when a
+         *       session is reestablished. It can result in a collision in
+         *       state engine. This will be handled by ignoring the
+         *       #MQTTStateCollision status from the state engine. The publish
```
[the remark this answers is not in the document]
leegeth: I see the concern that you are raising. I think it can be made more readable by depending on a different API to get the state for ack by using MQTT_CalculateStatePublish. Please let me know if there are any thoughts regarding that.

libraries/standard/mqtt/src/mqtt.c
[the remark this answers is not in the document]
leegeth: I will handle the second part of the change as a different PR as that will be independent of this change.

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -202,8 +342,24 @@ static bool validateTransitionAck( MQTTPublishState_t currentState,
...
+             *    This is the normal state transition without any connection
+             *    interuptions.
+             * 2. MQTTPubCompPending -> MQTTPubCompPending : Resending a PUBREL for
+             *    packets in state #MQTTPubCompPending can result in this
+             *    transition to the same state.
+             *    This can happen in the below state transition.
+             *    1. A TCP connection failure happened before receiving a PUBCOMP
+             *       for an outgoing PUBREL.
+             *    2. Reestablish MQTT session.
+             *    3. Resend the un-acked PUBREL will result in this transition
```
muneebahmed10, line 359: Nit: (grammar), I'd change the tenses in each of these steps to agree with each other:
```suggestion
             *    1. A TCP connection failure happens before receiving a PUBCOMP
             *       for an outgoing PUBREL.
             *    2. MQTT session is reestablished.
             *    3. Resending the un-acked PUBREL results in this transition
```

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -284,31 +486,48 @@ static MQTTStatus_t addRecord( MQTTPubAckInfo_t * records,
...
     size_t availableIndex = recordCount;
+    bool validEntryFound = false;
 
     assert( packetId != MQTT_PACKET_ID_INVALID );
     assert( qos != MQTTQoS0 );
 
-    /* Start from end so first available index will be populated. */
+    /* Check if we have to compact the records. This is known by checking if
+     * the last spot in the array is filled. */
+    if( records[ ( recordCount - 1U ) ].packetId != MQTT_PACKET_ID_INVALID )
```
muneebahmed10, line 496: Don't really need parentheses
```suggestion
    if( records[ recordCount - 1U ].packetId != MQTT_PACKET_ID_INVALID )
```

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -571,26 +954,10 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
...
-        else
-        {
-            status = MQTTIllegalState;
-            LogError( ( "Invalid transition from state %s to state %s.",
-                        MQTT_State_strerror( currentState ),
-                        MQTT_State_strerror( newState ) ) );
-        }
+        /* Validate state transition and update state record. */
+        status = updateStateAck( records, recordIndex, packetId, currentState, &newState );
+        *pNewState = newState;
```
muneebahmed10, line 960: It looks like now this function will set `pNewState` even if updating the state fails, when previously it would only do so if the return status was MQTTSuccess. However, it won't set it if the return status is MQTTBadParameter (due to the possibility of `pNewState == NULL`). Even though it sets it to MQTTStateNull on error, should we add an `if( status == MQTTSuccess)` check here to only set it on success? That would also mean `updateStateAck` wouldn't need to take a pointer parameter

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -448,85 +895,23 @@ MQTTStatus_t MQTT_UpdateStatePublish( MQTTContext_t * pMqttContext,
...
+        /* Validate state transition and update state records. */
+        mqttStatus = updateStatePublish( pMqttContext,
+                                         recordIndex,
+                                         packetId,
+                                         opType,
+                                         qos,
+                                         currentState,
+                                         &newState );
+
+        /* Output parameter. */
```
muneebahmed10, line 907: Same as above, we could add a status check if we don't want to set it for errors

libraries/standard/mqtt/src/mqtt_state.c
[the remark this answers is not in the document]
leegeth: Addressed it

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -186,8 +192,7 @@ static MQTTStatus_t updateStateAck( MQTTPubAckInfo_t * records,
  * @param[in] opType Send or Receive.
  * @param[in] qos 0, 1, or 2.
  * @param[in] currentState Current state of the publish record.
- * @param[in,out] pNewState Input/Output pointer of the new state of the
- * publish record.
+ * @param[in,out] newState New state of the publish record.
```
dan4thewin, line 195: Should this still be [in,out] ?

leegeth: in. thanks.

PR #1009, opened 2020-06-22
Fix _sendMqttMessage implementation to return network error for QoS 0…

aggarw13:
Fix issue causing `IotMqtt_PublishSync`to always return success on network disconnection for when called for **QoS 0**

Related GitHub issue: https://github.com/aws/amazon-freertos/issues/2155

Note: This also fixes the similar behavior of `IotMqtt_SubscribeSync`, and `IotMqtt_UnSubscribeSync` on network disconnection/network send failure


libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -803,12 +803,22 @@ static IotMqttError_t _sendMqttMessage( _mqttOperation_t * pMqttOperation,
...
+
     IotMqtt_Assert( pMqttOperation != NULL );
 
     /* Send the SUBSCRIBE packet. */
     if( ( flags & MQTT_INTERNAL_FLAG_BLOCK_ON_SEND ) == MQTT_INTERNAL_FLAG_BLOCK_ON_SEND )
     {
         _IotMqtt_ProcessSend( IOT_SYSTEM_TASKPOOL, pMqttOperation->job, pMqttOperation );
+
+        /* Update return status value if operation object status changed after job execution. */
+        if( pMqttOperation->u.operation.status != cachedStatus )
```
abhidixi11, line 818: This may be causing double free issues.  Please try moving code to check u.operation.status in PublishAsync() ( for QOS0 and MQTT_INTERNAL_FLAG_BLOCK_ON_SEND )

aggarw13: Have used a different approach of not utilizing the complex "operations" infrastructure when dealing with the simplistic case of synchronous QoS 0 request.

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
```diff
@@ -282,6 +282,24 @@ static size_t _sendSuccess( IotNetworkConnection_t pSendContext,
...
+static size_t _sendFailure( IotNetworkConnection_t pSendContext,
+                            const uint8_t * pMessage,
+                            size_t messageLength )
+{
+    /* Silence warnings about unused parameters. */
+    ( void ) pSendContext;
+    ( void ) pMessage;
+    ( void ) messageLength;
+
+    /* This function returns the message length to simulate a successful send. */
```
muneebahmed10, line 297:
```suggestion
    /* This function returns 0 to simulate a failed send. */
```

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
```diff
@@ -687,12 +705,15 @@ TEST_GROUP_RUNNER( MQTT_Unit_API )
     RUN_TEST_CASE( MQTT_Unit_API, DisconnectAlreadyDisconnected );
     RUN_TEST_CASE( MQTT_Unit_API, PublishQoS0Parameters );
     RUN_TEST_CASE( MQTT_Unit_API, PublishQoS0MallocFail );
+    RUN_TEST_CASE( MQTT_Unit_API, PublishSyncQoS0WithNetworkFailure );
```
muneebahmed10, line 708: Nit: `PublishQoS0SyncWithNetworkFailure` seems like it would fit better with the other test names

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1759,6 +1782,101 @@ IotMqttError_t IotMqtt_PublishAsync( IotMqttConnection_t mqttConnection,
...
+    else if( _IotMqtt_ValidatePublish( mqttConnection->awsIotMqttMode,
+                                       pPublishInfo,
+                                       flags,
+                                       pCallbackInfo,
+                                       pPublishOperation ) == false )
+    {
+        status = IOT_MQTT_BAD_PARAMETER;
+    }
+    else if( ( ( flags & MQTT_INTERNAL_FLAG_BLOCK_ON_SEND )
+               == MQTT_INTERNAL_FLAG_BLOCK_ON_SEND ) &&
```
leegeth, line 1861: Are we expecting that the flag MQTT_INTERNAL_FLAG_BLOCK_ON_SEND  is set from application? Isn't this an internal flag and I don't see it in the public docs as well https://docs.aws.amazon.com/freertos/latest/lib-ref/c-sdk/mqtt/mqtt_constants.html#mqtt_constants_flags

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -280,6 +280,35 @@ static void _handleConnectFailure( IotMqttConnection_t pMqttConnection,
...
+ * @param[in] mqttConnection The MQTT connection to use for the PUBLISH.
+ * @param[in] pPublishInfo MQTT publish parameters.
+ * @param[in] flags Flags which modify the behavior of this function.
+ * @param[in] pCallbackInfo Asynchronous notification of this function's completion (`NULL` to disable).
+ * @param[out] pPublishOperation Set to a handle by which this operation may be
+ * referenced after this function returns. This reference is invalidated once
+ * the publish operation completes.
+ *
+ * @return This function will return #IOT_MQTT_STATUS_PENDING upon success for
+ * QoS 1 publishes
```
leegeth, line 294: How about the return codes when there are errors?

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -280,6 +280,35 @@ static void _handleConnectFailure( IotMqttConnection_t pMqttConnection,
...
+                                     uint32_t flags,
+                                     const IotMqttCallbackInfo_t * pCallbackInfo,
+                                     IotMqttOperation_t * const pPublishOperation );
+
+/**
+ * @brief Sends a Publish message at QoS 0 synchronously.
+ * @param[in] mqttConnection The MQTT connection to use for the PUBLISH.
+ * @param[in] pPublishInfo MQTT publish parameters.
+ *
+ * @return returns #IOT_MQTT_SUCCESS upon success.
```
leegeth, line 307: Same comment about the return values for error cases.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1750,6 +1763,16 @@ IotMqttError_t IotMqtt_PublishAsync( IotMqttConnection_t mqttConnection,
...
         }
 
+        if( ( pPublishInfo->qos == IOT_MQTT_QOS_0 ) && ( pOperation != NULL ) )
+        {
+            if( ( ( flags & MQTT_INTERNAL_FLAG_BLOCK_ON_SEND )
+                  == MQTT_INTERNAL_FLAG_BLOCK_ON_SEND ) &&
+                ( pOperation->u.operation.status == IOT_MQTT_NETWORK_ERROR ) )
+            {
+                status = IOT_MQTT_NETWORK_ERROR;
+            }
```
leegeth, line 1773: In what case, the execution would reach here? Isn't it always be going to _publishQos0Sync as per line 1864.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1652,27 +1681,11 @@ IotMqttError_t IotMqtt_PublishAsync( IotMqttConnection_t mqttConnection,
...
     uint8_t ** pPacketIdentifierHigh = NULL;
 
-    /* Check that IotMqtt_Init was called. */
-    if( _checkInit() == false )
-    {
-        status = IOT_MQTT_NOT_INITIALIZED;
-    }
-    else if( _IotMqtt_ValidatePublish( mqttConnection->awsIotMqttMode,
-                                       pPublishInfo,
+    /* Create a PUBLISH operation. */
```
leegeth, line 1684: Better to have asserts in the internal function to validate the arguments.

libraries/standard/mqtt/src/iot_mqtt_api.c
```diff
@@ -1759,6 +1782,101 @@ IotMqttError_t IotMqtt_PublishAsync( IotMqttConnection_t mqttConnection,
...
+                                       pPublishOperation ) == false )
+    {
+        status = IOT_MQTT_BAD_PARAMETER;
+    }
+    else if( ( ( flags & MQTT_INTERNAL_FLAG_BLOCK_ON_SEND )
+               == MQTT_INTERNAL_FLAG_BLOCK_ON_SEND ) &&
+             ( pPublishInfo->qos == IOT_MQTT_QOS_0 ) )
+    {
+        status = _publishQos0Sync( mqttConnection,
+                                   pPublishInfo );
```
leegeth, line 1865: What is the callback behavior in this case? Are we completely ignoring the callback behavior? What will happen to applications already present and expecting a callback to mark the completion of publishAsync?

libraries/standard/mqtt/src/iot_mqtt_api.c
aggarw13: Good catch. Will remove this logic, was added in the initial revision of the fix in the PR, but had forgotten to remove it.

Regarding your point that the flag is internal facing, unfortunately, the `IotMqtt_PublishAsync` API will accept the flag even if the user accidentally sets the bit associated with this flag in the `flags` input parameter. (This is the behavior of `v4_beta` branch)

libraries/standard/mqtt/src/iot_mqtt_api.c
aggarw13: Yes, will remove this logic as mentioned in https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1009#discussion_r445779256

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
aggarw13: Changed

libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
aggarw13: Changed

libraries/standard/mqtt/src/iot_mqtt_api.c
aggarw13: Added documentation.

libraries/standard/mqtt/src/iot_mqtt_api.c
aggarw13: Added assert checks.

libraries/standard/mqtt/src/iot_mqtt_api.c
leegeth: I see that it is a possible behavior for the async API. Isn't this flag supposed to be set internally though for a publish sync API call for QoS0? Are you missing that set logic in this PR?

libraries/standard/mqtt/src/iot_mqtt_api.c
aggarw13: Added

libraries/standard/mqtt/src/iot_mqtt_api.c
leegeth: As discussed, I see that the flag is set from publish sync API

libraries/standard/mqtt/src/iot_mqtt_api.c
leegeth: As discussed offline, as this path is hit only for sync API, callbacks are not going to be present.

aggarw13: We talked separately to clarify that a callback parameter is not expected for the sync Publish case as firstly, the `MQTT_INTERNAL_FLAG_BLOCK_ON_SEND` flag is not public-facing and thus, the customer cannot call the `IotMqtt_PublishAsync` function for a QoS 0 synchronous operation, and secondly, the sync Publish Qos 0 operation is requested by the `IotMqtt_PublishSync` function that passes the internal `MQTT_INTERNAL_FLAG_BLOCK_ON_SEND` flag to the `IotMqtt_PublishAsync` function.

Basically, the changes made do not change the behavior of the function (for any customer application depending on this function to be affected)!

PR #1014, opened 2020-06-24
Small checks for MQTT_GetPacketId and MQTT_GetIncomingPacketTypeAndLength

sarenameas:
*Description of changes:*

- Intentionally check for integer overflow in MQTT_GetPacketId. This helps with alleviating flags in bound checking tools and any unwanted behavior that might result from simply just allowing overflow and handling it after.
- Add NULL parameter check for public function MQTT_GetIncomingPacketTypeAndLength
- Add unit tests for the last bullet point.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -1468,9 +1468,12 @@ uint16_t MQTT_GetPacketId( MQTTContext_t * pContext )
     if( pContext != NULL )
     {
         packetId = pContext->nextPacketId;
-        pContext->nextPacketId++;
 
-        if( pContext->nextPacketId == 0U )
+        if( pContext->nextPacketId == UINT16_MAX )
+        {
+            pContext->nextPacketId = 1;
+        }
+        else
         {
             pContext->nextPacketId++;
         }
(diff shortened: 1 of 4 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -2046,35 +2106,17 @@ MQTTStatus_t MQTT_GetIncomingPacketTypeAndLength( MQTTTransportRecvFunc_t readFu
...
-            }
-        }
-        else
-        {
-            LogError( ( "Incoming packet invalid: Packet type=%u",
-                        pIncomingPacket->type ) );
-            status = MQTTBadResponse;
-        }
-    }
-    else if( bytesReceived == 0 )
```
muneebahmed10, line 2071: I would suggest having this be something like `else if( ( bytesReceived == 0 ) && ( status != MQTTBadParameter ) )`, so that we don't have to add another helper function.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1468,9 +1468,12 @@ uint16_t MQTT_GetPacketId( MQTTContext_t * pContext )
     if( pContext != NULL )
     {
         packetId = pContext->nextPacketId;
-        pContext->nextPacketId++;
 
-        if( pContext->nextPacketId == 0U )
+        if( pContext->nextPacketId == UINT16_MAX )
```
abhidixi11, line 1472: I think we missed it earlier, it would be nice to add a reason here the the packet-Id 0 is  invalid.

sarenameas: Noted, I will add with the next wave of MQTT changes.

PR #1015, opened 2020-06-26
Add CBMC proof for HTTPClient_AddRangeHeader

yourslab:
This CBMC proof for HTTPClient_AddRangeHeader has no errors or warnings provided the following fixes are made below:
- Must check that `rangeStartOrlastNbytes != INT_MIN` because a signed 32-bit int cannot represent the absolute value of INT_MIN when calling `convertInt32ToAscii(...)`
- Must check that `pRequestHeaders->headersLen <= pRequestHeaders->bufferLen`, or another buffer overflow is possible exactly like what was found inHTTPClient_AddHeader.


libraries/standard/http/src/http_client.c
```diff
@@ -1465,6 +1470,13 @@ HTTPStatus_t HTTPClient_AddRangeHeader( HTTPRequestHeaders_t * pRequestHeaders,
                     rangeStartOrlastNbytes, rangeEnd ) );
         returnStatus = HTTP_INVALID_PARAMETER;
     }
+    else if( rangeStartOrlastNbytes == INT32_MIN )
```
sarenameas, line 1473: Do these need unit tests? Are these in a separate PR?

PR #1016, opened 2020-06-26
Bring down complexity of sendHttpData to 8

yourslab:
Brings down complexity of sendHttpData to 8

All credits go to @muneebahmed10 for this one! Thanks man!


libraries/standard/http/src/http_client.c
```diff
@@ -797,7 +797,7 @@ static int httpParserOnBodyCallback( http_parser * pHttpParser,
     /* MISRA Rule 11.8 flags casting away the const qualifier in the pointer
      * type. This rule is suppressed because when the body is of transfer
      * encoding chunked, the body must be copied over the chunk headers that
-     * preceed it. This is done to have a contigous response body. This does
+     * preceed it. This is done to have a contiguous response body. This does
```
muneebahmed10, line 800: Just realized there were two typos in this line
```suggestion
     * precede it. This is done to have a contiguous response body. This does
```

yourslab: Nice! Good catch.

PR #1018, opened 2020-06-28
Add unit test script for CI

nateglims:
*Description of changes:* Add a shell script to run unit tests in CI. Unit tests in this case are binaries generated by CMake with the `-DBUILD_TESTS=1` option set and have `utest` in their name. This distinction is to differentiate from platform integration tests, service integration tests and demos.


ci/ci_utest.sh
```diff
@@ -0,0 +1,25 @@
...
+
+    if [ $? -ne "0" ] ; then
+        STATUS=1
+    fi
+
+done
+
+if [ "$STATUS" = "1" ] ; then
+    exit 1
+fi
```
yourslab, line 25: Clean 👌🏼👌🏼👌🏼

ci/ci_utest.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
sarenameas, line 1: I am cool with the ci folder, but maybe this should be under tools/ci?

nateglims: Done

PR #1025, opened 2020-06-29
Handle session present error case

leegeth:
*Description of changes:*
Handle an error case when a session present flag is received from
broker when a clean session is requested.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -1148,9 +1148,8 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
     {
         if( ( cleanSession == true ) && ( *pSessionPresent == true ) )
         {
-            LogError( ( "Received CONNACK with session present from broker."
-                        " This should not happen as a clean session is requested"
-                        " in CONNECT." ) );
+            LogError( ( "Unexpected session present flag in CONNACK response from broker."
+                        " CONNECT request with clean session was made with broker." ) );
             status = MQTTBadResponse;
         }
     }
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -229,6 +229,7 @@ static MQTTStatus_t sendPublish( MQTTContext_t * pContext,
  *
  * @param[in] pContext Initialized MQTT context.
  * @param[in] timeoutMs Timeout for waiting for CONNACK packet.
+ * @param[in] cleanSession Clean session flag set by application.
  * @param[out] pIncomingPacket List of MQTT subscription info.
  * @param[out] pSessionPresent Whether a previous session was present.
  * Only relevant if not establishing a clean session.
@@ -240,6 +241,7 @@ static MQTTStatus_t sendPublish( MQTTContext_t * pContext,
  */
 static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
                                     uint32_t timeoutMs,
+                                    bool cleanSession,
                                     MQTTPacketInfo_t * pIncomingPacket,
                                     bool * pSessionPresent );
 
@@ -1064,6 +1066,7 @@ static MQTTStatus_t sendPublish( MQTTContext_t * pContext,
 
 static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
                                     uint32_t timeoutMs,
+                                    bool cleanSession,
                                     MQTTPacketInfo_t * pIncomingPacket,
                                     bool * pSessionPresent )
 {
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -997,6 +1000,19 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
...
 
+    /* If a clean session is requested, a session present should not be set by
+     * broker. */
+    if( status == MQTTSuccess )
+    {
+        if( ( cleanSession == true ) && ( *pSessionPresent == true ) )
+        {
+            LogError( ( "Received CONNACK with session present from broker."
+                        " This should not happen as a clean session is requested"
+                        " in CONNECT." ) );
```
aggarw13, line 1011: Probably should be a warning instead of an error as this represents an issue in the broker,
Also, recommend making the message concise as:
```suggestion
            LogWarn( ( "Unexpected session present flag in CONNACK response from broker."
                        " CONNECT request with clean session was made with broker." ) );
```

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -815,6 +827,20 @@ void test_MQTT_Connect_happy_path()
...
+    /* Request to establish a session if present and session present is received
+     * from broker. */
+    mqttContext.connectStatus = MQTTNotConnected;
+    connectInfo.cleanSession = false;
+    sessionPresent = true;
+    MQTT_GetIncomingPacketTypeAndLength_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_GetIncomingPacketTypeAndLength_ReturnThruPtr_pIncomingPacket( &incomingPacket );
+    MQTT_DeserializeAck_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_DeserializeAck_ReturnThruPtr_pSessionPresent( &sessionPresent );
+    status = MQTT_Connect( &mqttContext, &connectInfo, NULL, timeout, &sessionPresentResult );
```
muneebahmed10, line 840: The other tests pass `sessionPresent` to MQTT_Connect. I'd swap these two around and possibly change the name of the return pointer to `sessionPresentExpected` or something to be consistent with the other tests

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -691,6 +692,17 @@ void test_MQTT_Connect_receiveConnack( void )
     MQTT_DeserializeAck_ExpectAnyArgsAndReturn( MQTTBadResponse );
     status = MQTT_Connect( &mqttContext, &connectInfo, NULL, timeout, &sessionPresent );
     TEST_ASSERT_EQUAL_INT( MQTTBadResponse, status );
+
+    /* Session present is received when a clean session is requested. */
```
aggarw13, line 696: (Nit): Can be clearer with "Test case when broker sends session present flag in response to a clean session connection request. "

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -691,6 +692,17 @@ void test_MQTT_Connect_receiveConnack( void )
...
+    mqttContext.transportInterface.recv = transportRecvSuccess;
+    connectInfo.cleanSession = true;
+    sessionPresent = true;
+    MQTT_GetIncomingPacketTypeAndLength_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_GetIncomingPacketTypeAndLength_ReturnThruPtr_pIncomingPacket( &incomingPacket );
+    MQTT_DeserializeAck_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTT_DeserializeAck_ReturnThruPtr_pSessionPresent( &sessionPresent );
```
muneebahmed10, line 703: I'd add a variable so the pointer that gets returned by MQTT_DeserializeAck isn't identical to the one passed to MQTT_Connect

libraries/standard/mqtt/src/mqtt.c
leegeth: I think I still would prefer the error as it is a condition in which a connection cannot be continued with. I like the sentences you suggested :)

libraries/standard/mqtt/utest/mqtt_utest.c
leegeth: ok.

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -999,6 +1013,7 @@ void test_MQTT_Connect_happy_path()
     MQTT_GetIncomingPacketTypeAndLength_ReturnThruPtr_pIncomingPacket( &incomingPacket );
     MQTT_DeserializeAck_ExpectAnyArgsAndReturn( MQTTSuccess );
     MQTT_DeserializeAck_ReturnThruPtr_pSessionPresent( &sessionPresentExpected );
+    MQTT_PubrelToResend_ExpectAnyArgsAndReturn( MQTT_PACKET_TYPE_INVALID );
```
aggarw13, line 1016: Why was this needed to be added?

leegeth: Rebase brought the resend logic, so a call to function MQTT_PubrelToResend will be made if a session is present.

PR #1026, opened 2020-06-30
Add intermediate script for complexity

nateglims:
*Description of changes:* Moved complexity to a subscript so we can compensate for it's questionable exit code practices.

### DIFF
--- a/tools/ci/complexity.sh
+++ b/tools/ci/complexity.sh
@@ -4,7 +4,7 @@ function usage {
     cat <<USAGE
 Usage: $0 [files...]
 
-An script to run complexity in CI. This is needed to avoid issues with
+A script to run complexity in CI. This is needed to avoid issues with
 exit codes.
 
 USAGE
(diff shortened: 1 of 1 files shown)


tools/ci/complexity.sh
```diff
@@ -0,0 +1,27 @@
...
+exit codes.
+
+USAGE
+}
+
+if [ $# -eq 0 ] ; then
+    usage
+    exit 1
+fi
+
```
gkwicker, line 17: You could also consider checking to make sure that `complexity` is installed first, e.g.:

```sh
which complexity>/dev/null
if [ $? != 0 ]
then
    echo "complexity is not installed"
    exit 1
fi
```

nateglims: done

tools/ci/complexity.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+function usage {
+    cat <<USAGE
+Usage: $0 [files...]
+
+A script to run complexity in CI. This is needed to avoid issues with
```
sarenameas, line 7: This is uncommented

nateglims: It's all inside a here doc: https://en.wikipedia.org/wiki/Here_document#Unix_shells

sarenameas: thank you

PR #1029, opened 2020-07-01
JSON validator

dan4thewin:
This is a strict JSON document validator with key-value pair search.  A small ephemeral stack is used for nesting.  Search accepts queries for nested keys with a single call.  See the doc comments within for specifics.

### DIFF
--- a/libraries/standard/json/include/json.h
+++ b/libraries/standard/json/include/json.h
@@ -19,8 +19,8 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-#ifndef JSON_H
-#define JSON_H
+#ifndef JSON_H_
+#define JSON_H_
 
 #include <stddef.h>
 
@@ -37,7 +37,7 @@ typedef enum JSONStatus
  * @brief Parse a buffer to determine if it contains a valid JSON document.
  *
  * @param[in] buf  The buffer to parse.
- * @param[in] length  The size of the buffer.
+ * @param[in] max  The size of the buffer.
  *
  * @return #JSONSuccess if the buffer contents are valid JSON;
  * #JSONIllegalDocument if the buffer contents are NOT valid JSON;
@@ -52,13 +52,13 @@ typedef enum JSONStatus
  * contain an object or array, define JSON_VALIDATE_COLLECTIONS_ONLY.
  */
 JSONStatus_t JSON_Validate( const char * buf,
-                            size_t length );
+                            size_t max );
 
 /**
  * @brief Find a key in a JSON object and output a pointer to its value.
  *
  * @param[in] buf  The buffer to search.
- * @param[in] length  size of the buffer.
+ * @param[in] max  size of the buffer.
  * @param[in] queryKey  The key to search for.
  * @param[in] queryKeyLength  Length of the key.
  * @param[in] separator  A character between a key and a sub-key in queryKey.
@@ -71,8 +71,9 @@ JSONStatus_t JSON_Validate( const char * buf,
  * key strings joined by a separator.
  *
  * For example, if buf contains '{"foo":"abc","bar":{"foo","xyz"}}', then a 
- * search for 'foo' would output 'abc' and a search for 'bar.foo' would
- * output 'xyz' (given separator is specified as '.').
+ * search for 'foo' would output 'abc', 'bar' would output '{"foo","xyz"}',
+ * and a search for 'bar.foo' would output 'xyz' (given separator is
+ * specified as '.').
  *
  * On success, the pointer to the value points to a location in buf.  No null
  * termination is done for the value.  For valid JSON it is safe to place
@@ -101,11 +102,11 @@ JSONStatus_t JSON_Validate( const char * buf,
  * key and its value.  To validate the entire JSON document, use JSON_Validate().
  */
 JSONStatus_t JSON_Search( char * buf,
-                          size_t length,
+                          size_t max,
                           char * queryKey,
                           size_t queryKeyLength,
                           char separator,
                           char ** outValue,
                           size_t * outValueLength );
 
-#endif /* ifndef JSON_H */
+#endif /* ifndef JSON_H_ */
--- a/libraries/standard/json/include/json.h
+++ b/libraries/standard/json/include/json.h
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#ifndef JSON_H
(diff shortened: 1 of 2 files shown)


libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
+#include <assert.h>
+#include <ctype.h>
+#include <stddef.h>
+#include <stdint.h>
+#include "json.h"
+
+typedef enum bool
+{
+    true = 1, false = 0
+} bool;
```
courage, line 10: I think it would be good to choose a name that is not a keyword in C++. Being C++ compatible isn't a requirement, but in this case it seems easy enough to do, and it might save someone a porting headache down the line

libraries/standard/json/include/json.h
```diff
@@ -0,0 +1,111 @@
...
+ * @param[in] separator  A character between a key and a sub-key in queryKey.
+ * @param[out] outValue  A pointer to receive the address of the value found.
+ * @param[out] outValueLength  A pointer to receive the length of the value found.
+ *
+ * The JSON document must contain an object (e.g., '{"key":"value"}'). Any
+ * value may also be an object and so forth to a maximum depth.  A search
+ * may descend through nested objects when the queryKey contains matching
+ * key strings joined by a separator.
+ *
+ * For example, if buf contains '{"foo":"abc","bar":{"foo","xyz"}}', then a
```
courage, line 73: can you also document what happens if I search the example for 'bar'? I'm guessing a pointer to "{\"foo\",\"xyz\"}}", but there's no reason I should have to guess.

libraries/standard/json/include/json.h
```diff
@@ -0,0 +1,111 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#ifndef JSON_H
+#define JSON_H
```
yourslab, line 23: Nit: I think our convention is to have a trailing _ for include guards: `JSON_H` -> `JSON_H_`

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
+#include <assert.h>
```
yourslab, line 1: Missing copyright header

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
...
+
+    *start = i;
+    return ret;
+}
+
+/**
+ * @brief Find a key in a JSON object and output a pointer to its value.
+ *
+ * @param[in] buf  The buffer to search.
+ * @param[in] max  size of the buffer.
```
courage, line 885: is there any semantic difference between the max vs. length names? it looks to me like just an inconsistency, but maybe I'm missing something

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
...
+ *
+ * @return #JSONSuccess if the buffer contents are a valid JSON collection;
+ * #JSONIllegalDocument if the buffer contents are NOT valid JSON;
+ * #JSONMaxDepthExceeded if object and array nesting exceeds a threshold;
+ * #JSONPartial if the buffer contents are potentially valid but incomplete.
+ */
+#ifndef JSON_MAX_DEPTH
+    #define JSON_MAX_DEPTH    32
+#endif
+static JSONStatus_t match( const char * buf,
```
courage, line 711: i would have guessed this would be called skipCollection, which seems more indicative of what it does than "match". Is there a reason for the name difference?

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
...
+static bool nextKeyValuePair( const char * buf,
+                              size_t * start,
+                              size_t max,
+                              size_t * key,
+                              size_t * keyLength,
+                              size_t * value,
+                              size_t * valueLength )
+{
+    bool ret = true;
+    size_t i = *start, save = *start;
```
courage, line 835: I think it would be easier to read this function if you split "save" into two separate variables called something like "keyStart" and "valueStart". I got a little confused while reading because I didn't initially guess what the name "save" meant, and I also didn't notice that it was used twice.

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
...
+ * @brief Advance buffer index beyond one or more scalars.
+ *
+ * @param[in] buf  The buffer to parse.
+ * @param[in,out] start  The index at which to begin.
+ * @param[in] max  The size of the buffer.
+ *
+ * @return true if a scalar was present;
+ * false otherwise.
+ */
+static bool skipScalars( const char * buf,
```
courage, line 670: return value here doesn't seem to be used. other functions whose return value would be ignored (e.g. skipExponent) seem to return void. would be nice to either always return bool or always use void when the return value won't be used.

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
...
+ * @brief Advance buffer index beyond a UTF-8 code point.
+ *
+ * @param[in] buf  The buffer to parse.
+ * @param[in,out] start  The index at which to begin.
+ * @param[in] max  The size of the buffer.
+ *
+ * @return true if a valid code point was present;
+ * false otherwise.
+ */
+static bool skipUTF8MultiByte( const char * buf,
```
courage, line 68: this is a bit horrendous, but you may have to deal with the security considerations section of https://tools.ietf.org/html/rfc3629#page-10. I probably wrote the bug exploited in the "widespread virus attacking Web servers in 2001". I'd prefer not be responsible for a second version of the same attack

libraries/standard/json/src/json.c
dan4thewin: changed

libraries/standard/json/include/json.h
dan4thewin: added

libraries/standard/json/include/json.h
dan4thewin: fixed

libraries/standard/json/src/json.c
dan4thewin: added

libraries/standard/json/src/json.c
dan4thewin: No difference.  I made all the names the same.

libraries/standard/json/src/json.c
dan4thewin: changed

libraries/standard/json/src/json.c
dan4thewin: changed to use void for 3 functions

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1014 @@
+#include <assert.h>
+#include <ctype.h>
+#include <stddef.h>
+#include <stdint.h>
+#include "json.h"
+
+typedef enum bool
```
abhidixi11, line 7: Should we have these guards for C99 /C++ ?
#if defined( __cplusplus ) || ( defined( __STDC_VERSION__ ) && ( __STDC_VERSION__ >= 199901L ) )
    #include <stdbool.h>
#elif !defined( bool )
    typedef enum bool
    {
         true = 1, false = 0
    } bool;
#endif

dan4thewin: I've changed bool to bool_.  I think I'd like to keep this enum strictly local.

libraries/standard/json/src/json.c
dan4thewin: Challenge accepted!  Tell me if you think I've got it covered.

libraries/standard/json/src/json.c
```diff
@@ -0,0 +1,1173 @@
...
+    size_t i = 0U, start = 0U, keyLength = 0U;
+    char * p = buf;
+    size_t tmp = max;
+
+    assert( ( buf != NULL ) && ( max > 0U ) );
+    assert( ( queryKey != NULL ) && ( queryKeyLength != 0U ) );
+    assert( ( outValue != NULL ) && ( outValueLength != NULL ) );
+    assert( *queryKey != '\0' );
+
+    while( i < queryKeyLength )
```
courage, line 1147: Probably not too interesting for many documents, but I think this loop will not work quite right for queries that are either an empty string or end with a dot. E.g. if my document is '{"": {"": "I feel empty on the outside."}}', then I would expect querying for:

'' -> '{"": "I feel empty on the outside."}'.
'.' -> "I feel empty on the outside."

but the current code would return:

'' -> JSONPartial (this will happen no matter what is in the document!)
'.' -> '{"": "I feel empty on the outside."}' (missing a level of dereferencing)

dan4thewin: I think the assert would trip for the empty string - clearly I was thinking that was not a valid use case.  It seems to me that empty keys might be like escaped NULs -  they should probably be explicitly disallowed for the sake of security.

PR #1038, opened 2020-07-06
Fix for PINGRESP Deserializing

leegeth:
*Description of changes:*
A NULL check for pointer to remaining data is not required for PINGRESP. PINGRESP doesn't have any remaining data to point to as per MQTT 3.1.1 spec. A NULL for buffer should be valid in this case.

### DIFF
--- a/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
+++ b/libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
@@ -1354,13 +1354,15 @@ void test_MQTT_DeserializeAck_pingresp( void )
     uint8_t buffer[ 10 ] = { 0 };
 
     /* Bad remaining length. */
+    ( void ) memset( ( void * ) &mqttPacketInfo, 0x00, sizeof( mqttPacketInfo ) );
     mqttPacketInfo.type = MQTT_PACKET_TYPE_PINGRESP;
     mqttPacketInfo.remainingLength = MQTT_PACKET_PINGRESP_REMAINING_LENGTH + 1;
     status = MQTT_DeserializeAck( &mqttPacketInfo, &packetIdentifier, &sessionPresent );
     TEST_ASSERT_EQUAL_INT( MQTTBadResponse, status );
 
-    /* Process a valid PINGRESP. */
+    /* Process a valid PINGRESP with remaining data buffer NULL. */
     mqttPacketInfo.remainingLength = MQTT_PACKET_PINGRESP_REMAINING_LENGTH;
+    mqttPacketInfo.pRemainingData = NULL;
     status = MQTT_DeserializeAck( &mqttPacketInfo, NULL, NULL );
     TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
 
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -1354,14 +1354,21 @@ void test_MQTT_DeserializeAck_pingresp( void )
...
+    ( void ) memset( ( void * ) &mqttPacketInfo, 0x00, sizeof( mqttPacketInfo ) );
     mqttPacketInfo.type = MQTT_PACKET_TYPE_PINGRESP;
-    mqttPacketInfo.pRemainingData = buffer;
     mqttPacketInfo.remainingLength = MQTT_PACKET_PINGRESP_REMAINING_LENGTH + 1;
     status = MQTT_DeserializeAck( &mqttPacketInfo, &packetIdentifier, &sessionPresent );
     TEST_ASSERT_EQUAL_INT( MQTTBadResponse, status );
 
-    /* Process a valid PINGRESP. */
+    /* Process a valid PINGRESP with remaining data buffer NULL. */
+    mqttPacketInfo.remainingLength = MQTT_PACKET_PINGRESP_REMAINING_LENGTH;
```
aggarw13, line 1364: Is it possible that the buffer would be NULL but the remaining length be > 0? If not, should there be an internal assert to check that the `remainingLength` value is 0 for a `NULL` buffer?

leegeth: It is already checked here. https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/mqtt/src/mqtt_lightweight.c#L1158. MQTT_PACKET_PINGRESP_REMAINING_LENGTH is 0

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -1354,14 +1354,21 @@ void test_MQTT_DeserializeAck_pingresp( void )
...
+    mqttPacketInfo.pRemainingData = NULL;
+    status = MQTT_DeserializeAck( &mqttPacketInfo, NULL, NULL );
+    TEST_ASSERT_EQUAL_INT( MQTTSuccess, status );
+
+    /* Process a valid PINGRESP with remaining data buffer non-NULL. */
     mqttPacketInfo.remainingLength = MQTT_PACKET_PINGRESP_REMAINING_LENGTH;
+    mqttPacketInfo.pRemainingData = buffer;
```
aggarw13, line 1371: Should such a case of 0 remaining length value but a valid buffer be supported as a valid input parameter?

leegeth: We are not enforcing the buffer to be NULL for PINGRESP.  A non-NULL buffer pointer will just be ignored in this case

leegeth: As discussed, I agree that this additional test is not required and will be giving a wrong indication. Updated the PR to remove the test.

aggarw13: We discussed offline and decided that this test case is not required as it does not represent the intended use of the API.

PR #1039, opened 2020-07-07
CBMC Proofs for MQTT_SerializeConnect and MQTT_DeserializePublish

sarenameas:
*Description of changes:*
- Add allocators and isValid routines for all structures
- Add mqtt_config.h to disable logging.
- Add harness for MQTT_DeserializePublish
- Add harness for MQTT_SerializeConnect


libraries/standard/mqtt/cbmc/include/mqtt_config.h
```diff
@@ -0,0 +1,60 @@
...
+ * context maintains.
+ */
+#define MQTT_STATE_ARRAY_MAX_COUNT          10U
+
+/**
+ * @brief MQTT client identifier.
+ *
+ * No two clients may use the same client identifier simultaneously.
+ */
+#define MQTT_CLIENT_IDENTIFIER              "cbmcclient"
```
yourslab, line 57: Should this be defined in the `mqtt_config.h`?

sarenameas: Copied from the unit test config, it might be out of date?

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
```diff
@@ -18,13 +18,139 @@
...
+}
+
+bool isValidMqttPublishInfo( const MQTTPublishInfo_t * pPublishInfo )
+{
+    bool isValid = true;
+
+    if( pPublishInfo != NULL )
+    {
+        bool validQos = ( ( pPublishInfo->qos >= MQTTQoS0 ) ||
+                          ( pPublishInfo->qos <= MQTTQoS2 ) );
```
yourslab, line 85: I believe this should be `&&`

sarenameas: Thanks

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
```diff
@@ -18,13 +18,139 @@
...
+
+bool isValidMqttConnectInfo( const MQTTConnectInfo_t * pConnectInfo )
+{
+    bool isValid = true;
+
+    if( pConnectInfo != NULL )
+    {
+        isValid = isValid && ( pConnectInfo->clientIdentifierLength < CBMC_MAX_OBJECT_SIZE );
+        isValid = isValid && ( pConnectInfo->userNameLength < CBMC_MAX_OBJECT_SIZE );
+        isValid = isValid && ( pConnectInfo->passwordLength < CBMC_MAX_OBJECT_SIZE );
```
yourslab, line 124: Does `isValid &= ( pConnectInfo->passwordLength < CBMC_MAX_OBJECT_SIZE );` also work?

sarenameas: No that is a bitwise and

.gitignore
```diff
@@ -18,3 +18,4 @@ build/
 **/cbmc/proofs/**/html
 **/cbmc/proofs/**/logs
 **/cbmc/proofs/**/TAGS-*
+**/cbmc/proofs/**/report
```
yourslab, line 21: `**/cbmc/proofs/**/gotos` might also be a good thing to add here

libraries/standard/mqtt/cbmc/include/mqtt_config.h
yourslab: Possible because it's not being used anywhere in the MQTT library. For the demos, we have this defined in `demo_config.h`.

libraries/standard/mqtt/cbmc/include/mqtt_cbmc_state.h
```diff
@@ -18,11 +18,92 @@
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
+#ifndef MQTT_CBMC_STATE_H_
+#define MQTT_CBMC_STATE_H_
+
+#include <stdbool.h>
+
+#include "mqtt.h"
```
markrtuttle, line 26: Not sure you want the project header file pulled into a proof header file, but probably okay

libraries/standard/mqtt/cbmc/include/mqtt_config.h
```diff
@@ -0,0 +1,40 @@
...
+/* Configure name and log level for the MQTT library. */
+#define LIBRARY_LOG_NAME     "MQTT"
+#define LIBRARY_LOG_LEVEL    LOG_NONE
+
+#include "logging_stack.h"
+
+/************ End of logging configuration ****************/
+
+/* Set network context to a socket (int). */
+typedef int NetworkContext_t;
```
markrtuttle, line 26: Where did this come from?  Is this the hack of setting NetworkContext to an int (or int *) because in the real code it is typedef'd to an SSL struct?   I would document what is going on here, an that this is replacing a definition in some header file.

libraries/standard/mqtt/cbmc/include/mqtt_config.h
```diff
@@ -0,0 +1,40 @@
...
+ * @brief The maximum number of MQTT PUBLISH messages that may be pending
+ * acknowledgement at any time.
+ *
+ * QoS 1 and 2 MQTT PUBLISHes require acknowledgement from the server before
+ * they can be completed. While they are awaiting the acknowledgement, the
+ * client must maintain information about their state. The value of this
+ * macro sets the limit on how many simultaneous PUBLISH states an MQTT
+ * context maintains.
+ */
+#define MQTT_STATE_ARRAY_MAX_COUNT    ( 10U )
```
markrtuttle, line 38: Is this protocol definition or a bound for the sake of proof?

This file feels like it is mixing MQTT protocol configuration with MQTT proof bounds and stubs.  If that is the case, I would put all of the MQTT stuff in one place (eg the state header).  It is important for a newcomer to be able to go to exactly one place (or two well-documented places) to find all the proof assumptions (eg, queue bounds) and stubs.

libraries/standard/mqtt/cbmc/proofs/MQTT_GetIncomingPacketTypeAndLength/MQTT_GetIncomingPacketTypeAndLength_harness.c
```diff
@@ -34,13 +34,14 @@ void harness()
      * MQTT_GetIncomingPacketTypeAndLength(). */
     NetworkContext_t networkContext;
 
-    __CPROVER_assume( networkContext != NULL );
```
markrtuttle, line 37: NetworkContext_t is typedef'd above to an int and not to a pointer.

libraries/standard/mqtt/cbmc/proofs/MQTT_GetIncomingPacketTypeAndLength/MQTT_GetIncomingPacketTypeAndLength_harness.c
```diff
@@ -34,13 +34,14 @@ void harness()
...
-
     /* MQTT_GetIncomingPacketTypeAndLength() will set only the remainingLength
      * field in the input MQTTPacketInfo_t structure. */
-    MQTTPacketInfo_t * pIncomingPacketInfo = mallocCanFail( sizeof( MQTTPacketInfo_t ) );
+    MQTTPacketInfo_t * pIncomingPacket = NULL;
+
+    pIncomingPacket = allocateMqttPacketInfo( pIncomingPacket );
```
markrtuttle, line 41: Did you mean to to change the name from PacketInfo to Packet?  They seem like two different things.

libraries/standard/mqtt/cbmc/proofs/MQTT_Init/MQTT_Init_harness.c
```diff
@@ -25,16 +25,17 @@
...
 void harness()
 {
-    MQTTContext_t context;
-    MQTTTransportInterface_t transportInterface;
-    MQTTApplicationCallbacks_t callbacks;
-    MQTTFixedBuffer_t networkBuffer;
+    MQTTContext_t * pContext = mallocCanFail( sizeof( MQTTContext_t ) );
+    MQTTTransportInterface_t * pTransportInterface = mallocCanFail( sizeof( MQTTContext_t ) );
+    MQTTApplicationCallbacks_t * pCallbacks = mallocCanFail( sizeof( MQTTApplicationCallbacks_t ) );
+    MQTTFixedBuffer_t * pNetworkBuffer = mallocCanFail( sizeof( MQTTFixedBuffer_t ) );
```
markrtuttle, line 35: This is different from the style you use above of initializing pointers to NULL at declaration, then assigning the struct pointer below.  I thought the former was the group style.

libraries/standard/mqtt/cbmc/proofs/MQTT_Init/MQTT_Init_harness.c
```diff
@@ -25,16 +25,17 @@
...
 
-    MQTT_Init( nondet_bool() ? NULL : &context,
-               nondet_bool() ? NULL : &transportInterface,
-               nondet_bool() ? NULL : &callbacks,
-               nondet_bool() ? NULL : &networkBuffer );
+    MQTT_Init( pContext,
+               pTransportInterface,
+               pCallbacks,
+               pNetworkBuffer );
```
markrtuttle, line 40: You know, you could just use the mallocCanFail invocations in the MQTT_Init invocation and forget all about the type declarations above.

libraries/standard/mqtt/cbmc/proofs/MQTT_Init/Makefile
```diff
@@ -8,6 +8,7 @@ REMOVE_FUNCTION_BODY +=
 UNWINDSET +=
 
 PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROOF_SOURCES += $(PROOFDIR)/../../sources/mqtt_cbmc_state.c
```
markrtuttle, line 11: I like this style

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/MQTT_SerializeConnect_harness.c
```diff
@@ -0,0 +1,75 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file MQTT_SerializeConnect_harness.c
+ * @brief Implements the proof harness for MQTT_SerializeConnect function.
+ */
+#include "mqtt.h"
+#include "mqtt_cbmc_state.h"
+
```
markrtuttle, line 28: Are these defines below proof assumptions?  Would it make sense to define these in the Makefile?  Or at least #ifdef/#endif these defines so they can be overridden from the Makefile.  The most important thing, though, it to have one system for collecting proof assumptions in one place and use it everywhere.

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/MQTT_SerializeConnect_harness.c
```diff
@@ -0,0 +1,75 @@
...
+    * fit into the MQTTFixedBuffer_t. It is a violation of the API to call
+    * MQTT_SerializeConnect without first calling MQTT_GetConnectPacketSize(). */
+    if( pConnectInfo != NULL )
+    {
+        status = MQTT_GetConnectPacketSize( pConnectInfo, pWillInfo, &remainingLength, &packetSize );
+    }
+
+    if( status == MQTTSuccess )
+    {
+        MQTT_SerializeConnect( pConnectInfo, pWillInfo, remainingLength, pBuffer );
```
markrtuttle, line 73: Because status is initialized to success, this means SerializeConnect can be called with pConnectInfo == NULL, is that what you intended?

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/Makefile
```diff
@@ -0,0 +1,17 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+HARNESS_ENTRY=harness
+HARNESS_FILE=MQTT_SerializeConnect_harness
+
+DEFINES +=
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET += encodeRemainingLength.0:9
```
markrtuttle, line 11: My memory is that even 5 was enough for this loop unwinding, but I haven't read this version of the code.

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
```diff
@@ -18,13 +18,139 @@
  * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
-#include <stddef.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include "mqtt_cbmc_state.h"
+#include "mqtt.h"
```
markrtuttle, line 24: You are including this mqtt.h both here and in mqtt_cbmc_state.h.  I prefer removing it from mqtt_cbmc_state.h.  I like this style here.

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
```diff
@@ -18,13 +18,139 @@
...
+    }
+
+    if( pPacketInfo != NULL )
+    {
+        __CPROVER_assume( pPacketInfo->remainingLength < CBMC_MAX_OBJECT_SIZE );
+        pPacketInfo->pRemainingData = mallocCanFail( pPacketInfo->remainingLength );
+    }
+
+    return pPacketInfo;
+}
```
markrtuttle, line 46: Really, really nice.

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
```diff
@@ -18,13 +18,139 @@
...
+{
+    bool isValid = true;
+
+    if( pPacketInfo != NULL )
+    {
+        isValid = pPacketInfo->remainingLength < CBMC_MAX_OBJECT_SIZE;
+    }
+
+    return isValid;
+}
```
markrtuttle, line 58: This means that packetInfo == NULL is a valid packetInfo.  I kind of like this, but CBMC proof projects are inconsistent from project to project about whether valid means nonnull.  I like this style if you can make it work, but it needs to be used consistently.

libraries/standard/mqtt/cbmc/include/mqtt_cbmc_state.h
sarenameas: This is for the MQTT struct types

libraries/standard/mqtt/cbmc/include/mqtt_config.h
sarenameas: It can be typedef to whatever we want. We don't use it in the network interface stub.

libraries/standard/mqtt/cbmc/include/mqtt_config.h
sarenameas: I added this config header because I needed to disable logging in the library. This symbol is required for MQTT and must be defined. Before we took the definition from the plaintext demo. 
This is a typical count for all of the demos and not for the sake of bounding a proof.

libraries/standard/mqtt/cbmc/proofs/MQTT_GetIncomingPacketTypeAndLength/MQTT_GetIncomingPacketTypeAndLength_harness.c
sarenameas: This is deleted.

libraries/standard/mqtt/cbmc/proofs/MQTT_GetIncomingPacketTypeAndLength/MQTT_GetIncomingPacketTypeAndLength_harness.c
sarenameas: Yes because that is the how the parameter is named in the function. This information about the packet contains the actual packet contents in a pointer to a buffer it's fields.

libraries/standard/mqtt/cbmc/proofs/MQTT_Init/MQTT_Init_harness.c
sarenameas: I will change back to the other style for consistency.

libraries/standard/mqtt/cbmc/proofs/MQTT_Init/MQTT_Init_harness.c
sarenameas: yes, but for consistency I will just leave it.

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/MQTT_SerializeConnect_harness.c
sarenameas: Yes thanks for the advice. I left these during experimentation and I am not using them, so I will delete them.

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/MQTT_SerializeConnect_harness.c
sarenameas: Yes because I want coverage of the pConnectInfo null check.

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/Makefile
sarenameas: The  RemainingLength to encode will never be larger than 327700, so 3 is enough times to divide by 128 to reach 0 when the loop would end.

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
sarenameas: I will take it out of here and leave int mqtt_cbmc_state.h because mqtt_cbmc_state.h has types from this file in it.

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
sarenameas: Yes a NULL packet info is a valid packet info.

libraries/standard/mqtt/cbmc/proofs/MQTT_Init/MQTT_Init_harness.c
```diff
@@ -29,10 +29,15 @@
 
 void harness()
 {
-    MQTTContext_t * pContext = mallocCanFail( sizeof( MQTTContext_t ) );
-    MQTTTransportInterface_t * pTransportInterface = mallocCanFail( sizeof( MQTTContext_t ) );
-    MQTTApplicationCallbacks_t * pCallbacks = mallocCanFail( sizeof( MQTTApplicationCallbacks_t ) );
-    MQTTFixedBuffer_t * pNetworkBuffer = mallocCanFail( sizeof( MQTTFixedBuffer_t ) );
+    MQTTContext_t * pContext = NULL;
```
sarenameas, line 32: Initialized to NULL for consistency with other proofs.

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
```diff
@@ -21,7 +21,6 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include "mqtt_cbmc_state.h"
-#include "mqtt.h"
```
sarenameas: already included in the .h file.

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/Makefile
```diff
@@ -8,7 +8,7 @@ DEFINES +=
 INCLUDES +=
 
 REMOVE_FUNCTION_BODY +=
-UNWINDSET += encodeRemainingLength.0:9
+UNWINDSET += encodeRemainingLength.0:3
```
sarenameas, line 11: log_128_(327700) = ~2.6

libraries/standard/mqtt/cbmc/include/mqtt_config.h
markrtuttle: I really think a comment saying this is a stub would be valuable.

libraries/standard/mqtt/cbmc/include/mqtt_config.h
markrtuttle: I really think a one-line comment saying "10 is a typical value used in demos" would be valuable (we get hammered so hard for bounding lengths, it is relief to be able to say this is not a limitation of the proof).

libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/Makefile
markrtuttle: Not sure why this comment is here, is this in response to a comment I made about replacing 9 with 5?

libraries/standard/mqtt/cbmc/include/mqtt_config.h
sarenameas: ok

libraries/standard/mqtt/cbmc/include/mqtt_config.h
sarenameas: yes thank you

libraries/standard/mqtt/cbmc/include/mqtt_config.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**************************************************/
+/******* DO NOT CHANGE the following order ********/
+/**************************************************/
+
+/* Logging related header files are required to be included in the following order:
+ * 1. Include the header file "logging_levels.h".
+ * 2. Define LIBRARY_LOG_NAME and  LIBRARY_LOG_LEVEL.
+ * 3. Include the header file "logging_stack.h".
+ */
```
dan4thewin, line 12: I feel like I've seen this before - is this a copy?  Can we include the original?  Also - this text is rather unfriendly.  It would be better to explain why a given order is necessary.

libraries/standard/mqtt/cbmc/proofs/MQTT_DeserializeAck/Makefile
```diff
@@ -0,0 +1,17 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
```
dan4thewin, line 2: Apache??

libraries/standard/mqtt/cbmc/proofs/MQTT_DeserializeAck/Makefile
```diff
@@ -0,0 +1,17 @@
...
+HARNESS_FILE=MQTT_DeserializeAck_harness
+
+DEFINES +=
+INCLUDES +=
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET +=
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROOF_SOURCES += $(PROOFDIR)/../../sources/mqtt_cbmc_state.c
```
dan4thewin, line 14: sources, not src?

libraries/standard/mqtt/cbmc/proofs/MQTT_DeserializeAck/Makefile
```diff
@@ -0,0 +1,17 @@
...
+PROJECT_SOURCES += $(SRCDIR)/libraries/standard/mqtt/src/mqtt_lightweight.c
+
+include ../Makefile.common
```
dan4thewin, line 17: I see there are more Makefiles below with few differences, so please consider my questions asked for all, and likewise questions about any other repetitious content.

libraries/standard/mqtt/cbmc/proofs/MQTT_GetIncomingPacketTypeAndLength/MQTT_GetIncomingPacketTypeAndLength_harness.c
dan4thewin: So, it has type of MQTTPacketInfo_t but we call it pIncomingPacket because the parameter has that name for MQTT_Get...()?  Is this autogenerated? Does it matter if they match?  Do we have inconsistencies in our API?

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
dan4thewin: ELI5, please.

libraries/standard/mqtt/cbmc/proofs/MQTT_DeserializeAck/Makefile
sarenameas: That is the layout from the CBMC templates. Changing it here would warrant a change in the templates.

libraries/standard/mqtt/cbmc/proofs/MQTT_DeserializeAck/Makefile
sarenameas: This is auto copied from the CBMC templates.

libraries/standard/mqtt/cbmc/proofs/MQTT_DeserializeAck/Makefile
sarenameas: I can discuss with Mark about the makefile text.

libraries/standard/mqtt/cbmc/include/mqtt_config.h
sarenameas: The proofs need to have their own mqtt_config.h to disable logging.

libraries/standard/mqtt/cbmc/proofs/MQTT_GetIncomingPacketTypeAndLength/MQTT_GetIncomingPacketTypeAndLength_harness.c
sarenameas: All similiar functions in the API have the parameter name "pIncomingPacket" there doesn't appear to be an inconsistency. It doesn't really matter if they match, but I was just saying that this "pIncomingPacket" parameter name that was chosen I guess I can see the arguments for it's name.

libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
sarenameas: @markrtuttle  All isvalid routines in this file have this same style.

@dan4thewin A NULL packet info is valid because we want the proof, that calls this fuction, to accept a NULL packet info to pass into the function to be proved. This allows us to get full coverage and make sure the the function returns an error for NULL parameters.

dan4thewin: I understand that logic, but I would have thought a simple NULL check would be in a unit test.  Is that an integral part of a this proof?

PR #1041, opened 2020-07-08
Repair loop names mangled by the new cbmc.

markrtuttle:
This pull request prepares CBMC proofs in the development branch for the upgrade of CBMC used in CI.  This pull request changes two loop names mangled by the new CBMC.


libraries/standard/http/cbmc/proofs/HTTPClient_AddRangeHeader/Makefile
```diff
@@ -9,7 +9,9 @@ INCLUDES +=
 
 REMOVE_FUNCTION_BODY +=
 # Maximum value of a 32 bit signed integer is 2,147,483,647, which is 10 digits.
-UNWINDSET += strncmp.0:5 convertInt32ToAscii.0:11 convertInt32ToAscii.1:11
+UNWINDSET += strncmp.0:5
+UNWINDSET += __CPROVER_file_local_http_client_c_convertInt32ToAscii.0:11
+UNWINDSET += __CPROVER_file_local_http_client_c_convertInt32ToAscii.1:11
```
yourslab, line 14: This is for the new CBMC version right? Do we need to do this for all functions from project sources moving forward?

markrtuttle: I just noticed this comment.  I don't think so.  I believe this patch is sufficient for the sdk.  This repairs the old proofs for the use of the --export-file-local-symbols flag, and future proofs are already being written to use it.

PR #1069, opened 2020-07-22
Fix failing MQTT unit test in Release builds

aggarw13:
### Problem
The `test_MQTT_ProcessLoop_handleIncomingAck_Error_Paths` unit test fails in Release build (`-DCMAKE_BUILD_TYPE=Release`) but does not fail for Debug builds.

### Cause
The failure occurs from the network buffer not being initialized, thereby, causing the `MQTTFixedBuffer.size` value to be zero. Somehow, the Debug builds actually default initialize the stack variable to a value of `128`, causing the tests to pass.

### Corrective Action
We will update the CI checks to run unit tests with Release builds.

### DIFF
--- a/libraries/standard/mqtt/utest/mqtt_utest.c
+++ b/libraries/standard/mqtt/utest/mqtt_utest.c
@@ -12,80 +12,80 @@
 /**
  * @brief A valid starting packet ID per MQTT spec. Start from 1.
  */
-#define MQTT_FIRST_VALID_PACKET_ID          ( 1 )
+#define MQTT_FIRST_VALID_PACKET_ID             ( 1 )
 
 /**
  * @brief A PINGREQ packet is always 2 bytes in size, defined by MQTT 3.1.1 spec.
  */
-#define MQTT_PACKET_PINGREQ_SIZE            ( 2U )
+#define MQTT_PACKET_PINGREQ_SIZE               ( 2U )
 
 /**
  * @brief A packet type not handled by MQTT_ProcessLoop.
  */
-#define MQTT_PACKET_TYPE_INVALID            ( 0U )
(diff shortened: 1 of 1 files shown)


libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -1589,6 +1592,7 @@ void test_MQTT_ProcessLoop_handleIncomingAck_Error_Paths( void )
 
     setupTransportInterface( &transport );
     setupCallbacks( &callbacks );
+    setupNetworkBuffer( &networkBuffer );
```
aggarw13, line 1595: This was the _fix_ for the `test_MQTT_ProcessLoop_handleIncomingAck_Error_Paths` test

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -12,80 +12,80 @@
...
  * multiple iterations spent in the process loop for coverage.
  */
-#define MQTT_SAMPLE_TIMEOUT_MS              ( 1U )
+#define MQTT_SAMPLE_PROCESS_LOOP_TIMEOUT_MS    ( 1U )
 
 /**
  * @brief Zero timeout in the process loop implies one iteration.
  */
-#define MQTT_NO_TIMEOUT_MS                  ( 0U )
+#define MQTT_ZERO_PROCESS_LOOP_TIMEOUT_MS      ( 0U )
```
muneebahmed10, line 51: I though `MQTT_NO_TIMEOUT_MS` was already pretty clear enough as is. I think this change mainly makes lines longer instead of clarifying anything

aggarw13: Fair enough. Will revert it!

PR #1073, opened 2020-07-23
Update mqtt.c send and recv functions to handle more bytes and expected.

sarenameas:
- Add some documentation to application callbacks.
- Update recvExact and sendPacket to handle the case of the transport layer returning more bytes than expected.
- Add unit tests for changes.
- Fix MISRA flags left in mqtt_lightweight.c and straggling pFixedBuffer variables.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -963,6 +963,7 @@ static MQTTStatus_t receiveSingleIteration( MQTTContext_t * pContext,
     MQTTPacketInfo_t incomingPacket;
 
     assert( pContext != NULL );
+    assert( pContext->networkBuffer.pBuffer != NULL );
 
     status = MQTT_GetIncomingPacketTypeAndLength( pContext->transportInterface.recv,
                                                   pContext->transportInterface.pNetworkContext,
@@ -991,17 +992,9 @@ static MQTTStatus_t receiveSingleIteration( MQTTContext_t * pContext,
     }
     else
     {
-        if( pContext->networkBuffer.pBuffer != NULL )
-        {
-            /* Receive packet. Remaining time is recalculated before calling this
-             * function. */
-            status = receivePacket( pContext, incomingPacket, remainingTimeMs );
-        }
-        else
-        {
-            LogError( ( "The MQTT context's networkBuffer must not be NULL." ) );
-            status = MQTTBadParameter;
-        }
+        /* Receive packet. Remaining time is recalculated before calling this
+         * function. */
+        status = receivePacket( pContext, incomingPacket, remainingTimeMs );
     }
 
     /* Handle received packet. If no data was read then this will not execute. */
@@ -1795,20 +1788,24 @@ MQTTStatus_t MQTT_ProcessLoop( MQTTContext_t * pContext,
     MQTTGetCurrentTimeFunc_t getTimeStampMs = NULL;
     uint32_t entryTimeMs = 0U, remainingTimeMs = timeoutMs, elapsedTimeMs = 0U;
 
-    if( ( pContext != NULL ) && ( pContext->callbacks.getTime != NULL ) )
-    {
-        getTimeStampMs = pContext->callbacks.getTime;
-        entryTimeMs = getTimeStampMs();
-        status = MQTTSuccess;
-        pContext->controlPacketSent = false;
-    }
-    else if( pContext == NULL )
+    if( pContext == NULL )
     {
         LogError( ( "MQTT Context cannot be NULL." ) );
     }
-    else
+    else if( pContext->callbacks.getTime == NULL )
     {
         LogError( ( "MQTT Context must set callbacks.getTime." ) );
+    }
+    else if( pContext->networkBuffer.pBuffer == NULL )
+    {
+        LogError( ( "The MQTT context's networkBuffer must not be NULL." ) );
+    }
+    else
+    {
+        getTimeStampMs = pContext->callbacks.getTime;
+        entryTimeMs = getTimeStampMs();
+        pContext->controlPacketSent = false;
+        status = MQTTSuccess;
     }
 
     while( status == MQTTSuccess )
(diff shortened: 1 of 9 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -55,7 +50,7 @@
  * @brief param[in] pBufferToSend Buffer to be sent to network.
  * @brief param[in] bytesToSend Number of bytes to be sent.
  *
- * @return Total number of bytes sent; -1 if there is an error.
+ * @return Total number of bytes sent, or negative number on network error.
```
sarenameas, line 53: Consistency with recvExact()

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -326,33 +322,41 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
     /* Loop until the entire packet is sent. */
-    while( bytesRemaining > 0UL )
+    while( ( bytesRemaining > 0UL ) && ( sendError == false ) )
     {
         bytesSent = pContext->transportInterface.send( pContext->transportInterface.pNetworkContext,
                                                        pIndex,
                                                        bytesRemaining );
 
-        if( bytesSent > 0 )
+        if( bytesSent <= 0 )
```
sarenameas, line 331: By first checking if unsigned bytesSent is less than zero, we avoid possible loss of information from casting bytesSent to unsigned in the second else case on line 337. 
By doing the error checks first we do same thing as checking the success condition (  ( bytesSend > 0 ) && (bytesSent <= bytesRemaining) ).

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -326,33 +322,41 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+        {
+            LogError( ( "Transport send failed. Error code=%d.", bytesSent ) );
+            totalBytesSent = bytesSent;
+            sendError = true;
+        }
+        else if( ( size_t ) bytesSent > bytesRemaining )
+        {
+            LogError( ( "Transport send returned more bytes than expected: "
+                        "BytesExpected=%lu, BytesReceived=%d",
```
leegeth, line 340: Minor: Please end the log message with a '.'

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -326,33 +322,41 @@ static int32_t sendPacket( MQTTContext_t * pContext,
...
+                        bytesRemaining,
+                        bytesSent ) );
+            sendError = true;
```
leegeth, line 343: What would the function return in this case?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -433,19 +437,37 @@ static int32_t recvExact( const MQTTContext_t * pContext,
...
         {
             LogError( ( "Network error while receiving packet: ReturnCode=%d",
                         bytesRecvd ) );
             totalBytesRecvd = bytesRecvd;
             receiveError = true;
         }
+        else if( bytesRecvd > ( int32_t ) bytesRemaining )
+        {
+            LogError( ( "Transport receive returned more bytes than expected: "
+                        "BytesExpected=%lu, BytesReceived=%d",
```
leegeth, line 450: Add a '.' at the end of logging message.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -433,19 +437,37 @@ static int32_t recvExact( const MQTTContext_t * pContext,
...
+             * function will be less than bytesToRecv, which will be seen as an
+             * error. */
+        }
+        else
+        {
+            bytesRemaining -= ( size_t ) bytesRecvd;
+            totalBytesRecvd += ( int32_t ) bytesRecvd;
+            pIndex += bytesRecvd;
+            LogDebug( ( "BytesReceived=%d, BytesRemaining=%lu, "
+                        "TotalBytesReceived=%d",
```
leegeth, line 466: Please add a '.' at the end of logging message.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -965,9 +988,17 @@ static MQTTStatus_t receiveSingleIteration( MQTTContext_t * pContext,
     }
     else
     {
-        /* Receive packet. Remaining time is recalculated before calling this
-         * function. */
-        status = receivePacket( pContext, incomingPacket, remainingTimeMs );
+        if( pContext->networkBuffer.pBuffer != NULL )
```
leegeth, line 991: do we need to check it here as we have already asserted this in line 563?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1051,7 +1082,7 @@ static MQTTStatus_t sendPublish( MQTTContext_t * pContext,
                             pContext->networkBuffer.pBuffer,
                             headerSize );
 
-    if( bytesSent < 0 )
+    if( ( bytesSent < 0 ) || ( ( size_t ) bytesSent != headerSize ) )
```
leegeth, line 1085: Isn't the check ( ( size_t ) bytesSent != headerSize ) just enough here?

libraries/standard/mqtt/src/mqtt.c
sarenameas: Yes because the pattern of assert() is for developer errors. It is to follow a function where the condition has been checked previously outside of the assert.

libraries/standard/mqtt/src/mqtt.c
sarenameas: My reasoning here is that if bytesSent is a negative number, then casting it immediately to unsigned will lose information.

libraries/standard/mqtt/src/mqtt.c
sarenameas: Sure

libraries/standard/mqtt/src/mqtt.c
sarenameas: Let me leave the same comments as the recvExcat.

libraries/standard/mqtt/src/mqtt.c
sarenameas: Sure fixed everywhere it is missing in mqtt library.

libraries/standard/mqtt/src/mqtt.c
leegeth: I realized github hid the lines and I was taking a look at a wrong line number.

sarenameas: Per offline conversation we have decided to move this check to the public function scope.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -433,18 +441,35 @@ static int32_t recvExact( const MQTTContext_t * pContext,
...
-            bytesRemaining -= ( size_t ) bytesRecvd;
-            totalBytesRecvd += ( int32_t ) bytesRecvd;
-            pIndex += bytesRecvd;
+            LogError( ( "Network error while receiving packet: ReturnCode=%d.",
+                        bytesRecvd ) );
+            totalBytesRecvd = bytesRecvd;
+            receiveError = true;
         }
-        else
+        else if( bytesRecvd > ( int32_t ) bytesRemaining )
```
cobusve, line 451: How is it possible for the receive function to return more bytes than you ask for? The idea for this pattern is to limit the received bytes to fit into your buffer, basically thou shalt not receive more than this, so you should be guaranteed that this condition is never true.

sarenameas: If there is a bug in the applications implementation.

sarenameas: per offline conversation this will become an assert.

PR #1074, opened 2020-07-23
Exclude mutual auth demo from default targets if no credentials provided

yourslab:
It used to be that the mutual auth demos were not added as a target when no credentials are provided. Now, we simply remove it from the ALL target so that `make` will no longer build the mutual auth demos. However, a user can always see the mutual auth targets through `make help`.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -7,6 +7,34 @@ if( NOT DEFINED PLATFORM_NAME )
     endif()
 endif()
 
+# This function will add a demo target to the list of default targets
+# if all AWS Credentials are defined.
+function(check_aws_credentials demo_name)
+    if(AWS_IOT_ENDPOINT AND CLIENT_CERT_PATH AND CLIENT_PRIVATE_KEY_PATH)
+        target_compile_definitions(
+            ${demo_name} PRIVATE
+                AWS_IOT_ENDPOINT="${AWS_IOT_ENDPOINT}"
+                CLIENT_CERT_PATH="${CLIENT_CERT_PATH}"
+                CLIENT_PRIVATE_KEY_PATH="${CLIENT_PRIVATE_KEY_PATH}"
+        )
+        set(CMAKE_REQUIRED_DEFINITIONS -DAWS_IOT_ENDPOINT -DCLIENT_CERT_PATH -DCLIENT_PRIVATE_KEY_PATH)
+    endif()
+
+    set(FILES_TO_CHECK "demo_config.h")
+    list(APPEND CMAKE_REQUIRED_INCLUDES "${CMAKE_CURRENT_LIST_DIR};${LOGGING_INCLUDE_DIRS}")
+    unset(HAVE_AWS_ENDPOINT CACHE)
+    unset(HAVE_CLIENT_CERT CACHE)
+    unset(HAVE_PRIVATE_KEY CACHE)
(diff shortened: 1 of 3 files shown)


demos/CMakeLists.txt
```diff
@@ -7,6 +7,34 @@ if( NOT DEFINED PLATFORM_NAME )
...
+    list(APPEND CMAKE_REQUIRED_INCLUDES "${CMAKE_CURRENT_LIST_DIR};${LOGGING_INCLUDE_DIRS}")
+    unset(HAVE_AWS_ENDPOINT CACHE)
+    unset(HAVE_CLIENT_CERT CACHE)
+    unset(HAVE_PRIVATE_KEY CACHE)
+    check_symbol_exists(AWS_IOT_ENDPOINT ${FILES_TO_CHECK} HAVE_AWS_ENDPOINT)
+    check_symbol_exists(CLIENT_CERT_PATH ${FILES_TO_CHECK} HAVE_CLIENT_CERT)
+    check_symbol_exists(CLIENT_PRIVATE_KEY_PATH ${FILES_TO_CHECK} HAVE_PRIVATE_KEY)
+    
+    if(NOT(HAVE_AWS_ENDPOINT AND HAVE_CLIENT_CERT AND HAVE_PRIVATE_KEY))
+        message("To run ${demo_name}, define AWS_IOT_ENDPOINT, CLIENT_CERT_PATH, CLIENT_PRIVATE_KEY_PATH in demos/${demo_name}/demo_config.h.")
```
muneebahmed10, line 33: This path looks wrong
```suggestion
        message("To run ${demo_name}, define AWS_IOT_ENDPOINT, CLIENT_CERT_PATH, CLIENT_PRIVATE_KEY_PATH in demos/mqtt/${demo_name}/demo_config.h.")
```

demos/CMakeLists.txt
```diff
@@ -7,6 +7,34 @@ if( NOT DEFINED PLATFORM_NAME )
...
+function(check_aws_credentials demo_name)
+    if(AWS_IOT_ENDPOINT AND CLIENT_CERT_PATH AND CLIENT_PRIVATE_KEY_PATH)
+        target_compile_definitions(
+            ${demo_name} PRIVATE
+                AWS_IOT_ENDPOINT="${AWS_IOT_ENDPOINT}"
+                CLIENT_CERT_PATH="${CLIENT_CERT_PATH}"
+                CLIENT_PRIVATE_KEY_PATH="${CLIENT_PRIVATE_KEY_PATH}"
+        )
+        set(CMAKE_REQUIRED_DEFINITIONS -DAWS_IOT_ENDPOINT -DCLIENT_CERT_PATH -DCLIENT_PRIVATE_KEY_PATH)
+    endif()
```
muneebahmed10, line 21: Question: Is this so that parameters passed from the command line will be counted as well? I think the current code ignores command line parameters when determining build targets so that's a good change to have.

yourslab: Yup, this will allow the user to set the credentials without having to modify the demo config header.

yourslab: One thing I learned is that you can automatically set macros by just passing a CMake command line parameter: `-DSAMPLE_MACRO=1`

However, it doesn't work for strings unless you set it manually with `target_compile_definitions`.

demos/CMakeLists.txt
yourslab: Unfortunately, can't set it to mqtt because it's also being used for http so I am just going to be using ${demo_name}/demo_config.h

PR #1077, opened 2020-07-24
Update MQTT_Init API function to accept parameters by value

aggarw13:
Improve `MQTT_Init` API signature by passing input parameters by value instead of by pointers.
**Reason**: Passing by pointer usually gives the opinion of either:
- An output parameter
- An object who memory value will be stored by a state object 
In `MQTT_Init` function, the input parameters' values are _copied_ in the internal implementation of the function. There is no need to pass the parameters by pointer.


aggarw13:
Closing the PR due to the advantage of avoiding copy operations in stack when passing parameters by pointers (Had gone into a zone of past-time C++ habits of Object-Oriented Design 😛 )

PR #1078, opened 2020-07-24
Integ tests/mqtt restore session

aggarw13:
Add integration tests for a "session present" MQTT connection with the broker.  The test verifies that the MQTT library completes the PUBLISH operation, that was interrupted in a previous MQTT connection, in the restored session connection.


libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -264,6 +272,7 @@ static MQTTPublishInfo_t incomingInfo;
  * Session present response is obtained from the CONNACK from broker.
  */
 static void establishMqttSession( MQTTContext_t * pContext,
+                                  bool initContext,
```
muneebahmed10, line 275: 1. This parameter name doesn't match the comment or the one in the function definition, `initializeContext`.
2. Is this even necessary? Whenever `establishMqttSession` is called, this flag seems to have the same value as `createCleanSession`

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -780,13 +803,12 @@ void test_MQTT_Subscribe_Publish_With_Qos_2( void )
...
 void test_MQTT_Connect_LWT( void )
 {
-    int secondTcpSocket;
     NetworkContext_t secondNetworkContext = { 0 };
     bool sessionPresent;
     MQTTContext_t secondContext;
 
-    /* Establish a second TCP connection with the server endpoint, then
-     * a TLS session. The server info and credentials can be reused. */
+    /*Establish a second TCP connection with the server endpoint, then
```
muneebahmed10, line 810: Is this due to uncrustify? It looks wrong.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -795,23 +817,27 @@ void test_MQTT_Connect_LWT( void )
...
                            &context, TEST_MQTT_LWT_TOPIC, MQTTQoS0 ) );
 
-    /* Abruptly terminate TCP connection. */
+    /*Abruptly terminate TCP connection. */
     ( void ) Openssl_Disconnect( &secondNetworkContext );
 
-    /* Run the process loop to receive the LWT. Allow some more time for the
+    /*Run the process loop to receive the LWT.Allow some more time
+     *
+     * for the
```
muneebahmed10, line 833: I think something's wrong with this comment

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -374,10 +387,19 @@ static void eventCallback( MQTTContext_t * pContext,
     assert( pContext != NULL );
     assert( pPacketInfo != NULL );
 
+    if( pPacketInfo->type == disconnectOnPacketType )
+    {
+        /* Terminate MQTT connection with server for session restoration test. */
+        TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_Disconnect( &context ) );
+
+        /* Graceless disconnect. Terminate TLS session and TCP connection. */
```
muneebahmed10, line 395: Don't think it's graceless since we already sent the disconnect packet.
```suggestion
        /* Graceful disconnect. Terminate TLS session and TCP connection. */
```

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -836,15 +864,78 @@ void test_MQTT_ProcessLoop_KeepAlive( void )
...
+    /* Publish to a topic with Qos 2. */
+    TEST_ASSERT_EQUAL( MQTTSuccess, publishToTopic(
+                           &context, TEST_MQTT_TOPIC, MQTTQoS2 ) );
+
+    /* Disconnect on receiving PUBREC so that we aren't able to complete the QoS 2 PUBLISH in the current connection. */
+    disconnectOnPacketType = MQTT_PACKET_TYPE_PUBREC;
+
+    TEST_ASSERT_EQUAL( MQTTSendFailed,
+                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
+    /*TEST_ASSERT_TRUE( receivedPubRec ); */
```
muneebahmed10: Remove commented out code
```suggestion
```

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -836,15 +864,78 @@ void test_MQTT_ProcessLoop_KeepAlive( void )
...
+    TEST_ASSERT_FALSE( receivedPubRel );
```
muneebahmed10, line 911: Does this line need to be here? We receive a pubrel for an incoming publish, be we disconnect during the QoS 2 flow of an outgoing publish. Is it a protocol error for the client to receive both an incoming publish and pubrel before it receives a pubrec for the outgoing publish? If not, I don't think we need this assert.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Fixed

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Fixed

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Removed

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Good point, can use the `createCleanSession` flag instead

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Yes, doesn't need to be here. It is an artifact from an older state of the code 😉

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -832,19 +855,88 @@ void test_MQTT_Connect_LWT( void )
...
 
-    /* Sleep until control packet needs to be sent. */
+    /*Sleep until control packet needs to be sent. */
     Clock_SleepMs( MQTT_KEEP_ALIVE_INTERVAL_SECONDS * 1000 );
     TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_ProcessLoop( &context, MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
 
     TEST_ASSERT_NOT_EQUAL( 0, context.pingReqSendTimeMs );
     TEST_ASSERT_NOT_EQUAL( connectPacketTime, context.lastPacketTime );
-    /* Test that the ping was sent within 1.5 times the keep alive interval. */
+    /*Test that the ping was sent within 1.5 times the keep alive interval. */
```
muneebahmed10, line 875: Some comments still need to be fixed

libraries/standard/mqtt/integration-test/mqtt_system_test.c
muneebahmed10: This comment still needs to be fixed

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -832,19 +855,88 @@ void test_MQTT_Connect_LWT( void )
...
+
+/**
+ * @brief Verifies the behavior of the MQTT library in a restored session connection with the broker.
+ * Tests that the library resends PUBREL packets to the broker in a restored session for an incomplete
+ * PUBLISH operation in a previous connection.
+ */
+void test_MQTT_Connect_Restore_Session( void )
+{
+    /* Graceless disconnect. Terminate TLS session and TCP connection. */
+    ( void ) Openssl_Disconnect( &networkContext );
```
muneebahmed10, line 888: I don't think a graceless disconnect is required, there's just little point in sending the disconnect packet if the session will be discarded anyway. Can you edit the comment to clarify that?

aggarw13: Good point. Updated

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Fixed

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Fixed to avoid use of _graceless_

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -848,3 +877,68 @@ void test_MQTT_ProcessLoop_KeepAlive( void )
...
+
+    /* Publish to a topic with Qos 2. */
+    TEST_ASSERT_EQUAL( MQTTSuccess, publishToTopic(
+                           &context, TEST_MQTT_TOPIC, MQTTQoS2 ) );
+
+    /* Disconnect on receiving PUBREC so that we are not able to complete the QoS 2 PUBLISH in the current connection. */
+    disconnectOnPacketType = MQTT_PACKET_TYPE_PUBREC;
+    TEST_ASSERT_EQUAL( MQTTSendFailed,
+                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
+    TEST_ASSERT_FALSE( receivedPubComp );
```
leegeth, line 916: This test, covers the path for re sending the PUBREL ack, there are a few more cases to be covered for restoring the session.
1. Resend Unacked PUBLISHES with dup flag (QoS1 and QoS2)
2. Receive duplicate PUBREL (QoS2)
3. Receive duplicate PUBLISHES(QoS1 and QoS2)

muneebahmed10: 1. How would we test that the DUP flag is set? I don't think we have access to the publish packet that's created.
2. Isn't that testing more that the broker is up to spec rather than our client? Though I agree we should test it for the purpose of ensuring the state engine works fine.

leegeth: 1. I believe you are asking about DUP flag for re sending the unacked publishes. It will be application's responsibility to do that. In this case, integration test will be able to do that.
2. Yea. We have logic for handling the states when duplicate publishes or acks are received. We would need to test that in our integration tests.

aggarw13: Will take care of the other cases of restore session logic in succeeding PRs

aggarw13: Added a test for case 2. of receiving duplicate `PUBREL` packet in restored session

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -942,3 +942,89 @@ void test_MQTT_Connect_Restore_Session( void )
...
+    /* Resume the incomplete incoming QoS 2 PUBLISH transaction from the previous MQTT connection. */
+    TEST_ASSERT_EQUAL( MQTTSuccess,
+                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
+
+    /* Make sure that the broker resent the PUBREL packet on session restoration. */
+    TEST_ASSERT_TRUE( receivedPubRel );
+
+    /* Make sure that the library sent a PUBCOMP packet in response to the PUBREL packet
+     * from the server to complete the incoming PUBLISH QoS2 transaction. */
+    TEST_ASSERT_EQUAL( MQTTPubCompSend, context.incomingPublishRecords[ 0 ].publishState );
```
muneebahmed10, line 1027: I don't think this test should be here. While we don't erase the old packet state and only reset the id in our implementation to save time, I don't think we should always expect this to be the case. We could always decide to completely zero out the record, so we shouldn't expect old artifacts to remain after the record has been removed.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -942,3 +942,89 @@ void test_MQTT_Connect_Restore_Session( void )
...
+    TEST_ASSERT_EQUAL( 0u, context.incomingPublishRecords[ 0 ].packetId );
+    TEST_ASSERT_EQUAL( true, context.controlPacketSent );
```
muneebahmed10, line 1029: Why not just `TEST_ASSERT_TRUE( context.controlPacketSent );`?

libraries/standard/mqtt/integration-test/mqtt_system_test.c
leegeth: I agree with Muneeb's comment. We should only be validating it against the invalid packet Id.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -319,11 +325,15 @@ static void establishMqttSession( MQTTContext_t * pContext,
      * function will be used to calculate intervals in MQTT library.*/
     callbacks.getTime = Clock_GetTimeMs;
 
-    /* Initialize MQTT library. */
-    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_Init( pContext,
-                                               &transport,
-                                               &callbacks,
-                                               &networkBuffer ) );
+    /* Clear the state of the MQTT context when creating a clean session. */
+    if( createCleanSession == true )
```
leegeth, line 329: Can we always control the MQTT_Init logic based on createCleanSession?  How about adding a different flag to control the MQTT_Init?

muneebahmed10: He initially had a separate flag, but I asked him to remove it since its value was always the same as createCleanSession. Since the only time we don't want to call MQTT_Init is when we don't want a clean session, I don't see why they should be separate.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: `true` is platform dependent..

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Updated.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
muneebahmed10: I don't understand. You used `TEST_ASSERT_TRUE` 5 lines above this. Why not here?

libraries/standard/mqtt/integration-test/mqtt_system_test.c
leegeth: I think we won't be able to associate it always with a clean sesion flag. How about a client having QoS0 publishes only, but has active subscriptions? Reestablishing a session just means the subscriptions being continued in the broker. In this case reinitiailizing the context doesn't really matter. Just the client Id is what matters.

muneebahmed10: If we want to test that situation later, we can always add it then. Right now, a separate parameter is unnecessary given our current tests.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -848,3 +877,152 @@ void test_MQTT_ProcessLoop_KeepAlive( void )
...
+}
+
+/**
+ * @brief Verifies the behavior of the MQTT library on receiving a duplicate
+ * PUBREL packet from the broker in a restored session connection.
+ * Tests that the library sends a PUBCOMP packet in response to the broker for the
+ * incoming QoS 2 PUBLISH operation that was incomplete in a previous connection
+ * of the same session.
+ */
+void test_MQTT_Connect_Restore_Session_Received_Duplicate_PubRel( void )
```
muneebahmed10, line 953: I'm not too fond of the name, since receiving a pubrel is entirely dependent on the broker. From the client's perspective, what we're testing is whether an incomplete incoming publish can successfully resume. What about `test_MQTT_Restore_Session_Complete_Incoming_Publish`?

libraries/standard/mqtt/integration-test/mqtt_system_test.c
leegeth: I don't totally agree with this change. I like it to be separate init logic and separate clean session flag for anyone not so familiar with the code to easily reuse the function for any new integration tests. However I am ok to proceed with the change that we may need to modify it with a different test. I am good to disagree and commit.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -848,3 +877,154 @@ void test_MQTT_ProcessLoop_KeepAlive( void )
...
+
+    /* Verify that the session was resumed. */
+    TEST_ASSERT_TRUE( persistentSession );
+
+    /* Resume the incomplete QoS 2 PUBLISH in previous MQTT connection. */
+    TEST_ASSERT_EQUAL( MQTTSuccess,
+                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
+
+    /* Test that the MQTT library has completed the QoS 2 publish by sending the PUBREL flag. */
+    TEST_ASSERT_TRUE( receivedPubComp );
```
dan4thewin, line 943: Is this an assert on a global variable?  I think this smells like action-at-a-distance.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -848,3 +877,154 @@ void test_MQTT_ProcessLoop_KeepAlive( void )
...
+     * to start a persistent session with the broker. */
+
+    /* Create the TLS+TCP connection with the broker. */
+    TEST_ASSERT_EQUAL( OPENSSL_SUCCESS, Openssl_Connect( &networkContext,
+                                                         &serverInfo,
+                                                         &opensslCredentials,
+                                                         TRANSPORT_SEND_RECV_TIMEOUT_MS,
+                                                         TRANSPORT_SEND_RECV_TIMEOUT_MS ) );
+    TEST_ASSERT_NOT_EQUAL( -1, networkContext.socketDescriptor );
+    TEST_ASSERT_NOT_NULL( networkContext.pSsl );
```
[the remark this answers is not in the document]
aggarw13: We don't have tests for the transport interface implementations currently..so, this is just an extra layer of checking that the connection was actually established.
I agree that these asserts shouldn't _ideally_ exist

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Yes, it is a global variable. The MQTT callback type, `MQTTEventCallback_t` doesn;t provide a way to supply user-context..so, had to use to use global variable for keeping track of test state....

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Good suggestion. Updated

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -848,3 +877,154 @@ void test_MQTT_ProcessLoop_KeepAlive( void )
...
 }
+
+/**
+ * @brief Verifies the behavior of the MQTT library in a restored session connection with the broker.
+ * Tests that the library resends PUBREL packets to the broker in a restored session for an incomplete
+ * PUBLISH operation in a previous connection.
+ */
+void test_MQTT_Connect_Restore_Session( void )
+{
+    /* Terminate TLS session and TCP connection network connection to discard current MQTT session
```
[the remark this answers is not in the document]
aggarw13: Fixed

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Based on our separate conversation, I have updated all test cases to check the state of the global variable before calling `ProcessLoop` and expecting that the state has changed. This will guard against unexpected global variable state changes

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -366,10 +376,20 @@ static void eventCallback( MQTTContext_t * pContext,
...
+        /* Terminate TLS session and TCP connection to test session restoration
+         * across network connection. */
+        ( void ) Openssl_Disconnect( &networkContext );
+    }
+
     /* Handle incoming publish. The lower 4 bits of the publish packet
      * type is used for the dup, QoS, and retain flags. Hence masking
      * out the lower bits to check if the packet is publish. */
-    if( ( pPacketInfo->type & 0xF0U ) == MQTT_PACKET_TYPE_PUBLISH )
+    else if( ( pPacketInfo->type & 0xF0U ) == MQTT_PACKET_TYPE_PUBLISH )
```
gkwicker, line 392: `else if` with no braces makes introducing bugs during maintenance very easy so we should avoid it.  This is a pretty long function anyway, so you might consider splitting it into two: one to handle lines 387-417 and another for lines 421-500.

aggarw13: Broken down the `eventCallback` function

PR #1086, opened 2020-07-29
Integ tests/mqtt retain flag

aggarw13:
Add an integration test for the MQTT feature of _retaining_ PUBLISH messages on the broker with the `retain` flag

MQTT spec information about the **RETAIN** flag: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/mqtt-v3.1.1.pdf (Page 34)


libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -1312,3 +1385,64 @@ void test_MQTT_Restore_Session_Duplicate_Incoming_Publish_Qos2( void )
...
+    /* Publish to a topic with the "retain" flag set. */
+    TEST_ASSERT_EQUAL( MQTTSuccess, publishToTopic( &context,
+                                                    TEST_MQTT_TOPIC,
+                                                    true,  /* setRetainFlag */
+                                                    false, /* isDuplicate */
+                                                    MQTTQoS1,
+                                                    MQTT_GetPacketId( &context ) ) );
+    /* Complete the QoS 1 PUBLISH operation. */
+    TEST_ASSERT_EQUAL( MQTTSuccess,
+                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
```
muneebahmed10, line 1404: Should we assert that we received the puback?

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -1312,3 +1385,64 @@ void test_MQTT_Restore_Session_Duplicate_Incoming_Publish_Qos2( void )
...
+                                                    MQTTQoS1,
+                                                    MQTT_GetPacketId( &context ) ) );
+
+    /* Complete the QoS 1 PUBLISH operation. */
+    TEST_ASSERT_EQUAL( MQTTSuccess,
+                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
+
+    /* Again, subscribe to the same topic that we just published to.
+     * We don't expect the broker to send the message to us (as we
+     *  PUBLISHed without a retain flag set).. */
```
muneebahmed10, line 1438: Looks like there's an extra space and period
```suggestion
     * PUBLISHed without a retain flag set). */
```

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -593,18 +611,17 @@ static MQTTStatus_t unsubscribeFromTopic( MQTTContext_t * pContext,
                              globalUnsubscribePacketIdentifier );
 }
 
-
 static MQTTStatus_t publishToTopic( MQTTContext_t * pContext,
                                     const char * pTopic,
+                                    bool setRetainFlag,
```
muneebahmed10, line 616: You know, if we're going to pass in all these parameters to this function, why not just pass in an `MQTTPublishInfo_t`? The only parts that aren't affected by the parameters here are `pPayload` and `payloadLength`, which this function could still set if you wanted.

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -534,6 +545,13 @@ static void eventCallback( MQTTContext_t * pContext,
             incomingInfo.pPayload = malloc( pPublishInfo->payloadLength );
             TEST_ASSERT_NOT_NULL( incomingInfo.pPayload );
             memcpy( ( void * ) incomingInfo.pPayload, pPublishInfo->pPayload, pPublishInfo->payloadLength );
+
+            /* Update the global variable if the incoming PUBLISH packet
+             * represents a retained message. */
+            if( pPublishInfo->retain == true )
+            {
+                receivedRetainedMessage = true;
+            }
```
muneebahmed10, line 554: If `receivedRetainedMessage` is not expected to keep its value across publishes, then this could just be:
```suggestion
            receivedRetainedMessage = pPublishInfo->retain;
```

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: If we pass the `MQTTPublishInfo_t`, then this function wouldn't be required anymore!
The main utility that this function provides is commonizing the logic of assigning values to the `publishInfo` struct

libraries/standard/mqtt/integration-test/mqtt_system_test.c
aggarw13: Good catch

libraries/standard/mqtt/integration-test/mqtt_system_test.c
muneebahmed10: In that case, I'm wondering how much benefit this function actually provides any more, since every invocation of it uses almost as many lines as just setting the parameters of the publish info. Since a lot of these parameters are similar across different calls to this function, why don't we create a const `globalPublishInfoInitializer` to copy into a new struct each time, and only change the parameters we want to change for each publish?

Not a requirement for this PR, but a suggestion for how we can improve this for the future.

aggarw13: Yes, I did think about that, but didn't want to increase the scope of this PR further 🙂

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -1312,3 +1382,69 @@ void test_MQTT_Restore_Session_Duplicate_Incoming_Publish_Qos2( void )
...
+    TEST_ASSERT_TRUE( receivedPubAck );
+
+    /* Subscribe to the same topic that we published the message to.
+     * The broker should send the "retained" message with the "retain" flag set. */
+    TEST_ASSERT_EQUAL( MQTTSuccess, subscribeToTopic(
+                           &context, TEST_MQTT_TOPIC, MQTTQoS1 ) );
+    TEST_ASSERT_FALSE( receivedSubAck );
+    TEST_ASSERT_EQUAL( MQTTSuccess,
+                       MQTT_ProcessLoop( &context, 2 * MQTT_PROCESS_LOOP_TIMEOUT_MS ) );
+    TEST_ASSERT_TRUE( receivedSubAck );
```
dan4thewin, line 1412: I see this pattern a lot in this code.  We don't have any guarantees about broker latency, do we?

aggarw13: I would have to look into the Mosquitto configuration options to check whether there is any latency guarantee we can get. But yes, if we don;t have, then these tests are dependent on the responsiveness of the broker..

aggarw13: We discussed offline, and Dan mentioned that he was concerned about the timeout value introducing flaky behavior of these network-dependent integration tests. 
As these tests are run on hosts (or instances) with ethernet connection, a flaky behavior of these tests is less likely (based on the runs of the `v4_beta` branch that have been fairly stable).  Anyway, we will get concrete data once we have enabled nightly CI runs for the demos. We don't expect a flaky behavior when running against a locally owned Docker container running the Mosquitto broker

PR #1087, opened 2020-07-29
SHADOW: Add support for named shadows

andresmanelli:
## Description of changes

The shadow name parameter was added to every function that needed that information in order to create the correct topic.
A NULL value falls back to the Classic Shadow topics.

**Note**: I know the interface is changed with this PR, so I am not sure if it can be integrated in the code base as-is, so please let me know if the changes aren't aligned with the contributing guidelines or if there is something I could improve.

### DIFF
--- a/include/aws_iot_jobs_interface.h
+++ b/include/aws_iot_jobs_interface.h
@@ -61,8 +61,8 @@ extern "C" {
  * \param pClient the client to use
  * \param qos the qos to use
  * \param thingName the name of the thing to subscribe to
- * \param jobId the job id to subscribe to. To subscribe to messages not related to 
- *   a job the jobId passed should be null. The jobId can also be "+" to subscribe to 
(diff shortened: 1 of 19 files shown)


dan4thewin:
Hi, @andresmanelli .  Thank you for the PR.  The changes do alter existing API calls, so that would require people using the library to update their code even if they don't use named shadows.  One way to make these kind of changes without breaking existing code is to add new functions with slightly different names, then change the old API functions to call the new ones.

The shadow library is actively undergoing refactoring, so this version will have no new features, only bug fixes.  You can see the new code in this PR: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1081 .  Your feedback would be welcome.  The next iteration of the C SDK is on the development branch: https://github.com/aws/aws-iot-device-sdk-embedded-C/tree/development .  Please join in the fun.

PR #1089, opened 2020-07-30
Return suback when server refuses subscription

muneebahmed10:
*Description of changes:*
The application should be aware of when a subscription is refused by the broker. The app callback should be called when a suback is deserialized with the status `MQTTServerRefused`

### DIFF
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -248,14 +248,11 @@ static void handleIncomingPublish( MQTTPublishInfo_t * pPublishInfo,
  *
  * @param[in] pMqttContext MQTT context pointer.
  * @param[in] pPacketInfo Packet Info pointer for the incoming packet.
- * @param[in] packetIdentifier Packet identifier of the incoming packet.
- * @param[in] pPublishInfo Deserialized publish info pointer for the incoming
- * packet.
+ * @param[in] pDeserialized Deserialized information from the incoming packet.
  */
 static void eventCallback( MQTTContext_t * pMqttContext,
                            MQTTPacketInfo_t * pPacketInfo,
-                           uint16_t packetIdentifier,
-                           MQTTPublishInfo_t * pPublishInfo );
+                           MQTTDeserializedInfo_t * pDeserialized );
 
 /**
  * @brief Sends an MQTT CONNECT packet over the already connected TCP socket.
@@ -604,20 +601,24 @@ static void handleIncomingPublish( MQTTPublishInfo_t * pPublishInfo,
 
 static void eventCallback( MQTTContext_t * pMqttContext,
                            MQTTPacketInfo_t * pPacketInfo,
-                           uint16_t packetIdentifier,
-                           MQTTPublishInfo_t * pPublishInfo )
+                           MQTTDeserializedInfo_t * pDeserialized )
 {
+    uint16_t packetIdentifier;
+
     assert( pMqttContext != NULL );
     assert( pPacketInfo != NULL );
+    assert( pDeserialized != NULL );
+
+    packetIdentifier = pDeserialized->packetIdentifier;
 
     /* Handle incoming publish. The lower 4 bits of the publish packet
      * type is used for the dup, QoS, and retain flags. Hence masking
      * out the lower bits to check if the packet is publish. */
     if( ( pPacketInfo->type & 0xF0U ) == MQTT_PACKET_TYPE_PUBLISH )
     {
-        assert( pPublishInfo != NULL );
+        assert( pDeserialized->pPublishInfo != NULL );
         /* Handle incoming publish. */
-        handleIncomingPublish( pPublishInfo, packetIdentifier );
+        handleIncomingPublish( pDeserialized->pPublishInfo, packetIdentifier );
     }
     else
     {
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -248,11 +248,11 @@ static void handleIncomingPublish( MQTTPublishInfo_t * pPublishInfo,
  *
  * @param[in] pMqttContext MQTT context pointer.
  * @param[in] pPacketInfo Packet Info pointer for the incoming packet.
- * @param[in] pDeserialized Deserialized information from the incoming packet.
+ * @param[in] pDeserializedInfo Deserialized information from the incoming packet.
  */
 static void eventCallback( MQTTContext_t * pMqttContext,
                            MQTTPacketInfo_t * pPacketInfo,
-                           MQTTDeserializedInfo_t * pDeserialized );
+                           MQTTDeserializedInfo_t * pDeserializedInfo );
 
 /**
  * @brief Sends an MQTT CONNECT packet over the already connected TCP socket.
@@ -601,24 +601,24 @@ static void handleIncomingPublish( MQTTPublishInfo_t * pPublishInfo,
 
 static void eventCallback( MQTTContext_t * pMqttContext,
                            MQTTPacketInfo_t * pPacketInfo,
-                           MQTTDeserializedInfo_t * pDeserialized )
+                           MQTTDeserializedInfo_t * pDeserializedInfo )
 {
     uint16_t packetIdentifier;
 
     assert( pMqttContext != NULL );
     assert( pPacketInfo != NULL );
-    assert( pDeserialized != NULL );
+    assert( pDeserializedInfo != NULL );
 
-    packetIdentifier = pDeserialized->packetIdentifier;
+    packetIdentifier = pDeserializedInfo->packetIdentifier;
 
     /* Handle incoming publish. The lower 4 bits of the publish packet
      * type is used for the dup, QoS, and retain flags. Hence masking
      * out the lower bits to check if the packet is publish. */
     if( ( pPacketInfo->type & 0xF0U ) == MQTT_PACKET_TYPE_PUBLISH )
     {
-        assert( pDeserialized->pPublishInfo != NULL );
+        assert( pDeserializedInfo->pPublishInfo != NULL );
         /* Handle incoming publish. */
-        handleIncomingPublish( pDeserialized->pPublishInfo, packetIdentifier );
+        handleIncomingPublish( pDeserializedInfo->pPublishInfo, packetIdentifier );
     }
     else
     {
--- a/libraries/standard/mqtt/include/mqtt.h
+++ b/libraries/standard/mqtt/include/mqtt.h
@@ -128,7 +128,7 @@ struct MQTTDeserializedInfo
 {
     uint16_t packetIdentifier;
     MQTTPublishInfo_t * pPublishInfo;
-    MQTTStatus_t status;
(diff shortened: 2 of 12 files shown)


libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -1605,6 +1605,15 @@ void test_MQTT_ProcessLoop_handleIncomingAck_Error_Paths( void )
...
                             MQTTBadResponse, false, NULL );
 
+    /* Verify that MQTTServerRefused is propagated when SUBACK indicates a
+     * server refusal. */
+    currentPacketType = MQTT_PACKET_TYPE_SUBACK;
+    isEventCallbackInvoked = false;
+    expectProcessLoopCalls( &context, MQTTServerRefused, MQTTStateNull,
+                            MQTTIllegalState, MQTTBadResponse, MQTTStateNull,
+                            MQTTServerRefused, false, NULL );
+    TEST_ASSERT_TRUE( isEventCallbackInvoked );
```
aggarw13, line 1615: Can you also add the same test for the `UNSUBACK` packet as the new behavior will invoke the callback for server rejections in both subscribe and unscubscribe requests?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -938,7 +938,7 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
             break;
     }
 
-    if( ( status == MQTTSuccess ) && ( invokeAppCallback == true ) )
+    if( invokeAppCallback == true )
     {
         appCallback( pContext, pIncomingPacket, packetIdentifier, NULL );
```
aggarw13, line 943: I think it would be helpful to update the callback signature to provide the de-serialization status for the application, as it may not be obvious to the customer about calling `MQTT_DeserializeAck` API to match the rejected subscription request with the packet ID when there are multiple "in-flight" requests. However, addition of a separate parameter would be a noise for all other packet types (PING, PUBLISH acks)....

libraries/standard/mqtt/utest/mqtt_utest.c
muneebahmed10: No, since deserializing UNSUBACK can't return `MQTTServerRefused`, only `MQTTBadResponse` (or `MQTTBadParameter` if we called it improperly)

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: That's an option I considered, and weighed the drawback of adding another parameter to the callback with having the application call the deserializer. I thought this would be simpler, but figured it'd be discussed in this PR so a consensus could be reached. I can update this PR to update the callback if that's the direction we want to go

muneebahmed10: Updated the callback

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -123,6 +124,13 @@ struct MQTTContext
     bool waitingForPingResp;
 };
 
+struct MQTTDeserializedInfo
+{
+    uint16_t packetIdentifier;
+    MQTTPublishInfo_t * pPublishInfo;
+    MQTTStatus_t status;
```
aggarw13, line 131: (Nit) Suggestion to make the meaning of the variable clear
```suggestion
    MQTTStatus_t deserializationStatus;
```

libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -496,11 +493,16 @@ static void handleAckEvents( MQTTPacketInfo_t * pPacketInfo,
 
 static void eventCallback( MQTTContext_t * pContext,
                            MQTTPacketInfo_t * pPacketInfo,
-                           uint16_t packetIdentifier,
-                           MQTTPublishInfo_t * pPublishInfo )
+                           MQTTDeserializedInfo_t * pDeserialized )
 {
+    MQTTPublishInfo_t * pPublishInfo;
```
aggarw13, line 498: Default initialization of stack variables?
```suggestion
    MQTTPublishInfo_t * pPublishInfo = NULL;
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -865,9 +874,12 @@ static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
 
     if( status == MQTTSuccess )
     {
+        pDeserialized->packetIdentifier = packetIdentifier;
```
aggarw13, line 877: `pDeserialized->pPublishInfo = NULL`?

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -865,9 +874,12 @@ static MQTTStatus_t handlePublishAcks( MQTTContext_t * pContext,
...
+        pDeserialized->status = status;
+
         /* Invoke application callback to hand the buffer over to application
          * before sending acks. */
-        appCallback( pContext, pIncomingPacket, packetIdentifier, NULL );
+        appCallback( pContext, pIncomingPacket, pDeserialized );
```
aggarw13, line 882: If the app callback is invoked within this function, the `MQTTDeserializedInfo_t * pDeserialized` output parameter shouldn't be required

libraries/standard/mqtt/utest/mqtt_utest.c
aggarw13: So, can you add a test for invoking the callback with `MQTTBadResponse`  for UNSUBACK?

libraries/standard/mqtt/integration-test/mqtt_system_test.c
muneebahmed10: This isn't necessary since the pointer is unconditionally set to `pDeserialized->pPublishInfo` seven lines later, after the asserts. It doesn't hurt to add it though so I made the change

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Yes, it's not required, the reason I added it was that this function is called by `handleIncomingAck`, which already declared an `MQTTDeserializedInfo_t` on the stack. Even though the struct shouldn't be very large, I didn't think it was necessary to declare another one in this function, and just passed the first one in via pointer.

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: `handleIncomingAck` already sets `deserializedInfo.pPublishInfo = NULL;` before this function is called, so I didn't think it necessary to set it again.

libraries/standard/mqtt/include/mqtt.h
muneebahmed10: I'm unsure about this since the name of the struct is already `MQTTDeserializedInfo`, so I thought it was already clear what the status is referring to. Do you think it still should be changed?

libraries/standard/mqtt/src/mqtt.c
aggarw13: Yes, I think it should be set as this function shouldn't be aware of the implementation details of its calling function. More importantly, this function shouldn't depend on the calling function for setting the parameter correctly.

libraries/standard/mqtt/src/mqtt.c
aggarw13: I don't think we should have the API be sending such a parameter. It seems like the common function `handleIncomingAcks` should be broken for handling the Publish and PinResponse/Subscribe Acks separately

libraries/standard/mqtt/include/mqtt.h
aggarw13: Thinking from the customer's perspective, I don't think it may be obvious about what `status` the member of the `MQTTDeserializedInfo` represents. Does it mean a status value returned by the server? or does it mean the status value of our de-serialization effort in the library? 
As this member currently only represents whether the server rejected or accepted the request, I think it would be better to have a more descriptive name like `ackStatus` .

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Removed the pointer parameter

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Handled when pointer was removed

libraries/standard/mqtt/include/mqtt.h
muneebahmed10: Renamed it to `deserializationResult`. I'm unsure about `ackStatus`, since I'm not sure if we want to only set it when receiving an ack, or always set it before calling the app callback. In this PR I always set it before calling the app callback, so I named it accordingly; if it's supposed to be specific to acks then it can be renamed again

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -55,13 +58,11 @@ typedef uint32_t (* MQTTGetCurrentTimeFunc_t )( void );
  *
  * @param[in] pContext Initialized MQTT context.
  * @param[in] pPacketInfo Information on the type of incoming MQTT packet.
- * @param[in] packetIdentifier Packet identifier of incoming PUBLISH packet.
- * @param[in] pPublishInfo Incoming PUBLISH packet parameters.
+ * @param[in] pDeserialized Deserialized information from incoming packet.
```
aggarw13, line 61: Now that this PR updates this event callback, can you add documentation to mention that the event callback would only be called for either `MQTTSuccess` or `MQTTServerRefused` de-serialization statuses? (May be good to also mention that `MQTTServerRefused` is only passed for `SUBACK` packet events)
I ask so that we don't forget to mention that during whenever the documentation task for all MQTT library is taken up...

muneebahmed10: Added a `@note`

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -904,6 +916,8 @@ static MQTTStatus_t handleIncomingAck( MQTTContext_t * pContext,
     assert( pContext->appCallback != NULL );
 
     appCallback = pContext->appCallback;
+    /* We did not receive a publish. */
+    deserializedInfo.pPublishInfo = NULL;
```
aggarw13, line 920: This logic can be moved to the if-block after line 961 for a better logical placement.

muneebahmed10: I mainly put this out here since to reduce complexity since this will always be NULL, but the complexity of this function is 5 so it doesn't really matter. Moved to the if

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -247,14 +247,11 @@ static void handleIncomingPublish( MQTTPublishInfo_t * pPublishInfo,
...
- * @param[in] packetIdentifier Packet identifier of the incoming packet.
- * @param[in] pPublishInfo Deserialized publish info pointer for the incoming
- * packet.
+ * @param[in] pDeserialized Deserialized information from the incoming packet.
  */
 static void eventCallback( MQTTContext_t * pMqttContext,
                            MQTTPacketInfo_t * pPacketInfo,
-                           uint16_t packetIdentifier,
-                           MQTTPublishInfo_t * pPublishInfo );
+                           MQTTDeserializedInfo_t * pDeserialized );
```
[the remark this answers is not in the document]
muneebahmed10: Renamed

PR #1091, opened 2020-07-31
Add draft implementation of the subscription manager module

aggarw13:
#### Subscription Manager For MQTT 
Subscription Manager will be utilized in demo that exhibits connection sharing in a single-threaded demo

### DIFF
--- a/demos/mqtt/mqtt_demo_connection_sharing/CMakeLists.txt
+++ b/demos/mqtt/mqtt_demo_connection_sharing/CMakeLists.txt
@@ -1,6 +1,6 @@
 include(CheckSymbolExists)
 
-set( DEMO_NAME "mqtt_demo_mutual_auth" )
+set( DEMO_NAME "mqtt_demo_connection_sharing" )
 
 # Demo target.
 add_executable(${DEMO_NAME})
@@ -12,6 +12,7 @@ target_sources(
     ${DEMO_NAME}
     PRIVATE
         "${DEMO_NAME}.c"
+        subscription-manager/mqtt_subscription_manager.c
 )
 
 target_link_libraries(
@@ -33,22 +34,5 @@ target_include_directories(
     ${DEMO_NAME}
     PUBLIC
         ${LOGGING_INCLUDE_DIRS}
-)
-
-# Download the Amazon Root CA certificate.
-message( "Downloading the Amazon Root CA certificate..." )
-file(MAKE_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/certificates)
-execute_process(
-    COMMAND curl --url https://www.amazontrust.com/repository/AmazonRootCA1.pem
-    -o ${CMAKE_CURRENT_LIST_DIR}/certificates/AmazonRootCA1.crt
-)
-
-# Copy the certificates and client key to the binary directory.
-add_custom_command(
-    TARGET
-        ${DEMO_NAME}
-    POST_BUILD
-        COMMAND ${CMAKE_COMMAND} -E copy_directory
-        "${CMAKE_CURRENT_LIST_DIR}/certificates"
-        "$<TARGET_FILE_DIR:${DEMO_NAME}>/certificates"
+        ${CMAKE_CURRENT_LIST_DIR}/subscription-manager
 )
--- a/demos/mqtt/mqtt_demo_connection_sharing/CMakeLists.txt
+++ b/demos/mqtt/mqtt_demo_connection_sharing/CMakeLists.txt
@@ -1,0 +1,54 @@
+include(CheckSymbolExists)
+
+set( DEMO_NAME "mqtt_demo_mutual_auth" )
+
+# Demo target.
+add_executable(${DEMO_NAME})
+
+# Add to default target if all required macros needed to run this demo are defined
+check_aws_credentials(${DEMO_NAME})
+
+target_sources(
+    ${DEMO_NAME}
+    PRIVATE
+        "${DEMO_NAME}.c"
+)
+
+target_link_libraries(
+    ${DEMO_NAME}
+    PRIVATE
+        mqtt
+        clock_posix
+        openssl_posix
+        transport_reconnect_posix
+)
+
+target_include_directories(
+    mqtt
+    PUBLIC
+        ${CMAKE_CURRENT_LIST_DIR}
+)
+
+target_include_directories(
+    ${DEMO_NAME}
+    PUBLIC
+        ${LOGGING_INCLUDE_DIRS}
+)
+
+# Download the Amazon Root CA certificate.
+message( "Downloading the Amazon Root CA certificate..." )
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/certificates)
+execute_process(
+    COMMAND curl --url https://www.amazontrust.com/repository/AmazonRootCA1.pem
+    -o ${CMAKE_CURRENT_LIST_DIR}/certificates/AmazonRootCA1.crt
+)
+
+# Copy the certificates and client key to the binary directory.
+add_custom_command(
+    TARGET
+        ${DEMO_NAME}
+    POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy_directory
+        "${CMAKE_CURRENT_LIST_DIR}/certificates"
+        "$<TARGET_FILE_DIR:${DEMO_NAME}>/certificates"
+)
--- a/demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
+++ b/demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
@@ -1,0 +1,308 @@
+/*
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include <string.h>
+#include "mqtt_subscription_manager.h"
+
+#define MAX_SUBSCRIPTION_CALLBACK_RECORDS    5
+
+typedef struct SubscriptionManager_Record
+{
+    const char * pTopicFilter;
+    size_t topicFilterLength;
+    bool requiresExactMatch;
+    SubscriptionManager_Callback_t callback;
+} SubscriptionManager_Record_t;
+
+static SubscriptionManager_Record_t callbackRecordList[ MAX_SUBSCRIPTION_CALLBACK_RECORDS ] = { 0 };
+
+static const recordListSize = sizeof( callbackRecordList ) / sizeof( SubscriptionManager_Record_t );
+static size_t recordListCount = 0u;
+static size_t maxRecordIndex = 0u;
+
+
+
+static bool matchEndWildcards( const char * pTopicFilter,
+                               uint16_t topicNameLength,
+                               uint16_t topicFilterLength,
+                               uint16_t nameIndex,
+                               uint16_t filterIndex,
+                               bool * pMatch )
+{
+    bool status = false, endChar = false;
+
+    /* Determine if the last character is reached for both topic name and topic
+     * filter for the '#' wildcard. */
+    endChar = ( nameIndex == ( topicNameLength - 1U ) ) && ( filterIndex == ( topicFilterLength - 3U ) );
+
+    if( endChar == true )
+    {
+        /* Determine if the topic filter ends with the '#' wildcard. */
+        status = ( pTopicFilter[ filterIndex + 2U ] == '#' );
+    }
+
+    if( status == false )
+    {
+        /* Determine if the last character is reached for both topic name and topic
+         * filter for the '+' wildcard. */
+        endChar = ( nameIndex == ( topicNameLength - 1U ) ) && ( filterIndex == ( topicFilterLength - 2U ) );
+
+        if( endChar == true )
+        {
+            /* Filter "sport/+" also matches the "sport/" but not "sport". */
+            status = ( pTopicFilter[ filterIndex + 1U ] == '+' );
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,308 @@
...
+        availableIndex++;
+    }
+
+    if( availableIndex == recordListSize )
+    {
+        /* Record list is full. */
+    }
+    else
+    {
+        /* Should the topic string be copied? */
```
muneebahmed10, line 264: To accommodate varying use cases yes, so the `pTopicFilter` member of the `SubscriptionManager_Record_t` would need to be an array. However, since it's just a demo and you control everything, it's up to you; if you only use constant strings then I don't think it's a major issue as long as there's a comment calling that out.

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,308 @@
...
+    do
+    {
+        pRecord = &callbackRecordList[ matchingRecordIndex ];
+
+        if( ( topicFileterLength == pRecord->pTopicFilter ) &&
+            ( strncmp( pTopicFilter, pRecord->pTopicFilter, topicFileterLength ) == 0 ) )
+        {
+            recordFound = true;
+        }
+    } while( ( matchingRecordIndex < recordListSize ) && ( recordFound == false ) );
```
muneebahmed10, line 298: What if more than one application registers a different callback to the same topic filter? Then this would only remove the first one.

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,308 @@
...
+ */
+void SubscriptionManager_DispatchHandler( uint16_t packetIdentifier,
+                                          MQTTPublishInfo_t * pPublishInfo )
+{
+    assert( pPublishInfo != NULL );
+    assert( packetIdentifier != MQTT_PACKET_ID_INVALID );
+
+    size_t recordListCursor = 0u;
+
+    while( findRecordForTopic( pPublishInfo->pTopicName,
```
muneebahmed10, line 236: I don't see why `findRecordForTopic` is necessary. Since you're iterating over every entry in the list, why not just use a `for` loop in this function?

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,308 @@
...
+
+    /* Check for an exact match. */
+    if( topicNameLength == topicFilterLength )
+    {
+        status = ( strncmp( pTopicName, pTopicFilter, topicNameLength ) == 0 );
+    }
+
+    /* If  an exact match is required, return the result of the comparison above.
+     * Otherwise, attempt to match with MQTT wildcards in topic filters. */
+    if( requiresExactMatch == false )
```
muneebahmed10, line 194: Would we ever want to set this to true? I see you don't and that makes sense to me, so I think this function could just be removed and `topicFilterMatch` used in its place

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,308 @@
...
+
+#include <string.h>
+#include "mqtt_subscription_manager.h"
+
+#define MAX_SUBSCRIPTION_CALLBACK_RECORDS    5
+
+typedef struct SubscriptionManager_Record
+{
+    const char * pTopicFilter;
+    size_t topicFilterLength;
```
muneebahmed10, line 30: This can be a `uint16_t`, since filter lengths must fit in 2 bytes

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: Currently, the subscription manager is geared toward(s) a single-threaded demo that will showcase connection sharing for multiple libraries (like Jobs, Shadow) that need different callbacks for processing their topic-specific PUBLISH messages. 

However, even in the case of a multi-threaded use-case, a single application's call to remove its callback shouldn't remove another application's callback for the same topic filter. In that case, the subscription manager would need to store some thread/application-specific information in the records.

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: I took the topic matching logic from the `v4_beta` branch [here](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/iot_mqtt_subscription.c#L316-L321). 
Looking at the call-site of this function in the `v4_beta` implementation, it seems like this parameter was added for a different use-case (of updating registered callback against the same topic for which subscription already exists).

I will remove the `requiresExactMatch` logic

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: Yeah, I think for the single-threaded demo purpose, I will keep it simple and avoid the memory allocation problem (currently, the record list is statically allocated for a fixed size, but with local copy of the topic filter, it would require a max limit on the topic filter size, and thereby, increase the record list structure size)

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,288 @@
...
+ */
+
+#include <string.h>
+#include "mqtt_subscription_manager.h"
+
+#define MAX_SUBSCRIPTION_CALLBACK_RECORDS    5
+
+typedef struct SubscriptionManager_Record
+{
+    const char * pTopicFilter;
```
leegeth, line 29: Which module is responsible for topic filter memory? Will application be responsible for all topic filter buffers if it subscribes to different topics?

demos/mqtt/mqtt_demo_connection_sharing/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+include(CheckSymbolExists)
+
+set( DEMO_NAME "mqtt_demo_mutual_auth" )
```
leegeth, line 3: mqtt_demo_mutual_auth ?

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -231,18 +205,25 @@ void SubscriptionManager_DispatchHandler( uint16_t packetIdentifier,
...
-                               &recordListCursor ) != false )
+    /* Iterate through record list to find matching topics, and invoke their callbacks. */
+    while( listIndex < recordListCount )
     {
-        callbackRecordList[ recordListCursor ].callback( packetIdentifier, pPublishInfo );
+        if( matchTopic( pPublishInfo->pTopicName,
+                        pPublishInfo->topicNameLength,
+                        callbackRecordList[ listIndex ].pTopicFilter,
+                        callbackRecordList[ listIndex ].topicFilterLength,
+                        callbackRecordList[ listIndex ].requiresExactMatch ) == true )
```
muneebahmed10, line 217: This parameter doesn't exists in the struct anymore

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,288 @@
...
+    return status;
+}
+
+/*-----------------------------------------------------------*/
+
+static bool matchTopic( const char * pTopicName,
+                        const uint16_t topicNameLength,
+                        const char * pTopicFilter,
+                        const uint16_t topicFilterLength,
+                        bool requiresExactMatch )
```
muneebahmed10, line 179: This parameter can be deleted

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,288 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#include <string.h>
+#include "mqtt_subscription_manager.h"
+
+#define MAX_SUBSCRIPTION_CALLBACK_RECORDS    5
```
leegeth, line 25: I think this would need to be the default value. we would need to wrap with #ifndef to this define

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,288 @@
...
+        {
+            maxRecordIndex = availableIndex;
+        }
+    }
+
+    return recordAdded;
+}
+
+void SubscriptionManager_RemoveCallbackRecord( const char * pTopicFilter,
+                                               uint16_t topicFileterLength )
```
muneebahmed10, line 263: Fileter -> Filter, here and several other places

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: Good point, removed

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: Will probably look to move it to the `demo_config.h`

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,288 @@
...
+        else
+        {
+            /* Check for matching wildcards. */
+            matchFound = matchWildcards( pTopicFilter,
+                                         pTopicName,
+                                         topicNameLength,
+                                         filterIndex,
+                                         &nameIndex,
+                                         &status );
+        }
```
leegeth, line 152: Don't we have to check for the topics beginning with $ as well?

muneebahmed10: I think the '$' prefix reservation is specific to AWS IoT, so I'd think no

leegeth: http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718107, "Topics beginning with $", it is part of the spec. Not really specific to AWS IoT.

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,288 @@
...
+    while( listIndex < recordListCount )
+    {
+        if( matchTopic( pPublishInfo->pTopicName,
+                        pPublishInfo->topicNameLength,
+                        callbackRecordList[ listIndex ].pTopicFilter,
+                        callbackRecordList[ listIndex ].topicFilterLength,
+                        callbackRecordList[ listIndex ].requiresExactMatch ) == true )
+        {
+            /* Invoke the callback associated with the record as the topics match. */
+            callbackRecordList[ listIndex++ ].callback( packetIdentifier, pPublishInfo );
```
leegeth, line 220: I think this will result in an infinite loop if matchTopic returns false

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,288 @@
...
+    /* Iterate through the records list to find the matching record. */
+    do
+    {
+        pRecord = &callbackRecordList[ matchingRecordIndex ];
+
+        if( ( topicFileterLength == pRecord->pTopicFilter ) &&
+            ( strncmp( pTopicFilter, pRecord->pTopicFilter, topicFileterLength ) == 0 ) )
+        {
+            recordFound = true;
+        }
```
leegeth, line 278: possible Infinite loop. matchingRecordIndex needs to be incremented.

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
leegeth: It is a good point that Muneeb mentioned. What is the purpose of this Subscription manager? Is it only for a single task use case? Is there a plan to improve it to support it for connection sharing between multiple tasks?
If we are extending it to multiple tasks, yea we would need some way to identify the task from which it is called.

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: Yes, the single-threaded application will be responsible to maintain the lifetime of the topic string memory that is provides to the subscription manager. (Ideally, in a multi-threaded use-case, the data should be copied into the record, but I am thinking of keeping it simple for the single-threaded demo)

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: Good catch! Fixed

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: Good catch, fixed!

demos/mqtt/mqtt_demo_connection_sharing/CMakeLists.txt
aggarw13: Fixed

demos/mqtt/mqtt_demo_connection_sharing/subscription-manager/mqtt_subscription_manager.c
aggarw13: The spec mentions that "_The Server MUST NOT match Topic Filters starting with a wildcard character (# or +) with Topic Names beginning with a $ character_".
The client does not need to perform any special handling for topics starting with "$". Our service (Jobs, Shadow, Provisioning, etc) subscribe to these topics (starting with "$") and the topic matching on receiving incoming PUBLISHes only involves performing an exact character match for the starting "$" in the topic 🙂

leegeth: how does the topic matching happens if there are 2 subscriptions in the client for topics “+/monitor/Clients” and “$SYS/monitor/Clients”?

aggarw13: Good point, have added logic to avoid wildcard match for incoming PUBLISH topic names starting with the "$" character

sarenameas:
So this will be closed?

PR #1094, opened 2020-08-04
Add subscription manager for MQTT

aggarw13:
**Description of changes:**
Add a subscription manager library (under `/demos` folder) for handling multiple subscriptions to using the same MQTT connection context.

**Purpose**
The subscription manager allows application to register different callbacks for the subscribed topic filters so that incoming PUBLISH messages on the MQTT connection can be dispatched to the specific callback of the matching topic filter. 

**Note:** This PR sanitizes the draft implementation of the library shown in the draft PR https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1091

**Testing**: A follow-up PR #1098 exhibits the use of the subscription manager in an MQTT demo (and thus, the subscription manager logic is tested through the demo)


demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,439 @@
...
+    {
+        /* The record for the topic filter already exists. */
+        LogError( ( "Failed to register callback: Record for topic filter already exists: TopicFilter=%.*s",
+                    topicFilterLength,
+                    pTopicFilter ) );
+    }
+    else
+    {
+        /* Should the topic string be copied? */
+        callbackRecordList[ availableIndex ].pTopicFilter = pTopicFilter;
```
muneebahmed10, line 387: Is the subscription manager supposed to be specific to one demo? If so, I'm wondering why it is in its own directory. If not, I think this should be changed to a static buffer and memcpy, since if anyone tries to tweak the demo then they'll see issues due to this code that's in a completely different directory

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,439 @@
...
+ * subscription manager.
+ */
+#ifndef MAX_SUBSCRIPTION_CALLBACK_RECORDS
+    #define MAX_SUBSCRIPTION_CALLBACK_RECORDS    5
+#endif
+
+/**
+ * @brief The registry to store records of topic filters and their subscription callbacks.
+ */
+static SubscriptionManager_Record_t callbackRecordList[ MAX_SUBSCRIPTION_CALLBACK_RECORDS ] = { 0 };
```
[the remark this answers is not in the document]
aggarw13: Yes, that is the intended use of the manager here as a singleton. The alternative would be to use a _context_ for the subscription manager that the application would pass for multiple instances of subscription manager.
I think for the purposes of both a single and multi-threaded demo, a singleton subscription manager is suffice for now. If we have a need for multiple subscription managers in the future, we can expose the context to be application provided

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,439 @@
...
+    assert( pTopicName != NULL );
+    assert( topicNameLength != 0 );
+    assert( pTopicFilter != NULL );
+    assert( topicFilterLength != 0 );
+
+    bool status = false;
+
+    /* Check for an exact match if the incoming topic name and the registered
+     * topic filter length match. */
+    if( topicNameLength == topicFilterLength )
```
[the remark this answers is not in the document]
aggarw13: Good point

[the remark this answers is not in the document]
aggarw13: Have updated the logic to accommodate the case

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,439 @@
...
+void SubscriptionManager_DispatchHandler( MQTTContext_t * pContext,
+                                          MQTTPublishInfo_t * pPublishInfo )
+{
+    assert( pPublishInfo != NULL );
+    assert( pContext != NULL );
+
+    size_t listIndex = 0u;
+
+    /* Iterate through record list to find matching topics, and invoke their callbacks. */
+    while( listIndex < MAX_SUBSCRIPTION_CALLBACK_RECORDS )
```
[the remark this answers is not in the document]
aggarw13: Updated to use `for` loop

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,439 @@
...
+        if( pTopicName[ nameIndex ] == pTopicFilter[ filterIndex ] )
+        {
+            /* Handle special corner cases regarding wildcards at the end of
+             * topic filters, as documented by the MQTT protocol spec. */
+            matchFound = matchEndWildcards( pTopicFilter,
+                                            topicFilterLength,
+                                            topicNameLength,
+                                            nameIndex,
+                                            filterIndex,
+                                            &status );
```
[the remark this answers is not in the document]
aggarw13: Yes, that's a good point. I had most of the code as-is from the v4_beta branch [here](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/v4_beta/libraries/standard/mqtt/src/iot_mqtt_subscription.c#L148-L279).
Have updated the names to be related to their meaning.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: The subscription manager is meant to be used by other service client libraries (like Shadow), and is thus, placed in its own folder along with creation of its library target so that demo targets can link against it.
Is your concern that the customer may tweak the demo to use a stack based string pointer instead of a global, and thus, run into memory-out-of scope access problems? 
Would it be fine to mitigate the concern by adding comments in the demo code about keeping the topic filters as global string literals?

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
```diff
@@ -0,0 +1,111 @@
...
+ * 3. Include the header file "logging_stack.h".
+ */
+
+/* Include header that defines log levels. */
+#include "logging_levels.h"
+
+/* Logging configuration for the transport interface implementation which uses
+ * OpenSSL and Sockets. */
+#ifndef LIBRARY_LOG_NAME
+    #define LIBRARY_LOG_NAME     "Transport_OpenSSL_Sockets"
```
muneebahmed10, line 47: Seems like a weird name for this tbh

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
muneebahmed10: Yes, we should at least document this somewhere, such as where SubscriptionManager_Record_t is typedef'ed. We should mention the limitation of using a const char * instead of a static buffer

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,441 @@
...
+    assert( pTopicFilter != NULL );
+    assert( topicFilterLength != 0 );
+
+    bool status = false;
+
+    /* Check for an exact match if the incoming topic name and the registered
+     * topic filter length match. */
+    if( topicNameLength == topicFilterLength )
+    {
+        status = ( strncmp( pTopicName, pTopicFilter, topicNameLength ) == 0 ) ? true : false;
```
muneebahmed10, line 294: If we're not subjecting this to MISRA, then I think we should leave out the extraneous ternary
```suggestion
        status = ( strncmp( pTopicName, pTopicFilter, topicNameLength ) == 0 );
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    /* Match against wildcard characters in topic filter only if the incoming
+     * PUBLISH topic name does not start with a "$" character. */
+    if( ( status == false ) && ( pTopicName[ 0 ] != '$' ) )
```
muneebahmed10, line 299: See [Topics beginning with $](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718108). What if the topic name is `$SYS/monitor` and the topic filter is `$SYS/#`. Do we not want to match this as well?

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,441 @@
...
+    /* Search for an available spot in the list to store the record */
+    while( ( availableIndex < MAX_SUBSCRIPTION_CALLBACK_RECORDS ) &&
+           ( callbackRecordList[ availableIndex ].pTopicFilter != NULL ) )
+    {
+        /* Check if a record for the topic filter already exists in the registry. */
+        if( ( callbackRecordList[ availableIndex ].topicFilterLength == topicFilterLength ) &&
+            ( strncmp( pTopicFilter, callbackRecordList[ availableIndex ].pTopicFilter, topicFilterLength )
+              == 0 ) )
+        {
+            recordExists = true;
```
muneebahmed10, line 366: I think we mentioned this in the other PR, but we ought to document that this doesn't allow multiple callbacks for the same topic filter. I think it's plausible that more than one application wants to receive messages from the same topic filter, in which case we should point to the code that needs to be changed in order to allow it.

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
```diff
@@ -0,0 +1,111 @@
...
+ * @param[in] pTopicFilter The topic filter to register the callback for.
+ * @param[in] topicFilterLength The length of the topic filter string.
+ * @param[in] callback The callback to be registered for the topic filter.
+ *
+ * @return true if registration of the callback is successful; otherwise, false
+ * if the either the registry is full OR a registered callback already exists for
+ * the topic filter in the subscription manager.
+ */
+bool SubscriptionManager_RegisterCallback( const char * pTopicFilter,
+                                           uint16_t topicFileterLength,
```
muneebahmed10, line 97:
```suggestion
                                           uint16_t topicFilterLength,
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @brief Utility to remove the callback registered for a topic filter from the
+ * subscription manager.
+ *
+ * @param[in] pTopicFilter The topic filter to remove from the subscription manager.
+ * @param[in] topicFilterLength The length of the topic filter string.
+ */
+void SubscriptionManager_RemoveCallback( const char * pTopicFilter,
+                                         uint16_t topicFileterLength );
```
muneebahmed10, line 108:
```suggestion
                                         uint16_t topicFilterLength );
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,441 @@
...
+
+/*-----------------------------------------------------------*/
+
+void SubscriptionManager_RemoveCallback( const char * pTopicFilter,
+                                         uint16_t topicFilterLength )
+{
+    size_t matchingRecordIndex = 0u;
+    bool recordFound = false;
+    SubscriptionManager_Record_t * pRecord = NULL;
+
```
muneebahmed10, line 411: Should probably `assert( pTopicFilter != NULL );`

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,441 @@
...
+            recordExists = true;
+        }
+
+        availableIndex++;
+    }
+
+    if( ( availableIndex == MAX_SUBSCRIPTION_CALLBACK_RECORDS ) )
+    {
+        /* The record list is full. */
+        LogError( ( "Unable to register callback: Registry list is full: TopicFilter=%.*s",
```
muneebahmed10, line 375: In this case might be helpful to log what the max capacity actually is

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
aggarw13: Oops, copy-pasta, it is.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: I prefer that we keep it to be MISRA compliant, as we could consider making this a library (as subscription manager is needed by all MQTT based service client library demos)

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Added a `@note` in the documentation of the `RegisterCallback` function

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Added

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
aggarw13: Fixed

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Added a `@note` in the API documentation of the function in the header file

aggarw13: I don't think that adding documentation for supporting multiple callback registration per topic filter is necessarily required. (The source code is available for viewing and tweaking). 
If we really think  that there can be a customer need for multiple callbacks per topic filter, then we should just support that in the subscription manager. From the perspective of a single threaded application, I don;t see a reason why multiple callbacks per topic filter would be needed.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Good catch! Have updated the logic to allow wildcard matching for reserved topics when wildcards are at a non-beginning position

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+     * topic filter length match. */
+    if( topicNameLength == topicFilterLength )
+    {
+        status = ( strncmp( pTopicName, pTopicFilter, topicNameLength ) == 0 ) ? true : false;
+    }
+
+    /* If an exact match was not found, match against wildcard characters in topic filter.
+     * Note: According to the MQTT 3.1.1 specification, incoming PUBLISH topic names starting
+     * the "$" character cannot be matched against topic filter starting with a wildcard. */
+    if( ( status == false ) && ( ( pTopicName[ 0 ] != '$' ) &&
```
aggarg, line 300: Not sure I understand this condition. Are you trying to prevent the following:

```c
! ( ( pTopicName[ 0 ] == '$' ) &&  ( ( pTopicFilter[ 0 ] == '+' ) || ( pTopicFilter[ 0 ] == '#' ) ) )
```

i.e. If the topic on which the message is received, starts with `$` then it should match the registered topic filter starting with `#` or `+`.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+                               uint16_t topicFilterLength,
+                               uint16_t topicNameLength,
+                               uint16_t nameIndex,
+                               uint16_t filterIndex )
+{
+    bool matchFound = false, endChar = false;
+
+    /* Determine if the last character is reached for the topic name, and the
+     * second last character for the topic filter to match for the '#' wildcard. */
+    endChar = ( nameIndex == ( topicNameLength - 1U ) ) &&
```
aggarg, line 142: The name `endChar` seems a bit misleading.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+    if( pTopicFilter[ filterIndex ] == '+' )
+    {
+        /* Move topic name index to the end of the current level.
+         * This is identified by '/'. */
+        while( ( *pNameIndex < topicNameLength ) && ( pTopicName[ *pNameIndex ] != '/' ) )
+        {
+            ( *pNameIndex )++;
+        }
+
+        ( *pNameIndex )--;
```
aggarg, line 192: Please write a comment why this move back is needed.

- If the break condition is `( *pNameIndex < topicNameLength )`, the corresponding index increment is done in `topicFilterMatch`.
- If the break condition is `( pTopicName[ *pNameIndex ] != '/' )`, we move back the index on the '/' character. 

In the second case, do we match the '/' character in the topicName and filter or both the indexes are just skipped in the `topicFilterMatch`? In other words, would filter `/topic/+abc` match `/topic/part/bc`?

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+ * @brief Match a topic name and topic filter allowing the use of wildcards.
+ *
+ * @param[in] pTopicName The topic name to check.
+ * @param[in] topicNameLength Length of the topic name.
+ * @param[in] pTopicFilter The topic filter to check.
+ * @param[in] topicFilterLength Length of topic filter.
+ *
+ * @return `true` if the topic name and topic filter match; `false` otherwise.
+ */
+static bool topicFilterMatch( const char * pTopicName,
```
aggarg, line 109: For consistency with the other three functions, please change the name to matchTopicFilter.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+/**
+ * @brief Handle special corner cases regarding wildcards at the end of topic
+ * filters, as documented by the MQTT protocol spec.
+ *
+ * @param[in] pTopicFilter The topic filter containing the wildcard.
+ * @param[in] topicFilterLength Length of the topic filter being examined.
+ * @param[in] topicNameLength Length of the topic name being examined.
+ * @param[in] nameIndex Index of the topic name being examined.
+ * @param[in] filterIndex Index of the topic filter being examined.
+ * @param[out] pMatch Whether the topic filter and topic name match.
```
aggarg, line 68: There is no pMatch.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+    #define MAX_SUBSCRIPTION_CALLBACK_RECORDS    5
+#endif
+
+/**
+ * @brief The registry to store records of topic filters and their subscription callbacks.
+ */
+static SubscriptionManager_Record_t callbackRecordList[ MAX_SUBSCRIPTION_CALLBACK_RECORDS ] = { 0 };
+
+/**
+ * @brief Handle special corner cases regarding wildcards at the end of topic
```
aggarg, line 60: The description is vague. Please elaborate the description.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+ * @return Returns whether the topic filter and the topic name match.
+ */
+static bool matchEndWildcards( const char * pTopicFilter,
+                               uint16_t topicFilterLength,
+                               uint16_t topicNameLength,
+                               uint16_t nameIndex,
+                               uint16_t filterIndex );
+
+/**
+ * @brief Attempt to match characters in a topic filter by wildcards.
```
aggarg, line 79: Please elaborate the description. If the current character in the topic filter is '+', it consumes one level form the topic name. If the If the current character in the topic filter is '#', it assumes that the topic name matches and stops matching.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+                    topicFilterLength,
+                    pTopicFilter ) );
+    }
+    else
+    {
+        LogWarn( ( "Attempted to remove callback for un-registered topic filter: TopicFilter=%.*s",
+                   topicFilterLength,
+                   pTopicFilter ) );
+    }
+}
```
aggarg, line 448:
```suggestion
}
/*-----------------------------------------------------------*/

```

demos/mqtt/subscription-manager/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+set( LIBRARY_NAME "mqtt_subscription_manager" )
```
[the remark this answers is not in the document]
aggarw13: This is being added as a library instead of a demo binary so that multiple demos can call it (and thereby, link against it)

demos/mqtt/subscription-manager/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+# Library target.
+add_library( ${LIBRARY_NAME}
+            "${LIBRARY_NAME}.c" )
+
+target_link_libraries(
+    ${LIBRARY_NAME}
+    PRIVATE
+        mqtt
```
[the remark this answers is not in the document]
aggarw13: The `mqtt` library does not have any dependency on the subscription manager, but subscription manager depends on the mqtt for its types like `MQTTPublishInfo_t`. 
Practically, only a header dependency on mqtt exists for the subscription manager; so a linking dependency is not strictly required. (I can remove this linking dependency if you think that will be better, but the header include path dependency would definitely remain)

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Updated description to mention the special cases handled in the utility

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Removed

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
```diff
@@ -0,0 +1,117 @@
...
+/* Logging related header files are required to be included in the following order:
+ * 1. Include the header file "logging_levels.h".
+ * 2. Define LIBRARY_LOG_NAME and  LIBRARY_LOG_LEVEL.
+ * 3. Include the header file "logging_stack.h".
+ */
+
+/* Include header that defines log levels. */
+#include "logging_levels.h"
+
+/* Logging configuration for the transport interface implementation which uses
```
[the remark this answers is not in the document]
aggarw13: Fixed

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Couldn't find an appropriate name, so have removed the use of `endChar` altogether to instead have the conditional expression at `if`

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Added description for the matching logic performed in the function

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: Thanks for the suggestion. 

For your question on:
> In the second case, do we match the '/' character in the topicName and filter or both the indexes are just skipped in the topicFilterMatch? In other words, would filter /topic/+abc match /topic/part/bc?

The `topicFilterMatch` will match the `/` character in the next iteration of the loop after both the nameIndex and filterIndex indices are incremented, thus the invalid topic filter `/topic/+abc` cannot match `/topic/part/bc`

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+{
+    assert( pTopicFilter != NULL );
+    assert( topicFilterLength != 0 );
+
+    bool recordAdded = false;
+    size_t availableIndex = 0u;
+    bool recordExists = false;
+
+    /* Search for an available spot in the list to store the record */
+    while( ( availableIndex < MAX_SUBSCRIPTION_CALLBACK_RECORDS ) &&
```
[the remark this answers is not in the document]
aggarw13: Good catch about the possibility of adding multiple callbacks in the registry when gaps exist in the records list. I have updated the logic in a slightly modified form of the one you have suggested (primarily to end the loop when `recordExists` is set to true)

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+    else if( recordExists == true )
+    {
+        /* The record for the topic filter already exists. */
+        LogError( ( "Failed to register callback: Record for topic filter already exists: TopicFilter=%.*s",
+                    topicFilterLength,
+                    pTopicFilter ) );
+    }
+    else
+    {
+        /* Should the topic string be copied? */
```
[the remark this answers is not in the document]
aggarw13: Removed the comment

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,448 @@
...
+    bool recordFound = false;
+    SubscriptionManager_Record_t * pRecord = NULL;
+
+    /* Iterate through the records list to find the matching record. */
+    do
+    {
+        pRecord = &callbackRecordList[ matchingRecordIndex ];
+
+        if( ( topicFilterLength == pRecord->topicFilterLength ) &&
+            ( strncmp( pTopicFilter, pRecord->pTopicFilter, topicFilterLength ) == 0 ) )
```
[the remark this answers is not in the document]
aggarw13: Yes, good catch of the bug! Have updated it per your suggestion for better readability

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
```diff
@@ -0,0 +1,144 @@
...
+
+/* Enumeration type for return status value from Subscription Manager API. */
+typedef enum SubscriptionManagerStatus
+{
+    SUBSCRIPTION_MANAGER_INVALID = 0,
+
+    /**
+     * @brief Success return value from Subscription Manager API.
+     */
+    SUBSCRIPTION_MANAGER_SUCCESS = 1,
```
muneebahmed10, line 67: Nit: Usually 0 means success, is there any special reason for this?

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,487 @@
...
+
+    /* Check for an exact match if the incoming topic name and the registered
+     * topic filter length match. */
+    if( topicNameLength == topicFilterLength )
+    {
+        status = ( strncmp( pTopicName, pTopicFilter, topicNameLength ) == 0 ) ? true : false;
+    }
+
+    /* Determine if topic name starts with '$'. */
+    topicNameStartsWithDollar = ( pTopicName[ 0 ] == '$' ) ? true : false;
```
muneebahmed10, line 312: This one seems a little unnecessary, since `pTopicName[ 0 ] == '$'` actually has fewer characters than `topicNameStartsWithDollar == true`, so it doesn't even save space.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,487 @@
...
+    if( recordExists == true )
+    {
+        /* The record for the topic filter already exists. */
+        LogError( ( "Failed to register callback: Record for topic filter already exists: TopicFilter=%.*s",
+                    topicFilterLength,
+                    pTopicFilter ) );
+
+        returnStatus = SUBSCRIPTION_MANAGER_RECORD_EXISTS;
+    }
+    else if( ( availableIndex == MAX_SUBSCRIPTION_CALLBACK_RECORDS ) )
```
muneebahmed10, line 416: I think the second set of parenthesis are extraneous
```suggestion
    else if( availableIndex == MAX_SUBSCRIPTION_CALLBACK_RECORDS )
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
aggarw13: @dan4thewin has generally suggested to avoid using 0 as a success value as it could be a statically allocated default value assigned for variables and result in false positive validation in tests. We had followed the same philosophy for the simplied taskpool's enumeration values: https://github.com/FreeRTOS/FreeRTOS/blob/lts-development/FreeRTOS-Plus/Source/FreeRTOS-IoT-Libraries-LTS-Beta1/abstractions/platform/freertos/private/taskpool/taskpool_freertos.h#L43-L64

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarw13: That's a valid point. Have removed the local variable

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -0,0 +1,483 @@
...
+
+    /* Determine if topic filters starts with a wildcard. */
+    topicFilterStartsWithWildcard = ( ( pTopicFilter[ 0 ] == '+' ) ||
+                                      ( pTopicFilter[ 0 ] == '#' ) ) ? true : false;
+
+    /* If an exact match was not found, match against wildcard characters in topic filter.
+     * Note: According to the MQTT 3.1.1 specification, incoming PUBLISH topic names starting
+     * the "$" character cannot be matched against topic filter starting with a wildcard.
+     * i.e. for example, "$SYS/sport" cannot be matched with "#" or "+/sport" topic filters. */
+    if( ( status == false ) && !( ( ( pTopicName[ 0 ] == '$' ) ) &&
```
muneebahmed10, line 318: Too many parentheses?
```suggestion
    if( ( status == false ) && !( ( pTopicName[ 0 ] == '$' ) &&
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.h
```diff
@@ -67,13 +67,14 @@ typedef enum SubscriptionManagerStatus
...
+     * @brief Failure return value due to registry being full.
      */
     SUBSCRIPTION_MANAGER_REGISTRY_FULL = 2,
 
     /**
-     * @brief Failure return value due an existing record in registry being present
-     * for the topic filter that is requested for a new callback registration.
+     * @brief Failure return value due an existing record in registry being
+     * present for the topic filter that is requested for a new callback
+     * registration.
```
muneebahmed10, line 77: Possible rewording:
```suggestion
     * @brief Failure return value due to an already existing record in the
     * registry for a new callback registration's requested topic filter.
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -307,18 +310,23 @@ static bool matchTopic( const char * pTopicName,
...
+
+        /* Determine if topic filter starts with a wildcard. */
+        topicFilterStartsWithWildcard = ( ( pTopicFilter[ 0 ] == '+' ) ||
+                                          ( pTopicFilter[ 0 ] == '#' ) ) ? true : false;
+
+        /* Note: According to the MQTT 3.1.1 specification, incoming PUBLISH topic names
+         * starting the "$" character cannot be matched against topic filter starting with
+         * a wildcard, i.e. for example, "$SYS/sport" cannot be matched with "#" or
+         * "+/sport" topic filters. */
+        if( !( ( ( pTopicName[ 0 ] == '$' ) ) && ( topicFilterStartsWithWildcard == true ) ) )
```
muneebahmed10, line 326:
```suggestion
        if( !( ( pTopicName[ 0 ] == '$' ) && ( topicFilterStartsWithWildcard == true ) ) )
```
or even
```suggestion
        if( ( pTopicName[ 0 ] != '$' ) || ( topicFilterStartsWithWildcard == false ) )
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -149,25 +149,28 @@ static bool matchWildcardsSpecialCases( const char * pTopicFilter,
     bool matchFound = false;
 
     /* Determine if the last character is reached for the topic name, and the
-     * third last character for the topic filter to match for the '#' wildcard. */
+     * third last character is reached for the topic filter. */
```
muneebahmed10, line 152:
```suggestion
     * third to last character is reached for the topic filter. */
```

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -149,25 +149,28 @@ static bool matchWildcardsSpecialCases( const char * pTopicFilter,
...
+         * as well as "sport/tennis" topics. */
         matchFound = ( ( pTopicFilter[ filterIndex + 1U ] == '/' ) &&
                        ( pTopicFilter[ filterIndex + 2U ] == '#' ) ) ? true : false;
     }
     else
     {
         /* Determine if the last character is reached for the topic name and,
-         * the second last character for the topic filter to match for the
-         * '+' wildcard. */
+         * the second last character is reached for the topic filter. */
```
muneebahmed10, line 166:
```suggestion
         * the second to last character for the topic filter. */
```

aggarg: I think you did not want to remove "is reached".

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
aggarg: I would prefer the first of the two suggestions as it more clearly expresses the intention which is - do not match the topic starting with $ with filters starting with wildcards.

demos/mqtt/subscription-manager/mqtt_subscription_manager.c
muneebahmed10: That was intentional, I didn't think was necessary to repeat the "is reached" verb for both the 'last character' and 'second to last character' subjects. This is a really minor point though so it's fine if it's changed back.

PR #1096, opened 2020-08-04
Minor hygiene updates for logging

aggarw13:
Changes include:
* Revert of recent change made in #1095 as `__FUNCTION__` is not ISO C specified. Address the concern of long `__FILE__` macro in a modified way to only log the file name (instead of the absolute file path) in the log message as metadata.
* Update default log level of unit tests for MQTT and HTTP libraries to be `LOG_ERROR`

### DIFF
--- a/demos/logging-stack/logging_stack.h
+++ b/demos/logging-stack/logging_stack.h
@@ -34,6 +34,7 @@
 /* Standard Include. */
 #include <stdio.h>
 #include <stdint.h>
+#include <string.h>
 
 /* Macro to extract only the file name from file path to use for metadata in
  * log messages. */
--- a/demos/logging-stack/logging_stack.h
+++ b/demos/logging-stack/logging_stack.h
@@ -35,9 +35,13 @@
 #include <stdio.h>
 #include <stdint.h>
 
+/* Macro to extract only the file name from file path to use for metadata in
+ * log messages. */
+#define __FILENAME__           ( strrchr( __FILE__, '/' ) ? strrchr( __FILE__, '/' ) + 1 : __FILE__ )
+
 /* Metadata information to prepend to every log message. */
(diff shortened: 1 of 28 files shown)


demos/logging-stack/logging_stack.h
```diff
@@ -34,10 +34,15 @@
 /* Standard Include. */
 #include <stdio.h>
 #include <stdint.h>
+#include <string.h>
+
+/* Macro to extract only the file name from file path to use for metadata in
+ * log messages. */
+#define __FILENAME__           ( strrchr( __FILE__, '/' ) ? strrchr( __FILE__, '/' ) + 1 : __FILE__ )
```
muneebahmed10, line 41: I think we should avoid defining macros that begin with double underscore, since those can be reserved,

libraries/standard/mqtt/utest/mqtt_config.h
```diff
@@ -20,7 +20,7 @@
 #endif
 
 #ifndef LIBRARY_LOG_LEVEL
-    #define LIBRARY_LOG_LEVEL    LOG_NONE
+    #define LIBRARY_LOG_LEVEL    LOG_ERROR
```
muneebahmed10, line 23: Does this mean unit test results will be intertwined with error logs, or will they still be separate?. What's the advantage of this?

aggarw13: It is a change based on @cobusve's code expressivity feedback about always having the error level as the default logging verbosity, and instead override with CLI/Build configuration instead of the other way round.

demos/logging-stack/logging_stack.h
aggarw13: Great point. Removed the double underscores

PR #1097, opened 2020-08-05
Improve format specifiers in logs

muneebahmed10:
*Description of changes:*
Remove C99 %hhu format specifiers, add casts to unsigned long wherever %lu is used for size_t, and log strings instead of status codes

### DIFF
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -716,7 +716,7 @@ static int establishMqttSession( MQTTContext_t * pMqttContext,
     if( mqttStatus != MQTTSuccess )
     {
         returnStatus = EXIT_FAILURE;
-        LogError( ( "MQTT init failed with status %s.", MQTT_Status_strerror( mqttStatus ) ) );
+        LogError( ( "MQTT init failed: Status=%s.", MQTT_Status_strerror( mqttStatus ) ) );
     }
     else
     {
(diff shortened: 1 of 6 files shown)


demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -686,7 +686,7 @@ static int establishMqttSession( MQTTContext_t * pMqttContext,
     if( mqttStatus != MQTTSuccess )
     {
         returnStatus = EXIT_FAILURE;
-        LogError( ( "MQTT init failed with status %u.", mqttStatus ) );
+        LogError( ( "MQTT init failed with status %s.", MQTT_Status_strerror( mqttStatus ) ) );
```
aggarw13, line 689: (Nit) Log messages could be crisper (to follow the logging message format) as
```suggestion
        LogError( ( "MQTT init failed: Status=%s.", MQTT_Status_strerror( mqttStatus ) ) );
```

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -1383,7 +1382,7 @@ MQTTStatus_t MQTT_GetConnectPacketSize( const MQTTConnectInfo_t * pConnectInfo,
         LogError( ( "The Will Message length must not exceed %d. "
                     "pWillInfo->payloadLength=%lu.",
                     UINT16_MAX,
```
aggarw13, line 1384: Should the UINT16_MAX value be logged as `%u`?
```suggestion
        LogError( ( "The Will Message length must not exceed %u. "
                    "pWillInfo->payloadLength=%lu.",
                    UINT16_MAX,
```

muneebahmed10: In gcc's `stdint.h` I see this:
```
#define UINT8_MAX         255
#define UINT16_MAX        65535
#define UINT32_MAX        4294967295U
#define UINT64_MAX        18446744073709551615ULL
```
`UINT16_MAX` doesn't seem to be declared unsigned as `UINT32_MAX` is, so I think it's best to leave as is.

PR #1101, opened 2020-08-06
Use MQTT_PublishToResend() in the basic_tls and mutual_auth demos

sarenameas:
***Description of changes:***
For demonstration purposes MQTT_PublishToResend() is now used in the mqtt_basic_tls_demo and the mqtt_mutual_auth_demo. These demos use a Qos > 0. 

***Extra changes:***
Updated the README.md in tools/spell for clarity on how to use the scripts.

### DIFF
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -500,14 +500,17 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
     uint8_t index = 0U;
     MQTTStateCursor_t cursor = MQTT_STATE_CURSOR_INITIALIZER;
     uint16_t packetIdToResend = MQTT_PACKET_ID_INVALID;
+    bool foundPacketId = false;
 
     assert( pMqttContext != NULL );
     assert( outgoingPublishPackets != NULL );
 
-    /* For the sake of demonstrating how to use MQTT_PublishToResend(), the
-     * outgoingPublishPackets array is iterated through for the next pending
-     * publish packet to be resent. If the application requires increased
-     * efficiency in the look up of the packetIdToResend, then a hashmap of
+    /* MQTT_PublishToResend() provides a packet ID of the next PUBLISH packet
+     * that should be resent. In accordance with the MQTT v3.1.1 spec,
+     * MQTT_PublishToResend() preserves the ordering of when the original
+     * PUBLISH packets were sent. The outgoingPublishPackets array is searched
+     * through for the associated packet ID. If the application requires
+     * increased efficiency in the look up of the packet ID, then a hashmap of
      * packetId key and PublishPacket_t values may be used instead. */
     packetIdToResend = MQTT_PublishToResend( pMqttContext, &cursor );
 
@@ -517,6 +520,7 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
         {
             if( outgoingPublishPackets[ index ].packetId == packetIdToResend )
             {
+                foundPacketId = true;
                 outgoingPublishPackets[ index ].pubInfo.dup = true;
 
                 LogInfo( ( "Sending duplicate PUBLISH with packet id %u.",
@@ -542,8 +546,19 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
             }
         }
 
-        /* Get the next packetID to be resent. */
-        packetIdToResend = MQTT_PublishToResend( pMqttContext, &cursor );
+        if( foundPacketId == false )
+        {
+            LogError( ( "Packet id %u requires resend, but was not found in "
+                        "outgoingPublishPackets.",
+                        packetIdToResend ) );
+            returnStatus = EXIT_FAILURE;
+            break;
+        }
+        else
+        {
+            /* Get the next packetID to be resent. */
+            packetIdToResend = MQTT_PublishToResend( pMqttContext, &cursor );
+        }
     }
 
     return returnStatus;
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -511,7 +511,7 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
      * packetId key and PublishPacket_t values may be used instead. */
     packetIdToResend = MQTT_PublishToResend( pMqttContext, &cursor );
 
-    while( packetIdToResend != 0 )
+    while( packetIdToResend != MQTT_PACKET_ID_INVALID )
     {
         for( index = 0U; index < MAX_OUTGOING_PUBLISHES; index++ )
         {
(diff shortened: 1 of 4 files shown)


demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -51,6 +51,7 @@
 
 /* MQTT API header. */
```
leegeth, line 52: headers

sarenameas: Thanks

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -497,39 +498,52 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
     int returnStatus = EXIT_SUCCESS;
     MQTTStatus_t mqttStatus = MQTTSuccess;
     uint8_t index = 0U;
+    MQTTStateCursor_t cursor = MQTT_STATE_CURSOR_INITIALIZER;
+    uint16_t packetIdToResend = 0;
```
aggarw13, line 502: (nit)
```suggestion
    uint16_t packetIdToResend = MQTT_PACKET_ID_INVALID;
```

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -497,39 +498,52 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
...
-     * received, the publish is removed from the array. */
-    for( index = 0U; index < MAX_OUTGOING_PUBLISHES; index++ )
+    /* For the sake of demonstrating how to use MQTT_PublishToResend(), the
+     * outgoingPublishPackets array is iterated through for the next pending
+     * publish packet to be resent. If the application requires increased
+     * efficiency in the look up of the packetIdToResend, then a hashmap of
+     * packetId key and PublishPacket_t values may be used instead. */
+    packetIdToResend = MQTT_PublishToResend( pMqttContext, &cursor );
+
+    while( packetIdToResend != 0 )
```
aggarw13, line 514: (Nit for code expressivity)
```suggestion
    while( packetIdToResend != MQTT_PACKET_ID_INVALID )
```

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -497,39 +498,52 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
...
+                    break;
+                }
+                else
+                {
+                    LogInfo( ( "Sent duplicate PUBLISH successfully for packet id %u.\n\n",
+                               outgoingPublishPackets[ index ].packetId ) );
+                }
             }
         }
+
```
aggarw13, line 544: What about the case where the packet ID provided by `MQTT_PublishToResend` isn't present in the array? We should probably consider that as an error state

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -497,39 +498,52 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
...
+    assert( pMqttContext != NULL );
     assert( outgoingPublishPackets != NULL );
 
-    /* Resend all the QoS2 publishes still in the array. These are the
-     * publishes that hasn't received a PUBREC. When a PUBREC is
-     * received, the publish is removed from the array. */
-    for( index = 0U; index < MAX_OUTGOING_PUBLISHES; index++ )
+    /* For the sake of demonstrating how to use MQTT_PublishToResend(), the
+     * outgoingPublishPackets array is iterated through for the next pending
+     * publish packet to be resent. If the application requires increased
```
aggarw13, line 509: The comment may not be clear to people with the context of the utility of the MQTT function and how it is used in this demo app. It would be more helpful to provide some information about what this function does. 
```suggestion
    /* The MQTT_PublishToResend() provides the next PUBLISH packet that should be resent keeping the message 
       * ordering preservation mentioned in the MQTT 3.1.1 specification. The provided publish packet is then searched in the
       * outgoingPublishPackets array to find it's associated PUBLISH message for attempting a re-publish operation.
```

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
sarenameas: Thanks

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
sarenameas: Thanks

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
sarenameas: Good idea

sarenameas:
> I'm generally concerned that the use of `MQTT_PublishToResend` seems superfluous when the `outgoingPublishPackets` array already does the same work of maintaining PUBLISHes for resending?
> 
> We probably should show `MQTT_PublishToResend` in a way that makes its utility more meaningful as the only way of storing PUBLISH packet ID order; otherwise, the use of the utility carries no need for the customer...

What do you suggest? The application must still own their publishes and associate them packet IDs.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
sarenameas: Sure

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -511,7 +511,7 @@ static int handlePublishResend( MQTTContext_t * pMqttContext )
      * packetId key and PublishPacket_t values may be used instead. */
     packetIdToResend = MQTT_PublishToResend( pMqttContext, &cursor );
 
-    while( packetIdToResend != 0 )
+    while( packetIdToResend != MQTT_PACKET_ID_INVALID )
```
aggarw13, line 514: Can you make similar updates in other demos as well?

aggarw13:
>>I'm generally concerned that the use of MQTT_PublishToResend seems superfluous when the outgoingPublishPackets array already does the same work of maintaining PUBLISHes for resending?

>>We probably should show MQTT_PublishToResend in a way that makes its utility more meaningful as the only way of storing PUBLISH packet ID order; otherwise, the use of the utility carries no need for the customer...

>What do you suggest? My understanding is that the application must still own their publish packets and associate them packet IDs.

I spoke with @leegeth separately, and after some thinking through, we realized that the `outgoingPublishPackets` array in the current implementation actually does not preserve the message ordering. Thus, using the `MQTT_PublishToResend` is a must in these demo applications. Thus, we need to provide the correct documentation to suggest the value using this utility for preserving the message ordering

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
sarenameas: was not done

PR #1105, opened 2020-08-10
DRAFT: Update proof templates, add proof flag --pointer-primitive-check

markrtuttle:
This pull request updates the proof template submodule to [include checks missing from continuous integration](https://github.com/awslabs/aws-templates-for-cbmc-proofs/pull/20), and added a flag --pointer-primitive-check to the proofs to check the soundness of assumptions by checking that all pointers in pointer primitives are valid or null.

### DIFF
--- a/libraries/standard/http/cbmc/proofs/Makefile-project-defines
+++ b/libraries/standard/http/cbmc/proofs/Makefile-project-defines
@@ -27,3 +27,7 @@ INCLUDES += -I$(SRCDIR)/platform/include
 
 # Preprocessor definitions -D...
 DEFINES += -Dhttp_EXPORTS
+
+# Ensure that all assumptions are sound by checking that "all pointers
+# in pointer primitives are valid or null"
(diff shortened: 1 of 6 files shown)


libraries/standard/http/cbmc/proofs/Makefile-project-defines
```diff
@@ -27,3 +27,7 @@ INCLUDES += -I$(SRCDIR)/platform/include
 
 # Preprocessor definitions -D...
 DEFINES += -Dhttp_EXPORTS
+
+# Ensure that all assumptions are sound by checking that "all pointers
+# in pointer primitives are valid or null"
+CHECKFLAGS += --pointer-primitive-check
```
sarenameas, line 33: Shouldn't this be in makefile.common?

sarenameas: Per offline conversation we are using it here first but cbmc team wishes not to break other projects.

PR #1108, opened 2020-08-11
[PR from CLI tool] Add new MQTT API for obtaining SUBACK codes

aggarw13:
### Problem
AWS IoT can reject a subscription request due to throttling, and thus, the application needs to know which topic filter (out of the list of topic filters) that were requested for subscription in the same SUBSCRIBE packet have been rejected. The application can retry subscription after back-off.

### Solution
Add a new API, `MQTT_GetSubAckPayload`, to obtain the server returned codes of subscription statuses of all topic filters requested in the SUBSCRIBE packet. 

This utility will help application to determine rejected topic filter(s) after it is notified about a failed subscription response through the event callback from the MQTT library. Once the failed topic filter is determined, then the application can re-request subscription of the rejected topic filter.


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -950,15 +950,15 @@ static MQTTStatus_t readSubackStatus( size_t statusCount,
         /* MQTT 3.1.1 defines the following values as status codes. */
         switch( subscriptionStatus )
         {
-            case 0x00:
-            case 0x01:
-            case 0x02:
+            case MQTTQoS0:
+            case MQTTQoS1:
+            case MQTTQoS2:
```
muneebahmed10, line 955: Please revert this, this is a MISRA 10.3 violation
```suggestion
            case 0x00:
            case 0x01:
            case 0x02:
```

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -2189,6 +2189,63 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * pIncomingPacket,
     return status;
 }
 
+/*-----------------------------------------------------------*/
+
+MQTTStatus_t MQTT_GetSubAckPayload( MQTTPacketInfo_t * pSubackPacket,
```
muneebahmed10, line 2194: The first parameter can be `const`
```suggestion
MQTTStatus_t MQTT_GetSubAckPayload( const MQTTPacketInfo_t * pSubackPacket,
```

libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -2189,6 +2189,63 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * pIncomingPacket,
...
+        status = MQTTBadParameter;
+    }
+    else
+    {
+        /* According to the MQTT 3.1.1 protocol specification, the "Remaining Length" field is a
+         * length of the variable header (2 bytes) plus the length of the payload.
+         * Therefore, we add 2 positions for the starting address of the payload, and
+         * subtract 2 bytes from the remaining length for the length of the payload.*/
+        *pPayloadStart = pSubackPacket->pRemainingData + sizeof( uint16_t );
+        *pPayloadSize = pSubackPacket->remainingLength - sizeof( uint16_t );
```
muneebahmed10, line 2242: Since `sizeof` return a `size_t` but `pPayloadSize` points to a `uint16_t`, this needs an explicit cast

libraries/standard/mqtt/include/mqtt_lightweight.h
```diff
@@ -564,6 +566,26 @@ MQTTStatus_t MQTT_DeserializeAck( const MQTTPacketInfo_t * pIncomingPacket,
...
+ * @param[in] pSubackPacket The SUBACK packet whose payload is to be parsed.
+ * @param[out] pPayloadStart This is populated with the starting address
+ * of the payload (or return codes for topic filters) in the SUBACK packet.
+ * @param[out] pPayloadSize This is populated with the size of the payload
+ * in the SUBACK packet. It represents the number of topic filters whose
+ * SUBACK status is present in the packet.
+ *
+ * @return  #MQTTBadParameter, or #MQTTSuccess.
+ */
+MQTTStatus_t MQTT_GetSubAckPayload( MQTTPacketInfo_t * pSubackPacket,
```
muneebahmed10, line 585: Should this be in `mqtt_lightweight.h`? Although it doesn't use an `MQTTContext_t`, I thought the purpose of creating a new API function was so the application wouldn't have to mix lightweight and managed API. If that's the case, I'd expect this to be in `mqtt.h`

libraries/standard/mqtt/include/mqtt_lightweight.h
```diff
@@ -40,21 +40,23 @@
...
+#define MQTT_PACKET_TYPE_SUBSCRIBE      ( ( uint8_t ) 0x82U )    /**< @brief SUBSCRIBE (client-to-server). */
+#define MQTT_PACKET_TYPE_SUBACK         ( ( uint8_t ) 0x90U )    /**< @brief SUBACK (server-to-client). */
+#define MQTT_PACKET_TYPE_UNSUBSCRIBE    ( ( uint8_t ) 0xA2U )    /**< @brief UNSUBSCRIBE (client-to-server). */
+#define MQTT_PACKET_TYPE_UNSUBACK       ( ( uint8_t ) 0xB0U )    /**< @brief UNSUBACK (server-to-client). */
+#define MQTT_PACKET_TYPE_PINGREQ        ( ( uint8_t ) 0xC0U )    /**< @brief PINGREQ (client-to-server). */
+#define MQTT_PACKET_TYPE_PINGRESP       ( ( uint8_t ) 0xD0U )    /**< @brief PINGRESP (server-to-client). */
+#define MQTT_PACKET_TYPE_DISCONNECT     ( ( uint8_t ) 0xE0U )    /**< @brief DISCONNECT (client-to-server). */
+
+/* SUBACK return code for failure. */
+#define MQTT_SUBACK_STATUS_FAILURE      ( ( uint8_t ) 0X80 )
```
muneebahmed10, line 59: Nit: use lowercase x to match the other hex constants
```suggestion
#define MQTT_SUBACK_STATUS_FAILURE      ( ( uint8_t ) 0x80 )
```

libraries/standard/mqtt/src/mqtt_lightweight.c
aggarw13: Hmmm..That's unfortunate

libraries/standard/mqtt/src/mqtt_lightweight.c
aggarw13: Added casting to `uint16_t`

libraries/standard/mqtt/src/mqtt_lightweight.c
aggarw13: Have added `MQTT_QOS_X` macros to make these case statements expressive

libraries/standard/mqtt/include/mqtt_lightweight.h
aggarw13: Okay, (imo) the function provides a packet-layer parsing utility which logically belongs to the lightweight API. 
However, based on suggestion from @abhidixi11 as well, have moved it to the `mqtt.h` layer

libraries/standard/mqtt/include/mqtt_lightweight.h
```diff
@@ -99,9 +106,9 @@ typedef enum MQTTStatus
  */
 typedef enum MQTTQoS
 {
-    MQTTQoS0 = 0, /**< Delivery at most once. */
-    MQTTQoS1 = 1, /**< Delivery at least once. */
-    MQTTQoS2 = 2  /**< Delivery exactly once. */
+    MQTTQoS0 = MQTT_QOS_0, /**< Delivery at most once. */
+    MQTTQoS1 = MQTT_QOS_1, /**< Delivery at least once. */
+    MQTTQoS2 = MQTT_QOS_2  /**< Delivery exactly once. */
```
muneebahmed10, line 111: I really don't see the point in having a macro and enum for QoS, it's just confusing to me.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1954,6 +1954,62 @@ uint16_t MQTT_GetPacketId( MQTTContext_t * pContext )
...
+        LogError( ( "Invalid parameter: pSubackPacket->pRemainingData is NULL" ) );
+        status = MQTTBadParameter;
+    }
+
+    /* A SUBACK must have a remaining length of at least 3 to accommodate the
+     * packet identifier and at least 1 return code. */
+    else if( pSubackPacket->remainingLength < 3U )
+    {
+        LogError( ( "Invalid parameter: Packet remaining length is invalid: "
+                    "Remaining length should be greater than 2" ) );
```
muneebahmed10, line 1995: Why not log what the remaining length actually is?

libraries/standard/mqtt/include/mqtt_lightweight.h
```diff
@@ -40,21 +40,28 @@
...
+#define MQTT_PACKET_TYPE_PINGRESP       ( ( uint8_t ) 0xD0U )    /**< @brief PINGRESP (server-to-client). */
+#define MQTT_PACKET_TYPE_DISCONNECT     ( ( uint8_t ) 0xE0U )    /**< @brief DISCONNECT (client-to-server). */
+
+/* QoS values in MQTT 3.1.1 specification. */
+#define MQTT_QOS_0                      ( ( uint8_t ) 0x00 )
+#define MQTT_QOS_1                      ( ( uint8_t ) 0x01 )
+#define MQTT_QOS_2                      ( ( uint8_t ) 0x02 )
+
+/* SUBACK return code for failure. */
+#define MQTT_SUBACK_STATUS_FAILURE      ( ( uint8_t ) 0x80 )
```
muneebahmed10, line 64: Would it make sense to move this macro to mqtt.h along with `MQTT_GetSubAckPayload`? I feel like those should be in the same file, if they're expected to be used together

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1954,6 +1954,62 @@ uint16_t MQTT_GetPacketId( MQTTContext_t * pContext )
...
+    }
+    else if( pPayloadSize == NULL )
+    {
+        LogError( ( "Invalid parameter: pPayloadSize is NULL." ) );
+        status = MQTTBadParameter;
+    }
+    else if( pSubackPacket->type != MQTT_PACKET_TYPE_SUBACK )
+    {
+        LogError( ( "Invalid parameter: Input packet is not a SUBACK packet: "
+                    "ExpectedType=%u, InputType=%u", MQTT_PACKET_TYPE_SUBACK, pSubackPacket->type ) );
```
muneebahmed10, line 1981: It'd make more sense to log these in hex
```suggestion
                    "ExpectedType=%02x, InputType=%02x", MQTT_PACKET_TYPE_SUBACK, pSubackPacket->type ) );
```

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -406,6 +406,26 @@ MQTTStatus_t MQTT_ReceiveLoop( MQTTContext_t * pContext,
  */
 uint16_t MQTT_GetPacketId( MQTTContext_t * pContext );
 
+/**
+ * @brief Parses the payload of an MQTT SUBACK packet that contains the status
+ * code of subscription request made for topic filters in the equivalent
+ * subscribe packet.
+ *
+ * This function can be used to iterate over the status code
```
muneebahmed10, line 414: Can it? It doesn't seem so

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -406,6 +406,26 @@ MQTTStatus_t MQTT_ReceiveLoop( MQTTContext_t * pContext,
  */
 uint16_t MQTT_GetPacketId( MQTTContext_t * pContext );
 
+/**
+ * @brief Parses the payload of an MQTT SUBACK packet that contains the status
+ * code of subscription request made for topic filters in the equivalent
+ * subscribe packet.
```
muneebahmed10, line 412: Nit: status code should be plural:
```suggestion
 * @brief Parses the payload of an MQTT SUBACK packet that contains status codes
 * corresponding to topic filter subscription requests from the original
 * subscribe packet.
```

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -406,6 +406,26 @@ MQTTStatus_t MQTT_ReceiveLoop( MQTTContext_t * pContext,
...
+ * @param[in] pSubackPacket The SUBACK packet whose payload is to be parsed.
+ * @param[out] pPayloadStart This is populated with the starting address
+ * of the payload (or return codes for topic filters) in the SUBACK packet.
+ * @param[out] pPayloadSize This is populated with the size of the payload
+ * in the SUBACK packet. It represents the number of topic filters whose
+ * SUBACK status is present in the packet.
+ *
+ * @return #MQTTBadParameter, or #MQTTSuccess.
+ */
+MQTTStatus_t MQTT_GetSubAckPayload( const MQTTPacketInfo_t * pSubackPacket,
```
muneebahmed10, line 425: Nit: Maybe rename to `MQTT_GetSubackStatuses`, `MQTT_GetSubackStatusCodes` or something similar? I think it would be helpful to the customer to be clearer in what the payload actually is.

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -406,6 +406,26 @@ MQTTStatus_t MQTT_ReceiveLoop( MQTTContext_t * pContext,
...
+ *
+ * This function can be used to iterate over the status code
+ *
+ * @param[in] pSubackPacket The SUBACK packet whose payload is to be parsed.
+ * @param[out] pPayloadStart This is populated with the starting address
+ * of the payload (or return codes for topic filters) in the SUBACK packet.
+ * @param[out] pPayloadSize This is populated with the size of the payload
+ * in the SUBACK packet. It represents the number of topic filters whose
+ * SUBACK status is present in the packet.
+ *
```
muneebahmed10, line 422: Maybe we should add a note here about what values each status code can take, and what meaning they have.

libraries/standard/mqtt/src/mqtt_lightweight.c
muneebahmed10: I think the macros are unnecessary, it already seems clear what the values mean, especially due to the log statement that's there.

libraries/standard/mqtt/include/mqtt.h
aggarw13: Oops, this is an artifact of an alternative approach idea I had in mind of the API. Will remove it.

libraries/standard/mqtt/include/mqtt_lightweight.h
aggarw13: As this is a packet level information, I think that this is the appropriate location of the macro.

libraries/standard/mqtt/src/mqtt.c
aggarw13: Updated the log

libraries/standard/mqtt/include/mqtt.h
aggarw13: Have renamed the API to use `StatusCodes` instead of `Payload`. I agree that the name would be helpful for customers. I had originally used payload as I felt that it was appropriate for the lightweight API which represents packet level information for its APIs.

libraries/standard/mqtt/include/mqtt.h
aggarw13: Added

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -116,6 +116,17 @@ struct MQTTPubAckInfo
...
+/**
+ * @brief The status codes in the SUBACK response to a subscription request.
+ */
+typedef enum MQTTSubAckStatus
+{
+    MQTTSubAckSuccessQos0 = 0x00, /**< @brief Success with a maximum delivery at QoS 0 . */
+    MQTTSubAckSuccessQos1 = 0x01, /**< @brief Success with a maximum delivery at QoS 1. */
+    MQTTSubAckSuccessQos2 = 0x02, /**< @brief Success with a maximum delivery at QoS 2. */
+    MQTTSubAckFailure = 0x80      /**< @brief Failure. */
+} MQTTSubAckStatus_t;
```
muneebahmed10, line 128: This still seems unnecessary to me given we already have MQTTQoS_t, and list all status codes in the function documentation. Will disagree and commit since otherwise I don't have issues with this PR

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1963,6 +1963,64 @@ uint16_t MQTT_GetPacketId( MQTTContext_t * pContext )
...
+        status = MQTTBadParameter;
+    }
+    else
+    {
+        /* According to the MQTT 3.1.1 protocol specification, the "Remaining Length" field is a
+         * length of the variable header (2 bytes) plus the length of the payload.
+         * Therefore, we add 2 positions for the starting address of the payload, and
+         * subtract 2 bytes from the remaining length for the length of the payload.*/
+        *pPayloadStart = pSubackPacket->pRemainingData + ( ( uint16_t ) sizeof( uint16_t ) );
+        *pPayloadSize = pSubackPacket->remainingLength - ( ( uint16_t ) sizeof( uint16_t ) );
```
muneebahmed10, line 2016: This is still a MISRA 10.3 violation: 
Implicit conversion of "pSubackPacket->remainingLength - 2UL" from essential type "unsigned 64-bit int" to different or narrower essential type "unsigned 16-bit int".

I think the subtraction might promote it to a 64 bit type, so the cast needs to be applied to the whole expression. Try something like:
```suggestion
        *pPayloadSize = ( uint16_t ) ( pSubackPacket->remainingLength - sizeof( uint16_t ) );
```

libraries/standard/mqtt/include/mqtt.h
```diff
@@ -116,6 +116,17 @@ struct MQTTPubAckInfo
     MQTTPublishState_t publishState; /**< @brief The current state of the publish process. */
 };
 
+/**
+ * @brief The status codes in the SUBACK response to a subscription request.
+ */
+typedef enum MQTTSubAckStatus
```
muneebahmed10, line 122: This results in two 2.3 violations since `MQTTSubAckStatus` and `MQTTSubAckStatus_t` are both defined but never used. Might need to add comments to suppress that.

aggarw13: Will address that in a separate PR. Thanks for pointing out.

PR #1109, opened 2020-08-11
[PR from CLI tool] Fix pointer logging warnings in MQTT library

aggarw13:
Fix build warnings with `-Wall -Wextra` flags for pointer logging in MQTT files.

*Description of changes:*
* Use `(void *)` cast for logging non-function pointers
* Remove logging of function pointers as ISO C does not support `(void *)` cast of function pointers (Only POSIX does)

### DIFF
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -1055,8 +1055,8 @@ static MQTTStatus_t validateSubscribeUnsubscribeParams( const MQTTContext_t * pC
     {
         LogError( ( "Argument cannot be NULL: pContext=%p, "
                     "pSubscriptionList=%p.",
-                    pContext,
-                    pSubscriptionList ) );
+                    ( void * ) pContext,
+                    ( void * ) pSubscriptionList ) );
         status = MQTTBadParameter;
     }
     else if( subscriptionCount == 0UL )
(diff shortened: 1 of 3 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1377,20 +1377,29 @@ MQTTStatus_t MQTT_Init( MQTTContext_t * pContext,
...
-             ( pTransportInterface->recv == NULL ) || ( pTransportInterface->send == NULL ) )
+    else if( getTimeFunction == NULL )
     {
-        LogError( ( "Function pointers cannot be NULL: getTimeFunction=%p, userCallback=%p, "
-                    "transportRecv=%p, transportRecvSend=%p",
-                    getTimeFunction,
-                    userCallback,
-                    pTransportInterface->recv,
-                    pTransportInterface->send ) );
+        LogError( ( "Invalid parameter: getTimeFunction is NULL" ) );
```
muneebahmed10, line 1387: What's the purpose in separating these parameters? It seems more repetitive this way, and it doesn't match the condition earlier in this function, at line 1374

aggarw13: The earlier implementation was printing the function pointers under a single if condition so that the reader could detect which of the function pointers is null. However, casting function pointers to `void *` for printing is not supported by ISO C.
GCC gives the following warning when casting function pointers to `void *`
```
/home/ubuntu/Repos/aws-iot-device-sdk-embedded-C/libraries/standard/mqtt/src/mqtt.c:1387:21: warning: format ‘%p’ expects argument of type ‘void *’, but argument 2 has type ‘MQTTGetCurrentTimeFunc_t {aka unsigned int (*)(void)}’ [-Wformat=]
         LogError( ( "Invalid parameter: getTimeFunction is NULL: %p", getTimeFunction ) );
                     ^
/home/ubuntu/Repos/aws-iot-device-sdk-embedded-C/demos/logging-stack/logging_stack.h:49:40: note: in definition of macro ‘SdkLog’
     #define SdkLog( string )    printf string
                                        ^~~~~~
/home/ubuntu/Repos/aws-iot-device-sdk-embedded-C/libraries/standard/mqtt/src/mqtt.c:1387:9: note: in expansion of macro ‘LogError’
         LogError( ( "Invalid parameter: getTimeFunction is NULL: %p", getTimeFunction ) );
```

PR #1111, opened 2020-08-11
Change return code when cancelling already completed jobs

beckmjef:
*Expected behavior:*
When a keepalive ping does not receive a PINGRESP, the connection is
closed and all memory is cleaned up (if no other active jobs).

*Observed behavior:*
The keepalive job does not get properly cleaned up when this disconnect
occurs, resulting in a memory leak since the connection does not get
cleaned up with the dangling reference.

*Steps to reproduce:*
Add a long delay in the demo before closing the connection. During this
time, remove internet connectivity. This does not reproduce with server
side or client side initiated disconnects, only with a failed PINGRESP.

> I TEST:       Forcing connection disconnect
> D TASKPOOL:   Timer thread started for taskpool 0x7f3b8b5e7940.
> D TASKPOOL:   Scheduling job from timer event.
> D TASKPOOL:   No further timer events to process. Exiting timer thread.
> D MQTT:       Keep-alive job started.
> D MQTT:       Sending PINGREQ.
> D NET:        (Network connection 0x18f1870) Sending 2 bytes.
> D MQTT:       PINGREQ sent. Scheduling check for PINGRESP in 1000 ms.
> D MQTT:       Next keep-alive job in 1000 ms.
> D TASKPOOL:   Timer thread started for taskpool 0x7f3b8b5e7940.
> D TASKPOOL:   Scheduling job from timer event.
> D TASKPOOL:   No further timer events to process. Exiting timer thread.
> D MQTT:       Keep-alive job started.
> D MQTT:       Checking for PINGRESP.
> E MQTT        Failed to receive PINGRESP within 1000 ms.
> W TASKPOOL:   Attempt to cancel a job that is already executing, or canceled.
> I NET:        TLS session terminated.
> I NET:        Connection closed.
> I MQTT:       Network connection closed.
> D MQTT:       Reference count changed from 2 to 1.

The logs are missing the following line:

> D MQTT:       Keep-alive job canceled and cleaned up.

This log is skipped along with the code to clean up the keepalive job
due to the condition where the keepalive job has already completed by
the time the disconnect callbacks have finished occurring. In this case,
the taskpool's attempt to cancel the job fails because it has already
completed, and the code to cleanup the already completed job is not run.

*Description of changes:*

The proposed fix is to change the behavior of the taskpool's TryCancel
routine to return success in the case that a job does not need to be
actually cancelled since it has already run to completion.


beckmjef:
After some digging, the root cause of the memory leak was the dangling reference due to the absence of an IotMqtt_Disconnect.

We are implementing a reconnect feature using the disconnectCallback, and in cases where the callback param, ```(pParam->u.disconnectReason != IOT_MQTT_DISCONNECT_CALLED)```, IotMqtt_Disconnect needs to be called in order to release the reference accumulated by the call to IotMqtt_Connect.

This leads to multiple invocations of _IotMqtt_CloseNetworkConnection(), and multiple calls of the disconnectCallback for the same connection, but the connection is now being destroyed and network buffers free'd successfully.

PR #1114, opened 2020-08-11
[PR from CLI tool] Remove unused enum value from subscription manager

aggarw13:
Remove unused `SUBSCRIPTION_MANAGER_INVALID` enum value (that was intended to be a sentinel value) from the Subscription Manager API.

### DIFF
--- a/demos/mqtt/subscription-manager/mqtt_subscription_manager.c
+++ b/demos/mqtt/subscription-manager/mqtt_subscription_manager.c
@@ -379,7 +379,7 @@ SubscriptionManagerStatus_t SubscriptionManager_RegisterCallback( const char * p
     assert( topicFilterLength != 0 );
     assert( callback != NULL );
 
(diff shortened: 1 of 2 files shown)


demos/mqtt/subscription-manager/mqtt_subscription_manager.c
```diff
@@ -379,7 +379,7 @@ SubscriptionManagerStatus_t SubscriptionManager_RegisterCallback( const char * p
     assert( topicFilterLength != 0 );
     assert( callback != NULL );
 
-    SubscriptionManagerStatus_t returnStatus = SUBSCRIPTION_MANAGER_INVALID;
+    SubscriptionManagerStatus_t returnStatus;
```
aggarw13, line 382: Ideally, the status value should be initialized with a failure value, however, no value in the `SubscriptionManagerStatus_t` represents a general failure.
As the `SUBSCRIPTION_MANAGER_SUCCESS` has a value of **1**, the risk of default initialization causing an false success return value, is mitigated.

PR #1118, opened 2020-08-12
(Attempt to) Fix keep-alive job memory leak on PINGRESP timeout

aggarw13:
### Problem
In the case of a wait timeout of the PINGRESP packet from the server (due to a network disconnection, for example), the keep-alive job's self-clean up attempt fails, and thus, there is a memory leak. 

#### Cause
The `IotTaskPool_TryCancel` function that the keep-alive job calls to cancel itself ( while the job is executing in taskpool ) returns failure due to which the keep-alive job does not free its memory.

#### Fix
Update keep-alive job's clean-up operation to also clean-up its memory if its taskpool reported job status is `IOT_TASKPOOL_STATUS_COMPLETED`

### DIFF
--- a/libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
+++ b/libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
@@ -460,15 +460,15 @@ static size_t _receivePingresp( IotNetworkConnection_t pReceiveContext,
     ( void ) pReceiveContext;
 
     /* Receive of PINGRESP should only ever request 1 byte. */
-    if( bytesRequested == 1 )
-    {
-        /* Write a byte of PINGRESP. */
-        *pBuffer = pPingresp[ receiveIndex ];
-        bytesReceived = 1;
+    // if( bytesRequested == 1 )
+    // {
+    //     /* Write a byte of PINGRESP. */
+    //     *pBuffer = pPingresp[ receiveIndex ];
+    //     bytesReceived = 1;
 
-        /* Alternate the byte of PINGRESP to write. */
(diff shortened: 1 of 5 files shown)


libraries/standard/mqtt/test/unit/iot_tests_mqtt_api.c
```diff
@@ -692,46 +693,46 @@ TEST_TEAR_DOWN( MQTT_Unit_API )
...
-    RUN_TEST_CASE( MQTT_Unit_API, SerializeDisconnectChecks );
-    RUN_TEST_CASE( MQTT_Unit_API, SerializePingReqChecks );
-    RUN_TEST_CASE( MQTT_Unit_API, LightweightConnack );
-    RUN_TEST_CASE( MQTT_Unit_API, LightweightSuback );
-    RUN_TEST_CASE( MQTT_Unit_API, LightweightUnsuback );
-    RUN_TEST_CASE( MQTT_Unit_API, LightweightPingresp );
-    RUN_TEST_CASE( MQTT_Unit_API, LightweightPuback );
-    RUN_TEST_CASE( MQTT_Unit_API, DeserializePublishChecks );
-    RUN_TEST_CASE( MQTT_Unit_API, GetIncomingMQTTPacketTypeAndLengthChecks );
+    // RUN_TEST_CASE( MQTT_Unit_API, Init );
```
sarenameas, line 696: This is supposed to be draft PR, correct?

aggarw13:
Closing the PR as #1111 was resolved by author by making sure that `IotMqtt_Disconnect` is always called (once) per call matching call to `IotMqtt_Connect` for the connection memory to be destroyed

PR #1098, opened 2020-08-05
[PR from CLI tool] Add MQTT demo for connection sharing with subscription manager

aggarw13:
*MQTT demo for using subscription manager when sharing MQTT connection among multiple subscription topics*

﻿*Description of changes:*
* Add demo to showcase use of subscription manager for registering subscription callbacks in a single threaded demo
* Relocate **subscription manager** within the demo based on new decision about providing subscription manager specific references to demos (instead of a generic component utility that applies for all demos)
* Remove all topic matching functions from subscription manager, and instead update it to use the `MQTT_MatchTopic` function (that is added in #1119)

By submitting this pull request, I confirm that my contribution is made under the terms of the MIT license.

PR #1119, opened 2020-08-13
[PR from CLI tool]
 Add utility for MQTT topic filter and topic matching to library

aggarw13:
### Changes
Copy existing `matchTopic` from **subscription manager** to MQTT library. Add unit tests for all cases of topic matching to match MQTT protocol spec.

### Reason
The topic matching utility represents undifferentiated work for customers in meeting the MQTT specification for matching incoming PUBLISH topic name with a subscribed topic filter. The utility can be used in the application-defined MQTT callback for topic matching and routing to subsequent incoming PUBLISH handler. 
**Note:** We will show a use of the `MQTT_MatchTopic` function in the **MQTT connection sharing demo**, #1098


libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -2126,10 +2126,313 @@ void test_MQTT_Ping_error_path( void )
...
+    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_MatchTopic( pTopicName,
+                                                     strlen( pTopicName ),
+                                                     pTopicFilter,
+                                                     strlen( pTopicFilter ),
+                                                     &matchResult ) );
+    TEST_ASSERT_EQUAL( true, matchResult );
+
+    /* Edge case where topic filter starts with '#' and topic name starts with '$'. */
+    pTopicName = "$/test/match";
+    pTopicFilter = "#/test/match";
```
muneebahmed10, line 2391: This isn't a valid filter to begin with, so it could be returning false for reasons besides what we're testing for.
```suggestion
    pTopicFilter = "#";
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1963,6 +2188,71 @@ uint16_t MQTT_GetPacketId( MQTTContext_t * pContext )
...
+    else
+    {
+        /* Check for an exact match if the incoming topic name and the registered
+         * topic filter length match. */
+        if( topicNameLength == topicFilterLength )
+        {
+            *pIsMatch = ( strncmp( pTopicName, pTopicFilter, topicNameLength ) == 0 ) ? true : false;
+        }
+
+        if( status == false )
```
muneebahmed10, line 2231: This is always true since `MQTTSuccess == false == 0`
```suggestion
        if( *pIsMatch == false )
```

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -2126,10 +2126,313 @@ void test_MQTT_Ping_error_path( void )
...
+    pTopicName = "/test/match/";
+    pTopicFilter = "/test/match/+";
+    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_MatchTopic( pTopicName,
+                                                     strlen( pTopicName ),
+                                                     pTopicFilter,
+                                                     strlen( pTopicFilter ),
+                                                     &matchResult ) );
+    TEST_ASSERT_EQUAL( true, matchResult );
+
+    /* Edge case where filter ending with '/+' matches topic ending with ''. */
```
muneebahmed10, line 2279: This comment is a bit confusing and could be reworded
```suggestion
    /* Edge case where filter ending with '/+' should not match a topic ending with ''. */
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -301,6 +301,231 @@ static MQTTStatus_t validatePublishParams( const MQTTContext_t * pContext,
...
+/*-----------------------------------------------------------*/
+
+static bool matchWildcardsSpecialCases( const char * pTopicFilter,
+                                        uint16_t topicFilterLength,
+                                        uint16_t topicNameLength,
+                                        uint16_t nameIndex,
+                                        uint16_t filterIndex )
+{
+    /* The smallest topic filter accepted by this function is "/+". */
+    assert( topicFilterLength > 1 );
```
sarenameas, line 381: Are we no longer asserting on NULL pointers?

muneebahmed10: Also, do we even check this before calling this function? For example, with the function call of a topic name "a/" and topic filter "a":
```
MQTT_MatchTopic( "a/", 2, "a", 1, &match );
```
Unless I'm missing something, this will on line 2246 call 
```
matchTopicFilter( "a/", 2, "a", 1 );
```
which satisfies the conditions on line 480 and 484: 
```
while( ( 0 < 2 ) && ( 0 < 1 ) )
```
and 
```
if( pTopicName[ 0 ] == pTopicFilter[ 0 ] )
```
This will then call `matchWildcardsSpecialCases( "a", 1, 2, 0, 0 );`, triggering the assert since `topicFilterLength` = 1.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -301,6 +301,231 @@ static MQTTStatus_t validatePublishParams( const MQTTContext_t * pContext,
...
+ * @param[in] pTopicFilter The topic filter to check.
+ * @param[in] topicFilterLength Length of topic filter.
+ *
+ * @return `true` if the topic name and topic filter match; `false` otherwise.
+ */
+static bool matchTopicFilter( const char * pTopicName,
+                              uint16_t topicNameLength,
+                              const char * pTopicFilter,
+                              uint16_t topicFilterLength );
+
```
muneebahmed10, line 371: Arguments order of 
* `matchWildcardsSpecialCases`: topic filter/length, topic name length, name index, filter index <- most unintuitive
* `matchWildcards`: topic filter/length, topic name/length, filter index, name index
* `matchTopicFilter`: topic name/length, topic filter/length
* `MQTT_MatchTopic`: topic name/length, topic filter/length

I think some of these should be reordered for consistency

libraries/standard/mqtt/utest/mqtt_utest.c
aggarw13: Good catch

libraries/standard/mqtt/src/mqtt.c
aggarw13: I don't think that all the function signatures need to follow the same order of topic name followed by topic filter.
I have re-ordered the arguments to make them group together for topic filter and topic name (such that name index comes along with topic name where as filter index is grouped with topic filter params)

libraries/standard/mqtt/src/mqtt.c
aggarw13: > Are we no longer asserting on NULL pointers?

Thanks for catching. Have added the NULL-check asserts

aggarw13: > Also, do we even check this before calling this function? For example, with the function call of a topic name "a/" and topic filter "a":
> 
> ```
> MQTT_MatchTopic( "a/", 2, "a", 1, &match );
> ```
> 
> Unless I'm missing something, this will on line 2246 call
> 
> ```
> matchTopicFilter( "a/", 2, "a", 1 );
> ```
> 
> which satisfies the conditions on line 480 and 484:
> 
> ```
> while( ( 0 < 2 ) && ( 0 < 1 ) )
> ```
> 
> and
> 
> ```
> if( pTopicName[ 0 ] == pTopicFilter[ 0 ] )
> ```
> 
> This will then call `matchWildcardsSpecialCases( "a", 1, 2, 0, 0 );`, triggering the assert since `topicFilterLength` = 1.

Have removed the assert check for the topic filter length as there exists an if-check in the logic that covers the "/+" case
`if( ( nameIndex == ( topicNameLength - 1U ) ) && ( filterIndex == ( topicFilterLength - 2U ) ) )`

libraries/standard/mqtt/lexicon.txt
```diff
@@ -48,6 +49,8 @@ endif
 enums
 eventcallbackstub
 expectprocessloopcalls
+filterindex
+filterIndex
```
muneebahmed10, line 53: This one can be removed, no?
```suggestion
```

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -301,6 +301,235 @@ static MQTTStatus_t validatePublishParams( const MQTTContext_t * pContext,
...
+         * the second to last character for the topic filter. */
+        if( ( nameIndex == ( topicNameLength - 1U ) ) &&
+            ( filterIndex == ( topicFilterLength - 2U ) ) )
+        {
+            /* Determine if the topic filter contains "/+" as the last 2 characters.
+             * This covers the special case of topic matching when the topic name
+             * ends with '/' but the topic filter ends with "/+". Thus, for example,
+             * topic filter "sport/+" matches the "sport/" but not "sport". */
+            matchFound = ( ( pTopicFilter[ filterIndex ] == '/' ) &&
+                           ( pTopicFilter[ filterIndex + 1U ] == '+' ) ) ? true : false;
```
muneebahmed10, line 408: Question, what if the filter is "sport/#", and the topic is "sport/"? That should match, right? Is that case handled anywhere? Do we have a unit test for this?

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -2126,10 +2126,313 @@ void test_MQTT_Ping_error_path( void )
...
+    TEST_ASSERT_EQUAL( false, matchResult );
+
+    pTopicName = "test/match/level2";
+    pTopicFilter = "#/match/level2";
+    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_MatchTopic( pTopicName,
+                                                     strlen( pTopicName ),
+                                                     pTopicFilter,
+                                                     strlen( pTopicFilter ),
+                                                     &matchResult ) );
+    TEST_ASSERT_EQUAL( false, matchResult );
```
muneebahmed10, line 2426: Do we want to add some more unit tests for corner cases, e.g.
* topic filter: "test/match/lev#", topic name: "test/match/level1"  -> returns false
* topic filter: "test/match/#", topic name: "test/match/" -> returns true, I think
* topic filter: "test/match/lev+", topic name: "test/match/level1" -> returns false
* topic filter: "+/+", topic name: "/test" -> returns true

Just a suggestion. I'm unsure if all these specific tests need to pass, actually, since in the spec they're all "Non normative comments"

libraries/standard/mqtt/src/mqtt.c
aggarw13: Great catch! I verified that this corner case holds true for `#` wildcard within our demo, and have updated the code and added a unit test for the same.

libraries/standard/mqtt/utest/mqtt_utest.c
aggarw13: The unit tests cover all these cases with the following:
>topic filter: "test/match/lev#", topic name: "test/match/level1"
Covered here: https://github.com/aggarw13/aws-iot-device-sdk-embedded-C/blob/mqtt/add-match-topic-api-and-unit-tests/libraries/standard/mqtt/utest/mqtt_utest.c#L2411
Though your suggestion is a variant which I can add.

>topic filter: "test/match/#", topic name: "test/match/" -> returns true, I think
https://github.com/aggarw13/aws-iot-device-sdk-embedded-C/blob/mqtt/add-match-topic-api-and-unit-tests/libraries/standard/mqtt/utest/mqtt_utest.c#L2370-L2378
(Just added)

>topic filter: "test/match/lev+", topic name: "test/match/level1" 
Covered here: https://github.com/aggarw13/aws-iot-device-sdk-embedded-C/blob/mqtt/add-match-topic-api-and-unit-tests/libraries/standard/mqtt/utest/mqtt_utest.c#L2299-L2307
Your suggestion is a variant of this, which I can add.

>topic filter: "+/+", topic name: "/test"
Covered here: https://github.com/aggarw13/aws-iot-device-sdk-embedded-C/blob/mqtt/add-match-topic-api-and-unit-tests/libraries/standard/mqtt/utest/mqtt_utest.c#L2260-L2267

muneebahmed10: Ah I didn't see number 4. Please add 1 and 3 then, I want to make sure the tests still pass when the end of the topic name has not been reached.

aggarw13: Edit, after adding the variants for cases 1 and cases 3, another bug was found in the topic matching algorithm. Have updated it. Thanks for another suggestion!!

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -301,6 +301,245 @@ static MQTTStatus_t validatePublishParams( const MQTTContext_t * pContext,
...
+        ( *pNameIndex )--;
+    }
+
+    /* Check for '#' wildcard character and that the topic filter has the valid
+    * format for the '#' character.
+    * Note: '#' should be the last character in a topic filter. If the
+    * topic filter has multiple level, "/#" should be the last 2 characters. */
+    else if( ( pTopicFilter[ filterIndex ] == '#' ) &&
+             ( filterIndex == ( topicFilterLength - 1U ) ) && /* '#' should be the last character. */
+             ( ( filterIndex == 0u ) || ( pTopicFilter[ filterIndex - 1U ] == '/' ) ) )
```
muneebahmed10, line 461: This looks kind of excessive. Some intermediate boolean variables might be more helpful than comments here in keeping everything readable. If we're validating the filter the same time as we're matching, would a `isValidFilter()` function be helpful to call before we start matching?

Some suggestions to improve hygiene, don't necessarily have to be in this PR. I'm going to approve the PR with the assumption you've run the coverage tests and verified all these lines are covered

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -313,6 +313,247 @@ static MQTTStatus_t validatePublishParams( const MQTTContext_t * pContext,
...
+ *
+ * @param[in] pTopicFilter The topic filter containing the wildcard.
+ * @param[in] topicFilterLength Length of the topic filter being examined.
+ * @param[in] filterIndex Index of the topic filter being examined.
+ * @param[in] topicNameLength Length of the topic name being examined.
+ * @param[in] nameIndex Index of the topic name being examined.
+ *
+ * @return Returns whether the topic filter and the topic name match.
+ */
+static bool matchWildcardsSpecialCases( const char * pTopicFilter,
```
muneebahmed10, line 335: Iirc, in v4_beta wasn't this called `matchEndWildcards`? Why was it changed? Imo that was a more descriptive name since it immediately tells me that it checks for wildcards at the end of the filter.

aggarw13: The `matchEndWildcards` can be misleading as it may suggest that it covers _all_ cases of topic filters ending with wildcards. This function only covers some special cases, whereas the  `matchWildcards` function covers the _nominal_ cases of wildcards  (which includes wildcards at the end of the topic filter)

muneebahmed10: `matchWildcardsSpecialCases` gives less information as to what the function is doing that `matchEndWidcards`. The name could be improved imo. `matchEndWildcardsSpecialCases`?

aggarw13: Sure, will update

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -2126,10 +2126,396 @@ void test_MQTT_Ping_error_path( void )
...
+    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_MatchTopic( pTopicName,
+                                                     strlen( pTopicName ),
+                                                     pTopicFilter,
+                                                     strlen( pTopicFilter ),
+                                                     &matchResult ) );
+    TEST_ASSERT_EQUAL( true, matchResult );
+
+    pTopicName = "/test///level1";
+    pTopicFilter = "/test/+/+/level1";
+    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_MatchTopic( pTopicName,
```
muneebahmed10, line 2296: There's already a test for consecutive separators on line 2268. It'd make sense to delete one

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -2126,10 +2126,396 @@ void test_MQTT_Ping_error_path( void )
...
+    /* Edge case where topic filter starts with '#' and topic name starts with '$'. */
+    pTopicName = "$/test/match";
+    pTopicFilter = "#";
+    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_MatchTopic( pTopicName,
+                                                     strlen( pTopicName ),
+                                                     pTopicFilter,
+                                                     strlen( pTopicFilter ),
+                                                     &matchResult ) );
+    TEST_ASSERT_EQUAL( false, matchResult );
+
```
muneebahmed10, line 2462: The below five tests all test for invalid filters. Maybe separate them into another function so that this one doesn't grow too large?

libraries/standard/mqtt/utest/mqtt_utest.c
aggarw13: This is slightly a different case where it checks that multiple empty levels can be matched with the topic filter

libraries/standard/mqtt/utest/mqtt_utest.c
aggarw13: Separated

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -313,6 +313,253 @@ static MQTTStatus_t validatePublishParams( const MQTTContext_t * pContext,
...
+ *
+ * @note This function ASSUMES that the topic name been consumed in linear
+ * matching with the topic filer, but the topic filter has remaining characters
+ * to be matched.
+ *
+ * @param[in] pTopicFilter The topic filter containing the wildcard.
+ * @param[in] topicFilterLength Length of the topic filter being examined.
+ * @param[in] filterIndex Index of the topic filter being examined.
+ * @param[in] topicNameLength Length of the topic name being examined.
+ * @param[in] nameIndex Index of the topic name being examined.
```
yourslab, line 331: Extra params in the docstrings

PR #1120, opened 2020-08-13
Add cbmc pointer primitive check flag and mqtt_lightweight.c small refactor

sarenameas:
*Description of changes:*
- This pull request updates the proof template submodule to include checks missing from continuous integration, and added a flag --pointer-primitive-check to the proofs to check the soundness of assumptions by checking that all pointers in pointer primitives are valid or null.

- Update validateSerializeSubscribeParams(), MQTT_SerializePublish(), and MQTT_SerializePublishHeader() to check the packet size against the fixed buffer AFTER the NULL parameter checks. When a NULL subscribeInfo or publishInfo, is passed into these functions, we don't care about the remaining length, so the remaining length should NOT immediately be used in a possible overflow causing arithmetic operation. This helps with the CBMC proof that passes in a NULL parameter for coverage, but an unbounded remaining length.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -1874,7 +1874,6 @@ MQTTStatus_t MQTT_SerializePublishHeader( const MQTTPublishInfo_t * pPublishInfo
          *                               + Length of encoded remaining length
          *                               + Remaining length
          *                               - Payload Length.
-         * Payload length will be subtracted after verifying pPublishInfo parameter.
          */
         packetSize = 1U + remainingLengthEncodedSize( remainingLength )
                      + remainingLength
--- a/libraries/standard/mqtt/src/mqtt_lightweight.c
+++ b/libraries/standard/mqtt/src/mqtt_lightweight.c
@@ -1030,12 +1030,7 @@ static MQTTStatus_t validateSubscriptionSerializeParams( const MQTTSubscribeInfo
                                                          const MQTTFixedBuffer_t * pFixedBuffer )
 {
     MQTTStatus_t status = MQTTSuccess;
-
-    /* The serialized packet size = First byte
-     *  + length of encoded size of remaining length
-     *  + remaining length. */
-    size_t packetSize = 1U + remainingLengthEncodedSize( remainingLength )
-                        + remainingLength;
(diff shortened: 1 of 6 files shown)


libraries/standard/mqtt/src/mqtt_lightweight.c
```diff
@@ -1796,15 +1791,25 @@ MQTTStatus_t MQTT_SerializePublish( const MQTTPublishInfo_t * pPublishInfo,
...
         status = MQTTBadParameter;
     }
-    else if( packetSize > pFixedBuffer->size )
+    else
+    {
+        /* Length of serialized packet = First byte
+         *                                + Length of encoded remaining length
+         *                                + Remaining length. */
+        packetSize = 1U + remainingLengthEncodedSize( remainingLength )
+                     + remainingLength;
```
leegeth, line 1800: Just a consistency question, can this be similar to line 1062, for consistency reasons?

sarenameas: I had to move this to the else and check the success and size separately because the complexity score.

sarenameas: Oh let me try to check the size in here, then serializePublishCommon outside.

sarenameas: Tried:
```
    else
    {
        /* Length of serialized packet = First byte
         *                                + Length of encoded remaining length
         *                                + Remaining length. */
        packetSize = 1U + remainingLengthEncodedSize( remainingLength )
                     + remainingLength;

        if( packetSize > pFixedBuffer->size )
        {
            LogError( ( "Buffer size of %lu is not sufficient to hold "
                        "serialized PUBLISH packet of size of %lu.",
                        ( unsigned long ) pFixedBuffer->size,
                        ( unsigned long ) packetSize ) );
            status = MQTTNoMemory;
        }
    }

    if( status == MQTTSuccess )
    {
        /* Serialize publish with header and payload. */
        serializePublishCommon( pPublishInfo,
                                remainingLength,
                                packetId,
                                pFixedBuffer,
                                true );
    }
```
And the compleixty reached 9.

leegeth: Oh ok. Thanks for explaining this.

PR #1126, opened 2020-08-13
Download certs into demos/certificates with an option to disable downloading

yourslab:
Once the certs are downloaded, they are copied from `demos/certificates` into `build/bin/certificates`. Integration tests also need these certificates, so they are copied to  `build/bin/tests/certificates` whenever tests are built. 

You can also disable the downloading of certificates by passing the CMake flag: `-DDOWNLOAD_CERTS=0`


demos/CMakeLists.txt
```diff
@@ -35,11 +35,45 @@ function(check_aws_credentials demo_name)
...
+    execute_process(
+        COMMAND curl --url https://www.amazontrust.com/repository/AmazonRootCA1.pem
+        -o ${CMAKE_CURRENT_LIST_DIR}/certificates/AmazonRootCA1.crt
+    )
+    # Download the Mosquitto Root CA certificate.
+    message( "Downloading the Mosquitto Root CA certificate..." )
+    execute_process(
+        COMMAND curl --url https://test.mosquitto.org/ssl/mosquitto.org.crt
+        -o ${CMAKE_CURRENT_LIST_DIR}/certificates/mosquitto.org.crt
+    )
```
muneebahmed10, line 55: If our plan is to not use test.mosquitto.org by default any more, we might as well just remove this now.

demos/CMakeLists.txt
```diff
@@ -35,11 +35,39 @@ function(check_aws_credentials demo_name)
     endif()
 endfunction()
 
+function(copy_certs_into_target_dir demo_target)
```
aggarw13, line 38: Is this an empty function?

demos/CMakeLists.txt
```diff
@@ -35,11 +35,39 @@ function(check_aws_credentials demo_name)
...
+            "${CMAKE_BINARY_DIR}/bin/tests"
+    )
+endif()
+
 # Include each subdirectory that has a CMakeLists.txt file in it
 file(GLOB demo_dirs "${DEMOS_DIR}/*/*")
 foreach(demo_dir IN LISTS demo_dirs)
     if(IS_DIRECTORY "${demo_dir}" AND EXISTS "${demo_dir}/CMakeLists.txt")
-        get_filename_component( DEMO_EXE_NAME ${demo_dir} NAME )
+        get_filename_component( DEMO_TARGET_NAME ${demo_dir} NAME )
```
aggarw13, line 70: What is the purpose of this change?

demos/mqtt/mqtt_demo_mutual_auth/CMakeLists.txt
```diff
@@ -34,21 +34,3 @@ target_include_directories(
     PUBLIC
         ${LOGGING_INCLUDE_DIRS}
 )
```
aggarw13, line 36: It may be value to set the absolute path for the `ROOT_CA_CERT_PATH` macro when building the mutual auth demo so that there isn't a relative path dependency which affects the directory from which the `mqtt_demo_mutual_auth` binary is run.

(The macro could be renamed to `AMAZON_ROOT_CA_PATH` instead to be more expressive, and also, so that the generic `ROOT_CA_CERT` macros in other demos are not affected. )

```
# Set the AMAZON_ROOT_CA_CERT_PATH configuration for the demo to the absolute path of the 	
# copied Amazon Root CA certificate.	
target_compile_definitions(	
    ${DEMO_NAME} 	
    PRIVATE 	
    AMAZON_ROOT_CA_CERT_PATH="$<TARGET_FILE_DIR:${DEMO_NAME}>/certificates/AmazonRootCA1.crt" )
```

demos/CMakeLists.txt
yourslab: I just thought it was a better name, but this line is actually not needed at all, so I removed it.

demos/CMakeLists.txt
yourslab: Removed

demos/mqtt/mqtt_demo_mutual_auth/CMakeLists.txt
yourslab: We discussed offline that we should have the path to the certificate passed as a CMake flag instead.

demos/mqtt/mqtt_demo_mutual_auth/CMakeLists.txt
```diff
@@ -35,20 +35,21 @@ target_include_directories(
...
-        ${DEMO_NAME}
-    POST_BUILD
-        COMMAND ${CMAKE_COMMAND} -E copy_directory
-        "${CMAKE_CURRENT_LIST_DIR}/certificates"
-        "$<TARGET_FILE_DIR:${DEMO_NAME}>/certificates"
-)
+if(ROOT_CA_CERT_PATH)
+    target_compile_definitions(
+        ${DEMO_NAME} PRIVATE
+            ROOT_CA_CERT_PATH="${ROOT_CA_CERT_PATH}"
```
muneebahmed10, line 41: Even if the file path is passed as a cmake flag, will it still work properly from any directory if the given flag is a relative path and not an absolute path? If not, is it possible to parse this flag using `get_filename_component` and convert it into an absolute path first?

yourslab: Good idea!

demos/CMakeLists.txt
```diff
@@ -35,11 +35,41 @@ function(check_aws_credentials demo_name)
...
+file(COPY "${CMAKE_CURRENT_LIST_DIR}/certificates"
+     DESTINATION "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
+if(BUILD_TESTS)
+    file(COPY "${CMAKE_CURRENT_LIST_DIR}/certificates"
+         DESTINATION "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/tests")
+endif()
+
+# Set prefix to PWD if any path flags are relative
+if(DEFINED ENV{PWD})
+    if(NOT ROOT_CA_CERT_PATH MATCHES "/$")
```
aggarw13, line 58: What if these variables are not passed in the CMake command in which case they would (probably) have empty values? 
Would these generate invalid variables for the credential variables in that case?

aggarw13: I tested the `mqtt_demo_basic_tls` locally with not providing the `ROOT_CA_CERT_PATH` with the following command:
`cmake .. -DBUILD_TESTS="ON" -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS='-Wextra -Wall -O0 -ggdb'`

The build logs show that the absolute path logic is causing the `ROOT_CA_CERT_PATH` CMake variable to be defined even those it wasn't passed by me, and thus, it is building the demo target with an incorrect value (of `/home/ubuntu/Repos/aws-iot-device-sdk-embedded-C/build/` as can be seen in the logs)
```
╭─ubuntu@ip-172-31-24-209 ~/Repos/aws-iot-device-sdk-embedded-C/build  ‹development*› 
╰─➤  ./bin/mqtt_demo_basic_tls         
[INFO] [DEMO] [mqtt_demo_basic_tls.c:400] Establishing a TLS session to test.mosquitto.org:8883.
[DEBUG] [Sockets] [sockets_posix.c:170] Performing DNS lookup: Host=test.mosquitto.org.
[DEBUG] [Sockets] [sockets_posix.c:210] Attempting to connect to server: Host=test.mosquitto.org, IP address=5.196.95.208.
[DEBUG] [Sockets] [sockets_posix.c:225] Connected to IP address: 5.196.95.208.
[DEBUG] [Sockets] [sockets_posix.c:256] Established TCP connection: Server=test.mosquitto.org.

[DEBUG] [Transport_OpenSSL_Sockets] [openssl_posix.c:144] Attempting to open Root CA certificate: Path=/home/ubuntu/Repos/aws-iot-device-sdk-embedded-C/build/.
[ERROR] [Transport_OpenSSL_Sockets] [openssl_posix.c:195] PEM_read_X509 failed to parse root CA.
[ERROR] [Transport_OpenSSL_Sockets] [openssl_posix.c:474] Setting up credentials failed.
```

aggarw13: The fix is simple by just updating each of the nested if conditions to check if their respective CMake variables are defined

PR #1136, opened 2020-08-18
Hygiene changes in demo configuration

aggarw13:
Hygiene changes as follow-up of #1128
* Remove `localhost` default definition of `BROKER_ENDPOINT` to avoid developer inconvencience of understanding demo failure when they don't have local Mosquitto broker setup
* Supporting absolute path pre-pending to credential files for both demos and integration tests


demos/mqtt/mqtt_demo_basic_tls/demo_config.h
```diff
@@ -76,11 +75,6 @@
  * #define ROOT_CA_CERT_PATH         ".....insert here...."
  */
 
-/**
- * @brief Length of path to server certificate.
- */
-#define ROOT_CA_CERT_PATH_LENGTH    ( ( uint16_t ) ( sizeof( ROOT_CA_CERT_PATH ) - 1 ) )
```
muneebahmed10, line 82: I assume this is an unused macro

aggarw13: Yes, it is unused

PR #1137, opened 2020-08-19
Proof for MQTT_MatchTopic API

aggarw13:
* Add proof for `MQTT_MatchTopic` API function
* Fix issues in `matchWildcards` and `matchTopicFilter` functions (called by the API function) that are identified with proof


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -396,7 +401,8 @@ static bool matchEndWildcardsSpecialCases( const char * pTopicFilter,
      * "/#". This check handles the case to match filter "sport/#" with topic
      * "sport". The reason is that the '#' wildcard represents the parent and
      * any number of child levels in the topic name.*/
-    if( ( filterIndex == ( topicFilterLength - 3U ) ) &&
+    if( ( topicFilterLength >= 3U ) &&
```
aggarw13, line 404: Added `( topicFilterLength >= 3U )` condition to address overflow error flagged by CBMC proof for `( filterIndex == ( topicFilterLength - 3U ) )` condition

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -428,7 +434,7 @@ static bool matchWildcards( const char * pTopicName,
                             const char * pTopicFilter,
                             uint16_t topicFilterLength,
                             uint16_t * pNameIndex,
-                            uint16_t filterIndex,
+                            uint16_t * pFilterIndex,
```
aggarw13, line 437: Updated `filterIndex` to be an output parameter to be an output parameter to address corner case of matching first level between topic filter and topic name (when topic filter starts with `+`  and topic name has empty first level as  `/`) that resulted in underflow when calling `( *pNameIndex ) --` when input `*pNameIndex` is 0. 

By making `filterIndex` an output parameter, instead of decrementing the `*pNameIndex` value, we increment the `filterIndex` value and thereby proceed the indices of both topic name and topic filter in matching to avoid the underflow problem of the corner case

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -439,56 +445,87 @@ static bool matchWildcards( const char * pTopicName,
...
+
+        /* If the topic name and topic filter have child levels, then advance the
+         * filter index to the level separator in the topic filter, so that match
+         * can be performed in the next level.
+         * Note: The name index already points to the level separator in the topic
+         * name. */
+        else if( ( nextLevelExistsInTopicName == true ) &&
+                 ( nextLevelExistsinTopicFilter == true ) )
+        {
+            ( *pFilterIndex )++;
```
aggarw13, line 502: As mentioned in https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1137/files#r472797473, the filter index is incremented for the case when both topic name and topic filter have child levels by the presence of the level-separator, `/`, and thus, the underflow problem for the special case is avoided.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -439,56 +445,87 @@ static bool matchWildcards( const char * pTopicName,
...
+        }
+        else
+        {
+            /* If we have reached here, the the loop terminated on the
+             * ( *pNameIndex < topicNameLength) condition, which means that have
+             * reached past the end of the topic name, and thus, we decrement the
+             * index to the last character in the topic name.*/
+            ( *pNameIndex )--;
```
aggarw13, line 510: The decrement of `*pNameIndex` has been reduced to only the case when it has reached the end of the topic name. (In the earlier revision, this was decremented also for the case when the topic name contained a child level beyond the current level. The decrement for a special scenario in that case led to underflow)

muneebahmed10: Are the changes to this file necessary? 

`pNameIndex` is a pointer to a `uint16_t`, so per [6.2.5.9](http://www.open-std.org/JTC1/sc22/wg14/www/docs/n1256.pdf) of the C standard it can't under/overflow; computation regarding unsigned types is well defined, and for a uint16_t, is reduced modulo UINT16_MAX+1. While this decrement would turn a 0 to UINT16_MAX, we increment `nameIndex` back to 0 before the end of the `while` loop in `matchTopicFilter`. 

So, it seems to me like this case is already handled, and extra code doesn't seem necessary if a comment would be sufficient.

aggarw13: > pNameIndex is a pointer to a uint16_t, so per 6.2.5.9 of the C standard it can't under/overflow; computation regarding unsigned types is well defined, and for a uint16_t, is reduced modulo UINT16_MAX+1. While this decrement would turn a 0 to UINT16_MAX, we increment nameIndex back to 0 before the end of the while loop in matchTopicFilter.

CBMC flags underflow error for decrementing 0 to 65535 in `matchWildcards` and increment 65535 to 0 in `matchTopicFilter` as overflow which is a completely fair expectation of a function to not have such side-effects on its values. The functions are independently held accountable for their quality, it just so happens that the underflow in `matchWildcards` happens to be countered by the overflow in `matchTopicFilter`, and thus, the over logic works fine. However, we shouldn't depend on such a behavior for correct functionality of the topic matching algorithm

libraries/standard/mqtt/lexicon.txt
```diff
@@ -176,18 +176,17 @@ payloadlength
...
 pcontext
 pcurrentstate
 pcursor
 pdeserialized
 pdeserializedinfo
 pdestination
-ppacketidentifier
-ppingresp
 pem
+pfilterindex
```
nateglims, line 189: This has trailing whitespace.

aggarw13: Fixed

libraries/standard/mqtt/src/mqtt.c
aggarw13: Added.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -318,7 +318,11 @@ static MQTTStatus_t validatePublishParams( const MQTTContext_t * pContext,
                                            uint16_t packetId );
 
 /**
- * topic filter, this function handles the following 2 cases:
+ * @brief Performs matching for special cases of topic filter ending with
+ * wildcard character.
```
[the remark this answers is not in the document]
aggarw13: Have modified the statement slightly and added articles.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -439,56 +445,87 @@ static bool matchWildcards( const char * pTopicName,
...
+            ( *pNameIndex )++;
         }
 
-        /* '#' matches everything remaining in the topic name. It must be the
-         * last character in a topic filter. */
-        else if( ( pTopicFilter[ filterIndex ] == '#' ) &&
-                 ( filterIndex == ( topicFilterLength - 1U ) ) )
+        /* Determine if the topic filter contains a child level after the current level
+         * represented by the '+' wildcard. */
+        if( ( *pFilterIndex < ( topicFilterLength - 1U ) ) &&
```
sarenameas, line 479: If **pTopicFilter** = ['+'] and **topicFilterLength** = 1 and **\*pFilterIndex** = 0
Then pTopicFilter[ *pFilterIndex + 1U ] is an out of bounds reference if pTopicFilter is not NULL terminated.

sarenameas: Oh nevermind sorry I see that this is < and not <= ignore my comment.

libraries/standard/mqtt/cbmc/proofs/MQTT_MatchTopic/Makefile
```diff
@@ -0,0 +1,43 @@
...
+HARNESS_FILE=MQTT_MatchTopic_harness
+
+# The topic name/filter length are bounded, so that the loops in topic matching algorithmic
+# functions called by MQTT_MatchTopic can be unwounded to an expected
+# amount that won't make the proof run too long.
+MAX_TOPIC_NAME_FILTER_LENGTH=10
+
+DEFINES += -DMAX_TOPIC_NAME_FILTER_LENGTH=$(MAX_TOPIC_NAME_FILTER_LENGTH)
+INCLUDES +=
+INCLUDES +=
```
markrtuttle, line 32: Delete duplicate line?

libraries/standard/mqtt/cbmc/proofs/MQTT_MatchTopic/Makefile
```diff
@@ -0,0 +1,43 @@
...
+
+REMOVE_FUNCTION_BODY +=
+UNWINDSET += __CPROVER_file_local_mqtt_c_matchTopicFilter.0:$(MAX_TOPIC_NAME_FILTER_LENGTH)
+UNWINDSET += strncmp.0:$(MAX_TOPIC_NAME_FILTER_LENGTH)
+UNWINDSET += __CPROVER_file_local_mqtt_c_matchWildcards.0:$(MAX_TOPIC_NAME_FILTER_LENGTH)
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+PROOF_SOURCES += $(SRCDIR)/libraries/standard/mqtt/cbmc/sources/mqtt_cbmc_state.c
```
markrtuttle, line 40: Shouldn't this be PROJECT_SOURCES?

aggarw13: All Makefiles include this source have added it to the `PROOF_SOURCES` list
Example: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/mqtt/cbmc/proofs/MQTT_SerializeConnect/Makefile#L39

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: If this is a CBMC issue then I understand why it was changed; I had commented because I don't think we need to over engineer what already works as expected, since doing so can make the code harder to read, and can introduce mistakes, e.g. for MISRA.

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -439,56 +445,87 @@ static bool matchWildcards( const char * pTopicName,
...
             shouldStopMatching = true;
         }
+
+        /* If the topic name and topic filter have child levels, then advance the
+         * filter index to the level separator in the topic filter, so that match
+         * can be performed in the next level.
+         * Note: The name index already points to the level separator in the topic
+         * name. */
+        else if( ( nextLevelExistsInTopicName == true ) &&
+                 ( nextLevelExistsinTopicFilter == true ) )
```
muneebahmed10, line 500: Coverity flags this as a MISRA 14.3 violation, since it thinks `nextLevelExistsinTopicFilter` must equal `true`. Can you take a look and fix it?

aggarw13: Good catch, have fixed it.

aggarw13: Good catch, have fixed it.

libraries/standard/mqtt/cbmc/proofs/MQTT_MatchTopic/Makefile
markrtuttle: You are correct.

PR #1138, opened 2020-08-19
Reduce GNU complexity score of posix sockets wrapper

leegeth:
*Description of changes:*
Reduce GNU complexity score of posix sockets wrapper. 
The complexity score for the attemptConnection is reduced from 13 to 7 in this change.


platform/posix/transport/src/sockets_posix.c
```diff
@@ -143,29 +160,90 @@ static SocketStatus_t resolveHostName( const char * pHostName,
...
 
     assert( pListHead != NULL );
     assert( pHostName != NULL );
     assert( hostNameLength > 0 );
     assert( pTcpSocket != NULL );
 
-    netPort = htons( port );
-
-    LogDebug( ( "Performing DNS lookup: Host=%.*s.",
+    LogDebug( ( "Attempting to connect to: Host=%.*s.",
```
yourslab, line 246: This log seems to almost be the same as the one in connectToAddress. I believe it can just be removed.

leegeth: I think this debug log still adds value as this is the only log which will be printed out before hitting any of the error cases. In ConnectToAddress, it requires a valid socket to be created to get to the debug log. I think this log is still a good addition.  Log level is debug as well.

yourslab: Okay but is there any way we can distinguish both log messages further?
Maybe have one be `Attempting to connect to server` while the other can be `Attempting to connect to resolved IP from server`.

leegeth: Yea. Good suggestion. Will update the log in connectToSrever.

platform/posix/transport/src/sockets_posix.c
```diff
@@ -98,6 +98,23 @@ static SocketStatus_t attemptConnection( struct addrinfo * pListHead,
                                          int * pTcpSocket,
                                          int32_t maxAttempts );
 
+/**
+ * @brief Connect to server using the provided address record.
+ *
+ * @param[in] pAddrInfo Address record of the server.
```
aggarw13, line 104: `pAddrInfo` is also an output parameter, right?
```suggestion
 * @param[in, out] pAddrInfo Address record of the server.
```

platform/posix/transport/src/sockets_posix.c
```diff
@@ -143,29 +160,91 @@ static SocketStatus_t resolveHostName( const char * pHostName,
 }
 /*-----------------------------------------------------------*/
 
+static SocketStatus_t connectToAddress( struct sockaddr * pAddrInfo,
+                                        const char * pHostName,
+                                        size_t hostNameLength,
+                                        uint16_t port,
+                                        int tcpSocket )
+{
+    SocketStatus_t returnStatus = SOCKETS_SUCCESS;
```
aggarw13, line 169: It seems like pHostName is only passed for logging. I suggest that the hostname be logged in the calling code only instead to keep the API signature relevant only to its functionality

platform/posix/transport/src/sockets_posix.c
```diff
@@ -143,29 +160,91 @@ static SocketStatus_t resolveHostName( const char * pHostName,
...
+    uint16_t netPort = 0;
+
+    assert( pAddrInfo != NULL );
+    assert( pHostName != NULL );
+    assert( hostNameLength > 0 );
+
+    /* Convert port from host byte order to network byte order. */
+    netPort = htons( port );
+
+    if( pAddrInfo->sa_family == AF_INET )
```
dan4thewin, line 182: Since the return values of inet_ntop() are ignored, seems like we need an assert on sa_family.

platform/posix/transport/src/sockets_posix.c
```diff
@@ -143,29 +160,91 @@ static SocketStatus_t resolveHostName( const char * pHostName,
...
+                                        uint16_t port,
+                                        int tcpSocket )
+{
+    SocketStatus_t returnStatus = SOCKETS_SUCCESS;
+    int connectStatus = 0;
+    char resolvedIpAddr[ INET6_ADDRSTRLEN ];
+    socklen_t addrInfoLength;
+    uint16_t netPort = 0;
+
+    assert( pAddrInfo != NULL );
```
aggarw13, line 175: Assert for `tcpSocket`?

platform/posix/transport/src/sockets_posix.c
```diff
@@ -143,29 +160,91 @@ static SocketStatus_t resolveHostName( const char * pHostName,
...
+        LogWarn( ( "Failed to connect to server: Host=%.*s, IP address=%s.",
+                   ( int32_t ) hostNameLength,
+                   pHostName,
+                   resolvedIpAddr ) );
+        close( tcpSocket );
+        returnStatus = SOCKETS_CONNECT_FAILURE;
+    }
+    else
+    {
+        LogDebug( ( "Connected to IP address: %s.",
```
aggarw13, line 223:
```suggestion
        LogDebug( ( "Connected to server: Address=%s.",
```

platform/posix/transport/src/sockets_posix.c
```diff
@@ -236,25 +279,21 @@ static SocketStatus_t attemptConnection( struct addrinfo * pListHead,
...
             break;
         }
     }
 
-    if( pIndex == NULL )
+    if( returnStatus == SOCKETS_SUCCESS )
     {
-        /* Fail if no connection could be established. */
-        LogError( ( "Could not connect to any resolved IP address from %.*s.",
+        LogDebug( ( "Established TCP connection: Server=%.*s.\n",
```
dan4thewin, line 288: Is this message repeating the one in connectToAddress()?  Do we need both?

platform/posix/transport/src/sockets_posix.c
```diff
@@ -181,48 +260,12 @@ static SocketStatus_t attemptConnection( struct addrinfo * pListHead,
...
-                                sizeof( resolvedIpAddr ) );
-        }
-
-        LogDebug( ( "Attempting to connect to server: Host=%.*s, IP address=%s.",
-                    ( int32_t ) hostNameLength,
-                    pHostName,
-                    resolvedIpAddr ) );
-
-        connectStatus = connect( *pTcpSocket, pAddrInfo, addrInfoLength );
+        /* Attempt to connect to a resolved DNS address of the host. */
```
aggarw13, line 263:
```suggestion
        /* Attempt to connect to the resolved DNS address of the host. */
```

platform/posix/transport/src/sockets_posix.c
```diff
@@ -181,48 +260,12 @@ static SocketStatus_t attemptConnection( struct addrinfo * pListHead,
...
-                       pHostName,
-                       resolvedIpAddr ) );
-            close( *pTcpSocket );
-        }
-        else
-        {
-            LogDebug( ( "Connected to IP address: %s.",
-                        resolvedIpAddr ) );
             break;
         }
```
aggarw13, line 271: (Unrelated to the PR) Shouldn't these retry attempts use exponential back-off (with the `transport_reconnect` utility) ?

aggarw13: Nevermind, I see the attempts are across different DNS addresses here instead of re-attempts to the same address

platform/posix/transport/src/sockets_posix.c
```diff
@@ -143,29 +160,91 @@ static SocketStatus_t resolveHostName( const char * pHostName,
...
-    socklen_t addrInfoLength;
-    uint16_t netPort = 0;
-    int curAttempts = 0, connectStatus = 0;
-    char resolvedIpAddr[ INET6_ADDRSTRLEN ];
+    int curAttempts = 0;
 
     assert( pListHead != NULL );
     assert( pHostName != NULL );
     assert( hostNameLength > 0 );
     assert( pTcpSocket != NULL );
```
aggarw13, line 245: `assert( maxAttempts != 0 );`

aggarw13: `maxAttempts` should probably not be of signed type

platform/posix/transport/src/sockets_posix.c
leegeth: No. It is an input parameter

platform/posix/transport/src/sockets_posix.c
leegeth: Agree. Thanks

PR #1139, opened 2020-08-19
Add JSON Unit Tests

yourslab:
These unit tests give us full branch coverage over the JSON library. 


libraries/standard/json/src/json.c
```diff
@@ -366,7 +365,9 @@ static bool_ skipEscape( const char * buf,
     bool_ ret = false;
     size_t i = *start;
 
-    if( ( ( i + 1U ) < max ) && ( buf[ i ] == '\\' ) )
+    assert( buf[ i ] == '\\' );
```
yourslab, line 368: Moved the check to an assert @dan4thewin

libraries/standard/json/src/json.c
```diff
@@ -217,18 +217,17 @@ static bool_ skipUTF8( const char * buf,
 {
     bool_ ret = false;
 
-    if( *start < max )
+    assert( *start < max );
```
yourslab, line 220: Moved the check to an assert @dan4thewin

libraries/CMakeLists.txt
```diff
@@ -35,11 +35,11 @@ if( ${BUILD_TESTS} )
     # Add a target for running coverage on tests.
     add_custom_target(coverage
         COMMAND ${CMAKE_COMMAND} -P ${CMAKE_SOURCE_DIR}/tools/cmock/coverage.cmake
-        DEPENDS cmock unity http_utest mqtt_utest mqtt_lightweight_utest mqtt_state_utest http_send_utest
+        DEPENDS cmock unity http_utest mqtt_utest mqtt_lightweight_utest mqtt_state_utest http_send_utest json_utest
```
aggarw13, line 38: As a long term strategy, it would be good to probably add all these unit tests target within a list so that the coverage dependency is automatically added

libraries/standard/json/src/json.c
```diff
@@ -366,7 +365,9 @@ static bool_ skipEscape( const char * buf,
     bool_ ret = false;
     size_t i = *start;
```
aggarw13, line 367: `assert( *start < max )`?

aggarw13: Asserts for NULL check of `buf`?

libraries/standard/json/src/json.c
aggarw13: Why does this function require that the starting index be `\\`? From the documentation of the function, it wasn't obvious..

libraries/standard/json/src/json.c
```diff
@@ -836,6 +839,7 @@ static void skipObjectScalars( const char * buf,
  * @param[in] buf  The buffer to parse.
  * @param[in,out] start  The index at which to begin.
  * @param[in] max  The size of the buffer.
+ * @param[in] mode  The start marker of an array or object.
```
aggarw13, line 842:
```suggestion
 * @param[in] mode  The start marker of an array or object. Should be '[' or '{'.
```

libraries/standard/json/utest/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+include("../jsonFilePaths.cmake")
+project ("json unit test")
+cmake_minimum_required (VERSION 3.13)
```
aggarw13, line 3: (Unrelated to this PR) Are these macro calls about CMake versioning required in these unit test files? Should they instead exist in the master CMakeLists.txt file that includes these child Cmake files?

We need to also think from a broken repository perspective about how the minimum version would be enforced.

libraries/standard/json/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# Include filepaths for source and include.
+include(jsonFilePaths.cmake)
+
+# JSON library target.
+add_library( json
+             ${JSON_SOURCES} )
+
+# JSON public include path.
+target_include_directories( json PUBLIC ${JSON_INCLUDE_PUBLIC_DIRS}
+                                        ${LOGGING_INCLUDE_DIRS} )
```
aggarw13, line 10: The JSON library doesn't have any logging calls in it, so there shouldn't be this dependency

libraries/standard/json/utest/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+# =====================  Create your mock here  (edit)  ========================
+# ================= Create the library under test here (edit) ==================
+
+# list the files you would like to test here
+list(APPEND real_source_files
+            ${JSON_SOURCES}
+        )
+# list the directories the module under test includes
+list(APPEND real_include_directories
+            .
```
aggarw13, line 17: Are `.` and `${LOGGING_INCLUDE_DIRS}` dependencies required?

aggarw13: Same for `test_include_directories`

libraries/standard/json/src/json.c
yourslab: Seems a little contrary to the style of the other docstrings, but I don't mind. What do you think @dan4thewin?

libraries/standard/json/src/json.c
yourslab: Would it make sense to add these asserts everywhere else then? They seem a little redundant since we are making those asserts in the public-facing function. When does it make sense to have an assert @dan4thewin ?

libraries/CMakeLists.txt
yourslab: Good point. I will do this as part of a separate PR.

libraries/standard/json/utest/CMakeLists.txt
yourslab: I don't think this is necessary. We already have this as the minimum version in the root CMakeLists.txt file, so I will just remove this line and remove the rest in a separate PR.

libraries/standard/json/utest/CMakeLists.txt
yourslab: You are right. Forgot to remove these.

libraries/standard/json/CMakeLists.txt
yourslab: Thanks

libraries/standard/json/src/json.c
yourslab: Only makes sense to call skipEscape when the first character is an escape `"\\"`

yourslab:
codecov

yourslab:
@codecov-commenter

yourslab:
@codecov-commenter codecov

libraries/standard/json/src/json.c
leegeth: Why was this change made?

yourslab: because it is not falsifiable with any input so we cannot achieve 100% branch coverage without this change

leegeth: Oh ok. Got it. Thanks

libraries/standard/json/utest/json_utest.c
```diff
@@ -0,0 +1,1153 @@
...
+        memcpy( nestedObjectCur, NESTED_OBJECT_PREFIX, NESTED_OBJECT_PREFIX_LENGTH );
+        i += NESTED_OBJECT_PREFIX_LENGTH;
+        nestedObjectCur += NESTED_OBJECT_PREFIX_LENGTH;
+    }
+
+    memcpy( nestedObjectCur, NESTED_OBJECT_VALUE, NESTED_OBJECT_VALUE_LENGTH );
+    i += NESTED_OBJECT_VALUE_LENGTH;
+    nestedObjectCur += NESTED_OBJECT_VALUE_LENGTH;
+
+    while( i < len )
```
leegeth, line 384: Some comments about how this would result in correct number of closing brackets would be good.

libraries/standard/json/utest/json_utest.c
```diff
@@ -0,0 +1,1153 @@
...
+    "\":{\"" SECOND_QUERY_KEY "\" : \"\\\x20\"}}"
+#define UNKNOWN_ESCAPE_LENGTH    ( sizeof( UNKNOWN_ESCAPE ) - 1 )
+
+/* An unescaped control character is considered ILLEGAL. */
+#define UNESCAPED_CONTROL_CHAR            \
+    "{\"foo\":\"abc\",\"" FIRST_QUERY_KEY \
+    "\":{\"" SECOND_QUERY_KEY "\" : \"\x15\"}}"
+#define UNESCAPED_CONTROL_CHAR_LENGTH    ( sizeof( UNESCAPED_CONTROL_CHAR ) - 1 )
+
+/* Triggers the case in which i >= max for skipUTF8MultiByte.
```
leegeth, line 253: What is 'i' in i>=max?

libraries/standard/json/utest/json_utest.c
```diff
@@ -0,0 +1,1153 @@
...
+                              COMPLETE_QUERY_KEY,
+                              COMPLETE_QUERY_KEY_LENGTH,
+                              JSON_QUERY_SEPARATOR[ 0 ],
+                              &outValue,
+                              &outValueLength );
+    TEST_ASSERT_EQUAL( JSONSuccess, jsonStatus );
+    TEST_ASSERT_EQUAL( UNICODE_ESCAPE_SEQUENCES_BMP_LENGTH, outValueLength );
+    TEST_ASSERT_EQUAL_STRING_LEN( UNICODE_ESCAPE_SEQUENCES_BMP,
+                                  outValue,
+                                  UNICODE_ESCAPE_SEQUENCES_BMP_LENGTH );
```
leegeth, line 691: How about test cases in which  if only FIRST_QUERY_KEY is provided as the key?

libraries/standard/json/utest/json_utest.c
```diff
@@ -0,0 +1,1153 @@
...
+                              COMPLETE_QUERY_KEY,
+                              COMPLETE_QUERY_KEY_LENGTH,
+                              JSON_QUERY_SEPARATOR[ 0 ],
+                              &outValue,
+                              &outValueLength );
+    TEST_ASSERT_EQUAL( JSONIllegalDocument, jsonStatus );
+}
+
+/**
+ * @brief Test that a nested collection can only only have JSON_MAX_DEPTH levels
```
leegeth, line 1132: only only -> only

leegeth:
```suggestion
 * @brief Test that a nested collection can only have upto JSON_MAX_DEPTH levels
```

libraries/standard/json/utest/json_utest.c
yourslab: It shouldn't change the coverage but added an additional test case for this

libraries/standard/json/utest/json_utest.c
yourslab: The iterator to advance for each skip function (Added a comment stating this).

libraries/standard/json/utest/json_utest.c
```diff
@@ -746,6 +773,24 @@ void test_JSON_Search_Illegal_Documents( void )
...
     size_t outValueLength;
 
+    jsonStatus = JSON_Search( WHITE_SPACE,
+                              WHITE_SPACE_LENGTH,
+                              COMPLETE_QUERY_KEY,
+                              COMPLETE_QUERY_KEY_LENGTH,
+                              JSON_QUERY_SEPARATOR[ 0 ],
+                              &outValue,
+                              &outValueLength );
+    TEST_ASSERT_EQUAL( JSONIllegalDocument, jsonStatus );
```
leegeth, line 850: Why is the JSON_Search returning different return values for invalid JSON documents? 
JSONIllegalDocument on this line and JSONNotFound on line 810

yourslab: @dan4thewin mentioned that cases like these are expected when the user does not call JSON_Validate before calling JSON_Search, and the docs will be updated to say so

PR #1122, opened 2020-08-13
Fix doxygen warnings in MQTT Library documentation

sarenameas:
- Fix doxygen warnings throughout mqtt documentation.
- Add documentation to mqtt_lightweight.c static functions where missing (this was a warning)
- Add all words to the lexicon.txt
- Opportunistically Fix other code warnings found in the unit test and mqtt.c
- Ignore private functions in mqtt_state.h and .c

Doxygen configs and other infrastructure is to be in a separate PR.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1140, opened 2020-08-19
Change state status code and increase coverage

muneebahmed10:
*Description of changes:*
Changes the status code returned by the state engine from bad parameter to bad response if an unexpected packet is received. Also adds some debug logs to aid in debugging unexpected packets received from the network, and increases test coverage to 100%.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt_state.c
+++ b/libraries/standard/mqtt/src/mqtt_state.c
@@ -424,22 +424,17 @@ static size_t findInRecord( const MQTTPubAckInfo_t * records,
 {
     size_t index = 0;
 
+    assert( packetId != MQTT_PACKET_ID_INVALID );
+
     *pCurrentState = MQTTStateNull;
 
-    if( packetId == MQTT_PACKET_ID_INVALID )
+    for( index = 0; index < recordCount; index++ )
     {
-        index = recordCount;
-    }
-    else
-    {
-        for( index = 0; index < recordCount; index++ )
+        if( records[ index ].packetId == packetId )
         {
-            if( records[ index ].packetId == packetId )
-            {
-                *pQos = records[ index ].qos;
-                *pCurrentState = records[ index ].publishState;
-                break;
-            }
+            *pQos = records[ index ].qos;
+            *pCurrentState = records[ index ].publishState;
+            break;
         }
     }
 
@@ -964,7 +959,8 @@ MQTTStatus_t MQTT_UpdateStatePublish( MQTTContext_t * pMqttContext,
  * @param[in] opType Send or Receive.
  * @param[out] pNewState Updated state of the publish.
  *
- * @return #MQTTBadParameter, #MQTTIllegalState, or #MQTTSuccess.
+ * @return #MQTTBadParameter, #MQTTBadResponse, #MQTTIllegalState, or
+ * #MQTTSuccess.
  */
 MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
                                   uint16_t packetId,
@@ -978,13 +974,24 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
     MQTTQoS_t qos = MQTTQoS0;
     size_t recordIndex = MQTT_STATE_ARRAY_MAX_COUNT;
     MQTTPubAckInfo_t * records = NULL;
-    MQTTStatus_t status = MQTTBadParameter;
+    MQTTStatus_t status = MQTTBadResponse;
 
     if( ( pMqttContext == NULL ) || ( pNewState == NULL ) )
     {
         LogError( ( "Argument cannot be NULL: pMqttContext=%p, pNewState=%p.",
                     ( void * ) pMqttContext,
                     ( void * ) pNewState ) );
+        status = MQTTBadParameter;
+    }
+    else if( packetId == MQTT_PACKET_ID_INVALID )
+    {
+        LogError( ( "Packet ID must be nonzero." ) );
+        status = MQTTBadParameter;
+    }
+    else if( packetType > MQTTPubcomp )
+    {
+        LogError( ( "Invalid packet type." ) );
+        status = MQTTBadParameter;
     }
     else
     {
@@ -1019,7 +1026,7 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
     }
     else
     {
-        LogError( ( "No matching record found for publish %u.", packetId ) );
+        LogError( ( "No matching record found for publish ID %u.", packetId ) );
     }
 
     return status;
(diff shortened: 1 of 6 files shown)


libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -964,7 +959,8 @@ MQTTStatus_t MQTT_UpdateStatePublish( MQTTContext_t * pMqttContext,
  * @param[in] opType Send or Receive.
  * @param[out] pNewState Updated state of the publish.
  *
- * @return #MQTTBadParameter, #MQTTIllegalState, or #MQTTSuccess.
+ * @return #MQTTBadParameter, #MQTTBadResponse, #MQTTIllegalState, or
+ * #MQTTSuccess.
  */
```
aggarw13, line 964: Why is there doxygen documentation for these public facing functions inside the source file?

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -978,13 +974,24 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
...
+        status = MQTTBadParameter;
+    }
+    else if( packetId == MQTT_PACKET_ID_INVALID )
+    {
+        LogError( ( "Packet ID must be nonzero." ) );
+        status = MQTTBadParameter;
+    }
+    else if( packetType > MQTTPubcomp )
+    {
+        LogError( ( "Invalid packet type." ) );
```
aggarw13, line 993: Log the packet type value?

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -1019,7 +1026,7 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
     }
     else
     {
-        LogError( ( "No matching record found for publish %u.", packetId ) );
+        LogError( ( "No matching record found for publish ID %u.", packetId ) );
```
aggarw13, line 1029:
```suggestion
        LogError( ( "No matching record found for publish: PacketId=%u.", packetId ) );
```

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: It was added in #1120: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1122#discussion_r470146719

libraries/standard/mqtt/src/mqtt_state.c
```diff
@@ -978,13 +974,24 @@ MQTTStatus_t MQTT_UpdateStateAck( MQTTContext_t * pMqttContext,
     MQTTQoS_t qos = MQTTQoS0;
     size_t recordIndex = MQTT_STATE_ARRAY_MAX_COUNT;
     MQTTPubAckInfo_t * records = NULL;
-    MQTTStatus_t status = MQTTBadParameter;
+    MQTTStatus_t status = MQTTBadResponse;
```
aggarw13, line 977: What is the exact reason for this function to return `BadResponse`?  No path in this function is explicitly setting the `status` value as `BadResponse`.  
If the case for it to return `BadResponse` is when no record ID is found for the packet ID, a better return value would be `MQTTNoRecordFound` as this function by itself is not dealing with the transport interface directly...
If we are sticking with `BadResponse`, at least the `@return` documentation of the API should mention the case when it would return `BadResponse`

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -740,6 +742,23 @@ void test_MQTT_GetUnsubscribePacketSize( void )
                                             &packetSize );
     TEST_ASSERT_EQUAL_INT( MQTTBadParameter, status );
 
+    /* Verify packet size cannot exceed limit. Note the max remaining length of
+     * an MQTT packet is 2^28-1 = 268435455, or 256MiB. Since the only way to increase
+     * the subscribe packet size is with the topic filters of the subscriptions
+     * (the lengths of which are only 2 bytes), we need at least
+     * 2^28 / 2^16 = 2^12 = 4096 of them. */
+    for( i = 0; i < 4096; i++ )
```
aggarw13, line 750: Which specific line in `MQTT_GetUnsubscribePacketSize` does this provide coverage for?

muneebahmed10: An error log that can only be reached if the packet size would exceed that allowed by the spec: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/5ae868b852447d869b56c11f982af84a24be4993/libraries/standard/mqtt/src/mqtt_lightweight.c#L1772

libraries/standard/mqtt/src/mqtt_state.c
muneebahmed10: @leegeth and I had an offline conversation regarding this. This function is only used by the process loop to update states from packets received from the network. If the status here isn't set to bad parameter, it will need to be changed by the process loop before returning. Since this function isn't part of the public API, it seemed fine to change the status here instead and bubble it back to the process loop. I can update the documentation

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
aggarw13: Don't we disable log messages for coverage results? Or is coverage for that branch still required when logging is disabled?

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
```diff
@@ -709,6 +709,8 @@ void test_MQTT_GetUnsubscribePacketSize( void )
     size_t remainingLength = 0;
     size_t packetSize = 0;
     MQTTStatus_t status = MQTTSuccess;
+    MQTTSubscribeInfo_t fourThousandSubscriptions[ 4096 ] = { 0 };
+    int i;
```
aggarw13, line 713: Will using a non-bit-width specific integer type generate a warning with the `-Wextra -Wall` flags?

libraries/standard/mqtt/src/mqtt_state.c
aggarw13: One cannot necessarily call this function that is part of the public header file as non-public (even though we will _try_ to discourage its public visibility through doxygen).  The customer application could still call into it if they discover this function for their need. However, the changes related to that discussion is beyond the scope of this PR, so, I am fine with the `@return` documentation updates for this function

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: That's possible, I know the default logging was changed to log errors for unit tests in #1096, so this will result in 100% coverage without extra cmake flags. Either way, I think this test is worth it to copy here, since it's needed in test_MQTT_GetSubscribePacketSize for non-log coverage, and it makes sense to test them both with this test.

libraries/standard/mqtt/utest/mqtt_lightweight_utest.c
muneebahmed10: There weren't any warnings when running the tests, and I don't see why that would be a warning. `int32_t` would be typedef'ed to `int` anyway in most systems, right?

This test isn't new, it's copied from test_MQTT_GetSubscribePacketSize

aggarw13: I faintly remember that some compiler had complained about using `int`, but if you don't see any warning, that works.

PR #1141, opened 2020-08-19
Reduce the complexity score for openssl_connect

leegeth:
*Description of changes:*
Reduce the GNU complexity score for openssl_connect from 9 to 8.

### DIFF
--- a/platform/posix/transport/src/openssl_posix.c
+++ b/platform/posix/transport/src/openssl_posix.c
@@ -125,6 +125,16 @@ static int setCredentials( SSL_CTX * pSslContext,
 static void setOptionalConfigurations( SSL * pSsl,
                                        const OpensslCredentials_t * pOpensslCredentials );
 
+/**
+ * @brief Converts the sockets wrapper statuses to openssl statuses.
+ *
+ * @param[in] socketStatus Sockets wrapper status.
+ *
+ * @return #OPENSSL_SUCCESS, #OPENSSL_INVALID_PARAMETER, #OPENSSL_DNS_FAILURE,
+ * and #OPENSSL_CONNECT_FAILURE.
+ */
+static OpensslStatus_t convertToOpensslStatus( SocketStatus_t socketStatus );
+
 /*-----------------------------------------------------------*/
 
 #if ( LIBRARY_LOG_LEVEL == LOG_DEBUG )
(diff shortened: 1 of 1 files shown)


yourslab (dismissed):

platform/posix/transport/src/openssl_posix.c
```diff
@@ -125,6 +125,16 @@ static int setCredentials( SSL_CTX * pSslContext,
 static void setOptionalConfigurations( SSL * pSsl,
                                        const OpensslCredentials_t * pOpensslCredentials );
 
+/**
+ * @brief Converts the sockets wrapper statuses to openssl statuses.
```
aggarw13, line 129: Nitpick suggestion as this only takes a single status value as input parameter, not multiple status_es_
```suggestion
 * @brief Converts the sockets wrapper status to openssl status.
```

platform/posix/transport/src/openssl_posix.c
```diff
@@ -590,10 +617,8 @@ OpensslStatus_t Openssl_Disconnect( NetworkContext_t * pNetworkContext )
     /* Tear down the socket connection. */
     socketStatus = Sockets_Disconnect( pNetworkContext->socketDescriptor );
 
-    if( socketStatus == SOCKETS_INVALID_PARAMETER )
-    {
-        returnStatus = OPENSSL_INVALID_PARAMETER;
-    }
+    /* Convert socket wrapper status to openssl status. */
```
aggarw13, line 620: I see a NULL de-referencing bug in this function where a NULL `pNetworkContext` will be de-referenced in the call to `Sockets_Disconnect`

leegeth: Yea. That's a good find. Unrelated to this change though, but I think it definitely will be a good fix. I will fix it. Thanks.

gkwicker: Suggest refactoring for readability and to remove the possible `NULL` dereference at line 618:

```c
    SocketStatus_t socketStatus = SOCKETS_INVALID_PARAMETER:

    if( pNetworkContext != NULL )
    {
        if( pNetworkContext->pSsl != NULL )
        {
            /* SSL shutdown logic in here... */
        }
         /* Tear down the socket connection, pNetworkContext != NULL here. */ 
        socketStatus = Sockets_Disconnect( pNetworkContext->socketDescriptor );
    }
    else
    {
        LogError( ( "Parameter check failed: pNetworkContext is NULL." ) );
    }
    return convertToOpensslStatus( socketStatus );
```

leegeth: Thanks Gary. Addressed it.

platform/posix/transport/src/openssl_posix.c
```diff
@@ -590,37 +590,31 @@ OpensslStatus_t Openssl_Connect( NetworkContext_t * pNetworkContext,
 
 OpensslStatus_t Openssl_Disconnect( NetworkContext_t * pNetworkContext )
 {
-    OpensslStatus_t returnStatus = OPENSSL_SUCCESS;
-    SocketStatus_t socketStatus = SOCKETS_SUCCESS;
+    SocketStatus_t socketStatus = SOCKETS_INVALID_PARAMETER;
```
aggarw13, line 593: Using `SOCKETS_INVALID_PARAMETER` as the initialized value carries the knowledge that the `convertToOpensslStatus` translates `SOCKETS_INVALID_PARAMETER` to `OPENSSL_INVALID_PARAMETER`. 
If the translation would change (though quite unlikely), then the return value of this function for an invalid `pNetworkContext` would change..

leegeth: I have thought about it too, but I think it is  fair assumption to make that SOCKETS_INVALID_PARAMETER is converted to OPENSSL_INVALID_PARAMETER always. I have added comments to clarify that

PR #1142, opened 2020-08-20
Update proof for HTTPClient_ReadHeader to pass double pointer

yourslab:
I mistakenly used a regular pointer even though the parameter requires a double pointer. Although a pointer to a pointer can carry the same values as a regular pointer, so that's probably why CBMC didn't flag this.

### DIFF
--- a/libraries/standard/http/cbmc/sources/http_cbmc_state.c
+++ b/libraries/standard/http/cbmc/sources/http_cbmc_state.c
@@ -90,7 +90,7 @@ HTTPResponse_t * allocateHttpResponse( HTTPResponse_t * pResponse )
 
         pResponse->pBuffer = mallocCanFail( pResponse->bufferLen );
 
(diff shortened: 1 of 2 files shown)


libraries/standard/http/cbmc/sources/http_cbmc_state.c
```diff
@@ -90,7 +90,7 @@ HTTPResponse_t * allocateHttpResponse( HTTPResponse_t * pResponse )
 
         pResponse->pBuffer = mallocCanFail( pResponse->bufferLen );
 
-        __CPROVER_assume( headerOffset < pResponse->headersLen );
+        __CPROVER_assume( headerOffset <= pResponse->headersLen );
```
sarenameas, line 93: Should line 104 be `bodyOffset <= pResponse->bufferLen`?

PR #1143, opened 2020-08-20
Fix warnings in MQTT library and demo targets as CI enforces warnings as errors

aggarw13:
As CI has been updated to enforce warnings as errors with the `-Werror` for MQTT targets, we need to fix all warnings in MQTT library and demo targets to get passing CI checks


demos/mqtt/mqtt_demo_basic_tls/CMakeLists.txt
```diff
@@ -8,6 +8,9 @@ target_sources(
         "${DEMO_NAME}.c"
 )
 
+# Treat warnings as error for demo target.
+target_compile_options( ${DEMO_NAME} PUBLIC "-Werror" )
```
yourslab, line 12: What are we really trying to accomplish by passing this compile option by default? It seems more like an inconvenience for a customer who wants to experiment and make a few changes to the code that introduces warnings. Would it be better to just have CI pass this flag themselves?

aggarw13: Yeah, I agree. I will check with @nateglims to see whether they can enable it in CI.

PR #1148, opened 2020-08-21
Clear old records for clean sessions

muneebahmed10:
*Description of changes:*
Currently, `MQTT_Connect` does not clear records for a clean session, nor when the CONNACK's session present flag is set to false by the broker. Clients are required by the MQTT spec to discard state for clean sessions, and not clearing state when session present is false could lead to the state records being cluttered with dead publishes.

### DIFF
--- a/libraries/standard/mqtt/src/mqtt.c
+++ b/libraries/standard/mqtt/src/mqtt.c
@@ -277,14 +277,17 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
                                     bool * pSessionPresent );
 
 /**
- * @brief Resends pending acks for a re-established MQTT session.
+ * @brief Resends pending acks for a re-established MQTT session, or
+ * clears existing state records for a clean session.
  *
  * @param[in] pContext Initialized MQTT context.
+ * @param[in] sessionPresent Session present flag from the MQTT broker.
  *
- * @return #MQTTSendFailed if transport send failed;
+ * @return #MQTTSendFailed if transport send during resend failed;
  * #MQTTSuccess otherwise.
  */
-static MQTTStatus_t resendPendingAcks( MQTTContext_t * pContext );
+static MQTTStatus_t handleSessionResumption( MQTTContext_t * pContext,
+                                             bool sessionPresent );
 
 /**
  * @brief Serializes a PUBLISH message.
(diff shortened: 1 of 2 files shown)


libraries/standard/mqtt/src/mqtt.c
```diff
@@ -1558,25 +1561,39 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
...
+               ( status == MQTTSuccess ) )
+        {
+            status = sendPublishAcks( pContext, packetId, state );
+
+            packetId = MQTT_PubrelToResend( pContext, &cursor, &state );
+        }
+    }
+    else
+    {
+        ( void ) memset( pContext->outgoingPublishRecords,
```
aggarw13, line 1591: A comment stating that publish records are being cleared due to clean session would be good to have

libraries/standard/mqtt/src/mqtt.c
```diff
@@ -277,14 +277,17 @@ static MQTTStatus_t receiveConnack( const MQTTContext_t * pContext,
                                     bool * pSessionPresent );
 
 /**
- * @brief Resends pending acks for a re-established MQTT session.
+ * @brief Resends pending acks for a re-established MQTT session, or
+ * clears existing state records for a clean session.
  *
  * @param[in] pContext Initialized MQTT context.
+ * @param[in] sessionPresent Session present flag from the MQTT broker.
```
aggarw13, line 284:
```suggestion
 * @param[in] sessionPresent Session present flag that is received from the MQTT broker.
```

libraries/standard/mqtt/utest/mqtt_utest.c
```diff
@@ -1033,6 +1034,11 @@ void test_MQTT_Connect_happy_path()
     mqttContext.keepAliveIntervalSec = 0;
     connectInfo.cleanSession = true;
     sessionPresentExpected = false;
+    /* Populate some state records to make sure they are cleared. */
```
aggarw13, line 1037:
```suggestion
    /* Populate some state records to make sure they are cleared as we are establishing a clean session with the broker. */
```

libraries/standard/mqtt/src/mqtt.c
muneebahmed10: Thanks, I forgot to add one here

PR #1150, opened 2020-08-21
Doxygen templates and instructions for writing docs for a new library

sarenameas:
*Description of changes:*
- "docs" folder should be called "doc"
- Templates for style.css, layout.xml, pages.txt, and config.doxyfile
- Instructions for writing doxygen docs for a library in doxygen_guide.md
- Update all existing doxygen configs to the latest version

Related PRs:
- Scripts for keeping repeated files sytle.css and layout.xml are in a separate PR: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1151
- MQTT specific API docs are in a separate PR: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1149

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1149, opened 2020-08-21
Doxygen for MQTT API Only

sarenameas:
*Description of changes:*

- Doxygen for MQTT API (functions, constants, configurable macros, enums, structures, and function pointers)

Please run:
```console
cd libraries/standard/mqtt
doxygen doc/doxygen/config.doxyfile 
```

Expected directory structure for each library:
- doc
  - doxygen
    - style.css
    - pages.txt
    - layout.xml
    - config.doxyfile  

Note:
- Templates for style.css, pages.txt, layout.xml, and config.doxyfile is in a separate PR: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1150
- Scripts to verify the doxygen warnings and check the templates will be in a separate PR: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1151

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1151, opened 2020-08-21
CI - Doxygen warnings check scripts and template match checks

sarenameas:
*Description of changes:*
- Python script to generate doxygen for the single library. We are still deciding if this belongs with the library or not.
- Python script to generate doxygen for each library and check that the XML and CSS templates are the same.

Other PRs:
- MQTT Doxygen API code: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1149
- Templates and instructions for writing docs: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1150

### DIFF
--- a/tools/ci/ci_doxygen.py
+++ b/tools/ci/ci_doxygen.py
@@ -176,9 +176,9 @@ def check_transport_interface_match(results, abs_lib_path, lib_dir):
     # All lines except the second where the library/SDK name and version are will
     # need to match.
     line_count = 0
-    lib_file = open(lib_transport_interface_path, 'r')
+    lib_file = open(lib_transport_interface_path, "r")
     lib_lines = lib_file.readlines()
-    csdk_file = open(TRANSPORT_INTERFACE_PATH, 'r')
+    csdk_file = open(TRANSPORT_INTERFACE_PATH, "r")
     csdk_lines = csdk_file.readlines()
     same = True
     for line in lib_lines:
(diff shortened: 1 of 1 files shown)


tools/ci/ci_doxygen.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    Prints the Doxygen warnings found from running the configuration file in the
+    library under the abs_libs_path. Returns the number of warnings if there are
+    any.
+    """
+    lib_dirs = os.listdir(abs_libs_path)
+    for lib_dir in lib_dirs:
+        abs_lib_path = os.path.join(abs_libs_path, lib_dir)
+        if os.path.exists(os.path.join(abs_lib_path, "doc", "doxygen", "config.doxyfile")):
+            target_result = results.setdefault(lib_dir, {})
```
navinns, line 77: use elaborate key for dict, if multiple parent dir has same lib_dir in them then we will overwrite them

sarenameas: They should not, we expect them to be unique

libraries/standard/mqtt/tools/ci_doxygen.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/python3
```
muneebahmed10, line 1: Would the generic `#!/usr/bin/env python3` be better for compatibility, or does it not matter since these scripts are only intended to be used on specific systems?
```suggestion
#!/usr/bin/env python3
```

sarenameas: I did not know this is better for compatibility, thank you. These scripts are not intended for a specific system.

nateglims (dismissed):

sarenameas:
I need to turn these into github actions.

PR #1154, opened 2020-08-24
MISRA fixes for plain text transport interface implementation.

leegeth:
*Description of changes:*
MISRA fixes for plain text transport interface implementation.
MISRA C-2012 Rule 8.13 - Will be handled in a follow up PR.


sarenameas:
Add "errornumber" to lexicon.txt

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -88,6 +83,10 @@ static void logTransportError( void )
         case EPIPE:
             LogError( ( "The socket is shut down for writing, or the socket is connection-mode and is no longer connected. In the latter case, and if the socket is of type SOCK_STREAM or SOCK_SEQPACKET and the MSG_NOSIGNAL flag is not set, the SIGPIPE signal is generated to the calling thread." ) );
             break;
+
+        default:
+            LogError( ( "Unexpected error code: error number=%d", errorNumber ) );
```
yourslab, line 88: Would it be better to use errno=%d here since errno to show that it relates to errno.h?

leegeth: Yes. That is a good suggestion. Thanks

jainamt: Since this is posix specific, we can report a more meaningful error for customer by using ```strerror```

PR #1157, opened 2020-08-25
MISRA fixes for openssl transport implementation

leegeth:
*Issue #, if available:*

*Description of changes:*
MISRA fixes for openssl transport implementation.
Fixes all the MISRA violations except.
8.13 - To be fixed for all the transport implementations.


platform/posix/transport/src/openssl_posix.c
```diff
@@ -74,8 +74,8 @@
  *
  * @return 1 on success; -1, 0 on failure;
  */
-static int setRootCa( SSL_CTX * pSslContext,
-                      const char * pRootCaPath );
+static int32_t setRootCa( const SSL_CTX * pSslContext,
```
aggarw13, line 77: Why has const been set for this setter function but not for other setter functions for private key and client certificate?
Isn't the context expected to be modified when setting the root CA?

platform/posix/transport/src/openssl_posix.c
```diff
@@ -421,8 +435,13 @@ static void setOptionalConfigurations( SSL * pSsl,
         LogDebug( ( "Setting server name %s for SNI.",
                     pOpensslCredentials->sniHostName ) );
 
-        sslStatus = SSL_set_tlsext_host_name( pSsl,
-                                              pOpensslCredentials->sniHostName );
+        /* MISRA Rule 11.8 flags the following line for removing the const
+         * qualifier from the point to type. This rule is suppressed because
```
aggarw13, line 439:
```suggestion
         * qualifier from the pointed to type. This rule is suppressed because
```

platform/posix/transport/src/openssl_posix.c
leegeth: Misra tagged only this as the the openssl functions called within this function expects a pointer to a const ssl context, but that is not the same with other 2 functions.

platform/posix/transport/include/openssl_posix.h
```diff
@@ -120,7 +120,7 @@ typedef struct OpensslCredentials
      * @note By setting this to 0, OpenSSL uses the default value,
      * which is 16384 (SSL3_RT_MAX_PLAIN_LENGTH).
      */
-    size_t maxFragmentLength;
+    uint64_t maxFragmentLength;
```
aggarw13, line 123: Why is this type change needed?

leegeth: In most of the openssl functions in which maxFragmentLength is used, an explicit cast was required to uint64_t. Instead of explicitly casting it everywhere, changed the type to uint64_t.

aggarw13: Okay, I asked primarily in relation to the `alpnProtosLen` member which uses a `size_t` type. Should it be fine to change that member's type as well for consistency?

leegeth: Yea. That is a good point. I will change the type of that as well. Thanks.

cobusve, line 123: If this fragment Length is limited to 16384 then why is 64bit the most appropriate type here? It seems like this requires only 16 bits?

leegeth: Thanks. Resolved it.

platform/posix/transport/src/openssl_posix.c
```diff
@@ -460,10 +479,10 @@ OpensslStatus_t Openssl_Connect( NetworkContext_t * pNetworkContext,
 {
     SocketStatus_t socketStatus = SOCKETS_SUCCESS;
     OpensslStatus_t returnStatus = OPENSSL_SUCCESS;
-    int64_t verifyPeerCertStatus = X509_V_OK;
     int32_t sslStatus = 0;
     uint8_t sslObjectCreated = 0;
     SSL_CTX * pSslContext = NULL;
+    int32_t verifyPeerCertStatus = X509_V_OK;
```
aggarw13, line 478: Why was the bit-width reduced for this? It seems to be adding an explicit cast for the return value from `SSL_get_verify_result`...

leegeth: I have looked at the value range and the max value is ~75.  Also, our implementation only uses the information of whether the return value is X509_V_OK(0) or not. A type with the same width of the return type of SSL_get_verify_result is going to be of 64 bits and this was an opportunity we identified that we could optimize.

PR #1159, opened 2020-08-25
Misra fixes for posix clock utility

leegeth:
*Description of changes:*
Misra fixes for posix clock utility


platform/posix/clock_posix.c
```diff
@@ -37,24 +37,28 @@
...
-             + ( uint32_t ) ( timeSpec.tv_nsec / ( 1000 * 1000 ) );
+    timeMs = ( timeSpec.tv_sec * MILLISECONDS_PER_SECOND )
+             + ( timeSpec.tv_nsec / NANOSECONDS_PER_MILLISECOND );
 
-    return timeMs;
+    /* Libraries need only the lower 32 bits of the time in milliseconds, since
+     * this function is used only for a calculating the time difference.
+     * Also, the possible overflows of this time value is handled in the
+     * libraries. */
+    return ( uint32_t ) timeMs;
```
aggarw13, line 61: Does this cast not create a MISRA violation?

platform/posix/clock_posix.c
```diff
@@ -63,8 +67,9 @@ void Clock_SleepMs( uint32_t sleepTimeMs )
 {
     /* Convert parameter to timespec. */
     struct timespec sleepTime = { 0 };
-    sleepTime.tv_sec = sleepTimeMs / MILLISECONDS_PER_SECOND;
-    sleepTime.tv_nsec = ( sleepTimeMs % MILLISECONDS_PER_SECOND ) * NANOSECONDS_PER_MILLISECOND;
+
+    sleepTime.tv_sec = ( ( time_t ) sleepTimeMs / ( time_t ) MILLISECONDS_PER_SECOND );
+    sleepTime.tv_nsec = ( ( int64_t ) sleepTimeMs % MILLISECONDS_PER_SECOND ) * NANOSECONDS_PER_MILLISECOND;
```
aggarw13, line 72: Is the `int64_t` cast for just `sleepTimeMs` or for the `sleepTimeMs % MILLISECONDS_PER_SECOND` expression?

aggarw13: As the POSIX specification uses `long` as the type for `tv_nsec`, should we use that type here instead?
https://pubs.opengroup.org/onlinepubs/007908799/xsh/time.h.html

platform/posix/clock_posix.c
```diff
@@ -63,8 +67,9 @@ void Clock_SleepMs( uint32_t sleepTimeMs )
 {
     /* Convert parameter to timespec. */
     struct timespec sleepTime = { 0 };
-    sleepTime.tv_sec = sleepTimeMs / MILLISECONDS_PER_SECOND;
-    sleepTime.tv_nsec = ( sleepTimeMs % MILLISECONDS_PER_SECOND ) * NANOSECONDS_PER_MILLISECOND;
+
+    sleepTime.tv_sec = ( ( time_t ) sleepTimeMs / ( time_t ) MILLISECONDS_PER_SECOND );
```
aggarw13, line 71: Does this not work?
`sleepTime.tv_sec = ( time_t ) ( sleepTimeMs /  MILLISECONDS_PER_SECOND );`

platform/posix/clock_posix.c
leegeth: No. Misra doesn't complain about this as it is an explicit cast.

platform/posix/clock_posix.c
leegeth: Use of long is also a MISRA directive violation. We cannot directly use that.
int64_t cast is for sleepTimeMs

platform/posix/clock_posix.c
leegeth: No. The issue was that the result of a composite expression cannot be casted to a different type. So that is not a solution.

platform/posix/clock_posix.c
```diff
@@ -37,24 +37,28 @@
...
     struct timespec timeSpec;
 
     /* Get the MONOTONIC time. */
-    clock_gettime( CLOCK_MONOTONIC, &timeSpec );
+    ( void ) clock_gettime( CLOCK_MONOTONIC, &timeSpec );
 
     /* Calculate the milliseconds from timespec. */
-    timeMs = ( uint32_t ) ( timeSpec.tv_sec * 1000 )
-             + ( uint32_t ) ( timeSpec.tv_nsec / ( 1000 * 1000 ) );
+    timeMs = ( timeSpec.tv_sec * MILLISECONDS_PER_SECOND )
```
cobusve, line 54: This is a 32_bit multiplication. Seems like int64_t for timeMs is not required, this way will just use twice the memory with the same precision?

PR #1162, opened 2020-08-25
CBMC proofs for JSON library

dan4thewin:
The JSON library is divided into sets of functions with each set having a separate proof.  Each set is mocked in adjacent proofs.  In combination, the coverage is complete with no errors reported.


yourslab:
The `cbmc-batch.yaml` files are needed for CBMC checks to run on the newly created JSON CBMC proofs. I believe you need to force add the `cbmc-batch.yaml` files because the CBMC team made the decision to add them to the `.gitignore`.

markrtuttle:
I will not be able to review this before vacation, I've run out of time.

libraries/standard/json/cbmc/proofs/JSON_Validate/cbmc-viewer.json
```diff
@@ -0,0 +1,7 @@
+{ "expected-missing-functions":
+  [
+
+  ],
+  "proof-name": "JSON_Validate",
+  "proof-root": "libraries/standard/json/cbmc/proofs"
+}
```
sarenameas, line 7: Please add a new line

libraries/standard/json/cbmc/proofs/JSON_Search/glue.h
```diff
@@ -0,0 +1,27 @@
+#ifndef GLUE_H_
+#define GLUE_H_
+
+#include "json.h"
+#include "skipGeneric.h"
+
+bool_ skipAnyScalar( const char * buf,
```
sarenameas, line 7: Can we please document these?

dan4thewin: They are replacements for the functions of the same name from json.c, so the documentation is in json.c.  I'll add a comment to that effect.

libraries/standard/json/cbmc/proofs/JSON_Validate/cbmc-viewer.json
dan4thewin: Add it where?

sarenameas: Right here lol

libraries/standard/json/cbmc/proofs/JSON_Validate/JSON_Validate_harness.c
```diff
@@ -0,0 +1,45 @@
...
+    char * buf = NULL;
+    size_t max;
+    JSONStatus_t ret;
+
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    if( nondet_bool() )
+    {
+        buf = malloc( max );
+        __CPROVER_assume( __CPROVER_r_ok( buf, max ) );
```
yourslab, line 41: I'm not sure if these calls to `__CPROVER_r_ok` are needed since we are already mallocing it to the max size. Probably doesn't hurt though.

libraries/standard/json/cbmc/proofs/JSON_Validate/Makefile
```diff
@@ -0,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+HARNESS_FILE=JSON_Validate_harness
+
+CBMC_MAX_BUFSIZE=14
```
yourslab, line 6: Is there any reason to choose this magic number?

libraries/standard/json/cbmc/sources/skipGeneric.c
```diff
@@ -0,0 +1,33 @@
...
+                          size_t min )
+{
+    bool_ ret = nondet_bool() ? true : false;
+
+    if( ret == true )
+    {
+        size_t x;
+        __CPROVER_assume( x >= min );
+        __CPROVER_assume( x <= max );
+        __CPROVER_assume( *start <= max );
```
yourslab, line 20: Should we __CPROVER_assert that start is not NULL?

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
```diff
@@ -0,0 +1,53 @@
...
+
+bool_ skipAnyScalar( const char * buf,
+                     size_t * start,
+                     size_t max )
+{
+    /* min argument is 1 for a single character scalar like 0,
+     * or 2 for an empty double-quoted string, i.e., "". */
+    size_t min = 1;
+
+    if( ( *start < max ) && ( buf[ *start ] == '"' ) )
```
yourslab, line 11: Same comment about using __CPROVER_assert to check that start and buf are not NULL

libraries/standard/json/cbmc/proofs/JSON_Validate/Makefile
yourslab: Although, I cannot think of a workaround for this that doesn't involve using a magic number.

libraries/standard/json/cbmc/include/skipGeneric.h
```diff
@@ -0,0 +1,31 @@
...
+ * from json.c.
+ *
+ * @param[in] buf  The buffer to parse.
+ * @param[in,out] start  The index at which to begin.
+ * @param[in] max  The size of the buffer.
+ * @param[in] min  The smallest size required for a true result.
+ *
+ * @return true or false, nondeterministically
+ * if true, the index in start will increment by at least min
+ * but will not exceed max.
```
danielsn, line 24: If min > max, what happens?

libraries/standard/json/cbmc/proofs/JSON_Validate/JSON_Validate_harness.c
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    ret = JSON_Validate( buf, max );
```
danielsn, line 44: What are the post-conditions of this function?

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.h
```diff
@@ -0,0 +1,27 @@
+#ifndef GLUE_H_
+#define GLUE_H_
+
+#include "json.h"
+#include "skipGeneric.h"
+
+bool_ skipAnyLiteral( const char * buf,
```
danielsn, line 7: comments

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.h
```diff
@@ -0,0 +1,27 @@
+#ifndef GLUE_H_
```
danielsn, line 1: copyright

libraries/standard/json/cbmc/proofs/Makefile-json.common
```diff
@@ -0,0 +1,20 @@
...
+HARNESS_ENTRY=harness
+
+DEFINES += -DCBMC_MAX_BUFSIZE=$(CBMC_MAX_BUFSIZE)
+INCLUDES += -I$(PROOFDIR) -I$(CBMC_ROOT)/include
+
+PROOF_SOURCES += $(PROOFDIR)/$(HARNESS_FILE).c
+ifneq (,$(wildcard glue.c))
+PROOF_SOURCES += $(PROOFDIR)/glue.c
+PROOF_SOURCES += $(CBMC_ROOT)/sources/skipGeneric.c
+endif
```
danielsn, line 13: indent code inside an if

libraries/standard/json/cbmc/proofs/skipAnyLiteral/README.md
```diff
@@ -0,0 +1,10 @@
+skipAnyLiteral proof
+==============
+
+This directory contains a memory safety proof for skipAnyLiteral.
```
danielsn, line 4: Runtime? Coverage? Assumptions made?

libraries/standard/json/cbmc/proofs/skipCollection/skipCollection_harness.c
```diff
@@ -0,0 +1,43 @@
...
+    size_t start, max;
+    JSONStatus_t ret;
+
+    __CPROVER_assume( max > 0 );
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    buf = malloc( max );
+    __CPROVER_assume( __CPROVER_r_ok( buf, max ) );
+
+    ret = skipCollection( buf, &start, max );
```
danielsn, line 42: I know its probably a coding convention on your team, but why not just say
 `JSONStatus_t ret = skipCollection( buf, &start, max );`

libraries/standard/json/cbmc/proofs/skipEscape/skipEscape_harness.c
```diff
@@ -0,0 +1,51 @@
...
+
+typedef enum
+{
+    true = 1, false = 0
+} bool_;
+
+bool_ skipEscape( const char * buf,
+                  size_t * start,
+                  size_t max );
+
```
danielsn, line 37: isn't there a header for this?

libraries/standard/json/cbmc/proofs/skipEscape/skipEscape_harness.c
```diff
@@ -0,0 +1,51 @@
...
+    size_t start, max;
+    bool_ ret;
+
+    __CPROVER_assume( max > 0 );
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    buf = malloc( max );
+    __CPROVER_assume( __CPROVER_r_ok( buf, max ) );
+
+    ret = skipEscape( buf, &start, max );
```
danielsn, line 50: Same comments as all the other similar proofs.

libraries/standard/json/cbmc/proofs/skipNumber/skipNumber_harness.c
```diff
@@ -0,0 +1,51 @@
...
+                  size_t * start,
+                  size_t max );
+
+void harness()
+{
+    char * buf;
+    size_t start, max;
+    bool_ ret;
+
+    __CPROVER_assume( max > 0 );
```
danielsn, line 44: What would happen if max was 0. Does the actual function have a test against this? And does it have an assert against that?

libraries/standard/json/cbmc/proofs/skipString/Makefile
```diff
@@ -0,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+HARNESS_FILE=skipString_harness
+
+CBMC_MAX_BUFSIZE=14
```
danielsn, line 6: why?

libraries/standard/json/include/json.h
```diff
@@ -30,7 +30,9 @@ typedef enum
     JSONSuccess,
     JSONIllegalDocument,
     JSONMaxDepthExceeded,
-    JSONNotFound
+    JSONNotFound,
+    JSONNullParameter,
+    JSONBadParameter
```
danielsn, line 35: would it make sense to have a `JSON_MAX_ENUM` param?

libraries/standard/json/cbmc/sources/skipGeneric.c
```diff
@@ -0,0 +1,33 @@
...
+                          size_t max,
+                          size_t min )
+{
+    bool_ ret = nondet_bool() ? true : false;
+
+    if( ret == true )
+    {
+        size_t x;
+        __CPROVER_assume( x >= min );
+        __CPROVER_assume( x <= max );
```
danielsn, line 19: If `min < max`, this will become `assume(0)`, and block further verification

libraries/standard/json/cbmc/sources/skipGeneric.c
```diff
@@ -0,0 +1,33 @@
...
+ * See skipGeneric.h for docs
+ *
+ * Advance buffer index beyond some minimum value.
+ */
+static bool_ skipGeneric( const char * buf,
+                          size_t * start,
+                          size_t max,
+                          size_t min )
+{
+    bool_ ret = nondet_bool() ? true : false;
```
danielsn, line 13: What are the preconditions on buf and start? We should assert them.

libraries/standard/json/cbmc/sources/skipGeneric.c
```diff
@@ -0,0 +1,33 @@
+#include "skipGeneric.h"
+
+/**
+ * See skipGeneric.h for docs
+ *
+ * Advance buffer index beyond some minimum value.
+ */
+static bool_ skipGeneric( const char * buf,
```
danielsn, line 8: This file concerns me, because we're literally doing nothing involving the buffer. So even if there was a memory-safety problem, we wouldn't necessarily catch it.

libraries/standard/json/cbmc/proofs/skipUTF8/Makefile
```diff
@@ -0,0 +1,15 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+HARNESS_FILE=skipUTF8_harness
+
+CBMC_MAX_BUFSIZE=10
```
danielsn, line 6: why?

libraries/standard/json/src/json.c
```diff
@@ -41,11 +43,13 @@ static void skipSpace( const char * buf,
                        size_t * start,
                        size_t max )
 {
-    size_t i = *start;
+    size_t i;
+
+    assert( ( buf != NULL ) && ( start != NULL ) && ( max > 0U ) );
```
danielsn, line 48: `assert(__CPROVER_r_ok(buf,*start+max))` - or something similar

libraries/standard/json/src/json.c
```diff
@@ -30,6 +29,9 @@ typedef enum
     true = 1, false = 0
 } bool_;
 
+#define isdigit_( x )    ( ( x >= '0' ) && ( x <= '9' ) )
+#define isspace_( x )    ( ( x == ' ' ) || ( x == '\t' ) || ( x == '\n' ) || ( x == '\r' ) )
```
danielsn, line 33: What about `\v` and `\f`?

libraries/standard/json/src/json.c
```diff
@@ -160,9 +160,17 @@ static bool_ skipUTF8MultiByte( const char * buf,
                                 size_t max )
 {
     bool_ ret = false;
-    size_t i = *start, bitCount, j;
+    size_t i, bitCount, j;
     uint32_t value = 0;
-    uint8_t c = ( uint8_t ) buf[ i ];
+    uint8_t c;
+
+    assert( ( buf != NULL ) && ( start != NULL ) && ( max > 0U ) );
```
danielsn, line 167: `assert __CPROVER_r_ok` on the actual required size of the buffer

libraries/standard/json/src/json.c
```diff
@@ -160,9 +160,17 @@ static bool_ skipUTF8MultiByte( const char * buf,
...
+
+    i = *start;
+    assert( i < max );
+    assert( buf[ i ] < 0 );
+
+    c = ( 0x80U | ( buf[ i ] & 0x7F ) );
```
danielsn, line 173: Magic numbers. Can these be made into named constants?

libraries/standard/json/src/json.c
```diff
@@ -241,24 +249,25 @@ static bool_ skipUTF8( const char * buf,
  *
  * @return the integer value upon success or UINT8_MAX on failure.
  */
-static uint8_t hexToInt( uint8_t c )
+static uint8_t hexToInt( char c )
```
danielsn, line 252: Is this better than a 256 entry lookup table?

libraries/standard/json/src/json.c
```diff
@@ -297,14 +306,19 @@ static bool_ skipHexEscape( const char * buf,
...
 {
     bool_ ret = false;
-    size_t i = *start, end = *start + 6U;
+    size_t i, end;
     uint16_t value = 0U;
 
+    assert( ( buf != NULL ) && ( start != NULL ) && ( max > 0U ) );
+
+    i = *start;
+    end = i + 6U;
```
danielsn, line 315: why 6? Magic number alert!

libraries/standard/json/src/json.c
```diff
@@ -364,9 +378,13 @@ static bool_ skipEscape( const char * buf,
...
     bool_ ret = false;
-    size_t i = *start;
+    size_t i;
+
+    assert( ( buf != NULL ) && ( start != NULL ) && ( max > 0 ) );
+
+    i = *start;
 
-    if( ( ( i + 1U ) < max ) && ( buf[ i ] == '\\' ) )
+    if( ( i < ( max - 1 ) ) && ( buf[ i ] == '\\' ) )
```
danielsn, line 387: can `max` ever by `0`?

libraries/standard/json/src/json.c
```diff
@@ -394,7 +412,7 @@ static bool_ skipEscape( const char * buf,
             default:
 
                 /* a control character: (NUL,SPACE) */
-                if( ( ( uint8_t ) c ) < 0x20U )
+                if( ( c > '\0' ) && ( c < ' ' ) )
```
danielsn, line 415: You have macros for other character sets. One here would make things clearer.

libraries/standard/json/src/json.c
```diff
@@ -404,7 +422,11 @@ static bool_ skipEscape( const char * buf,
         }
     }
 
-    *start = i;
+    if( ret == true )
```
danielsn, line 425: redundant.

libraries/standard/json/src/json.c
```diff
@@ -423,7 +445,11 @@ static bool_ skipString( const char * buf,
                          size_t max )
 {
     bool_ ret = false;
-    size_t i = *start;
+    size_t i;
+
+    assert( ( buf != NULL ) && ( start != NULL ) && ( max > 0U ) );
```
danielsn, line 450: do we really need the `U` on `0`?

libraries/standard/json/src/json.c
```diff
@@ -446,7 +472,7 @@ static bool_ skipString( const char * buf,
                 }
             }
             /* An unescaped control character is not allowed. */
-            else if( ( ( uint8_t ) buf[ i ] ) < 0x20U )
+            else if( ( buf[ i ] >= 0 ) && ( buf[ i ] < 0x20 ) )
```
danielsn, line 475: magic constant

libraries/standard/json/src/json.c
```diff
@@ -481,6 +511,8 @@ static bool_ strnEq( const char * a,
 {
     size_t i;
 
+    assert( ( a != NULL ) && ( b != NULL ) );
```
danielsn, line 514: `assert(__CPROVER_r_ok(a,n) && __CPROVER_r_ok(b,n))` [Or possibly `n+1`, if we have a null terminator]

danielsn, line 514: if `n == 0`, is `NULL` allowed?

libraries/standard/json/src/json.c
```diff
@@ -564,18 +599,28 @@ static bool_ skipDigits( const char * buf,
                          size_t * start,
                          size_t max )
 {
-    size_t i = *start, save = *start;
+    bool_ ret = false;
+    size_t i, save;
+
+    assert( ( buf != NULL ) && ( start != NULL ) && ( max > 0U ) );
```
danielsn, line 605: is `*start > max` allowed?

libraries/standard/json/src/json.c
```diff
@@ -564,18 +599,28 @@ static bool_ skipDigits( const char * buf,
                          size_t * start,
                          size_t max )
 {
-    size_t i = *start, save = *start;
+    bool_ ret = false;
+    size_t i, save;
```
danielsn, line 603: `save` what?

libraries/standard/json/cbmc/proofs/JSON_Validate/Makefile
yourslab: The number of loop iterations depends on the size of the buffer. How do we bound the number of iterations without using a magic number?

libraries/standard/json/cbmc/proofs/JSON_Validate/cbmc-viewer.json
dan4thewin: Ahhh.  added.

libraries/standard/json/cbmc/proofs/JSON_Search/glue.h
dan4thewin: added comment about doc location

libraries/standard/json/cbmc/proofs/JSON_Validate/JSON_Validate_harness.c
dan4thewin: Looks like the template uses `--bounds-check --conversion-check --div-by-zero-check --float-overflow-check --nan-check --pointer-check --pointer-overflow-check --signed-overflow-check --undefined-shift-check --unsigned-overflow-check --unwinding-assertions`  - so keep or remove the _r_ok calls?

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
dan4thewin: There's already a __CPROVER_r_ok on the buffer just after the malloc calls in the harness.  The real skipAnyScalar has a (regular) assert within, and it's proof runs cleanly, by which I reason I do not need one here since I only need to duplicate the index manipulations for the purpose of proving JSON_Search().

libraries/standard/json/cbmc/include/skipGeneric.h
```diff
@@ -0,0 +1,31 @@
+#ifndef SKIPGENERIC_H_
```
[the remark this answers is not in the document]
dan4thewin: added

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
```diff
@@ -0,0 +1,64 @@
...
+ * @file JSON_Search_harness.c
+ * @brief Implements the proof harness for the JSON_Search function.
+ */
+
+#include <stdlib.h>
+#include "json.h"
+
+void harness()
+{
+    char * buf = NULL;
```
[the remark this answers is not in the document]
dan4thewin: I wanted to surface the behaviors so I could understand them better.

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
```diff
@@ -0,0 +1,64 @@
...
+    char * outValue;
+    size_t outValueLength;
+    JSONStatus_t ret;
+
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    if( nondet_bool() )
+    {
+        buf = malloc( max );
+        __CPROVER_assume( __CPROVER_r_ok( buf, max ) );
```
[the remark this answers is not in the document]
dan4thewin: It seemed to me this assume statement expressed the properties I expected from malloc explicitly.

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
```diff
@@ -0,0 +1,64 @@
...
+    char * buf = NULL;
+    size_t max;
+    char * queryKey = NULL;
+    size_t queryKeyLength;
+    char separator;
+    char * outValue;
+    size_t outValueLength;
+    JSONStatus_t ret;
+
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
```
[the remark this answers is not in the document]
dan4thewin: I don't understand the question.

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
```diff
@@ -0,0 +1,64 @@
...
+
+    __CPROVER_assume( queryKeyLength < CBMC_MAX_BUFSIZE );
+
+    if( nondet_bool() )
+    {
+        queryKey = malloc( queryKeyLength );
+        __CPROVER_assume( __CPROVER_r_ok( queryKey, queryKeyLength ) );
+    }
+
+    ret = JSON_Search( buf,
```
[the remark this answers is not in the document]
dan4thewin: We could assert that the value returned is one of the values within the enum return type - but, frankly, I thought that would already be proven.

I can make assertions about outValue and outValueLength in the success case.  Like this:
```
    if( ret == JSONSuccess )
    {
        __CPROVER_assert( ( outValue >= buf ) &&
                          ( ( outValue + outValueLength ) <= ( buf + max ) ),
                          "The output value is a sequence of characters within buf." );
    }
```
What do you think?

libraries/standard/json/cbmc/proofs/JSON_Search/README.md
```diff
@@ -0,0 +1,18 @@
...
+* skipCollection()
+* skipSpace()
+* skipSpaceAndComma()
+* skipString()
+
+To run the proof.
+* Add cbmc, goto-cc, goto-instrument, goto-analyzer, and cbmc-viewer
+  to your path.
+* Run "make".
+* Open html/index.html in a web browser.
```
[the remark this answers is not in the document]
dan4thewin: That's right out of the templates...  I didn't give any thought to reducing boilerplate (except in the Makefiles).  Perhaps we should change the templates first?

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
```diff
@@ -0,0 +1,53 @@
+#include "glue.h"
```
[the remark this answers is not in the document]
dan4thewin: added

libraries/standard/json/cbmc/include/skipGeneric.h
dan4thewin: It returns false.  (at least the modified version I'm adding now clearly returns false for that case...)

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
```diff
@@ -0,0 +1,53 @@
...
+
+bool_ skipAnyScalar( const char * buf,
```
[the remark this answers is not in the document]
dan4thewin: I've added a comment to all the glue files locating the function docs.

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
```diff
@@ -0,0 +1,53 @@
...
+
+    if( ( *start < max ) && ( buf[ *start ] == '"' ) )
+    {
+        min = 2;
+    }
+
+    return skipGeneric( buf, start, max, min );
+}
+
+JSONStatus_t skipCollection( const char * buf,
```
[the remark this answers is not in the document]
dan4thewin: The static version of this function in json.c has a function comment.  I've added a notice.

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
```diff
@@ -0,0 +1,53 @@
...
+                             size_t * start,
+                             size_t max )
+{
+    JSONStatus_t ret;
+
+    __CPROVER_assume( ( ret == JSONPartial ) || ( ret == JSONIllegalDocument ) || ( ret == JSONMaxDepthExceeded ) );
```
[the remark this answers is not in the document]
dan4thewin: By inspection of the original function (also listed in the function doc comment).

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
```diff
@@ -0,0 +1,53 @@
...
+{
+    /* min argument is 1 for a single space. */
+    skipGeneric( buf, start, max, 1 );
+}
+
+bool_ skipSpaceAndComma( const char * buf,
+                         size_t * start,
+                         size_t max )
+{
+    /* min argument is 1 for a single space or comma. */
```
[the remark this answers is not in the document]
dan4thewin: The real function will match 0 or more spaces, followed by a comma, followed by 0 or more spaces.  The spaces are still skipped if there is no comma.

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
```diff
@@ -0,0 +1,53 @@
...
+    return skipGeneric( buf, start, max, 1 );
+}
+
+bool_ skipString( const char * buf,
+                  size_t * start,
+                  size_t max )
+{
+    /* min argument is 2 for an empty double-quoted string, i.e., "". */
```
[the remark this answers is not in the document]
dan4thewin: No, a null terminator within the JSON disqualifies it as a legal JSON doc.

libraries/standard/json/cbmc/proofs/JSON_Search/glue.h
```diff
@@ -0,0 +1,27 @@
+#ifndef GLUE_H_
```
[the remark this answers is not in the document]
dan4thewin: I've added MIT headers to all .h files.

libraries/standard/json/cbmc/proofs/JSON_Validate/JSON_Validate_harness.c
dan4thewin: The only output is the return value, and it must be one of the enums of it's type.

libraries/standard/json/cbmc/proofs/JSON_Validate/Makefile
dan4thewin: 14 is large enough to hold a Unicode high and low surrogate pair escape in quotes, which may be why I chose it to start with, but some proofs had to be adjusted smaller due to memory constraints.

libraries/standard/json/cbmc/proofs/JSON_Validate/Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+UNWINDSET += JSON_Validate.0:$(CBMC_MAX_BUFSIZE)
+
+include ../Makefile-json.common
+
+$(PROOFDIR)/json.c: $(SRCDIR)/libraries/standard/json/src/json.c
+	sed -E '1s/^/#include "glue.h" /; s/^static //; s/(bool_|JSONStatus_t|void) skip(AnyLiteral|Collection|Number|Space|String)\b/&_/' $< > $@
```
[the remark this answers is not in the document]
dan4thewin: It changes the storage class of the functions from static to extern.  It also appends an underscore to a subset of functions so that their replacements will be used instead (the originals are proven in another proof).

Why?  The JSON validator is too big to prove all at once, so I did it in cumulative steps.

libraries/standard/json/cbmc/proofs/JSON_Validate/README.md
```diff
@@ -0,0 +1,17 @@
+JSON_Validate proof
+==============
+
+This directory contains a memory safety proof for JSON_Validate.
```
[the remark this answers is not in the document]
dan4thewin: The combinations of the proofs cover everything.  Most are short, seconds to minutes.  One is longer, a couple dozen minutes, I think.

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
```diff
@@ -0,0 +1,45 @@
+#include "glue.h"
```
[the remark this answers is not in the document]
dan4thewin: "There are only two hard things in Computer Science: cache invalidation and naming things."  - Phil Karlton

I suppose it means I relying on something already proven to permit me to make these replacements...  substituting in a prior conclusion.

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
```diff
@@ -0,0 +1,45 @@
...
+
+bool_ skipAnyLiteral( const char * buf,
+                      size_t * start,
+                      size_t max )
+{
+    /* min argument is 4 for the shortest literal, e.g., true or null. */
```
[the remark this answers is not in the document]
dan4thewin: Yes, unless the JSON standard adds a new shorter literal (vanishingly small probability).

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
```diff
@@ -0,0 +1,45 @@
...
+    return skipGeneric( buf, start, max, 4 );
+}
+
+JSONStatus_t skipCollection( const char * buf,
+                             size_t * start,
+                             size_t max )
+{
+    JSONStatus_t ret;
+
+    __CPROVER_assume( ( ret == JSONPartial ) || ( ret == JSONIllegalDocument ) || ( ret == JSONMaxDepthExceeded ) );
```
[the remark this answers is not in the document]
dan4thewin: Yes, two of the enums relate to parameter checking which is only done in the top JSON_* functions, and the only other enum, JSONNotFound, is for JSON_Search() and search().

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
```diff
@@ -0,0 +1,45 @@
...
+
+    /* min argument is 2 for an empty collection, i.e., {} or []. */
+    return ( skipGeneric( buf, start, max, 2 ) == true ) ? JSONSuccess : ret;
+}
+
+bool_ skipNumber( const char * buf,
```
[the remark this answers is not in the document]
dan4thewin: It would be more DRY, but I think it would complicate running the proof, since the total set of replacements is different for each proof.  This way, the compiler makes sure I get it right.

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
```diff
@@ -0,0 +1,45 @@
...
+                  size_t * start,
+                  size_t max )
+{
+    /* min argument is 1 for a single digit, e.g., 0. */
+    return skipGeneric( buf, start, max, 1 );
```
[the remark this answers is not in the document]
dan4thewin: It's expressing the shortest successful match the original function could return.  The skipGeneric() only uses the value as a lower bound.  A 0 or negative here would break parity with the original function.

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.h
dan4thewin: added

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.h
dan4thewin: added to all .h

libraries/standard/json/cbmc/proofs/Makefile-json.common
dan4thewin: changed

libraries/standard/json/cbmc/proofs/skipAnyLiteral/Makefile
```diff
@@ -0,0 +1,15 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+HARNESS_FILE=skipAnyLiteral_harness
+
+CBMC_MAX_BUFSIZE=6
```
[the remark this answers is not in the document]
dan4thewin: Coverage drops below 100% for smaller buffer values.  I think this is because the longest literal is `false`.

libraries/standard/json/cbmc/proofs/skipAnyLiteral/Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+UNWINDSET += skipAnyLiteral.0:$(CBMC_MAX_BUFSIZE)
+UNWINDSET += skipLiteral.0:$(CBMC_MAX_BUFSIZE)
+UNWINDSET += strnEq.0:$(CBMC_MAX_BUFSIZE)
+
+include ../Makefile-json.common
+
+$(PROOFDIR)/json.c: $(SRCDIR)/libraries/standard/json/src/json.c
+	sed 's/^static //' $< > $@
```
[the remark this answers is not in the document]
dan4thewin: Yes, although there'd need to be changes anyway if there were two files.  A static variable would violate the project's tenets.

libraries/standard/json/cbmc/proofs/skipAnyLiteral/README.md
dan4thewin: trivial runtime, full coverage, no assumptions (that I can think of)

I can add such verbiage to each readme, but it seems to me the report says it better.

libraries/standard/json/cbmc/proofs/skipAnyLiteral/skipAnyLiteral_harness.c
```diff
@@ -0,0 +1,51 @@
...
+ * @file skipAnyLiteral_harness.c
+ * @brief Implements the proof harness for the skipAnyLiteral function.
+ */
+
+#include <stdlib.h>
+
+typedef enum
+{
+    true = 1, false = 0
+} bool_;
```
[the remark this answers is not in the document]
dan4thewin: The type is meant to be private to the json.c file and is only used internally with static functions, thus the duplication here/

libraries/standard/json/cbmc/proofs/skipAnyLiteral/skipAnyLiteral_harness.c
```diff
@@ -0,0 +1,51 @@
...
+
+bool_ skipAnyLiteral( const char * buf,
```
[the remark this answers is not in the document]
dan4thewin: No, it's a static function in json.c (which is no longer static due to the sed script).  In other proofs where I'm replacing this function, I have it in a separate header.

libraries/standard/json/cbmc/proofs/skipAnyLiteral/skipAnyLiteral_harness.c
```diff
@@ -0,0 +1,51 @@
...
+void harness()
+{
+    char * buf;
+    size_t start, max;
+    bool_ ret;
+
+    __CPROVER_assume( max > 0 );
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    buf = malloc( max );
```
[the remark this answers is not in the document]
dan4thewin: No, not for the internal static functions, only for the top level API functions (those that start with JSON_ ).

libraries/standard/json/cbmc/proofs/skipAnyLiteral/skipAnyLiteral_harness.c
```diff
@@ -0,0 +1,51 @@
...
+    __CPROVER_assume( __CPROVER_r_ok( buf, max ) );
+
+    ret = skipAnyLiteral( buf, &start, max );
```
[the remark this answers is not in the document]
dan4thewin: if the function succeeded, then start <= max..  I'll add that to the appropriate harnesses.

libraries/standard/json/cbmc/sources/skipGeneric.c
dan4thewin: sure, I'll add that.

libraries/standard/json/cbmc/proofs/skipCollection/Makefile
```diff
@@ -0,0 +1,18 @@
...
+
+REMOVE_FUNCTION_BODY += strnEq
+UNWINDSET += skipCollection.0:$(CBMC_MAX_BUFSIZE)
+UNWINDSET += skipArrayScalars.0:$(CBMC_MAX_BUFSIZE)
+UNWINDSET += skipObjectScalars.0:$(CBMC_MAX_BUFSIZE)
+
+include ../Makefile-json.common
+
+$(PROOFDIR)/json.c: $(SRCDIR)/libraries/standard/json/src/json.c
+	sed -E '1s/^/#include "glue.h" /; s/^static //; s/(bool_|void) skip(AnyLiteral|Number|Space|SpaceAndComma|String)\b/&_/' $< > $@
```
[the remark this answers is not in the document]
dan4thewin: Same response as before ( libraries/standard/json/cbmc/proofs/JSON_Validate/Makefile )

libraries/standard/json/cbmc/proofs/skipCollection/skipCollection_harness.c
dan4thewin: We're writing ANSI C (i.e., C90).

libraries/standard/json/cbmc/proofs/skipEscape/skipEscape_harness.c
dan4thewin: As with the others, no.  This declaration is for a (normally) static function so there is no header.

libraries/standard/json/cbmc/proofs/skipNumber/skipNumber_harness.c
dan4thewin: Yes, there is an assert.  The top level functions return JSONBadParameter if max is 0.

libraries/standard/json/cbmc/proofs/skipNumber/skipNumber_harness.c
```diff
@@ -0,0 +1,51 @@
...
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    buf = malloc( max );
+    __CPROVER_assume( __CPROVER_r_ok( buf, max ) );
+
+    ret = skipNumber( buf, &start, max );
```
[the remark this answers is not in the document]
dan4thewin: Yes, I've added an assert on the value of start.

libraries/standard/json/cbmc/proofs/skipSpace/skipSpace_harness.c
```diff
@@ -0,0 +1,45 @@
...
+void harness()
+{
+    char * buf;
+    size_t start, max;
+
+    __CPROVER_assume( max > 0 );
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    buf = malloc( max );
+    __CPROVER_assume( __CPROVER_r_ok( buf, max ) );
```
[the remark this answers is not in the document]
dan4thewin: I thought this malloc could return NULL, and that this assume would prevent that.

libraries/standard/json/cbmc/proofs/skipString/Makefile
dan4thewin: For some of the proofs, I had to titrate a value that was large enough for complete coverage without tripping unwinding assertions and small enough to not exhaust memory.  I then tried to use the value in several proofs, but some had to be reduced further due to memory.

libraries/standard/json/cbmc/proofs/skipString/glue.h
```diff
@@ -0,0 +1,18 @@
+#ifndef GLUE_H_
```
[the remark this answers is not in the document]
dan4thewin: Some of the proofs have this glue.h, but only the one copy in the proof directory is used by that proof.

libraries/standard/json/cbmc/proofs/skipUTF8/Makefile
dan4thewin: See my answer here ( libraries/standard/json/cbmc/proofs/skipString/Makefile )

libraries/standard/json/cbmc/sources/skipGeneric.c
```diff
@@ -0,0 +1,33 @@
+#include "skipGeneric.h"
```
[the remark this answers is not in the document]
dan4thewin: There is only the one function which is used by the replacements from the proofs, but is not itself a direct replacement.  Since it is a utility function, I put it here.

libraries/standard/json/cbmc/sources/skipGeneric.c
dan4thewin: This servers to mimic the index manipulations for the replacement functions.  For every replacement function there is a proof which exercises the original function directly including the memory accesses.

libraries/standard/json/cbmc/sources/skipGeneric.c
dan4thewin: I thought this yielded some value x such that min <= x <= max.  You're saying that's not the case?

libraries/standard/json/include/json.h
dan4thewin: There are no iterators for this enum.

libraries/standard/json/src/json.c
dan4thewin: Not in the JSON spec.

libraries/standard/json/src/json.c
dan4thewin: I'd like to keep the proofs separate from the actual library, and not entangle things.

libraries/standard/json/src/json.c
dan4thewin: These are bit masks.  They're actually easier to read as hex (as opposed to something like LOW_ORDER_SEVEN_BITS_HIGH - ick).

libraries/standard/json/src/json.c
dan4thewin: I'll add a #define.

libraries/standard/json/src/json.c
dan4thewin: There is an assert at the beginning of this function, and every one like it, that requires max > 0.

libraries/standard/json/src/json.c
dan4thewin: Added

libraries/standard/json/src/json.c
dan4thewin: No, the value of i is changed and must be given back to the caller.

libraries/standard/json/src/json.c
dan4thewin: MISRA 
Rule 10.4 | Both operands of an operator in which the usual arithmetic conversions are performed shall have the same essential type category

Some places I can use a 0, but some places I need 0U, based on coverity scans.  I did change assignments from 0U to 0, since that generated no MISRA violations.

libraries/standard/json/src/json.c
dan4thewin: changed to use macro

libraries/standard/json/src/json.c
dan4thewin: no

libraries/standard/json/src/json.c
dan4thewin: save the starting index

libraries/standard/json/src/json.c
dan4thewin: yes

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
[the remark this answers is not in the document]
dan4thewin: Experimentation seems to show there is no builtin check on the enums.  I'll add them.

libraries/standard/json/cbmc/proofs/JSON_Validate/JSON_Validate_harness.c
danielsn: Best to remove them.

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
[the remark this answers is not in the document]
danielsn: Whoever set the tag for the repo to Apache2.0 should fix it then https://github.com/aws/aws-iot-device-sdk-embedded-C/tree/development

Screen Shot 2020-09-03 at 9 06 41 PM

I'll open an issue for that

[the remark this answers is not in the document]
danielsn: https://github.com/aws/aws-iot-device-sdk-embedded-C/issues/1195

libraries/standard/json/src/json.c
danielsn: Might be clearer to name it `save_starting_index` then.

libraries/standard/json/cbmc/proofs/JSON_Validate/JSON_Validate_harness.c
danielsn: And/or, add --pointer-primitive-check to the makefile.

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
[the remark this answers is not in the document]
danielsn: There is a builtin check, but it had some false positives if the harness wasn't written carefully, we had to turn it off from the default set of flags.  You might want to try turning it on for your repo, since its all new proofs.
```
 --enum-range-check           checks that all enum type expressions have values in the enum range
```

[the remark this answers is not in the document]
danielsn: But having the explicit check is also good documentation, so I'd do that too.

libraries/standard/json/cbmc/proofs/skipCollection/skipCollection_harness.c
danielsn: Fair enough

libraries/standard/json/cbmc/proofs/skipNumber/skipNumber_harness.c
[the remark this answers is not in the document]
danielsn: nice

libraries/standard/json/cbmc/sources/skipGeneric.c
```diff
@@ -10,22 +31,18 @@ static bool_ skipGeneric( const char * buf,
                           size_t max,
                           size_t min )
 {
-    bool_ ret = nondet_bool() ? true : false;
+    bool_ ret = false;
 
-    if( ret == true )
+    if( nondet_bool() && ( min <= max ) && ( *start <= max ) )
```
danielsn: This is easier to read, thanks!

libraries/standard/json/cbmc/proofs/skipString/glue.h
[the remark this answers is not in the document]
danielsn: That's probably OK then.

libraries/standard/json/cbmc/sources/skipGeneric.c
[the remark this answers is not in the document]
danielsn: OK

libraries/standard/json/cbmc/sources/skipGeneric.c
danielsn: Given the guard on line 38, this should be OK. Would it make sense to make `min <= max` an assert?

libraries/standard/json/cbmc/sources/skipGeneric.c
danielsn: I assume there is a relationship between `buf`, `*start`, `min` and `max`.  For e.g. could you do something like
`assert(__CPROVER_r_ok(buf + *start, max))`?

libraries/standard/json/src/json.c
danielsn: Might be worth calling this something like `json_isspace_` to note that its not the same as the standard `is_space` https://en.cppreference.com/w/cpp/string/byte/isspace

libraries/standard/json/src/json.c
danielsn: In some libraries we've worked on, they've added something like

```C
#ifdef CBMC
#   define PTR_IS_VALID(p,l) (l == 0 || __CPROVER_r_ok(p,l)
#else
#   define PTR_IS_VALID(p,l) (l == 0 || ptr != NULL)
#endif
```

Would something like that make sense here?

libraries/standard/json/src/json.c
danielsn: Fair.

libraries/standard/json/src/json.c
danielsn: I remember those days :)

danielsn (changes requested):

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
```diff
@@ -0,0 +1,79 @@
...
+
+#include <stdlib.h>
+#include "json.h"
+
+void harness()
+{
+    char * buf = NULL;
+    size_t max;
+    char * queryKey = NULL;
+    size_t queryKeyLength;
```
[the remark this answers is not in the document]
dan4thewin: returns JSONBadParameter

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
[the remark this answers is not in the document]
dan4thewin: added

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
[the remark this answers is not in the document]
dan4thewin: added `--pointer-primitive-check` and removed r_ok lines

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
```diff
@@ -0,0 +1,64 @@
...
+        queryKey = malloc( queryKeyLength );
+        __CPROVER_assume( __CPROVER_r_ok( queryKey, queryKeyLength ) );
+    }
+
+    ret = JSON_Search( buf,
+                       max,
+                       queryKey,
+                       queryKeyLength,
+                       separator,
+                       ( nondet_bool() ? &outValue : NULL ),
```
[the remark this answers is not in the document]
dan4thewin: The `&outValue` is in keeping with how the function is usually used.

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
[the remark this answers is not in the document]
dan4thewin: added

libraries/standard/json/cbmc/proofs/JSON_Search/glue.c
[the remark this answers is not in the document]
dan4thewin: added

libraries/standard/json/cbmc/proofs/JSON_Validate/JSON_Validate_harness.c
dan4thewin: added `--pointer-primitive-check`; removed `__CPROVER_assume( __CPROVER_r_ok( buf, max ) );` and the like.

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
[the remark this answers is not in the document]
dan4thewin: macros added

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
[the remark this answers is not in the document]
dan4thewin: I changed my mind.  I removed the glue files and added stubs.

libraries/standard/json/cbmc/proofs/skipAnyLiteral/skipAnyLiteral_harness.c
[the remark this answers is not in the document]
dan4thewin: I created a single header for all the normally static stuff.

libraries/standard/json/cbmc/proofs/skipAnyLiteral/skipAnyLiteral_harness.c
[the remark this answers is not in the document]
dan4thewin: added

libraries/standard/json/cbmc/proofs/skipEscape/skipEscape_harness.c
dan4thewin: And... now there is.

libraries/standard/json/cbmc/proofs/skipSpace/skipSpace_harness.c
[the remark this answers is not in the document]
dan4thewin: I removed the redundant line.  Only the JSON_Search() and JSON_Validate() functions deal with NULLs, and I already have code in the harness to exercise that, so I don't think I'll add these flags.  The library itself does not use malloc.

libraries/standard/json/cbmc/sources/skipGeneric.c
dan4thewin: The `min > max` is one of the false cases, which we want to exercise in addition to the true case.

libraries/standard/json/cbmc/sources/skipGeneric.c
dan4thewin: added

libraries/standard/json/src/json.c
dan4thewin: I'll add a comment.

libraries/standard/json/src/json.c
dan4thewin: If I were a customer reading this code, I think I would be more reassured by the familiar `assert( ptr != NULL)`.

libraries/standard/json/src/json.c
dan4thewin: changed

libraries/standard/json/cbmc/proofs/JSON_Validate/glue.c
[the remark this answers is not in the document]
dan4thewin: Using separate files now, no more glue.*

libraries/standard/json/cbmc/sources/skipGeneric.c
dan4thewin: added

libraries/standard/json/src/json.c
dan4thewin: I'd prefer to keep the proof separate.

libraries/standard/json/src/json.c
dan4thewin: I'd prefer to keep the proof separate.

libraries/standard/json/cbmc/include/json_annex.h
```diff
@@ -0,0 +1,89 @@
...
+    ( ( x == JSONPartial ) || ( x == JSONIllegalDocument ) || ( x == JSONMaxDepthExceeded ) )
+
+/* All possible return values for skipCollection() */
+#define skipCollectionEnum( x )  ( skipCollectionFailEnum( x ) || ( x == JSONSuccess ) )
+
+/* All possible return values for JSON_Validate() */
+#define jsonValidateEnum( x )  ( skipCollectionEnum( x ) || parameterEnum( x ) )
+
+/* All possible return values for JSON_Search() */
+#define jsonSearchEnum( x )  ( jsonValidateEnum( x ) || ( x == JSONNotFound ) )
```
danielsn, line 46: Nice.  Like the recursive grammar :)

libraries/standard/json/cbmc/proofs/JSON_Search/JSON_Search_harness.c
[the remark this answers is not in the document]
danielsn: OK

libraries/standard/json/cbmc/proofs/skipUTF8/skipUTF8_harness.c
```diff
@@ -0,0 +1,54 @@
...
+
+    /* max is the buffer length which must not exceed unwindings. */
+    __CPROVER_assume( max < CBMC_MAX_BUFSIZE );
+
+    /* buf must not be NULL */
+    buf = malloc( max );
+
+    ret = skipUTF8( buf, &start, max );
+
+    __CPROVER_assert( ( ret == true ) || ( ret == false ), "A bool_ value is returned.");
```
danielsn, line 47: ⛏️ this might make sense as a macro

libraries/standard/json/cbmc/proofs/skipString/Makefile
danielsn: Add a comment explaining this in the file.

danielsn (approved):

libraries/standard/json/cbmc/proofs/skipString/Makefile
dan4thewin: added

n9wxu (approved):

libraries/standard/json/cbmc/include/json_annex.h
dan4thewin: Thanks!

libraries/standard/json/cbmc/proofs/skipUTF8/skipUTF8_harness.c
dan4thewin: changed

dan4thewin:
Based on the two approvals received before the base branch changed, I've incorporated the approved changes into the coreJSON repo ( https://github.com/FreeRTOS/coreJSON ).

PR #1170, opened 2020-08-27
Fix integration test from change in MQTT_Publish behavior

aggarw13:
### Background
`MQTT_Connect` API behavior was changed in #1148 to clear state records on a new session establishment with the broker.

### Problem
The update to the behavior of the API exposed an issue with the `test_MQTT_Resend_Unacked_Publish_QoS1` and `test_MQTT_Resend_Unacked_Publish_QoS2` tests that caused them to fail. They were not attempting to resend PUBLISH packet in a persistent session, instead were re-attempting PUBLISH in a clean session.

### Fix
Update the integration tests to establish a persistent session with the broker for resending PUBLISH packets


libraries/standard/mqtt/integration-test/mqtt_system_test.c
```diff
@@ -1148,6 +1148,9 @@ void test_MQTT_Restore_Session_Complete_Incoming_Publish( void )
  */
 void test_MQTT_Resend_Unacked_Publish_QoS1( void )
 {
+    /* Start a persistent session with the broker. */
+    startPersistentSession();
```
muneebahmed10, line 1152: I think you can also replace the below calls to `OpenSSL_Connect()` and `establishMqttSession()` with `resumePersistentSession()` to assert if a session was actually resumed

PR #1175, opened 2020-08-28
Add unit tests for reconnect logic

yourslab:
Complete branch coverage for reconnect logic

Bulk of the code is at `reconnect_utest.c`.


platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+
+/* Called after each test method. */
+void tearDown()
+{
+}
+
+/* Called at the beginning of the whole suite. */
+void suiteSetUp()
+{
+}
```
aggarw13, line 31: If there is no definition for these, these can be removed

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
+#include <string.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include "/usr/include/errno.h"
```
aggarw13, line 4: This is part of the ISO C99 standard library at least (We don't have issues with having tests use C99)
```suggestion
#include <errno.h>
```

platform/posix/transport/utest/reconnect_utest.c
yourslab: Traditionally, we have never removed these even though they are not used. Should we decide to, we can remove all of the unused definitions as part of a separate PR.

platform/posix/transport/utest/reconnect_utest.c
yourslab: Keep in mind that this will require me to `set( CMAKE_C_FLAGS "-isystem /usr/include ${CMAKE_C_FLAGS}" )` because there are errno.h files that seem to just include itself... The right one is at /usr/include, and I'm not sure why the other ones exist. I've also updated the other ones.

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+
+#include "unity.h"
+
+/* Include paths for public enums, structures, and macros. */
+#include "transport_reconnect.h"
+
+#include "mock_unistd_api.h"
+#include "mock_stdio_api.h"
```
aggarw13, line 12: (Nit)
```suggestion
/* Include Mock files. */
#include "mock_unistd_api.h"
#include "mock_stdio_api.h"
```

platform/posix/transport/utest/CMakeLists.txt
```diff
@@ -152,3 +152,33 @@ create_test(${utest_name}
...
+# list the files you would like to test here
+set(real_source_files
+        ${RECONNECT_SOURCES}
+        )
+set(real_name "transport_reconnect_real")
+
+create_real_library(${real_name}
+                    "${real_source_files}"
+                    "${real_include_directories}"
+                    "${mock_name}"
```
aggarw13, line 167: This will include all mock files defined at the top of the file.  It seems like this file has become a mish-mash of multiple test targets. 
As the reconnect utilities may be possibly moved outside of transport (as it has no direct relation to transport layer but is purely a math retry utility), it would be better to separately define its test configuration with the mock dependencies directly related to it

platform/posix/transport/utest/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ include("../transportFilePaths.cmake")
 project ("transport unit test")
 cmake_minimum_required (VERSION 3.13)
 
+set( CMAKE_C_FLAGS "-isystem /usr/include ${CMAKE_C_FLAGS}" )
```
aggarw13, line 5: What is this for?

platform/posix/transport/utest/reconnect_utest.c
aggarw13: Okay, if that's a convention followed, then I am fine with this being here.

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+static void verifyReconnectParamsAfterReset( void )
+{
+    TEST_ASSERT_EQUAL( 0, reconnectParams.attemptsDone );
+    TEST_ASSERT_GREATER_OR_EQUAL_INT32( INITIAL_RECONNECT_BACKOFF_SECONDS,
+                                        reconnectParams.nextJitterMax );
+    TEST_ASSERT_LESS_THAN_INT32( ( INITIAL_RECONNECT_BACKOFF_SECONDS +
+                                   MAX_JITTER_VALUE_SECONDS ),
+                                 reconnectParams.nextJitterMax );
+}
+
```
aggarw13, line 54: Please add a visual break between the utility function and tests
```suggestion
/* ====================  Test Cases ========================== */
```

platform/posix/transport/utest/reconnect_utest.c
yourslab: The `mock_` prefix is sufficient enough for me.

aggarw13: We have a tendency to add comments for include files when it is obvious. Example:
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/development/libraries/standard/http/include/http_client.h#L8-L9

With this convention, I as a reader expect comments as grouping of header includes for showing dependency on modules

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+        {
+            expectedNextJitterMax += expectedNextJitterMax;
+        }
+        else
+        {
+            expectedNextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS;
+        }
+
+        /* Mock sleep, so the test can run faster. */
+        sleep_ExpectAnyArgsAndReturn( 0 );
```
aggarw13, line 94: What about mocking the `rand()` function? This test should ensure that the `sleep()` system API is called with the correct value.

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+ * @brief Test that #Transport_ReconnectBackoffAndSleep is able to double the
+ * max value of the next jitter and return false when all attempts are exhausted.
+ *
+ * @note #Transport_ReconnectParamsReset is expected to be called before
+ * #Transport_ReconnectBackoffAndSleep.
+ */
+void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
+{
+    bool retriesArePending;
+    int32_t expectedNextJitterMax = INITIAL_RECONNECT_BACKOFF_SECONDS;
```
aggarw13, line 75:
```suggestion
    uint32_t expectedNextJitterMax = INITIAL_RECONNECT_BACKOFF_SECONDS;
```

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+
+    Transport_ReconnectParamsReset( &reconnectParams );
+    verifyReconnectParamsAfterReset();
+
+    expectedNextJitterMax = reconnectParams.nextJitterMax;
+
+    while( reconnectParams.attemptsDone < MAX_RECONNECT_ATTEMPTS )
```
aggarw13, line 82: This test code seems to have copied the logic inside the function under test. Instead of using the while loop, there should instead be multiple calls to `Transport_ReconnectBackoffAndSleep` (with a small values for max attempts and max backoff value so that this test can be completed in 3 or so calls)

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+        else
+        {
+            expectedNextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS;
+        }
+
+        /* Mock sleep, so the test can run faster. */
+        sleep_ExpectAnyArgsAndReturn( 0 );
+        retriesArePending = Transport_ReconnectBackoffAndSleep( &reconnectParams );
+        TEST_ASSERT_TRUE( retriesArePending );
+        TEST_ASSERT_EQUAL( expectedNextJitterMax, reconnectParams.nextJitterMax );
```
aggarw13, line 97: Please add assertion on the `reconnectParams.attemptsDone` member after each call to the function under test

platform/posix/transport/utest/reconnect_utest.c
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * @brief Test that the value of the next max jitter is capped at some max threshold.
+ */
+void test_Transport_ReconnectBackoffAndSleep_Jitter_Has_Max( void )
+{
+    bool retriesArePending;
+
+    /* Initialize to 0 attempts, so the max value of next jitter will increase. */
+    reconnectParams.attemptsDone = 0;
```
aggarw13, line 118: Also, add  a test when the initialized `nextJitterMax` value is `MAX_RECONNECT_BACKOFF_SECONDS - 1`

platform/posix/transport/utest/reconnect_utest.c
yourslab: There is no way to change max attempts or backoff value without changing the definition in the header. Also, don't you think  these tests should run independent of the defined value?

platform/posix/utest/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
+project ("posix unit test")
```
aggarw13, line 1: Should this be "Reconnect unit test" specifically as POSIX is an umbrella term for all things platform related

platform/posix/utest/mocks/math_api.h
```diff
@@ -0,0 +1,13 @@
+#ifndef MATH_API_H_
```
aggarw13, line 1: Please add Copyright text to both mocks

platform/posix/utest/reconnect_utest.c
```diff
@@ -8,8 +8,11 @@
 /* Include paths for public enums, structures, and macros. */
 #include "transport_reconnect.h"
 
-#include "mock_unistd_api.h"
-#include "mock_stdio_api.h"
+#include "mock_time_api.h"
+#include "mock_math_api.h"
+
+/* Jitter upon resetting reconnect parameters. */
+#define JITTER    ( MAX_JITTER_VALUE_SECONDS / ( MAX_RECONNECT_ATTEMPTS ) )
```
aggarw13, line 15: What does this value represent exactly? The JITTER values are not supposed to be evenly spread between attempts

platform/posix/utest/reconnect_utest.c
```diff
@@ -45,19 +48,23 @@ int suiteTearDown( int numFailures )
...
-    TEST_ASSERT_LESS_THAN_INT32( ( INITIAL_RECONNECT_BACKOFF_SECONDS +
-                                   MAX_JITTER_VALUE_SECONDS ),
-                                 reconnectParams.nextJitterMax );
+    TEST_ASSERT_EQUAL( JITTER, reconnectParams.nextJitterMax - JITTER );
+    TEST_ASSERT_GREATER_OR_EQUAL_UINT32( INITIAL_RECONNECT_BACKOFF_SECONDS,
+                                         reconnectParams.nextJitterMax );
+    TEST_ASSERT_LESS_THAN_UINT32( ( INITIAL_RECONNECT_BACKOFF_SECONDS +
+                                    MAX_JITTER_VALUE_SECONDS ),
+                                  reconnectParams.nextJitterMax );
+    TEST_ASSERT_LESS_THAN_UINT32( MAX_RECONNECT_ATTEMPTS,
```
aggarw13, line 57: This is already covered on line 50 as the assertion against the value of `0`

platform/posix/utest/reconnect_utest.c
aggarw13: If it is just a test jitter value that is used as the return value by `rand`, it can be renamed as `TEST_RAND_RETURN_VALUE`

platform/posix/utest/CMakeLists.txt
yourslab: Agreed, but this file will also be reused for POSIX clock unit tests.

platform/posix/utest/reconnect_utest.c
```diff
@@ -45,19 +48,23 @@ int suiteTearDown( int numFailures )
...
 }
 
 /**
  * @brief Test that #Transport_ReconnectParamsReset initializes the #nextJitterMax
- * to be equal to the some configurable backoff plus jitter.
+ * to be equal to some configurable backoff plus jitter.
  */
 void test_Transport_ReconnectParamsReset_Sets_Jitter_Correctly( void )
 {
+    rand_ExpectAndReturn( JITTER );
```
aggarw13, line 67: `rand` is not required by `Transport_ReconnectParamsReset`

yourslab: Yes it is...

platform/posix/utest/reconnect_utest.c
```diff
@@ -74,6 +81,7 @@ void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
     bool retriesArePending;
     int32_t expectedNextJitterMax = INITIAL_RECONNECT_BACKOFF_SECONDS;
 
+    rand_ExpectAndReturn( JITTER );
```
aggarw13, line 84: This call is superfluous for the `Transport_ReconnectParamsReset` function

yourslab: So do you want me to mock rand or not?

platform/posix/utest/reconnect_utest.c
```diff
@@ -91,14 +99,18 @@ void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
         }
 
         /* Mock sleep, so the test can run faster. */
+        rand_ExpectAndReturn( 0 );
         sleep_ExpectAnyArgsAndReturn( 0 );
```
aggarw13: This should specify a return value for `rand` and ensure that the `returned value % currentJitterWindow` is passed to `sleep`. Without having these checks in the test, the unit test is not testing the algorithm
```suggestion
        rand_ExpectAndReturn( TEST_RAND_VALUE );
        sleep_ExpectAndReturn( ( TEST_RAND_VALUE % expectedNextJitterMax ), 0   );
```

platform/posix/utest/reconnect_utest.c
```diff
@@ -108,21 +120,49 @@ void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
...
 
     /* Initialize to 0 attempts, so the max value of next jitter will increase. */
-    reconnectParams.attemptsDone = 0;
+    reconnectParams.attemptsDone = 0U;
     reconnectParams.nextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS / 2U;
 
     /* The next max jitter doubles on every retry unless it is greater than or
      * equal to half of the max threshold. */
+    rand_ExpectAndReturn( 0 );
+    sleep_ExpectAnyArgsAndReturn( 0 );
```
aggarw13, line 137: This should have the same logic as I suggested in a previous comment
```suggestion
    rand_ExpectAndReturn( TEST_RAND_VALUE );
    sleep_ExpectAndReturn( ( TEST_RAND_VALUE % (MAX_RECONNECT_BACKOFF_SECONDS / 2U)) , 0 );
```

platform/posix/utest/reconnect_utest.c
```diff
@@ -108,21 +120,49 @@ void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
...
+    retriesArePending = Transport_ReconnectBackoffAndSleep( &reconnectParams );
+    TEST_ASSERT_TRUE( retriesArePending );
+    TEST_ASSERT_EQUAL( reconnectParams.nextJitterMax,
+                       MAX_RECONNECT_BACKOFF_SECONDS );
+    TEST_ASSERT_LESS_THAN_UINT32( MAX_RECONNECT_ATTEMPTS,
+                                  reconnectParams.attemptsDone );
```
aggarw13, line 143: This should test for the exact expected value of `attemptsDone`
```suggestion
    TEST_ASSERT_EQUAL( 1u, reconnectParams.attemptsDone );
```

platform/posix/utest/reconnect_utest.c
```diff
@@ -108,21 +120,49 @@ void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
...
  */
-void test_Transport_ReconnectBackoffAndSleep_Jitter_Has_Max( void )
+void test_Transport_ReconnectBackoffAndSleep_Lower_Bound_Jitter_To_Cap( void )
 {
     bool retriesArePending;
 
     /* Initialize to 0 attempts, so the max value of next jitter will increase. */
-    reconnectParams.attemptsDone = 0;
+    reconnectParams.attemptsDone = 0U;
     reconnectParams.nextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS / 2U;
```
aggarw13, line 132: To truly test the lower bound of capping, this should use the lowest value whose double is greater than the capped value (i.e. `( MAX_RECONNECT_BACKOFF_SECONDS / 2) + 1`  )
```suggestion
    reconnectParams.nextJitterMax = ( MAX_RECONNECT_BACKOFF_SECONDS / 2U ) + 1U;
```

platform/posix/utest/reconnect_utest.c
```diff
@@ -108,21 +120,49 @@ void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
...
+{
+    bool retriesArePending;
+
+    /* Initialize to 0 attempts, so the max value of next jitter will increase. */
+    reconnectParams.attemptsDone = 0U;
+    reconnectParams.nextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS - 1U;
+
+    /* The next max jitter doubles on every retry unless it is greater than or
+     * equal to half of the max threshold. */
+    rand_ExpectAndReturn( 0 );
```
aggarw13, line 160: Same comments for this test about the mock calls to `rand` and `sleep`, and also checking the `reconnectParams.attemptsDone` value after the call

platform/posix/utest/reconnect_utest.c
aggarw13: Oops my bad. In that case, there can be a global mock configuration for the `rand` in `setUp`

platform/posix/utest/reconnect_utest.c
aggarw13: Nevermind, I forgot that `Transport_ReconnectParamsReset` calls `rand`

platform/posix/transport/utest/reconnect_utest.c
aggarw13: Instead of going through all the iterations of reaching the max threshold value, the `reconnectParams` value can be updated to reach the max attempts (after 2 sequential calls to `Transport_ReconnectBackoffAndSleep` ) so that the "max attempts" reached case is simulated.

I prefer that the test code doesn't copy the implementation code!

platform/posix/utest/reconnect_utest.c
```diff
@@ -0,0 +1,168 @@
...
+            expectedNextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS;
+        }
+
+        /* Mock sleep, so the test can run faster. */
+        rand_ExpectAndReturn( 0 );
+        sleep_ExpectAnyArgsAndReturn( 0 );
+        retriesArePending = Transport_ReconnectBackoffAndSleep( &reconnectParams );
+        TEST_ASSERT_TRUE( retriesArePending );
+        TEST_ASSERT_EQUAL( expectedNextJitterMax, reconnectParams.nextJitterMax );
+        TEST_ASSERT_LESS_OR_EQUAL_UINT32( MAX_RECONNECT_ATTEMPTS,
```
aggarw13, line 107: This should check for the exact value of `attemptsDone` after each call

platform/posix/utest/reconnect_utest.c
yourslab: obfuscation

platform/posix/utest/reconnect_utest.c
yourslab: In either case, what you end up testing is the same...

platform/posix/transport/utest/reconnect_utest.c
yourslab: It is simply expecting each iteration of the retry... You simply do 2 sequential calls, and any test that simply increases attempts done will still pass.

platform/posix/utest/reconnect_utest.c
```diff
@@ -98,19 +99,17 @@ void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
             expectedNextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS;
         }
 
-        /* Mock sleep, so the test can run faster. */
-        rand_ExpectAndReturn( 0 );
-        sleep_ExpectAnyArgsAndReturn( 0 );
+        rand_ExpectAndReturn( RAND_RET_VAL );
```
aggarw13, line 102: We should test with a return value greater than the max jitter window so that it is verified that the `% reconnectParams.nextJitterMax` does scale the value down to the range of jitter window.

Thus, I suggest that the `RAND_RET_VAL` is set to a value greater than the `MAX_RECONNECT_BACKOFF_SECONDS`

platform/posix/utest/reconnect_utest.c
```diff
@@ -129,18 +128,17 @@ void test_Transport_ReconnectBackoffAndSleep_Lower_Bound_Jitter_To_Cap( void )
 
     /* Initialize to 0 attempts, so the max value of next jitter will increase. */
     reconnectParams.attemptsDone = 0U;
-    reconnectParams.nextJitterMax = MAX_RECONNECT_BACKOFF_SECONDS / 2U;
+    reconnectParams.nextJitterMax = ( MAX_RECONNECT_BACKOFF_SECONDS / 2U ) + 1;
 
     /* The next max jitter doubles on every retry unless it is greater than or
      * equal to half of the max threshold. */
```
aggarw13: This comment can be updated to be more relevant to the test. It can say something like "The next max jitter window should be set at the cap value, because double of the existing jitter value will cross the `MAX_RECONNECT_BACKOFF_SECONDS`"

Same for the next test case

platform/posix/utest/reconnect_utest.c
```diff
@@ -0,0 +1,165 @@
...
+ * max value of the next jitter and return false when all attempts are exhausted.
+ *
+ * @note #Transport_ReconnectParamsReset is expected to be called before
+ * #Transport_ReconnectBackoffAndSleep.
+ */
+void test_Transport_ReconnectBackoffAndSleep_Succeeds( void )
+{
+    bool retriesArePending;
+    int32_t expectedNextJitterMax = INITIAL_RECONNECT_BACKOFF_SECONDS;
+    int32_t expectedAttemptsDone = 0;
```
aggarw13, line 81:
```suggestion
    uint32_t expectedNextJitterMax = INITIAL_RECONNECT_BACKOFF_SECONDS;
    uint32_t expectedAttemptsDone = 0;
```

platform/posix/utest/reconnect_utest.c
```diff
@@ -0,0 +1,165 @@
...
+    verifyReconnectParamsAfterReset();
+}
+
+/**
+ * @brief Test that the value of the next max jitter has a lower bound that will
+ * then be capped at some max threshold.
+ */
+void test_Transport_ReconnectBackoffAndSleep_Lower_Bound_Jitter_To_Cap( void )
+{
+    bool retriesArePending;
```
aggarw13, line 127: (Nit) This could be a global and reset to false in `setUp`

aggarw13: Similarly, `reconnectParams.attemptsDone = 0U;` initialization can be done in `setUp`

platform/posix/utest/reconnect_utest.c
```diff
@@ -0,0 +1,172 @@
...
+/* ========================================================================== */
+
+/**
+ * @brief A helper method to make assertions on the #TransportReconnectParams_t
+ * object upon calling #Transport_ReconnectParamsReset.
+ */
+static void verifyReconnectParamsAfterReset( void )
+{
+    TEST_ASSERT_EQUAL( 0, reconnectParams.attemptsDone );
+    TEST_ASSERT_EQUAL( EXPECTED_NEXT_JITTER, reconnectParams.nextJitterMax - EXPECTED_NEXT_JITTER );
```
aggarw13, line 55: Suggestion to improve readability of this assert
```suggestion
    TEST_ASSERT_EQUAL( INITIAL_RECONNECT_BACKOFF_SECONDS + EXPECTED_NEXT_JITTER , reconnectParams.nextJitterMax );
```

PR #1186, opened 2020-09-02
Poll the socket before send/recv in plaintext transport implementation

yourslab:
The following changes are made for this PR:
- First, we call `select` on the socket before calling `send` or `recv`.
- `strerror` is used instead to print the error from `errno`.
- Unit tests are updated for coverage


leegeth (dismissed):

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -126,9 +126,11 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
     {
         logTransportError( errno );
 
-        /* Check if it was time out. Note that for most POSIX implementations,
-         * EAGAIN and EWOULDBLOCK have the same value. */
-        if( ( errno == EAGAIN ) || ( errno == EWOULDBLOCK ) )
+        /* Check if there was a timeout.
+         * Note that for most POSIX implementations, EAGAIN and EWOULDBLOCK have
+         * the same value. Therefore, branch coverage is suppressed to achieve
+         * full branch coverage for all UNIX systems. */
```
yourslab, line 132: I added some comments about why we are suppressing branch coverage. Let me know if you would like me to remove.

leegeth (dismissed):

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,72 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
 {
-    int32_t bytesReceived = 0;
+    int32_t bytesReceived = -1, selectStatus = -1, getTimeoutStatus = -1;
+    struct timeval transportTimeout, selectTimeout;
+    socklen_t transportTimeoutLen;
+    fd_set readfds;
 
-    bytesReceived = ( int32_t ) recv( pNetworkContext->socketDescriptor, pBuffer, bytesToRecv, 0 );
+    assert( pNetworkContext != NULL );
+    assert( pBuffer != NULL );
```
leegeth, line 78: can you also add an assert to make sure bytesToRecv >0

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,72 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
+    {
+        selectTimeout.tv_sec = transportTimeout.tv_sec;
+        selectTimeout.tv_usec = transportTimeout.tv_usec;
+    }
+
+    /* Check if there is data to read from the socket. */
+    selectStatus = select( pNetworkContext->socketDescriptor + 1,
+                           &readfds,
+                           NULL,
+                           &readfds,
```
leegeth, line 105: What is the purpose of this argument?

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,72 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
+    {
+        /* Make #select return immediately if getting the timeout failed. */
+        selectTimeout.tv_sec = 0;
+        selectTimeout.tv_usec = 0;
+    }
+    else
+    {
+        selectTimeout.tv_sec = transportTimeout.tv_sec;
+        selectTimeout.tv_usec = transportTimeout.tv_usec;
+    }
```
leegeth, line 99: can we use the transportTimeout itself? Do we need selectTimeout?

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,72 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
+                                          0 );
+    }
+    else if( selectStatus < 0 )
+    {
+        /* An error occurred while polling. */
+        bytesReceived = -1;
+    }
+    else
+    {
+        /* No data to receive at this time. */
```
leegeth, line 122: I think it is better to mention that this is the timeout case.

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -147,21 +145,72 @@ int32_t Plaintext_Send( const NetworkContext_t * pNetworkContext,
...
-        logTransportError( errno );
+        /* Make #select return immediately if getting the timeout failed. */
+        selectTimeout.tv_sec = 0;
+        selectTimeout.tv_usec = 0;
+    }
+    else
+    {
+        selectTimeout.tv_sec = transportTimeout.tv_sec;
+        selectTimeout.tv_usec = transportTimeout.tv_usec;
+    }
```
leegeth, line 175: All comments in Plaintext_Recv apply in this function as well.

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -147,21 +145,72 @@ int32_t Plaintext_Send( const NetworkContext_t * pNetworkContext,
...
+
+    /* Check if data can be written to the socket. */
+    selectStatus = select( pNetworkContext->socketDescriptor + 1,
+                           NULL,
+                           &writefds,
+                           &writefds,
+                           &selectTimeout );
```
leegeth, line 182: Just one thought. We currently have 2 timeouts. One set for the sockets and one for select. I think this can result in a worst case timeout of 2*timeout. I think it would be better to use only one. What are your thoughts?

yourslab: Back when we didn't use `select`, suppose the socket received data a millisecond before the timeout ended, then the data will likely not be received with the remaining time left, resulting in a timeout error.

In that sense, I think this new implementation is better. The timeout to receive data is no longer affected by how long the socket has been waiting on new data.

platform/posix/transport/src/plaintext_posix.c
yourslab: Removing `selectTimeout`

platform/posix/transport/src/plaintext_posix.c
yourslab: That is for `exceptfds`, which activates events from exceptions. Although, we probably don't want to send/recv when there are exceptions. I am setting it to `NULL`.

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,68 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
+    struct timeval transportTimeout;
+    socklen_t transportTimeoutLen;
+    fd_set readfds;
+
+    assert( pNetworkContext != NULL );
+    assert( pBuffer != NULL );
+    assert( bytesToRecv > 0 );
+
+    transportTimeoutLen = ( socklen_t ) sizeof( transportTimeout );
+    getTimeoutStatus = getsockopt( pNetworkContext->socketDescriptor,
```
aggarw13, line 82: Can you add a comment about what this call is doing? Is it obtaining the receive timeout value?

aggarw13 (changes requested):

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,68 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
+    {
+        transportTimeout.tv_sec = 0;
+        transportTimeout.tv_usec = 0;
+    }
 
-    bytesReceived = ( int32_t ) recv( pNetworkContext->socketDescriptor, pBuffer, bytesToRecv, 0 );
+    FD_ZERO( &readfds );
+    FD_SET( pNetworkContext->socketDescriptor, &readfds );
+    /* Check if there is data to read from the socket. */
+    selectStatus = select( pNetworkContext->socketDescriptor + 1,
```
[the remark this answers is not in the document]
yourslab: You are right that it is not necessarily required for a blocking socket, but I wanted to also handle the case where the user flags the socket as non-blocking.

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,68 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
+    getTimeoutStatus = getsockopt( pNetworkContext->socketDescriptor,
+                                   SOL_SOCKET,
+                                   SO_RCVTIMEO,
+                                   &transportTimeout,
+                                   &transportTimeoutLen );
+
+    /* Make #select return immediately if getting the timeout failed. */
+    if( getTimeoutStatus < 0 )
+    {
+        transportTimeout.tv_sec = 0;
```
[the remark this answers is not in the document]
yourslab: Yes, we still need to know whether there is data to send or receive using the `select` call.

leegeth (approved):

leegeth (dismissed):

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -113,30 +69,70 @@ int32_t Plaintext_Recv( const NetworkContext_t * pNetworkContext,
...
+        /* The socket is available for receiving data. */
+        bytesReceived = ( int32_t ) recv( pNetworkContext->socketDescriptor,
+                                          pBuffer,
+                                          bytesToRecv,
+                                          0 );
+    }
+    else if( selectStatus < 0 )
+    {
+        /* An error occurred while polling. */
+        bytesReceived = -1;
```
aggarw13, line 116: I'm concerned that with this change, an error return value will purely depend on the call to `select` instead of the `recv` call. 
Are there some error return values of `recv` that will be ignored with this approach?

Looking at the Linux man page of [`select`](https://man7.org/linux/man-pages/man2/select.2.html), I am hoping that the `EBADF` return value could catch errors encountered with _open_ file descriptors.
image

platform/posix/transport/utest/mocks/select_api.h
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @file select_api.h
+ * @brief This file is used to generate mocks for functions used from <sys/select.h>.
+ * Mocking select.h itself causes several errors from parsing its macros.
+ */
+
+#ifndef SELECT_API_H_
+#define SELECT_API_H_
+
+#include <sys/select.h>
```
aggarw13, line 31: Is this include required when the `select` declaration has already been provided in this file?

platform/posix/transport/utest/plaintext_utest.c
```diff
@@ -190,20 +222,62 @@ void test_Plaintext_Send_Network_Error( void )
...
+    select_ExpectAnyArgsAndReturn( 1 );
+    send_ExpectAnyArgsAndReturn( 0 );
+    bytesSent = Plaintext_Send( &networkContext,
+                                plaintextBuffer,
+                                BYTES_TO_SEND );
+    TEST_ASSERT_EQUAL( SEND_RECV_ERROR, bytesSent );
+}
+
+/**
+ * @brief Test that #Plaintext_Send returns 0 bytes when the socket has no activity.
```
aggarw13, line 255: This can instead be called the send timeout test

yourslab: Lovely

platform/posix/transport/utest/mocks/select_api.h
aggarw13: How does inclusion of the system file not causing conflicting declaration error for `select`?

aggarw13 (dismissed):

aggarw13 (approved):

n9wxu (approved):

PR #1187, opened 2020-09-02
Replace MQTT files in C-SDK with submodule to coreMQTT

aggarw13:
Use submodule pointer to **coreMQTT** repositriy, the Single Source of Truth (SSoT) for MQTT stack


demos/mqtt/mqtt_demo_plaintext/Makefile
```diff
@@ -1,11 +1,11 @@
 CC = gcc
 
 INCLUDE_DIRS = -I . \
-               -I ../../../libraries/standard/mqtt/include \
+               -I ../../../libraries/standard/coreMQTT/include \
```
leegeth, line 4: This is a good find. I don't think we need this file anymore. Do you know of a reason why we would need this file. This is out of the scope of this PR, but we may need to just remove it.

demos/shadow/shadow_demo_main/core_mqtt_config.h
```diff
@@ -0,0 +1,78 @@
+/*
```
leegeth, line 1: Why is this file added as a new file? Wasn't there a mqtt_config.h already?

demos/mqtt/mqtt_demo_plaintext/Makefile
aggarw13: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1182
Had created a PR to remove this unused file once I found it in change set 😃

demos/shadow/shadow_demo_main/core_mqtt_config.h
aggarw13: Nope, it wasn't present before in the Shadow demo. It was instead utilizing the mqtt_config.h file from the connection sharing demo due to configuration logic in the demo's CMakeLists.txt which I have removed as part of this PR. 
Removing the include path config logic from connection sharing demo's CMakeLists.txt exposed that the shadow demo lacked the mqtt config file

leegeth: Ok. Thanks. That was a good find and a fix.

cobusve (approved):

PR #1208, opened 2020-09-09
Update retry_utils.h to use enum instead of bool

sarenameas:
*Description of changes:*
* bool was used as the return value for the function RetryUtils_BackoffAndSleep(). RetryUtils_BackoffAndSleep() does not return a predicate and instead we replace the return with an enum of statuses.
* Update everywhere in the code this change affects.


platform/include/retry_utils.h
```diff
@@ -62,6 +53,15 @@
  */
 #define MAX_JITTER_VALUE_SECONDS         5U
 
+/**
+ * @brief Status for @ref RetryUtils_BackoffAndSleep.
+ */
+typedef enum RetryUtilsStatus
```
sarenameas, line 59: Enum update

dan4thewin (dismissed):

yourslab (approved):

yourslab (approved):

dan4thewin (approved):

PR #1233, opened 2020-09-11
Randomize client IDs in all demos to avoid collision

aggarw13:
Equivalent of #1214 for all MQTT demos


demos/mqtt/mqtt_demo_serializer/mqtt_demo_serializer.c
```diff
@@ -851,10 +890,20 @@ int main( int argc,
...
     ( void ) argv;
 
+    /* Get current time to seed pseudo random number generator.
+     * Note: The random number generator is used for adding randomness
+     * to the client ID strings to avoid client identifier collision
+     * when connecting to the MQTT broker. */
+    ( void ) clock_gettime( CLOCK_REALTIME, &tp );
+
+    /* Seed pseudo random number generator with nanoseconds. */
+    srand( tp.tv_nsec );
```
muneebahmed10, line 905: Is `srand(time(NULL));` not sufficient?

demos/mqtt/mqtt_demo_serializer/mqtt_demo_serializer.c
```diff
@@ -389,11 +413,26 @@ static int createMQTTConnectionWithBroker( NetworkContext_t * pNetworkContext,
      * gets disconnected. */
     mqttConnectInfo.cleanSession = true;
 
-    /* The client identifier is used to uniquely identify this MQTT client to
+    /* Generate a random number to prefix to the client ID string.
+     * Note: The randomization of the client ID is used to avoid
+     * client identifier collisions when connecting to the MQTT broker. */
```
muneebahmed10, line 418: I don't think we need a "Note" to call attention to something, since that's what comments are for anyway
```suggestion
    /* To avoid client identifier collisions when connecting to the MQTT broker,
     * we generate a random number to prefix to the client ID string. */
```

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -82,60 +84,74 @@
...
  * @brief Length of MQTT server host name.
  */
-#define BROKER_ENDPOINT_LENGTH              ( ( uint16_t ) ( sizeof( BROKER_ENDPOINT ) - 1 ) )
+#define BROKER_ENDPOINT_LENGTH                  ( ( uint16_t ) ( sizeof( BROKER_ENDPOINT ) - 1 ) )
 
 /**
  * @brief Length of client identifier.
  */
-#define CLIENT_IDENTIFIER_LENGTH            ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) - 1 ) )
+#define CLIENT_IDENTIFIER_LENGTH                ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) + MAX_RAND_NUMBER_DIGITS_FOR_CLIENT_ID - 1 ) )
```
muneebahmed10, line 109: This doesn't seem entirely accurate anymore
```suggestion
#define MAX_CLIENT_IDENTIFIER_LENGTH            ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) + MAX_RAND_NUMBER_DIGITS_FOR_CLIENT_ID - 1 ) )
```
or even:
```suggestion
#define MAX_CLIENT_IDENTIFIER_SIZE              ( ( uint16_t ) ( sizeof( CLIENT_IDENTIFIER ) + MAX_RAND_NUMBER_DIGITS_FOR_CLIENT_ID ) )
```
and that way, you won't need the `+ 1` in the array below

aggarw13: The +1 is added for the NULL character

demos/mqtt/mqtt_demo_serializer/mqtt_demo_serializer.c
aggarw13: Removed

demos/mqtt/mqtt_demo_serializer/mqtt_demo_serializer.c
aggarw13: Sure.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
muneebahmed10: I know, isn't it only used in one place?
```
char clientIdBuffer[ MAX_CLIENT_IDENTIFIER_LENGTH + 1u ];
```
Removing the `- 1` would allow you to remove the `+ 1` from the array.

aggarw13: Updated the macro name to `CLIENT_IDENTIFIER_BUFFER_SIZE`

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -869,9 +869,8 @@ static int establishMqttSession( MQTTContext_t * pMqttContext,
     MQTTStatus_t mqttStatus;
     MQTTConnectInfo_t connectInfo;
 
-    /* Buffer for storing client ID with random number.
-     * Note: The "+ 1U" is for the NULL character.*/
-    char clientIdBuffer[ MAX_CLIENT_IDENTIFIER_LENGTH + 1u ];
+    /* Buffer for storing client ID with random number. */
+    char clientIdBuffer[ MAX_CLIENT_IDENTIFIER_LENGTH ];
```
muneebahmed10:
```suggestion
    char clientIdBuffer[ CLIENT_IDENTIFIER_BUFFER_SIZE ];
```

muneebahmed10 (approved):

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
aggarw13: Was changed in https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1233/commits/ba24c0887ff413356328e21eaae9f0b3e3958a30

n9wxu (approved):

leegeth (approved):

demos/mqtt/mqtt_demo_serializer/mqtt_demo_serializer.c
cobusve: On my embedded device tv_nsec is always 0 because my tick rate is one tick per ms. Would that mean every device always gets the same ID?

nateglims:
For CI, we would actually prefer to be able to pass a client ID in without warnings. It's hard to correlate logs without controlling it, and it would give us parity with IoT Core based demos (we name the thing in CI).

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -82,60 +84,75 @@
     #define NETWORK_BUFFER_SIZE    ( 1024U )
 #endif
 
+/**
+ * @brief The largest random number to use in client identifier.
+ *
+ * @note Random number is added to MQTT client identifier to avoid client
+ * identifier collisions while connecting to MQTT broker.
+ */
+#define MAX_RAND_NUMBER_FOR_CLIENT_ID           ( 999u )
```
nateglims, line 93: Seeding considerations aside, 1/1000 isn't particularly good. If a customer were to take this demo and run with it as an application, at about 37-38 demos, there is a 50% chance of a collision.

Not being deterministic across starts of the same binary also seems problematic to a lot of customer use cases.

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -976,11 +996,26 @@ static int establishMqttSession( MQTTContext_t * pMqttContext,
...
-    connectInfo.clientIdentifierLength = CLIENT_IDENTIFIER_LENGTH;
+    connectInfo.clientIdentifierLength =
+        snprintf( clientIdBuffer, sizeof( clientIdBuffer ),
+                  "%d%s",
+                  randomNumForClientId,
+                  CLIENT_IDENTIFIER );
+    connectInfo.pClientIdentifier = clientIdBuffer;
+
+    LogDebug( ( "Created randomized client ID for MQTT connection: "
+                "ClientID={%.*s}",
```
nateglims, line 1016: This format appears to be dependent on the underlying log buffer size, the size of the input, and the previous code preventing malformed data being passed. Is the underlying log implementation defensive against buffer size issues?

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -976,11 +996,26 @@ static int establishMqttSession( MQTTContext_t * pMqttContext,
      * reestablish a session which was already present. */
     connectInfo.cleanSession = createCleanSession;
 
-    /* The client identifier is used to uniquely identify this MQTT client to
+    /* Generate a random number to prefix to the client ID string.
+     * The randomization of the client ID is used to avoid
+     * client identifier collisions when connecting to the MQTT broker. */
+    randomNumForClientId = ( rand() % ( MAX_RAND_NUMBER_FOR_CLIENT_ID + 1u ) );
```
nateglims, line 1002: In the man page, it is mentioned that the lower bits of rand are less random than the higher bits on many platforms. This suggests collisions may be further degraded in different environments.

nateglims (changes requested):
We need to reconsider this. I believe it may degrade both CI and customer use cases.

cobusve:
> For CI, we would actually prefer to be able to pass a client ID in without warnings. It's hard to correlate logs without controlling it, and it would give us parity with IoT Core based demos (we name the thing in CI).

I agree I think that randomizing ID's like this is a mistake, identity should be provided not randomly assigned. There are many problems with randomizing identities and this correllation is only one of them, and the way this is implemented also does not scale.

cobusve (changes requested):
I think we need to re-think this.

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
cobusve: Yes, this is called the "brithday attack", well spotted!

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -1324,6 +1360,12 @@ int main( int argc,
      * done only once in this demo. */
     returnStatus = initializeMqtt( &mqttContext, &networkContext );
 
+    /* Seed pseudo random number generator with current time.
+     * Note: The random number generator is used for adding randomness
+     * to the client ID strings to avoid client identifier collision
+     * when connecting to the MQTT broker. */
+    srand( time( NULL ) );
```
cobusve, line 1367: I think in CI it is very likely that we start 2 instances within 1 second, which will likely cause a lot of collisions.

n9wxu (changes requested):
Remove Random and use CLI to allow different client ID's.  This is also better for CI as it allows the automation to create explicit traceable clientID's and see the specific results on the server.

aggarw13:
Closing the PR based on feedback about using command line supplied client IDs instead of randomization in demo application

aggarw13:
> Remove Random and use CLI to allow different client ID's. This is also better for CI as it allows the automation to create explicit traceable clientID's and see the specific results on the server.

Providing client IDs from command line is an already supported feature for most demos (barring MQTT Subscription Manager and Shadow currently.) So, will update the 2 demos to also support command-line overriding, and then, we should be good with demos with the approach.

PR #1240, opened 2020-09-12
Use .dox extension for doxygen files rather than .txt

yourslab:
As suggested by @n9wxu, `.txt` files are too general to be used for doxygen documentation. It would be best to use the standard extension: `.dox`.


aggarw13:
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/master/docs/doxygen/config.doxyfile#L883
The `.txt` entry can be removed as well from the config file

aggarw13 (approved):

n9wxu (approved):

Issue #1247, opened 2020-09-14
mbedtls_ssl_write error not passed to aws_iot_sdk


raulgotor:
Hi, 

I'm working with the aws-iot-device-sdk in an ESP32. I have the system working for several hours with constant publishing every 5 seconds. At some random point, I start to get the error:

`E (41075) aws_iot: failed! mbedtls_ssl_write returned -0x50`,

In file network_mbedtls_wrapper.c, I found a comment that says:

`/* All other negative return values (besides MBEDTLS_ERR_SSL_WANT_READ && MBEDTLS_ERR_SSL_WANT_WRITE) indicate connection needs to be reset. Will be caught in ping request so ignored here */`

Problem is that publishing is failing, and this error doesn't seem to call the `disconnectHandler` member of the `IoT_Client_Init_Params` struct. 
I'm trying to catch the error at application level to manually restart the connection, but I've found out that at some point in the nesting functions, the error that `mbedtls_ssl_write` returns is lost, specifically at `aws_iot_mqtt_internal_send_packet`:

```c
IoT_Error_t aws_iot_mqtt_internal_send_packet(AWS_IoT_Client *pClient, size_t length, Timer *pTimer) {
 
        //..Ommited function fragment...//
 
	while(sent < length && !has_timer_expired(pTimer)) {
		rc = pClient->networkStack.write(&(pClient->networkStack),    // pClient→networkStack.write points to "mbedtls_ssl_write"
						 &pClient->clientData.writeBuf[sent],      // error last time seen is here
						 (length - sent),
						 pTimer,
						 &sentLen);
		if(SUCCESS != rc) {
			/* there was an error writing the data */
			break;
		}
		sent += sentLen;
	}
      
 
#ifdef _ENABLE_THREAD_SUPPORT_
 
        //I have _ENABLE_THREAD_SUPPORT_, so rc error value from "pClient->networkStack.write(&(pClient→networkStack)" is 
        //replaced by the result of "aws_iot_mqtt_client_unlock_mutex" which is 0
	rc = aws_iot_mqtt_client_unlock_mutex(pClient, &(pClient->clientData.tls_write_mutex));
	if(SUCCESS != rc) {
		FUNC_EXIT_RC(rc);
	}
#endif
 
	if(sent == length) {              //lengths here are different thought
		/* record the fact that we have successfully sent the packet */
		//countdown_sec(&c->pingTimer, c->clientData.keepAliveInterval);
[...]

PR #1297, opened 2020-09-22
MQTT v3: Only return a failure to unlock the mutex, if the write was successful.

sarenameas:
Similar to what aws_iot_mqtt_internal_cycle_read() does, I update aws_iot_mqtt_internal_send_packet() to return a failure on the mutex unlock only if the network write was successful. The network write failure has precedence over the possible and extremely rare mutex unlock failure.

Before this change, the success on the mutex unlock swallowed any possible network write failures.

This fixes issue: https://github.com/aws/aws-iot-device-sdk-embedded-C/issues/1247


sarenameas:
Travis was successful: https://travis-ci.org/github/aws/aws-iot-device-sdk-embedded-C/builds/729202109

aggarg (approved):

PR #1308, opened 2020-09-26
Add doxygen warning checks.

sarenameas:
### The Doxygen warning check do these steps:
- Recursively clone this repo.
- Run the Doxygen command on each of the library spoke repos. The warnings for the libraries are already checked in the library repo's themselves, so there is no need to check the output.
- Run the Doxygen command on this repo. Check the output for warnings. 

### Test:
- See commit 1a2b8db for the test that adds a warning in the CSDK doxygen. The check fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,7 @@ jobs:
           do
             cd $LIB
             echo $PWD
+            if [ ! -d docs/doxygen/ ]; then exit 0; fi
             doxygen docs/doxygen/config.doxyfile
             cd $ROOT
           done
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,7 +130,10 @@ jobs:
           for lib in $(find libraries/aws libraries/standard -maxdepth 1 -mindepth 1)
           do
             cd $lib
-            if [ ! -d docs/doxygen/ ]; then exit 0; fi
+            if [ ! -d docs/doxygen/ ]; then
+              echo "$lib has no Doxygen documentation."
+              exit 0
+            fi
             doxygen docs/doxygen/config.doxyfile
             cd $ROOT
           done
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,7 +115,7 @@ jobs:
             exit 0
           fi
   doxygen:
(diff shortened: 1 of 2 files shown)


nateglims (dismissed):

.github/workflows/ci.yml
```diff
@@ -114,3 +114,27 @@ jobs:
...
+        run: |
+          wget -qO- "http://doxygen.nl/files/doxygen-1.8.20.linux.bin.tar.gz" | sudo tar --strip-components=1 -xz -C /usr/local
+          sudo apt-get install -y libclang-9-dev
+      - name: Run Doxygen on all library submodules
+        run: |
+          ROOT=$PWD
+          for lib in $(find libraries/aws libraries/standard -maxdepth 1 -mindepth 1)
+          do
+            cd $lib
+            if [ ! -d docs/doxygen/ ]; then exit 0; fi
```
gkwicker, line 133: Should we log a message if there is nothing in `docs/doxygen`?

.github/workflows/ci.yml
```diff
@@ -114,3 +114,27 @@ jobs:
...
+            doxygen docs/doxygen/config.doxyfile
+            cd $ROOT
+          done
+      - name: Run Doxygen on CSDK and verify stdout is empty
+        run: |
+          doxygen docs/doxygen/config.doxyfile 2>&1 | tee doxyoutput.txt
+          if [[ "$(wc -c < doxyoutput.txt | bc)" = "0" ]]; then exit 0; else exit 1; fi
```
gkwicker, line 140: If `doxyoutput.txt` isn't empty, do we want to log it so that the developer can see the errors?

gkwicker (dismissed):
Approved, but there are a couple of questions in the comments.

.github/workflows/ci.yml
sarenameas: yes those are shown with tee above (which outputs to both the stdout and the text file).

sarenameas: image

.github/workflows/ci.yml
sarenameas: I can see that being useful I will add that

gkwicker (approved):

nateglims (approved):

PR #1309, opened 2020-09-30
Fix Doxygen Job in GitHub Actions

aggarw13:
Fix step in Doxygen job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,7 +138,7 @@ jobs:
           ROOT=$PWD
           for lib in $(find libraries/aws libraries/standard -maxdepth 1 -mindepth 1)
           do
-            cd $lib
+            cd $ROOT/$lib
             if [ ! -d docs/doxygen/ ]; then
               echo "$lib has no Doxygen documentation."
             else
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -138,13 +138,12 @@ jobs:
...
-            cd $lib
+            cd $ROOT/$lib
             if [ ! -d docs/doxygen/ ]; then
               echo "$lib has no Doxygen documentation."
-              exit 0
+            else
+              doxygen docs/doxygen/config.doxyfile
             fi
-            doxygen docs/doxygen/config.doxyfile
-            cd $ROOT
```
sarenameas, line 147: If we don;t change back to the ROOT, then the next step, "Run Doxygen on CSDK and verify stdout is empty", won't be running doxygen on the CSDK, it will be running in a library repo?

aggarw13: Good point. Fixed the `cd` problem and tested it with a deliberate doxygen issue in 7bc84c0

sarenameas: Also thank you for catching my mistake.

sarenameas (approved):

dan4thewin (approved):

PR #1312, opened 2020-10-02
Modify CBMC proofs to make assumptions about malloc explicit.

markrtuttle:
Some proofs assume that some pointers returned by malloc are not
NULL. This patch modifies those proofs to make these assumptions
explicit with __CPROVER_assume(pointer != NULL) for all such
pointers.

### DIFF
--- a/cbmc/proofs/mqtt_state.c
+++ b/cbmc/proofs/mqtt_state.c
@@ -36,7 +36,10 @@ void * malloc_can_fail( size_t size )
 
 void * allocate_opaque_type()
 {
-    return malloc( 1 ); /* consider using malloc(0) */
+    void * ptr = malloc( 1 );
+
+    __CPROVER_assume( ptr != NULL );
+    return ptr;
 }
 
 /****************************************************************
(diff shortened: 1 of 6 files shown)


n9wxu (approved):

n9wxu (approved):

cbmc/proofs/mqtt_state.c
```diff
@@ -36,7 +36,10 @@ void * malloc_can_fail( size_t size )
 
 void * allocate_opaque_type()
 {
-    return malloc( 1 ); /* consider using malloc(0) */
+    void * ptr = malloc( 1 );
+
+    __CPROVER_assume( ptr != NULL );
```
feliperodri, line 41: Is it safe to always assume this is different than null? Could you add a comment?

feliperodri (approved):
LGTM. I just left one suggestion for an extra comment.

abhidixi11:
Merged the code as CBMC proofs will not pass ThreadSanitizer test.

PR #1313, opened 2020-10-06
Add S3 Download Demo

sukhmanm:
*Description of changes:* This demo uses range requests to download an object file in chunks from a S3 bucket, using a pre-signed URL. 

- adding `getUrlPath()` and `getUrlAddress()` to common HTTP demo utilities, as they are used in the S3 Upload demo as well.
- adding python script `presigned_urls_gen.py` to common/src folder (used to generate pre-signed URLs).

By submitting this pull request, I confirm that my contribution is made under the terms of the MIT license.

PR #1318, opened 2020-10-09
Add S3 Upload Demo

sukhmanm:
*Description of changes:* This demo uses pre-signed URLs to upload a file to a specified S3 object, and then verify that the file was uploaded correctly by cross-checking the file sizes.

This PR parallels that for the [S3 download demo](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1313), and shares the files in `demos/http/common`.


demos/http/http_demo_s3_upload/demo_config.h
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * @brief Presigned PUT URL generated by python script in common/src/presigned_urls_gen.py
+ *
+ * @note This script requires AWS CLI to be configured. For instructions, see
+ * https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html
+ *
+ * Run this script and paste the output value IOT_DEMO_HTTPS_PRESIGNED_PUT_URL into
+ * S3_PRESIGNED_URL below.
+ */
+#define S3_PRESIGNED_PUT_URL    ""
```
yourslab, line 94: If you intend the user to fill this up, put it inside the comment block, then have a
```
#ifndef S3_PRESIGNED_URL
    #error "whatever"
#endif 
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+
+    /* Initialize TLS credentials. */
+    ( void ) memset( &opensslCredentials, 0, sizeof( opensslCredentials ) );
+    opensslCredentials.pRootCaPath = ROOT_CA_CERT_PATH;
+
+    /* ALPN is required when communicating to AWS IoT Core over port 443 through HTTP. */
+    if( HTTPS_PORT == 443 )
+    {
+        opensslCredentials.pAlpnProtos = IOT_CORE_ALPN_PROTOCOL_NAME;
+        opensslCredentials.alpnProtosLen = strlen( IOT_CORE_ALPN_PROTOCOL_NAME );
```
yourslab, line 207: We are discouraged from using `strlen` in demos. You are allowed to use it though in unit tests and integration tests. I know it's a hassle, but you can also create a IOT_CORE_ALPN_PROTOCOL_NAME_LENGTH macro.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    /* Retrieve the address location and length from the S3_PRESIGNED_PUT_URL. */
+    httpStatus = getUrlAddress( S3_PRESIGNED_PUT_URL,
+                                strlen( S3_PRESIGNED_PUT_URL ),
```
yourslab, line 212: Same here and whereever it's used

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+    if( opensslStatus == OPENSSL_SUCCESS )
+    {
+        returnStatus = EXIT_SUCCESS;
+    }
+    else
+    {
+        returnStatus = EXIT_FAILURE;
+    }
+
+    return returnStatus;
```
yourslab, line 254:
```suggestion
    return opensslStatus == OPENSSL_SUCCESS ? EXIT_SUCCESS : EXIT_FAILURE;
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+    char * contentRangeValStr = NULL;
+    size_t contentRangeValStrLength = 0;
+
+    /* Initialize the request object. */
+    requestInfo.pHost = pHost;
+    requestInfo.hostLen = strlen( pHost );
+    requestInfo.method = pMethod;
+    requestInfo.methodLen = strlen( pMethod );
+    requestInfo.pPath = pPath;
+    requestInfo.pathLen = strlen( pPath );
```
yourslab, line 280: I would recommend a structure for a string and its length since we cannot use `strlen`:
```
/**
 * @brief A pair containing a path string of the URI and its length.
 */
typedef struct httpPathStrings
{
    const char * httpPath;
    size_t httpPathLength;
} httpPathStrings_t;

/**
 * @brief A pair containing an HTTP method string and its length.
 */
typedef struct httpMethodStrings
{
    const char * httpMethod;
    size_t httpMethodLength;
} httpMethodStrings_t;
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+                                     &pPath,
+                                     &pathLen );
+
+            if( httpStatus != HTTP_SUCCESS )
+            {
+                LogError( ( "An error occurred in getUrlPath() on URL %s. Error code: %d",
+                            S3_PRESIGNED_GET_URL,
+                            httpStatus ) );
+                returnStatus = EXIT_FAILURE;
+            }
```
yourslab, line 579: I think it would be best to leave the error logging up to the function responsible for doing that work. You can then use a ternary operator here to make it look cleaner.

demos/http/http_demo_s3_upload/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+# Demo target.
+add_executable(
+    ${DEMO_NAME}
+        "${DEMO_NAME}.c"
+        "../common/src/http_demo_utils.c"
+        ${HTTP_SOURCES}
+        ${HTTP_THIRD_PARTY_SOURCES}
+)
+
```
yourslab, line 14: I think it would be helpful to download the Baltimore Cybertrust root CA similar to how we do it for the Amazon Root CA.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+
+/* Pointer to the data to upload.*/
+#ifndef DEMO_HTTP_UPLOAD_DATA
+    #define DEMO_HTTP_UPLOAD_DATA    "Hello World!"
+#endif
+
+/* The size of the data in bytes to upload. */
+#ifndef DEMO_HTTP_UPLOAD_DATA_SIZE
+    #define DEMO_HTTP_UPLOAD_DATA_SIZE    ( sizeof( DEMO_HTTP_UPLOAD_DATA ) - 1 )
+#endif
```
yourslab, line 106: hmm why do we need to wrap these in #ifndef?

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+ */
+static HTTPRequestInfo_t requestInfo;
+
+/**
+ * @brief Represents a response returned from an HTTP server.
+ */
+static HTTPResponse_t response;
+
+/* The host address string extracted from the pre-signed URL. */
+static char serverHost;
```
yourslab, line 135: Missing @brief tag

yourslab: Also, shouldn't this be a `char *`?

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+        if( pFileSizeStr == NULL )
+        {
+            LogError( ( "'/' not present in Content-Range header value: %s.",
+                        contentRangeValStr ) );
+        }
+
+        pFileSizeStr += sizeof( char );
+        *pFileSize = ( size_t ) strtoul( pFileSizeStr, NULL, 10 );
+
+        if( ( *pFileSize == 0 ) || ( *pFileSize == UINT32_MAX ) )
```
yourslab, line 359: What is wrong if the file size is equal to `UINT32_MAX`? Should we instead have a mechanism for detecting overflow?

demos/http/common/src/http_demo_utils.c
```diff
@@ -67,3 +70,109 @@ int32_t connectToServerWithBackoffRetries( TransportConnect_t connectFunction,
...
 }
+
+/*-----------------------------------------------------------*/
+
+HTTPStatus_t getUrlPath( const char * pUrl,
+                         size_t urlLen,
+                         const char ** pPath,
+                         size_t * pPathLen )
+{
+    /* http-parser status. Initialized to 0 to signify success. */
```
yourslab, line 81: Some reviewers may encourage you to initialize with failure

demos/http/common/include/http_demo_utils.h
```diff
@@ -49,3 +52,60 @@ typedef int32_t ( * TransportConnect_t )( NetworkContext_t * pNetworkContext );
...
+ * This function retrieves the location and length of the path from within the
+ * input the URL. The query is not included in the length returned.
+ *
+ * The URL MUST start with "http://" or "https://" to find the path.
+ *
+ * For example, if the URL is:
+ * pUrl = "https://www.somewebsite.com/path/to/item.txt?optionalquery=stuff"
+ *
+ * *pPath = "/path/to/item.txt?optionalquery=stuff"
+ * *pPathLen = 17
```
yourslab, line 68: For code examples, we normally use 
```
@code{c}
    ...
@endcode
```

demos/http/common/include/http_demo_utils.h
```diff
@@ -49,3 +52,60 @@ typedef int32_t ( * TransportConnect_t )( NetworkContext_t * pNetworkContext );
...
+ *
+ * @param[in] pUrl - URL string to parse.
+ * @param[in] urlLen - The length of the URL string input.
+ * @param[out] pPath - pointer within input url that the path starts at.
+ * @param[out] pPathLen - Length of the path.
```
yourslab, line 73: I don't think we ever add dashes for @param. Does it have an effect on the doxygen output? A space should be fine.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: Sounds good -- changed all instances of `strlen`.

demos/http/http_demo_s3_upload/demo_config.h
sarenameas: So these should be in comments then like the ROOT_CA_CERT_PATH

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
+/*
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
sarenameas, line 2: Maybe we can add `AWS IoT Device SDK for Embedded C V202009.00` to make is easier to change the version during release.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: Please change to length

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: Or in other words checkout what the other HTTP demos do

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,626 @@
...
+ * @return EXIT_FAILURE on failure; EXIT_SUCCESS on success.
+ */
+static int uploadS3ObjectFile( const TransportInterface_t * pTransportInterface,
+                               const char * pHost,
+                               const char * pMethod,
+                               const char * pPath );
+
+/*-----------------------------------------------------------*/
+
+static int32_t connectToServer( NetworkContext_t * pNetworkContext )
```
sarenameas, line 182: It's philosophical I was thinking more and it's weird to return EXIT_SUCCESS and EXIT_FAILURE from private functions. EXIT_SUCCESS and EXIT_FAILURE were meant for returning to the host OS that the program was successful. Perhaps you can follow the jobs demo and use bool or return the OpensslStatus_t.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: The standard states that it will return the maximum long integer if it detects overflow.

demos/http/http_demo_s3_upload/CMakeLists.txt
sukhmanm: Yes, good point. Updated the CMake to download the Baltimore root CA alongside Amazon CA1.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: It was supposed to be a char[] to extract the host name from the address pointer. Updated to fix that.

demos/http/common/include/http_demo_utils.h
sukhmanm: I think it was more an example of an input/output case rather than a code example, but I tried to make the comments a little clearer.

demos/http/common/include/http_demo_utils.h
sukhmanm: No, no effect. I had taken this bit from AFR and didn't realize the dashes caused inconsistencies -- removed them, thanks.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: That makes sense, although in this case `connectToServer` is a paramater for the common util `connectToServerWithBackoffRetries`, so it matches the `TransportConnect_t` typedef signature (which returns EXIT_SUCCESS/FAILURE). So I suppose `http_demo_utils.h` along with all the files that already use it would have to be updated -- is that something that should be done here?

demos/http/http_demo_s3_upload/demo_config.h
sukhmanm: I had put empty strings for CI purposes, but intend on removing them once they are configured (unless there's a better workaround?)

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: Sounds good -- moved all logging tasks to their respective functions.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: If each private function has a specific use case with no string parameters (in other words it knows exactly which string to reference, similar to mutual auth, and dissimilar to other demos where `sendHttpRequest` is used to send an array of different requests), and each string is only referenced once, are defining structs necessary? I had attempted to implement them but hadn't found much value in doing so, but if you think it would still be beneficial to show their use, I can definitely add them in.

demos/http/http_demo_s3_upload/demo_config.h
sarenameas: For see, I see you already catch this as a command line parameter: https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1318/files#diff-224512c4a5413ed757baf615abc71943a8b8dba575a61dddd52e6a1781f3c657R48

And I see you recently put them in `#ifndef` and `#endif`s, in this file; this works for CI. Although, instead of these blank default strings, we should have demo fail the compilation if the presigned URL is not present. That way the customer is alerted in the `#error` to use the python script to generate them.  Putting the presigned URL configs in the comments, CI can still configure on the command line build the definitions.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+    NetworkContext_t networkContext;
+
+    ( void ) argc;
+    ( void ) argv;
+
+    for( ; ; )
+    {
+        LogInfo( ( "HTTPS Client Synchronous S3 upload demo using pre-signed PUT URL:\n%s", S3_PRESIGNED_PUT_URL ) );
+
+        /**************************** Parse Signed URL. ******************************/
```
sarenameas, line 530:
```suggestion
        /**************************** Parse Presigned URL. ******************************/
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: Oh yea sorry, I see that this path is the rest of the presigned URL.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: Can we move this code to where it is needed in the `/*** Upload S3 Object File. ***/` section, like how you have the URL partsing for the GET url in the `/*** Verify S3 Object File Upload. ***/` section?

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+            returnStatus = ( httpStatus == HTTP_SUCCESS ) ? EXIT_SUCCESS : EXIT_FAILURE;
+        }
+
+        /**************************** Connect. ******************************/
+
+        /* Establish TLS connection on top of TCP connection using OpenSSL. */
+        if( returnStatus == EXIT_SUCCESS )
+        {
+            /* Attempt to connect to the HTTP server. If connection fails, retry after
+             * a timeout. Timeout value will be exponentially increased till the maximum
```
sarenameas, line 549:
```suggestion
             * a timeout. The timeout value will be exponentially increased till the maximum
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+             * attempts are reached or maximum timeout value is reached. The function
+             * returns EXIT_FAILURE if the TCP connection cannot be established to
+             * broker after configured number of attempts. */
```
sarenameas, line 552:
```suggestion
             * broker after the configured number of attempts. */
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+            returnStatus = connectToServerWithBackoffRetries( connectToServer,
+                                                              &networkContext );
+
+            if( returnStatus == EXIT_FAILURE )
+            {
+                /* Log error to indicate connection failure after all
```
sarenameas, line 558:
```suggestion
                /* Log an error to indicate connection failure after all
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+    else
+    {
+        LogError( ( "An error occurred in getting the file size from %s. Error=%s.",
+                    serverHost,
+                    HTTPClient_strerror( httpStatus ) ) );
+    }
+
+    return httpStatus;
+}
+
```
sarenameas, line 415:
```suggestion

/*-----------------------------------------------------------*/

```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+static HTTPStatus_t verifyS3ObjectFileSize( size_t * pFileSize,
+                                            const TransportInterface_t * pTransportInterface,
+                                            const char * pPath );
+
+/**
+ * @brief Send an HTTP PUT request based on a specified path to upload a file,
+ * then print the response received from the server.
+ *
+ * @param[in] pTransportInterface The transport interface for making network calls.
+ * @param[in] pPath The Request-URI to the objects of interest.
```
sarenameas, line 204:
```suggestion
 * @param[in] pPath The NULL-terminated Request-URI to the object of interest.
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+ * @return EXIT_FAILURE on failure; EXIT_SUCCESS on successful connection.
+ */
+static int32_t connectToServer( NetworkContext_t * pNetworkContext );
+
+/**
+ * @brief Retrieve and verify the size of the S3 object that is specified in pPath.
+ *
+ * @param[out] pFileSize The size of the S3 object.
+ * @param[in] pTransportInterface The transport interface for making network calls.
+ * @param[in] pPath The Request-URI to the objects of interest.
```
sarenameas, line 193:
```suggestion
 * @param[in] pPath The NULL-terminated Request-URI to the object of interest.
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: Oh yes, I see. I don't want you to change code you didn't write. I will write a SIM thanks.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+                   response.pBody ) );
+    }
+    else
+    {
+        LogError( ( "An error occurred in uploading the file."
+                    "Failed to send HTTP PUT request to %s%s: Error=%s.",
+                    serverHost, pPath, HTTPClient_strerror( httpStatus ) ) );
+    }
+
+    return httpStatus;
```
sarenameas, line 487: Can we return success only for a 200 response in the response.statusCode?

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+    }
+    else
+    {
+        LogError( ( "Failed to initialize HTTP request headers: Error=%s.",
+                    HTTPClient_strerror( httpStatus ) ) );
+    }
+
+    if( httpStatus == HTTP_SUCCESS )
+    {
+        LogInfo( ( "Received HTTP response from %s%s...",
```
sarenameas, line 469: Can we please change these printings of the response headers and body to LogDebug and then have a LogInfo() about the success from a 200 response?

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: I am suggesting this because this demo is targeted to S3 and its status codes are well known to us. We don't want to continue the demo if S3 doesn't accept the object.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+    if( httpStatus == HTTP_SUCCESS )
+    {
+        httpStatus = HTTPClient_InitializeRequestHeaders( &requestHeaders,
+                                                          &requestInfo );
+    }
+
+    if( httpStatus == HTTP_SUCCESS )
+    {
+        LogInfo( ( "Uploading file..." ) );
+        LogInfo( ( "Request Headers:\n%.*s",
```
sarenameas, line 451: Please log the request headers as LogDebug.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+        LogInfo( ( "Response Headers:\n%.*s",
+                   ( int32_t ) response.headersLen,
+                   response.pHeaders ) );
+        LogInfo( ( "Response Status:\n%u",
+                   response.statusCode ) );
+        LogInfo( ( "Response Body:\n%.*s\n",
+                   ( int32_t ) response.bodyLen,
+                   response.pBody ) );
+
+        httpStatus = HTTPClient_ReadHeader( &response,
```
sarenameas, line 349: I think we should check that 206 was found in response.statusCode before continuing in this function. If 206 was not returned from the server, then we cannot parse for the contentRangeValStr anyways.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+            httpStatus = verifyS3ObjectFileSize( &fileSize,
+                                                 &transportInterface,
+                                                 pPath );
+
+            returnStatus = ( httpStatus == HTTP_SUCCESS ) ? EXIT_SUCCESS : EXIT_FAILURE;
+        }
+
+        if( returnStatus == EXIT_SUCCESS )
+        {
+            /* Log message indicating an iteration completed successfully. */
```
sarenameas, line 610:
```suggestion
            /* Log a message indicating an iteration completed successfully. */
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+            LogInfo( ( "Demo completed successfully." ) );
+        }
+
+        /************************** Disconnect. *****************************/
+
+        /* End TLS session, then close TCP connection. */
```
sarenameas, line 616:
```suggestion
        /* End the TLS session, then close the TCP connection. */
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+                                             pPath );
+
+            returnStatus = ( httpStatus == HTTP_SUCCESS ) ? EXIT_SUCCESS : EXIT_FAILURE;
+        }
+
+        /******************* Verify S3 Object File Upload. ********************/
+
+        if( returnStatus == EXIT_SUCCESS )
+        {
+            /* Retrieve the path location from S3_PRESIGNED_GET_URL. This function returns the length of the path
```
sarenameas, line 588: Do you have your line width marker in your text editor set to 80?

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+ *
+ */
+int main( int argc,
+          char ** argv )
+{
+    /* Return value of main. */
+    int32_t returnStatus = EXIT_SUCCESS;
+    /* HTTPS Client library return status. */
+    HTTPStatus_t httpStatus = HTTP_SUCCESS;
+    /* The length of the path within the pre-signed URL. */
```
sarenameas, line 513:
```suggestion
    /* The length of the path within the pre-signed URL. This variable is defined
     * in order to store the length returned from parsing the URL, but it is
     * unused. The path used for the requests in this demo needs all the query
     * information following the location of the object, to the end of the S3 presigned
     * URL. */
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: Ah yes I had meant to move it when I moved `getUrlAddress `to `connectToServer` but missed it. Thanks for pointing it out.

demos/http/common/src/presigned_urls_gen.py
```diff
@@ -0,0 +1,38 @@
...
+    # Generate the URL to get 'key-name' from 'bucket-name'
+    for method in client_method_dict.keys():
+        url = s3.generate_presigned_url(
+            ClientMethod=client_method_dict[method],
+            Params={
+                'Bucket': bucket_name,
+                'Key': key_name
+            }
+        )
+        print('#define IOT_DEMO_HTTPS_PRESIGNED_' + method + '_URL' + '    ' + '\"' + url + '\"')
```
sarenameas, line 27:
```suggestion
        print('#define S3_PRESIGNED_' + method + '_URL' + '    ' + '\"' + url + '\"')
```

demos/http/common/src/presigned_urls_gen.py
```diff
@@ -0,0 +1,38 @@
...
+def main():
+    parser = argparse.ArgumentParser(description='S3 Presigned URL Generator. See README.md')
+    parser.add_argument('--bucket', action='store', required=True, dest='bucket_name', help='The name of the S3 bucket of interest.')
+    parser.add_argument('--key', action='store', required=True, dest='key_name', help='The name of the S3 Object in the bucket. This is referred to as a \'key\'')
+    args = parser.parse_args()
+
+    get_presigned_urls(args.bucket_name, args.key_name)
+
+if __name__ == '__main__':  # pragma: no cover
+    main()
```
sarenameas, line 38: Can you please run this on this script to fix any formatting. 
```shell
black -l 120 presigned_urls_gen.py
```

demos/http/common/src/presigned_urls_gen.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/python
+
+import boto3
+from botocore.client import Config
+import argparse
+
+def get_presigned_urls(bucket_name, key_name):
```
sarenameas, line 7: Let's increase the quality of our python scripts and add function doc strings. 

```suggestion
def get_presigned_urls(bucket_name, key_name) -> None:
    """
    Prints the presigned GET and PUT URLs assigned to the demo specific C macros,
    for the given object key in the given S3 bucket. The URLs are presigned with AWS's
    Signature Version 4.
    Args:
        bucket_name (str): S3 bucket
        key_name (str):  S3 object key
    """
        """
```

demos/http/common/src/presigned_urls_gen.py
```diff
@@ -0,0 +1,38 @@
...
+        url = s3.generate_presigned_url(
+            ClientMethod=client_method_dict[method],
+            Params={
+                'Bucket': bucket_name,
+                'Key': key_name
+            }
+        )
+        print('#define IOT_DEMO_HTTPS_PRESIGNED_' + method + '_URL' + '    ' + '\"' + url + '\"')
+
+def main():
```
sarenameas, line 29:
```suggestion
def main():
    """
    Generate demo C macro strings, on the console, for the input S3 bucket and object key.
    """
```

demos/http/http_demo_s3_upload/demo_config.h
```diff
@@ -0,0 +1,114 @@
...
+
+
+/**
+ * @brief Presigned GET URL generated by python script in common/src/presigned_urls_gen.py
+ *
+ * @note This script requires AWS CLI to be configured. For instructions, see
+ * https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html
+ *
+ * Run this script and paste the output value IOT_DEMO_HTTPS_PRESIGNED_GET_URL into
+ * S3_PRESIGNED_GET_URL below.
```
sarenameas, line 84:
```suggestion
 * Run this script and paste the output value into S3_PRESIGNED_GET_URL below.
```

demos/http/http_demo_s3_upload/demo_config.h
```diff
@@ -0,0 +1,114 @@
...
+
+
+/**
+ * @brief Presigned PUT URL generated by python script in common/src/presigned_urls_gen.py
+ *
+ * @note This script requires AWS CLI to be configured. For instructions, see
+ * https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html
+ *
+ * Run this script and paste the output value IOT_DEMO_HTTPS_PRESIGNED_PUT_URL into
+ * S3_PRESIGNED_PUT_URL below.
```
sarenameas, line 98:
```suggestion
 * Run this script and paste the S3_PRESIGNED_PUT_URL below.
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+ */
+static int32_t connectToServer( NetworkContext_t * pNetworkContext );
+
+/**
+ * @brief Retrieve and verify the size of the S3 object that is specified in pPath.
+ *
+ * @param[out] pFileSize The size of the S3 object.
+ * @param[in] pTransportInterface The transport interface for making network calls.
+ * @param[in] pPath The Request-URI to the objects of interest.
+ */
```
sarenameas, line 194: Please add `@return` for consistency.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,624 @@
...
+                                            const TransportInterface_t * pTransportInterface,
+                                            const char * pPath );
+
+/**
+ * @brief Send an HTTP PUT request based on a specified path to upload a file,
+ * then print the response received from the server.
+ *
+ * @param[in] pTransportInterface The transport interface for making network calls.
+ * @param[in] pPath The Request-URI to the objects of interest.
+ */
```
sarenameas, line 205: Please add `@return` docs here to thank you

demos/http/common/src/presigned_urls_gen.py
sukhmanm: Sure, sounds good. Will go through and update the script.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: Ah I clearly did not, but I do now :)

demos/http/http_demo_s3_upload/demo_config.h
sukhmanm: Sorry what I meant was I was trying to avoid build check failures as I already had the `#ifndef/#endif`s with `#error` statements present in `http_demo_s3_upload.c` and didn't want to hit the errors, but I now realize I should probably update `ci.yml` instead.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: Please note this that this parameter is expected to be NULL terminated.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sarenameas: Please note that this parameter is expected to be NULL terminated.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,561 @@
...
+ *
+ * @return The status of the file upload using a PUT request to the server: true
+ * on success, false on failure.
+ */
+static bool uploadS3ObjectFile( const TransportInterface_t * pTransportInterface,
+                                const char * pPath );
+
+/*-----------------------------------------------------------*/
+
+static int32_t connectToServer( NetworkContext_t * pNetworkContext )
```
sarenameas, line 223: You can separate this out with these parameters.
```suggestion
static int32_t connectToServer( NetworkContext_t * pNetworkContext,
                                const char* rootCaCertPath,
                                const char* alpnProtocol,
                                size_t alpnProtocolLen,
                                uint16_t port,
                                const char* serverHost,
                                size_t serverHostLength )
```

You can parse for the server host address using getUrlAddress() in the demo code, like you did for the download demo.

demos/http/http_demo_s3_upload/demo_config.h
```diff
@@ -0,0 +1,117 @@
...
+ * @note This script requires AWS CLI to be configured. For instructions, see
+ * https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html
+ *
+ * Run this script and paste the output S3_PRESIGNED_GET_URL below.
+ *
+ * #define S3_PRESIGNED_GET_URL         "...insert here..."
+ */
+#ifndef S3_PRESIGNED_GET_URL
+    #define S3_PRESIGNED_GET_URL    ""
+#endif
```
sarenameas, line 90: Can we delete these since they are in the comments.
```suggestion
```

demos/http/http_demo_s3_upload/demo_config.h
```diff
@@ -0,0 +1,117 @@
...
+ * @note This script requires AWS CLI to be configured. For instructions, see
+ * https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html
+ *
+ * Run this script and paste the output S3_PRESIGNED_PUT_URL below.
+ *
+ * #define S3_PRESIGNED_PUT_URL         "...insert here..."
+ */
+#ifndef S3_PRESIGNED_PUT_URL
+    #define S3_PRESIGNED_PUT_URL    ""
+#endif
```
sarenameas, line 105:
```suggestion
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,561 @@
...
+                                    const char * pPath );
+
+/**
+ * @brief Send an HTTP PUT request based on a specified path to upload a file,
+ * then print the response received from the server.
+ *
+ * @param[in] pTransportInterface The transport interface for making network
+ * calls.
+ * @param[in] pPath The Request-URI to the objects of interest. This string
+ * should be null-terminated.
```
sarenameas, line 213:
```suggestion
 * must be null-terminated.
```

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,561 @@
...
+static int32_t connectToServer( NetworkContext_t * pNetworkContext );
+
+/**
+ * @brief Retrieve and verify the size of the S3 object that is specified in
+ * pPath.
+ *
+ * @param[in] pTransportInterface The transport interface for making network
+ * calls.
+ * @param[in] pPath The Request-URI to the objects of interest. This string
+ * should be null-terminated.
```
sarenameas, line 198:
```suggestion
 * must be null-terminated.
```
Because you are using strlen

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,561 @@
...
+ *
+ * @note When using ALPN, port 443 must be used to connect to AWS IoT Core.
+ */
+#define IOT_CORE_ALPN_PROTOCOL_NAME               "\x0ex-amzn-http-ca"
+
+/**
+ * @brief Length of ALPN protocol name to be sent as part of the ClientHello
+ * message.
+ */
+#define IOT_CORE_ALPN_PROTOCOL_NAME_LENGTH        ( sizeof( IOT_CORE_ALPN_PROTOCOL_NAME ) - 1 )
```
sarenameas, line 103: I just realized we shouldnt have this becayse we are not talking to AWS IoT.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
```diff
@@ -0,0 +1,561 @@
...
+    ( void ) memset( &opensslCredentials, 0, sizeof( opensslCredentials ) );
+    opensslCredentials.pRootCaPath = ROOT_CA_CERT_PATH;
+
+    /* ALPN is required when communicating to AWS IoT Core over port 443 through
+     * HTTP. */
+    if( HTTPS_PORT == 443 )
+    {
+        opensslCredentials.pAlpnProtos = IOT_CORE_ALPN_PROTOCOL_NAME;
+        opensslCredentials.alpnProtosLen = IOT_CORE_ALPN_PROTOCOL_NAME_LENGTH;
+    }
```
sarenameas, line 248: I don't think that S3 needs an ALPN protocol.

demos/http/http_demo_s3_upload/http_demo_s3_upload.c
sukhmanm: Removed it, thanks

sarenameas (approved):

n9wxu (approved):

yinlonglyl:
@sukhmanm Is it possible to provide a S3 upload demo without using presiged URL, use the same way as the 'demos\http\http_demo_s3_download' done?

PR #1326, opened 2020-10-21
Remove const qualifier from send/recv in transport implementations and MQTT system test

yourslab:
Submodule pointers are updated to follow changes from changes from FreeRTOS/coreMQTT#86, FreeRTOS/coreHTTP#45, & FreeRTOS/FreeRTOS@398abba. The const qualifier is removed from send/recv because there are transport implementations that require a member of the network context to be modified such as in the case of mbedtls.

### DIFF
--- a/platform/posix/transport/include/openssl_posix.h
+++ b/platform/posix/transport/include/openssl_posix.h
@@ -177,7 +177,7 @@ OpensslStatus_t Openssl_Disconnect( const NetworkContext_t * pNetworkContext );
  *
  * @return Number of bytes received if successful; negative value on error.
  */
-int32_t Openssl_Recv( const NetworkContext_t * pNetworkContext,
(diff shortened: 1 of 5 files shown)


platform/posix/transport/include/openssl_posix.h
```diff
@@ -177,7 +177,7 @@ OpensslStatus_t Openssl_Disconnect( const NetworkContext_t * pNetworkContext );
  *
  * @return Number of bytes received if successful; negative value on error.
  */
-int32_t Openssl_Recv( const NetworkContext_t * pNetworkContext,
+int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
```
sarenameas, line 180: When MISRA is in CI we can make an annotation later for this I suppose.

sarenameas (approved):

dan4thewin (approved):

PR #1327, opened 2020-10-22
Make CBMC run with all property checking flags.

markrtuttle:
This pull request adds to the list of undefined behaviors that CBMC checks for, and it runs CBMC with the new model of malloc that allows malloc either to fail and return NULL or return a valid pointer.


feliperodri (approved):
LGTM

abhidixi11 (approved):

PR #1330, opened 2020-10-26
TRANSPORT_SEND_RECEIVE_TIMEOUT_MS was too short

gregbreen:
*Issue #, if available:*

None.

*Description of changes:*

In trying to connect to **eu-west-1**, from Australia, the default timeouts in some of the demos were too short. Suffering SSL handshake failures. MQTT mutual auth, shadows and subscription manager. Changed the other demos just for consistency. 

Cost me an hour to figure it out (assumed it was a certificate issue at first). Hopefully this saves someone else an hour. I think there's not much downside to a longer timeout in demo code.

### DIFF
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -167,7 +167,7 @@
 /**
  * @brief Transport timeout in milliseconds for transport send and receive.
  */
-#define TRANSPORT_SEND_RECV_TIMEOUT_MS      ( 200 )
+#define TRANSPORT_SEND_RECV_TIMEOUT_MS      ( 500 )
 
 /*-----------------------------------------------------------*/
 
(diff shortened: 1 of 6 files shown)


abhidixi11:
Hello @gregbreen  thanks for the PR, we should update timeout, at least make them consistent for all the demos.  Is 1000 MS the minimum timeout value that you have observed ? It seems bit high. Thanks.

gregbreen:
Hi @abhidixi11 . It can probably be lower than 1000. I picked 1000 because TRANSPORT_SEND_RECV_TIMEOUT_MS is defined in 19 files (not just the 6 files I changed), and it's either 1000 or 5000 in each of the other 13 files. (e,g. 1000 in HTTP demo plaintext, S3 download, S3 upload, etc).

The shadow demo didn't work at 100 or 200, but did work at 300. I didn't test 300 many times, so it could have been marginal. Aust to eu-west-1 might also not be worst case latency. So I think 500 is probably a good choice on the low side.

Please advise your preferred number. I'll update the PR and squash it back to one commit.

gregbreen:
@abhidixi11 are we able to square this one away?

abhidixi11:
Apologies for delay. Let's make it 500 for all demos.  That's reasonable. Thanks!

gregbreen:
All changed to 500ms.

abhidixi11 (approved):

gkwicker (approved):

PR #1341, opened 2020-10-28
Add region parameter to presigned_urls_gen.py and fix formatting

sukhmanm:
*Description of changes:* 
- If the region specified in a pre-signed URL doesn't match that of the target S3 bucket, the server will reject the request. Adding an optional --region parameter to presigned_urls_gen.py to avoid AWS CLI configuration conflicts/changes. 

### DIFF
--- a/demos/http/common/src/presigned_urls_gen.py
+++ b/demos/http/common/src/presigned_urls_gen.py
@@ -5,27 +5,30 @@ from botocore.client import Config
 import argparse
 
 
-def get_presigned_urls(bucket_name, key_name) -> None:
+def get_presigned_urls(bucket_name, key_name, region_name) -> None:
     """
-    Prints the presigned GET and PUT URLs assigned to the demo specific C macros,
-    for the given object key in the given S3 bucket. The URLs are presigned with AWS's
-    Signature Version 4.
+    Prints the presigned GET and PUT URLs assigned to the demo specific C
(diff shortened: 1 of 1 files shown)


sarenameas (dismissed):

demos/http/common/src/http_demo_utils.c
```diff
@@ -345,15 +345,17 @@ bool getS3ObjectFileSize( size_t * pFileSize,
     }
     else
     {
-        LogError( ( "Failed to read Content-Range header from HTTP response: Error=%s.",
-                    HTTPClient_strerror( httpStatus ) ) );
+        LogError( ( "Failed to read Content-Range header from HTTP response. Error: %s",
+                    ( httpStatus == HTTPSuccess ) ?
+                    "Unsuccessful response status code." : HTTPClient_strerror( httpStatus ) ) );
```
dan4thewin, line 350: This confuses me.  The log says "Unsuccessful" when the value was HTTPSuccess?

sukhmanm: The intention was to differentiate between a non-206 status code error and an HTTP error, as they both could lead to this statement, but after some more thought I agree that this is confusing -- I will remove these changes as Archit will be addressing the error logs in his PR (#1337).

dan4thewin (approved):

sarenameas (approved):

PR #1353, opened 2020-10-31
Fix bug in multithreaded S3 download demo

archigup:
The HTTPRequestHeaders_t that was being sent over the queue contained additional pointers that were no longer pointing into the correct buffer after being transferred. 
This caused runtime issues when compiled with LLVM.

To fix this issue, the response and request structs were made global variables so that they would share the same address in the main and HTTP threads, and thus the pointers would all remain valid.

### DIFF
--- a/demos/http/http_demo_s3_download_multithreaded/http_demo_s3_download_multithreaded.c
+++ b/demos/http/http_demo_s3_download_multithreaded/http_demo_s3_download_multithreaded.c
@@ -161,6 +161,24 @@ typedef struct ResponseItem
     uint8_t responseBuffer[ USER_BUFFER_LENGTH ];
 } ResponseItem_t;
 
+/* @brief Struct used for sending requests to the HTTP thread.
+ *
+ * This is used by both the main and HTTP threads, though since we create the
+ * HTTP thread without a shared memory space, each will have its own copy. We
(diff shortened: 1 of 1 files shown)


sarenameas (dismissed):

sarenameas (approved):

demos/http/http_demo_s3_download_multithreaded/http_demo_s3_download_multithreaded.c
```diff
@@ -161,6 +161,24 @@ typedef struct ResponseItem
     uint8_t responseBuffer[ USER_BUFFER_LENGTH ];
 } ResponseItem_t;
 
+/* @brief Struct used for sending requests to the HTTP thread.
+ *
+ * This is used by both the main and HTTP threads, though since we create the
+ * HTTP thread without a shared memory space, each will have its own copy. We
+ * have this as a global so that it will be located at the same address in the
+ * main and HTTP threads so that way pointers remain valid when copied over.
```
gkwicker, line 169: How do the main and HTTP threads synchronize?  What prevents one task from overwriting one of these values while the other one is reading them?  Is there any critical section protection?

gkwicker (commented):
One question - see my comment.

gkwicker (approved):

PR #1356, opened 2020-11-01
Fix catching the S3_PRESIGNED_PUT_URL in cmake.

sarenameas:
S3_PRESIGNED_PUT_URL was missed in the CMake check for macro definitions. This was keeping the S3 upload demo from building in the checks.

This PR also updates all submodules to the latest.

### DIFF
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -38,25 +38,20 @@ endfunction()
 # This function will add a demo target to the list of default targets
 # if pre-signed GET and PUT URLs are defined.
 function(check_presigned_urls demo_name)
-    if(S3_PRESIGNED_GET_URL AND S3_PRESIGNED_PUT_URL)
-        target_compile_definitions(
-            ${demo_name} PRIVATE
-                S3_PRESIGNED_GET_URL="${S3_PRESIGNED_GET_URL}"
-                S3_PRESIGNED_PUT_URL="${S3_PRESIGNED_PUT_URL}"
-        )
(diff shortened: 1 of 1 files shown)


abhidixi11 (dismissed):

nateglims (approved):

demos/CMakeLists.txt
```diff
@@ -55,6 +55,14 @@ function(check_presigned_urls demo_name)
         set(CMAKE_REQUIRED_DEFINITIONS -DS3_PRESIGNED_GET_URL)
     endif()
 
+    if(S3_PRESIGNED_GET_URL)
+        target_compile_definitions(
+            ${demo_name} PRIVATE
+                S3_PRESIGNED_PUT_URL="${S3_PRESIGNED_PUT_URL}"
+        )
+        set(CMAKE_REQUIRED_DEFINITIONS -DS3_PRESIGNED_PUT_URL)
+    endif()
```
yourslab, line 64: Should be able to put this in the last if block

yourslab: Or perhaps you meant to set this as `if(S3_PRESIGNED_PUT_URL)`. In that case, I think the `if(S3_PRESIGNED_GET_URL AND S3_PRESIGNED_PUT_URL)` can be removed to make it possible for the user to define one parameter in the header file and another through a CMake flag.

sarenameas: Yes thank you for the catch

sarenameas: Looks like I still need the first if case:
image

yourslab (approved):

aggarg (approved):

PR #1366, opened 2020-11-03
Automatically clone submodules of libraries upon building with cmake

yourslab:
Automatically clone submodules of library upon building with cmake

We used to just fail when the user had not cloned submodules

### DIFF
--- a/libraries/CMakeLists.txt
+++ b/libraries/CMakeLists.txt
@@ -2,25 +2,27 @@
 add_subdirectory(${3RDPARTY_DIR})
 
 # Macro utility to clone a library and source-related submodules.
-macro( clone_library working_dir )
-        find_package( Git REQUIRED )
-        execute_process( COMMAND rm -rf ${submodule_dir}
-                         COMMAND ${GIT_EXECUTABLE} submodule update --init ${submodule_dir}
-                         WORKING_DIRECTORY ${ROOT_DIR}
-                         RESULT_VARIABLE LIB_CLONE_RESULT )
-        if( NOT ${LIB_CLONE_RESULT} STREQUAL "0" )
-            message( FATAL_ERROR "Failed to clone submodule ${submodule_dir}." )
-        endif()
+macro( clone_library submodule_dir clone_source )
+    find_package( Git REQUIRED )
+    execute_process( COMMAND rm -rf ${submodule_dir}
+                        COMMAND ${GIT_EXECUTABLE} submodule update --init ${submodule_dir}
+                        WORKING_DIRECTORY ${ROOT_DIR}
+                        RESULT_VARIABLE LIB_CLONE_RESULT )
+    if( NOT ${LIB_CLONE_RESULT} STREQUAL "0" )
+        message( FATAL_ERROR "Failed to clone submodule ${submodule_dir}." )
+    endif()
+    if(${clone_source})
         # Sources may have submodules like from the depedency directory
         execute_process( COMMAND ${GIT_EXECUTABLE} submodule update --init -- source
-                         WORKING_DIRECTORY ${submodule_dir}
-                         RESULT_VARIABLE LIB_CLONE_RESULT )
+                            WORKING_DIRECTORY ${submodule_dir}
+                            RESULT_VARIABLE LIB_CLONE_RESULT )
         if( NOT ${LIB_CLONE_RESULT} STREQUAL "0" )
             message( FATAL_ERROR "Failed to submodules from sources." )
         endif()
+    endif()
 endmacro()
 
-set( SUBMODULE_DIRS
+set( LIBRARY_SUBMODULES
        ${MODULES_DIR}/standard/coreHTTP
        ${MODULES_DIR}/standard/coreJSON
        ${MODULES_DIR}/standard/coreMQTT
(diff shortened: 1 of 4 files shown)


platform/posix/transport/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ include( ${PLATFORM_DIR}/posix/posixFilePaths.cmake )
 if( NOT EXISTS ${MODULES_DIR}/standard/coreMQTT/source )
     # Attempt to clone coreMQTT.
     if( ${BUILD_CLONE_SUBMODULES} )
```
aggarw13, line 8: Is this logic redundant considering that if the `BUILD_CLONE_SUBMODULES` flag is set, it will automatically clone all submodules?

libraries/CMakeLists.txt
```diff
@@ -2,15 +2,33 @@
...
+        execute_process( COMMAND rm -rf ${submodule_dir}
+                         COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive ${submodule_dir}
                          WORKING_DIRECTORY ${ROOT_DIR}
-                         RESULT_VARIABLE MQTT_CLONE_RESULT )
+                         RESULT_VARIABLE LIB_CLONE_RESULT )
 
-        if( NOT ${MQTT_CLONE_RESULT} STREQUAL "0" )
-                message( FATAL_ERROR "Failed to clone coreMQTT submodule." )
+        if( NOT ${LIB_CLONE_RESULT} STREQUAL "0" )
+                message( FATAL_ERROR "Failed to clone submodule." )
```
aggarw13, line 13:
```suggestion
                message( FATAL_ERROR "Failed to clone submodule ${submodule_dir}." )
```

libraries/CMakeLists.txt
```diff
@@ -2,15 +2,33 @@
 add_subdirectory(${3RDPARTY_DIR})
 
 # Macro utility to clone the coreMQTT submodule.
-macro( clone_mqtt )
+macro( clone_library submodule_dir )
```
aggarw13, line 5: What is the reason of submoduling each spoke repository separately instead of submoduling all libraries as once with `git submodule update --init --recursive -- libraries/`?

platform/posix/transport/CMakeLists.txt
yourslab: I think if the `CMakeLists.txt` file from the platform directory is included, then transport_interface.h needs to be there for platform to build. Ideally, it would be best to keep platform independent from libraries, but this is another way around that. In this case, if the submodule is already cloned, nothing will happen.

libraries/CMakeLists.txt
yourslab: It has the added advantage of not having to do a recursive clone and only taking what is needed for each submodule's submodules.

platform/posix/transport/utest/CMakeLists.txt
```diff
@@ -6,17 +6,6 @@ cmake_minimum_required (VERSION 3.2.0)
...
-# The transport interface unit tests use the coreMQTT's transport_interface.h,
-# so make sure the submodule is cloned.
-if( NOT EXISTS ${MODULES_DIR}/standard/coreMQTT/source )
-    # Attempt to clone coreMQTT.
-    if( ${BUILD_CLONE_SUBMODULES} )
-        clone_mqtt()
-    else()
-        message( FATAL_ERROR "The required submodule coreMQTT does not exist. Either clone it manually, or set BUILD_CLONE_SUBMODULES to 1 to automatically clone it during build." )
-    endif()
-endif()
```
yourslab, line 18: This is redundant because the parent CMakeLists.txt already checks this

libraries/CMakeLists.txt
```diff
@@ -1,16 +1,42 @@
...
+            execute_process( COMMAND rm -rf ${submodule_dir}
+                        COMMAND ${GIT_EXECUTABLE} submodule update --init ${submodule_dir}
+                        WORKING_DIRECTORY ${ROOT_DIR}
+                        RESULT_VARIABLE LIB_CLONE_RESULT )
+            if( NOT ${LIB_CLONE_RESULT} STREQUAL "0" )
+                message( FATAL_ERROR "Failed to clone submodule ${submodule_dir}." )
+            endif()
+            if(${clone_source})
+                # Sources may have submodules like from the depedency directory
+                execute_process( COMMAND ${GIT_EXECUTABLE} submodule update --init -- source
```
aggarw13, line 16: This carries the knowledge about `source` directory of the individual repositories owned by us, and somewhat mixes it with the third-party repositories for which `source` is not relevant (thereby, requiring the need of providing the `0` parameter). I do not this that this commonization is clean.
Instead, it would be cleaner to just have this file update all submodules (owned by us) at once with the `--init --recursive` flag. As our individual submodules have been updated to disable cloning of the CBMC and Unity/Cmock leaf submodules, using the `--recursive` flag will be clean here.

libraries/CMakeLists.txt
```diff
@@ -1,16 +1,32 @@
...
+       ${MODULES_DIR}/standard/coreMQTT
+       ${MODULES_DIR}/aws/device-defender-for-aws-iot-embedded-sdk
+       ${MODULES_DIR}/aws/device-shadow-for-aws-iot-embedded-sdk
+       ${MODULES_DIR}/aws/jobs-for-aws-iot-embedded-sdk )
+
+foreach(submodule_dir IN ITEMS ${LIBRARY_SUBMODULES})
+    if( NOT EXISTS ${submodule_dir}/.git )
+        message( FATAL_ERROR "The required submodule {} does not exist. Either clone it manually, or set BUILD_CLONE_SUBMODULES to 1 to automatically clone it during build." )
+    endif()
+endforeach()
```
aggarw13, line 29: I don;t think this is required considering that the default behavior of CMake is to set the `BUILD_CLONE_SUBMODULES` flag to 1. Thus, this error would only occur when the customer is aware of this option, and has explicitly set it to `OFF`.

aggarw13 (approved):

n9wxu (approved):

PR #1383, opened 2020-11-09
Fix memory leak in tearDown of integration tests

yourslab:
If the assertion on the returned status of `MQTT_Disconnect` fails, then `Openssl_Disconnect` will never be called. This means that the memory will remain allocated while a new connection will be established for the next test case, causing a memory leak.

### DIFF
--- a/integration-test/mqtt/mqtt_system_test.c
+++ b/integration-test/mqtt/mqtt_system_test.c
@@ -826,6 +826,8 @@ void tearDown()
     /* Terminate TLS session and TCP connection. */
     ( void ) Openssl_Disconnect( &networkContext );
 
+    /* Make any assertions at the end that may prevent #Openssl_Disconnect from
+     * being called and ultimately cause a memory leak. */
     TEST_ASSERT_EQUAL( MQTTSuccess, status );
 }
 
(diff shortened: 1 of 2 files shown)


aggarw13 (dismissed):

aggarw13 (dismissed):

integration-test/mqtt/mqtt_system_test.c
```diff
@@ -819,10 +821,14 @@ void tearDown()
     }
 
     /* Terminate MQTT connection. */
-    TEST_ASSERT_EQUAL( MQTTSuccess, MQTT_Disconnect( &context ) );
+    status = MQTT_Disconnect( &context );
 
     /* Terminate TLS session and TCP connection. */
     ( void ) Openssl_Disconnect( &networkContext );
+
+    /* Make any assertions at the end that may prevent #Openssl_Disconnect from
```
gkwicker, line 829: Why are we throwing away the return value from `Openssl_Disconnect()`?  Shouldn't you also validate that in the same way that you're validating the result from `MQTT_Disconnect()`?  If you did this then you'd know that it `Openssl_Disconnect()` has been called and you could delete this comment.

gkwicker (commented):
See comments.

aggarw13 (approved):

gkwicker (approved):

PR #1391, opened 2020-11-17
Add GA to check links

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Adds a script to test links in the repository. For non-Markdown files, it extracts http and https links and tests each one using the Python requests library.

For Markdown files, [pandoc](https://github.com/jgm/pandoc) is used to convert from GitHub-flavored markdown to HTML, and beautifulsoup4 is used to parse the HTML and extract href tags. This lets us also verify anchors within the same document, and relative links to files in the repository.

Since change logs often contain many links to different pull requests or issues within the same GitHub repository, we're likely to run into rate limiting from GitHub. To minimize this, the [GitHub CLI](https://github.com/cli/cli) can be used to fetch a list of all PRs/issues for a repository, so we don't have to make any HTTP requests. If there are more than 5 links found for the PRs or issues in a repository, `gh` is attempted to use to fetch the list.

Dependencies include [beautifulsoup4](https://pypi.org/project/beautifulsoup4/), [requests](https://pypi.org/project/requests), and [termcolor](https://pypi.org/project/termcolor) from PyPI, [pandoc](https://github.com/jgm/pandoc) for .md files, and optionally, [gh](https://github.com/cli/cli).

### DIFF
--- a/tools/link-verifier/verify-links.py
+++ b/tools/link-verifier/verify-links.py
@@ -1,0 +1,353 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import time
+import argparse
+import re
+import subprocess
+import requests
+from bs4 import BeautifulSoup
+from termcolor import cprint
+from multiprocessing import Pool
+from functools import partial
+import shutil
+
+MARKDOWN_SEARCH_TERM = r'\.md$'
+# Regex to find a URL
+URL_SEARCH_TERM = r'https?://'
+# Some HTML tags that we choose to ignore
+IGNORED_LINK_SCHEMES = r'mailto:|ftp:|tel:'
+# Regexes to identify links to Github PRs or issues, which are very common in changelogs
+# and may result in rate limiting if each link is fetched manually.
+PULL_REQUEST_SEARCH = r'https://github.com/([A-Za-z0-9_.-]+)/([A-Za-z0-9_.-]+)/pull/(\d+)$'
+ISSUE_SEARCH = r'https://github.com/([A-Za-z0-9_.-]+)/([A-Za-z0-9_.-]+)/issues/(\d+)$'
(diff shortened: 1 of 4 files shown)


tools/link-verifier/verify-links.py
```diff
@@ -0,0 +1,347 @@
...
+def test_url(url):
+    """Tests a single url"""
+
+    global use_cache
+    global main_repo_list
+    status = ''
+    is_broken = False
+    # Test if link was already cached. If not, fail silently and send a request for the link.
+    try:
+        if use_cache:
```
sarenameas, line 183: Why can't we do an if/else here?

tools/link-verifier/verify-links.py
```diff
@@ -0,0 +1,347 @@
...
+            except Exception as e:
+                traceback.print_exc()
+                broken_links.append(link)
+                cprint(f'Error\t{link}', 'red')
+
+    # Return code > 0 to return error. This may return success if there are 256 broken links ¯\_(ツ)_/¯
+    num_broken = len(broken_links)
+    if num_broken > 0:
+        print(f'{num_broken} broken link' + ('s', '')[num_broken == 1])
+    sys.exit(num_broken)
```
sarenameas, line 343: Can you try to avoid the overflow?
```suggestion
    if num_broken > 0:
        print(f'{num_broken} broken link' + ('s', '')[num_broken == 1])
        sys.exit(1)
    sys.exit(0)
```

tools/link-verifier/verify-links.py
muneebahmed10: Good point, it's better to just check the key exists instead of `except`ing for a ValueError

sarenameas (approved):

aggarg (approved):

PR #1401, opened 2020-11-20
Replace retry-utils with submodule to FreeRTOS/backoffAlgorithm

aggarw13:
Use the new location of "retry utils" library which is the new `FreeRTOS/backoffAlgorithm` repository. 
Changes include:
* Adding the submodule
* Changing calling code of demos and tests to use the renamed API 
* Updating build files of demos/tests to use the submodule files


mww-aws:
It looks like this needs a re-base in order to get the proofs running.

.gitmodules
```diff
@@ -29,3 +29,7 @@
 	path = libraries/standard/corePKCS11
 	branch = main
 	url = https://github.com/FreeRTOS/corePKCS11.git
+[submodule "libraries/standard/backoffAlgorithm"]
+	path = libraries/standard/backoffAlgorithm
+	branch = main
+	url = git@github.com:FreeRTOS/backoffAlgorithm.git
```
leegeth, line 35: I think it is better to be https to be consistent with other submodules.

leegeth (approved):

cobusve (approved):

PR #1408, opened 2020-11-25
Update demos and tests for backoffAlgorithm API change

aggarw13:
The API of `FreeRTOS/backoffAlgorithm` library has changed to remove dependency on random number generator; instead require the caller to generate the random number and pass it to the `BackoffAlgorithm_GetNextBackoff` API for backoff period calculation. This PR updates the submodule pointer commit, and updates the demos and tests to use the simplied library API

### DIFF
--- a/demos/defender/defender_demo_json/mqtt_operations.c
+++ b/demos/defender/defender_demo_json/mqtt_operations.c
@@ -372,8 +372,9 @@ static bool connectToBrokerWithBackoffRetries( NetworkContext_t * pNetworkContex
         opensslCredentials.alpnProtosLen = ALPN_PROTOCOL_NAME_LENGTH;
     }
 
-    /* Seed pseudo random number generator (provided by ISO C standard library) for
-     * use by retry utils library when retrying failed connection attempts to broker. */
+    /* Seed pseudo random number generator used in the demo for
+     * backoff period calculation when retrying failed network operations
+     * with broker. */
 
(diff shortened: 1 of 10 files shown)


leegeth (dismissed):

demos/defender/defender_demo_json/mqtt_operations.c
```diff
@@ -331,9 +328,9 @@ static void mqttCallback( MQTTContext_t * pMqttContext,
 static bool handlePublishResend( MQTTContext_t * pMqttContext );
 /*-----------------------------------------------------------*/
 
-static int32_t generateRandomNumber()
+static uint32_t generateRandomNumber()
 {
-    return( rand() % ( INT32_MAX ) );
+    return( rand() & UINT32_MAX );
```
dan4thewin, line 333: Why the `& UINT32_MAX` ?  rand() returns a int between 0 and RAND_MAX, which would have to be less than UINT32_MAX (since it is an int), right?

aggarw13: Hmm, good point. Even though the docs say that `RAND_MAX` is platform dependent, you're right about the `int` type restricting to below `UINT32_MAX`.  I have removed `& UINT32_MAX`.

leegeth (approved):

dan4thewin (approved):

PR #1414, opened 2020-11-30
Support multiple transports in the same compilation unit

yourslab:
By removing the definition of the NetworkContext struct in the header file, we allow the application to define it. This allows an application writer to use multiple transports in the same compilation unit. That way, multiple .c files do not have to be created for each transport.

All these changes are tested to work in CI.

### DIFF
--- a/demos/lexicon.txt
+++ b/demos/lexicon.txt
@@ -125,6 +125,7 @@ deserialize
 deserialized
 deserializer
 deserializing
+desired
 dev
 developerguide
 devicepublickeyasciihex
@@ -304,6 +305,7 @@ mutex
 mutexes
 mxz
 necesarily
+networkcontext
 ni
 nist
 noninfringement
@@ -395,6 +397,7 @@ pouttcpportsarray
 poutudpportsarray
 poweron
 ppacketinfo
+pparams
 ppath
 ppathlen
 ppayload
@@ -533,6 +536,7 @@ toolchain
 topicbuffer
 topicfilterlength
 topiclength
+transport
 transportinterface
 transporttimeout
 txt
--- a/demos/lexicon.txt
+++ b/demos/lexicon.txt
@@ -536,7 +536,6 @@ toolchain
 topicbuffer
 topicfilterlength
 topiclength
-transport
 transportinterface
 transporttimeout
 txt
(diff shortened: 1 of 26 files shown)


platform/posix/transport/utest/plaintext_utest.c
```diff
@@ -55,8 +55,15 @@
...
-static ServerInfo_t serverInfo;
-static NetworkContext_t networkContext;
+/* Each compilation unit must define the NetworkContext struct. */
+struct NetworkContext
+{
+    PlaintextParams_t * pParams;
+};
+
+static ServerInfo_t serverInfo = { 0 };
+static NetworkContext_t networkContext = { 0 };
```
muneebahmed10, line 65: Don't know if you know this, but the `{ 0 }` isn't necessary since `static` is zero initialized anyway, but I guess it doesn't hurt readability to be explicit.

demos/lexicon.txt
```diff
@@ -533,6 +536,7 @@ toolchain
 topicbuffer
 topicfilterlength
 topiclength
+transport
```
muneebahmed10, line 539: Why would "transport" be dinged by spellcheck?

muneebahmed10 (approved):

platform/posix/transport/utest/plaintext_utest.c
yourslab: You're right, but it doesn't hurt to be explicit.

muneebahmed10 (dismissed):

muneebahmed10 (approved):

platform/posix/transport/utest/plaintext_utest.c
abhidixi11: This initialization is not C89 either, so not sure if we should use it, given that we having warnings with bool.

yourslab: I recall that tests do not have to adhere to C89. These initializers are used several times in the integration tests too.
Also, is the initialization really against C89?
https://stackoverflow.com/questions/63355760/how-standard-is-the-0-initializer-in-c89

cobusve (approved):

PR #1432, opened 2020-12-08
Update OpenSSL transport send/recv implementation

aggarw13:
The `OpenSSL_Send` and `OpenSSL_Recv` functions should return zero only when there is a timeout in sending/receiving bytes over the network which should be retried by the coreMQTT/coreHTTP libraries.

Based on the OpenSSL documentation of its [`SSL_write`](https://www.openssl.org/docs/man1.0.2/man3/SSL_write.html) and [`SSL_read`](https://www.openssl.org/docs/man1.0.2/man3/SSL_read.html) APIs, the `SSL_ERROR_WANT_READ` and `SSL_ERROR_WANT_WRITE` error codes can be returned by both the APIs. Also, the documentation of these error codes suggest that the read/write operation should be re-tried by calling the same OpenSSL API.
Thus, the transport interface implementations with `OpenSSL_Send` and `OpenSSL_Recv` should return 0 for those error codes so that the calling library can potentially retry the operation.


platform/posix/transport/src/openssl_posix.c
```diff
@@ -778,8 +782,20 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
         {
             sslError = SSL_get_error( pOpensslParams->pSsl, bytesSent );
 
-            LogError( ( "Failed to send data over network: SSL_write of OpenSSL failed: "
-                        "ErrorStatus=%s.", ERR_reason_error_string( sslError ) ) );
+            if( ( sslError == SSL_ERROR_WANT_WRITE ) || ( sslError == SSL_ERROR_WANT_READ ) )
```
muneebahmed10, line 785: If I recall this wasn't checked here because `SSL_write` was blocking and these values can't be returned. Is this necessary, @yourslab ?

platform/posix/transport/src/openssl_posix.c
```diff
@@ -726,9 +726,13 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
...
 
-            if( sslError == SSL_ERROR_WANT_READ )
+            if( ( sslError == SSL_ERROR_WANT_READ ) || ( sslError == SSL_ERROR_WANT_WRITE ) )
             {
-                /* There is no data to receive at this time. */
+                /* The OpenSSL documentation mentions that SSL_Write can have a return code of
+                 * SSL_ERROR_WANT_READ and SSL_ERROR_WANT_WRITE. Both of those codes mean that the
+                 * SSL_read() operation needs to be retried to complete the read operation.
+                 * Thus, setting the return value of this function as zero to represent that no
+                 * data was received from the network. */
```
muneebahmed10, line 735: @abhidixi11 and I had a discussion a while back about this block, and I believe that the `else` block below this should set `bytesReceived` to a negative value, otherwise it's possible that a non-retriable error may return 0.

platform/posix/transport/utest/openssl_utest.c
```diff
@@ -840,3 +865,28 @@ void test_Openssl_Recv_Network_Error( void )
...
+
+    /* Test when SSL_read returns SSL_ERROR_WANT_WRITE. */
+    opensslParams.pSsl = &ssl;
+    SSL_read_ExpectAnyArgsAndReturn( SSL_READ_WRITE_ERROR );
+
+    SSL_get_error_ExpectAnyArgsAndReturn( SSL_ERROR_WANT_WRITE );
+    bytesReceived = Openssl_Recv( &networkContext, opensslBuffer, BYTES_TO_RECV );
+    TEST_ASSERT_EQUAL( 0, bytesReceived );
+
+    /* Test when SSL_read returns SSL_ERROR_WANT_READ. */
```
muneebahmed10, line 885: The test below is already tested on line 852 and is duplicated here. It'd probably make sense to just move the above test into that block as well

platform/posix/transport/utest/openssl_utest.c
```diff
@@ -768,12 +768,37 @@ void test_Openssl_Send_Network_Error( void )
 
     /* Several errors can be returned from #SSL_get_error as mentioned here:
      * https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html */
-    SSL_get_error_ExpectAnyArgsAndReturn( SSL_ERROR_WANT_WRITE );
+    SSL_get_error_ExpectAnyArgsAndReturn( SSL_READ_WRITE_ERROR );
```
muneebahmed10, line 771: I don't believe this is a value able to be returned by SSL_get_error(). Might want to use something like SSL_ERROR_ZERO_RETURN instead

platform/posix/transport/src/openssl_posix.c
yourslab: Yah, this is blocking, so we don't expect to get these errors. However, we actually have renegotiation activated because of the `SSL_MODE_AUTO_RETRY` flag. In that case, we only want to check `SSL_ERROR_WANT_READ`.

platform/posix/transport/src/openssl_posix.c
yourslab: Although, I don't think we should necessarily change the error code returned by OpenSSL. We could probably have an else if like so:
```
else if( bytesReceived = 0 || sslError == SSL_ERROR_ZERO_RETURN )
```

platform/posix/transport/src/openssl_posix.c
aggarw13: Based on the documentation of [OpenSSL_write](https://www.openssl.org/docs/man1.0.2/man3/SSL_write.html), `SSL_ERROR_WANT_READ` can occur in a blocking case (unless the the auto retry mode is set)

platform/posix/transport/src/openssl_posix.c
```diff
@@ -726,9 +726,13 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
         {
             sslError = SSL_get_error( pOpensslParams->pSsl, bytesReceived );
 
-            if( sslError == SSL_ERROR_WANT_READ )
+            if( ( sslError == SSL_ERROR_WANT_READ ) || ( sslError == SSL_ERROR_WANT_WRITE ) )
```
yourslab, line 729: This is blocking, so we don't expect to get `SSL_ERROR_WANT_WRITE`. Only need to check `SSL_ERROR_WANT_READ` here.

platform/posix/transport/src/openssl_posix.c
aggarw13: I thought about that and wasn't sure about what negative value this should be set to. The expectation by the calling code could be to expect the negative error code sent by OpenSSL.
However, for the correctness in functional behavior of the transport send/recv functions, I will set the value to be explicitly negative as `-1`.

platform/posix/transport/src/openssl_posix.c
yourslab: If `SSL_MODE_AUTO_RETRY` is set, then we need to check `SSL_ERROR_WANT_READ`. `SSL_ERROR_WANT_WRITE` check should be removed though.

platform/posix/transport/src/openssl_posix.c
yourslab: I think the negative error code sent by OpenSSL should be returned. The only exception to this case is when OpenSSL returns `0`, which signifies a close notify, then transport recv should return `-1`.

platform/posix/transport/utest/openssl_utest.c
aggarw13: Thanks for catching. I have instead removed the older section as I prefer keeping the `0` and `-1` (or any negative value) return test cases as separate.

platform/posix/transport/src/openssl_posix.c
aggarw13: Oops, didn't notice that the auto retry mode was set. Have removed the check for `SSL_ERROR_WANT_READ`.

platform/posix/transport/src/openssl_posix.c
aggarw13: Removed check for `SSL_ERROR_WANT_WRITE`

platform/posix/transport/src/openssl_posix.c
yourslab: However, the auto retry requires us to check `SSL_ERROR_WANT_READ`. What needs to be removed is `SSL_ERROR_WANT_WRITE`.

aggarw13: Okay, updated.

yourslab (approved):

muneebahmed10 (dismissed):

platform/lexicon.txt
```diff
@@ -128,15 +146,17 @@ pbuffer
...
+pfilecontext
 pfilepath
 pformat
 phostname
-platformimagestate
 plaintext
-platfrom
 platformimagestate
+platformimagestate
+platfrom
```
gkwicker, line 159: Let's not put misspelled words in the lexicon – __platform__

platform/posix/transport/src/openssl_posix.c
```diff
@@ -778,8 +790,28 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
         {
             sslError = SSL_get_error( pOpensslParams->pSsl, bytesSent );
 
-            LogError( ( "Failed to send data over network: SSL_write of OpenSSL failed: "
-                        "ErrorStatus=%s.", ERR_reason_error_string( sslError ) ) );
+            if( sslError == SSL_ERROR_WANT_READ )
```
gkwicker, line 793: [The documentation for SSL_write()](https://www.openssl.org/docs/man1.1.1/man3/SSL_write.html) says that it will only return `SSL_ERROR_WANT_READ` (and also `SSL_ERROR_WANT_WRITE`) in __nonblocking__ BIO implementations.  Also note that both of these return values are possible for `SSL_read()` as well with a nonblocking BIO.  So these comments as well as the ones in `Openssl_Recv()` don't seem consistent with the documentation.

gkwicker (changes requested):
See comments

platform/posix/transport/src/openssl_posix.c
aggarw13: I had earlier checked against both `SSL_ERROR_WANT_READ` and `SSL_ERROR_WANT_WRITE` error codes, however, I was notified in [this comment](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1432#discussion_r537964636) that the transport implementation configured OpenSSL in blocking mode.

gkwicker: [According to the documentation](https://www.openssl.org/docs/man1.1.1/man3/SSL_write.html), `SSL_Write()` will __not__ return either `SSL_ERROR_WANT_READ` or `SSL_ERROR_WANT_WRITE` in blocking mode, so lines 793-814 can be deleted since they will never execute.

platform/posix/transport/src/openssl_posix.c
gkwicker: [According to the documentation](https://www.openssl.org/docs/man1.1.1/man3/SSL_read.html), this is not as simple as returning `SSL_ERROR_WANT_READ` in blocking mode.  In order to get that return value, one of the following must be true:

* Underlying BIO is non-blocking (this doesn't seem to be the case based on later comments).  In that case you also need to check for `SSL_ERROR_WANT_READ`.
* Underlying BIO is blocking __and__ `SSL_MODE_AUTO_RETRY` is disabled.

These comments and the initialization/configuration of this TLS need to be very clear and consistent with the OpenSSL documentation, so let's clarify this.

gkwicker (changes requested):
A couple of more questions

gkwicker (dismissed):

yourslab (approved):

gkwicker (approved):

yourslab (approved):

PR #1433, opened 2020-12-08
Support hostname verification and partial writes in Openssl wrapper

yourslab:
When the SNI hostname is set, we need to use `SSL_set1_host` to verify the hostname against the server's certificate. Also, when the payload was too large, OpenSSL would just return `SSL_WANT_WRITE` instead of trying to write as many bytes as it could, potentially causing an infinite loop in libraries. The `SSL_MODE_ENABLE_PARTIAL_WRITE` flag allows OpenSSL to support partial writes so that less bytes than the expected payload can be returned.

### DIFF
--- a/platform/posix/transport/src/openssl_posix.c
+++ b/platform/posix/transport/src/openssl_posix.c
@@ -608,7 +608,8 @@ OpensslStatus_t Openssl_Connect( NetworkContext_t * pNetworkContext,
     /* Setup credentials. */
     if( returnStatus == OPENSSL_SUCCESS )
     {
-        /* Set auto retry mode for the blocking calls to SSL_read and SSL_write.
+        /* Enable partial writes for blocking calls to SSL_write to allow a
+         * payload larger than the maximum fragment length.
          * The mask returned by SSL_CTX_set_mode does not need to be checked. */
 
         /* MISRA Directive 4.6 flags the following line for using basic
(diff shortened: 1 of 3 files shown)


platform/posix/transport/src/openssl_posix.c
```diff
@@ -223,22 +225,37 @@ static OpensslStatus_t convertToOpensslStatus( SocketStatus_t socketStatus )
...
+    /* Validate the hostname against the server's certificate. */
+    sslStatus = SSL_set1_host( pOpensslParams->pSsl,
+                               pServerInfo->pHostName );
 
     if( sslStatus != 1 )
     {
-        LogError( ( "SSL_set_fd failed to set the socket fd to SSL context." ) );
-        returnStatus = OPENSSL_API_ERROR;
+        LogError( ( "SSL_set1_host failed to validate the hostname "
+                    "in the server's certificate." ) );
```
muneebahmed10, line 242: Are we actually validating the hostname here, or just setting the hostname for openssl to verify? It looks like the handshake happens below this

yourslab: Yes, it's just setting it to validate later.

muneebahmed10 (approved):

cobusve (approved):

muneebahmed10 (dismissed):

cobusve (approved):

muneebahmed10 (approved):

Issue #1427, opened 2020-12-04
Build requires mosquitto.h but not found?


rpcme:
I did a clean build of tip of main and received:

```bash
[  5%] Building C object demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/jobs_demo_mosquitto.c.o
/home/ANT.AMAZON.COM/elberger/src/aws-iot-device-sdk-embedded-C/demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c:52:10: fatal error: mosquitto.h: No such file or directory
 #include <mosquitto.h>
          ^~~~~~~~~~~~~
compilation terminated.
demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/build.make:62: recipe for target 'demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/jobs_demo_mosquitto.c.o' failed
make[2]: *** [demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/jobs_demo_mosquitto.c.o] Error 1
CMakeFiles/Makefile2:766: recipe for target 'demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/all' failed
make[1]: *** [demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2
```

Is this expected for the build? I didn't set any environment variables.

yourslab:
Hi @rpcme 

The Jobs demo in C SDK uses the Mosquitto Client Library to communicate with AWS IoT Core. That package is usually called `libmosquitto-dev` if installing through `apt` or `mosquitto-dev` if installing through `yum`.

Best regards,
Oscar

rpcme:
@yourslab I didn't see this specifically called out in the documentation. Can you please add it?

yourslab:
That's a good point. I do prefer having it as a 3rdparty submodule too.

rpcme:
I was shown the documentation here:
https://github.com/aws/aws-iot-device-sdk-embedded-C/tree/main/demos/jobs/jobs_demo_mosquitto

Perhaps need a table with per-demo requirement links?

[...]

PR #1436, opened 2020-12-08
Update CMake to be more lenient with package requirements 

yourslab:
- Do not force any packages or libraries to be required
- If a certain package or library that a demo or test needs is not installed, exclude the demo or test from the `all` target
- Add table of required build dependencies for demos and tests in README
- If mosquitto library is not installed, build it from source
- Inform user about `BUILD_CLONE_SUBMODULES` flag in README


platform/CMakeLists.txt
```diff
@@ -1,10 +1,21 @@
...
-find_package(OpenSSL REQUIRED)
+# For static library builds, link against static library of OpenSSL
+if(NOT BUILD_SHARED_LIBS)
+    set( OPENSSL_USE_STATIC_LIBS TRUE )
+endif()
+find_package(OpenSSL)
+# Verify the minimum OpenSSL version required
+if( ${OPENSSL_FOUND} AND ( OPENSSL_VERSION MATCHES "0.9$" OR OPENSSL_VERSION MATCHES "1.0$" ) )
+    message( WARNING "OpenSSL 1.1.0 or later required: OpenSSL ${OPENSSL_VERSION} found." )
+endif()
```
muneebahmed10, line 9: Is it better to have these checks on the uppermost level of `platform/`, or keep them on the lowermost level of `platform/posix/transport/` as they were before?

demos/jobs/jobs_demo_mosquitto/README.md
```diff
@@ -8,6 +8,7 @@ To install curl and libmosquitto on a Debian or Ubuntu host, run:
 
     apt install curl libmosquitto-dev
 
-To build the latest version of libmosquitto with ALPN support,
-clone the repository at https://github.com/eclipse/mosquitto ,
-and follow the instructions in compiling.txt within the repo.
+Version 1.4.10 of libmosquitto is required to run this demo.
+To build the latest version of libmosquitto (1.6.12) for ALPN support,
```
muneebahmed10, line 12: Version 2.0.0. came out last week, so this is no longer accurate

README.md
```diff
@@ -233,8 +238,17 @@ The libraries in this SDK are not dependent on any operating system. However, th
 * C90 compiler
 * Although not a part of the ISO C90 standard, `stdint.h` is required for fixed-width integer types that include `uint8_t`, `int8_t`, `uint16_t`, `uint32_t` and `int32_t`, and constant macros like `UINT16_MAX`.
```
muneebahmed10, line 239: While we're at it, would you mind updating this for me?
```suggestion
* Although not a part of the ISO C90 standard, `stdint.h` is required for fixed-width integer types that include `uint8_t`, `int8_t`, `uint16_t`, `uint32_t` and `int32_t`, and constant macros like `UINT16_MAX`, while `stdbool.h` is required for boolean parameters in coreMQTT. For compilers that do not provide these header files, [coreMQTT](https://github.com/FreeRTOS/coreMQTT) provides the files [stdint.readme](https://github.com/FreeRTOS/coreMQTT/blob/main/source/include/stdint.readme) and [stdbool.readme](https://github.com/FreeRTOS/coreMQTT/blob/main/source/include/stdbool.readme), which can be renamed to `stdint.h` and `stdbool.h`, respectively, to provide the required type definitions.
```

platform/CMakeLists.txt
yourslab: I was contemplating this as well and decided it was better to have it in a single file. Because CMake makes everything global, I'm trying to keep all shared build dependencies in the same files to make things easier to manage. We ended up duplicating these checks in so many places. Also, seems like it should be possible to check if OpenSSL is available even though the wrapper isn't being included. What happens when there is no wrapper like in the case of pthreads? Open to your thoughts though.

demos/jobs/jobs_demo_mosquitto/README.md
yourslab, line 12:
```suggestion
For ALPN support, build the latest version of the first major release of libmosquitto (1.6.12).
```

demos/jobs/jobs_demo_mosquitto/README.md
```diff
@@ -8,6 +8,7 @@ To install curl and libmosquitto on a Debian or Ubuntu host, run:
...
+go to the libmosquitto directory,
```
yourslab, line 13:
```suggestion
To do this, go to the libmosquitto directory,
```

demos/jobs/jobs_demo_mosquitto/README.md
```diff
@@ -8,6 +8,7 @@ To install curl and libmosquitto on a Debian or Ubuntu host, run:
 
     apt install curl libmosquitto-dev
 
-To build the latest version of libmosquitto with ALPN support,
-clone the repository at https://github.com/eclipse/mosquitto ,
-and follow the instructions in compiling.txt within the repo.
+Version 1.4.10 of libmosquitto is required to run this demo.
```
yourslab, line 11:
```suggestion
libmosquitto 1.4.10 or any later version of the first major release is required to run this demo.
```

muneebahmed10 (commented):
Looks good, but had a few errors when trying it out. I saw this for defender:
```
-- Found OpenSSL: /usr/local/opt/openssl/lib/libcrypto.dylib (found version "1.1.1h")
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Looking for AWS_IOT_ENDPOINT
-- Looking for AWS_IOT_ENDPOINT - found
-- Looking for CLIENT_CERT_PATH
-- Looking for CLIENT_CERT_PATH - found
-- Looking for CLIENT_PRIVATE_KEY_PATH
-- Looking for CLIENT_PRIVATE_KEY_PATH - found
All required macros for defender_demo were found - Adding to default target.
```
However, defender also requires "THING_NAME", so I had to define that to get it to compile. Also, I ran into this error with jobs mosquitto when trying `make all`:
```
Scanning dependencies of target jobs_demo_mosquitto
[ 10%] Building C object demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/jobs_demo_mosquitto.c.o
/Users/ahmemune/Documents/DELETE_THIS_TEMP_PR/aws-iot-device-sdk-embedded-C/demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c:52:10: fatal error: mosquitto.h: No such file or directory
   52 | #include <mosquitto.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
make[2]: *** [demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/jobs_demo_mosquitto.c.o] Error 1
make[1]: *** [demos/jobs/jobs_demo_mosquitto/CMakeFiles/jobs_demo_mosquitto.dir/all] Error 2
make: *** [all] Error 2
```
Is that intentional? This PR makes it seem like it shouldn't be added to the target if mosquitto is not available

yourslab:
I've fixed the THING_NAME issue, but I believe the mosquitto problem only happens in Macs due to not finding the appropriate header files. However, we don't officially support Macs. There are ways around it, but also, I didn't get the same problem in my own Mac. Could it be a problem with how mosquitto is installed in your system?

muneebahmed10 (approved):

dan4thewin (approved):

yourslab:
This fixes issue #1427

PR #1453, opened 2020-12-10
Fix OTA PAL unit test mock

yanjos-dev:
CMock doesn't support variadic functions. Use an alias to replace the original function as a workaround to this issue.

### DIFF
--- a/platform/lexicon.txt
+++ b/platform/lexicon.txt
@@ -235,6 +235,7 @@ unistd
 utest
 utils
 v1
+variadic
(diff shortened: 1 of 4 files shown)


pvyawaha (approved):

nihaljamazon (commented):
#define snprintf                           snprintf_alias 
=> Should it be other way? 
#define snprintf_alias                           snprintf

nihaljamazon (approved):

PR #1475, opened 2020-12-11
Block until all data has been read or written in mbedTLS port

yourslab:
- `iot_tls_write` and `iot_tls_read` used to return an error when the actual number of sent/received bytes did not match the requested bytes. As such, an appropriate update is made to update these functions to block until the entire payload has been sent or received. This requires allocating new timers with configurable timeouts, `IOT_SSL_WRITE_RETRY_TIMEOUT_MS` and `IOT_SSL_READ_RETRY_TIMEOUT_MS`. Keep in mind that the dependency on the old yield timer is removed so that we never break out of the loop even though the yield timer has expired.
    - `iot_tls_write`: When the write timer expires and`MBEDTLS_ERR_SSL_WANT_WRITE` or `MBEDTLS_ERR_SSL_TIMEOUT` are returned by `mbedtls_ssl_write`.  `NETWORK_SSL_WRITE_TIMEOUT_ERROR` is appropriately returned by `iot_tls_write`.
    - `iot_tls_read`: When the read timer expires and `MBEDTLS_ERR_SSL_WANT_READ`, `MBEDTLS_ERR_SSL_WANT_WRITE`, or `MBEDTLS_ERR_SSL_TIMEOUT` are returned by `mbedtls_ssl_read`. `NETWORK_SSL_READ_TIMEOUT_ERROR` is appropriately returned by `iot_tls_read`.
- Another change is made so that when the number of sent/received bytes is larger than the number of requested bytes, then `NETWORK_SSL_WRITE_ERROR` or `NETWORK_SSL_READ_ERROR` is appropriately returned.
- In `aws_iot_mqtt_set_client_state`, whenever `_ENABLE_THREAD_SUPPORT_` was defined, `aws_iot_mqtt_client_lock_mutex` would return its status to `rc` rather than `threadRc`. While this won't make any difference in terms of functionality, this fix needs to be made for consistency and code hygiene.

### DIFF
--- a/platform/linux/mbedtls/network_mbedtls_wrapper.c
+++ b/platform/linux/mbedtls/network_mbedtls_wrapper.c
@@ -270,7 +270,7 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
 	}
 
 #ifdef ENABLE_IOT_DEBUG
-	if (mbedtls_ssl_get_peer_cert(&(tlsDataParams->ssl)) != NULL) {
+	if(mbedtls_ssl_get_peer_cert(&(tlsDataParams->ssl)) != NULL) {
 		IOT_DEBUG("  . Peer certificate information    ...\n");
 		mbedtls_x509_crt_info((char *) buf, sizeof(buf) - 1, "      ", mbedtls_ssl_get_peer_cert(&(tlsDataParams->ssl)));
 		IOT_DEBUG("%s\n", buf);
@@ -307,7 +307,7 @@ IoT_Error_t iot_tls_write(Network *pNetwork, unsigned char *pMsg, size_t len, Ti
 	init_timer(&writeTimer);
 	countdown_ms(&writeTimer, IOT_SSL_WRITE_RETRY_TIMEOUT_MS);
 
-	while (len > 0) {
+	while(len > 0) {
 		ret = mbedtls_ssl_write(pSsl, pMsg, len);
 
 		if(ret > 0) {
@@ -362,11 +362,11 @@ IoT_Error_t iot_tls_read(Network *pNetwork, unsigned char *pMsg, size_t len, Tim
 	init_timer(&readTimer);
 	countdown_ms(&readTimer, IOT_SSL_READ_RETRY_TIMEOUT_MS);
 
-	while (len > 0) {
+	while(len > 0) {
 		/* This read will timeout after IOT_SSL_READ_TIMEOUT if there's no data to be read */
 		ret = mbedtls_ssl_read(pSsl, pMsg, len);
 
-		if (ret > 0) {
+		if(ret > 0) {
 			if((size_t) ret > len) {
 				IOT_ERROR("More bytes read than requested\n\n");
 				return NETWORK_SSL_WRITE_ERROR;
@@ -379,11 +379,11 @@ IoT_Error_t iot_tls_read(Network *pNetwork, unsigned char *pMsg, size_t len, Tim
 			rxLen += ret;
 			pMsg += ret;
 			len -= ret;
-		} else if (ret == MBEDTLS_ERR_SSL_WANT_READ || 
+		} else if(ret == MBEDTLS_ERR_SSL_WANT_READ || 
 				ret == MBEDTLS_ERR_SSL_WANT_WRITE ||
 				ret == MBEDTLS_ERR_SSL_TIMEOUT) {
 			if(has_timer_expired(&readTimer)) {
-				if (rxLen == 0) {
+				if(rxLen == 0) {
 					return NETWORK_SSL_NOTHING_TO_READ;
 				} else {
 					return NETWORK_SSL_READ_TIMEOUT_ERROR;
--- a/platform/linux/mbedtls/network_mbedtls_wrapper.c
+++ b/platform/linux/mbedtls/network_mbedtls_wrapper.c
@@ -31,7 +31,9 @@ extern "C" {
 
 
 /* This is the value used for ssl read timeout */
-#define IOT_SSL_READ_TIMEOUT 10
+#ifndef IOT_SSL_READ_TIMEOUT
+	#define IOT_SSL_READ_TIMEOUT 10
+#endif
 
 /* This defines the value of the debug buffer that gets allocated.
  * The value can be altered based on memory constraints
@@ -289,7 +291,9 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
 /* When this much time has elapsed after receiving MBEDTLS_ERR_SSL_WANT_READ
  * or MBEDTLS_ERR_SSL_WANT_WRITE, then iot_tls_write will return
  * NETWORK_SSL_WRITE_TIMEOUT_ERROR. */
-#define IOT_SSL_WRITE_RETRY_TIMEOUT_MS 10
+#ifndef IOT_SSL_WRITE_RETRY_TIMEOUT_MS
+	#define IOT_SSL_WRITE_RETRY_TIMEOUT_MS 10
+#endif
 
 IoT_Error_t iot_tls_write(Network *pNetwork, unsigned char *pMsg, size_t len, Timer *timer, size_t *written_len) {
 	mbedtls_ssl_context *pSsl = &(pNetwork->tlsDataParams.ssl);
@@ -343,7 +347,9 @@ IoT_Error_t iot_tls_write(Network *pNetwork, unsigned char *pMsg, size_t len, Ti
 /* When this much time has elapsed after receiving MBEDTLS_ERR_SSL_WANT_READ,
  * MBEDTLS_ERR_SSL_WANT_WRITE, or MBEDTLS_ERR_SSL_TIMEOUT, then iot_tls_read
  * will return NETWORK_SSL_READ_TIMEOUT_ERROR. */
-#define IOT_SSL_READ_RETRY_TIMEOUT_MS 10
+#ifndef IOT_SSL_READ_RETRY_TIMEOUT_MS
+	#define IOT_SSL_READ_RETRY_TIMEOUT_MS 10
+#endif
 
 IoT_Error_t iot_tls_read(Network *pNetwork, unsigned char *pMsg, size_t len, Timer *timer, size_t *read_len) {
 	mbedtls_ssl_context *pSsl = &(pNetwork->tlsDataParams.ssl);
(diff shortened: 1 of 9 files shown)


platform/linux/mbedtls/network_mbedtls_wrapper.c
```diff
@@ -286,73 +286,116 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
 	return (IoT_Error_t) ret;
 }
 
+/* When this much time has elapsed after receiving MBEDTLS_ERR_SSL_WANT_READ
+ * or MBEDTLS_ERR_SSL_WANT_WRITE, then iot_tls_write will return
+ * NETWORK_SSL_WRITE_TIMEOUT_ERROR. */
+#define IOT_SSL_WRITE_RETRY_TIMEOUT_MS 10
```
leegeth, line 292: Why is this decided to be provided as a MACRO in .c file? Can't this be set in a config header?

platform/linux/mbedtls/network_mbedtls_wrapper.c
```diff
@@ -286,73 +286,116 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
...
 
-	if(isErrorFlag) {
-		return NETWORK_SSL_WRITE_ERROR;
-	} else if(has_timer_expired(timer) && written_so_far != len) {
-		return NETWORK_SSL_WRITE_TIMEOUT_ERROR;
+			txLen += ret;
+			pMsg += ret;
+			len -= ret;
+		} else if(ret == MBEDTLS_ERR_SSL_WANT_READ ||
+				ret == MBEDTLS_ERR_SSL_WANT_WRITE) {
```
leegeth, line 327: uncrustify?

platform/linux/mbedtls/network_mbedtls_wrapper.c
```diff
@@ -286,73 +286,116 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
...
 	}
 
+	*written_len = txLen;
 	return SUCCESS;
 }
 
+/* When this much time has elapsed after receiving MBEDTLS_ERR_SSL_WANT_READ,
+ * MBEDTLS_ERR_SSL_WANT_WRITE, or MBEDTLS_ERR_SSL_TIMEOUT, then iot_tls_read
+ * will return NETWORK_SSL_READ_TIMEOUT_ERROR. */
+#define IOT_SSL_READ_RETRY_TIMEOUT_MS 10
```
leegeth, line 346: same comment about the config

platform/linux/mbedtls/network_mbedtls_wrapper.c
```diff
@@ -286,73 +286,116 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
...
-		ret = mbedtls_ssl_read(ssl, pMsg, len);
-		if (ret > 0) {
 			rxLen += ret;
 			pMsg += ret;
 			len -= ret;
-		} else if (ret == 0 || (ret != MBEDTLS_ERR_SSL_WANT_READ && ret != MBEDTLS_ERR_SSL_WANT_WRITE && ret != MBEDTLS_ERR_SSL_TIMEOUT)) {
+		} else if(ret == MBEDTLS_ERR_SSL_WANT_READ || 
+				ret == MBEDTLS_ERR_SSL_WANT_WRITE ||
+				ret == MBEDTLS_ERR_SSL_TIMEOUT) {
+			if(has_timer_expired(&readTimer)) {
```
leegeth, line 384: is the indentation correct here?

platform/linux/mbedtls/network_mbedtls_wrapper.c
```diff
@@ -286,73 +286,116 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
...
+				if(rxLen == 0U) {
+					return NETWORK_SSL_NOTHING_TO_READ;
```
leegeth, line 386: Is this error code handled differently in calling code?

platform/linux/mbedtls/network_mbedtls_wrapper.c
yourslab: I don't like it either, but this file exclusively uses tabs... I really want to uncrustify all files in V3.

platform/linux/mbedtls/network_mbedtls_wrapper.c
yourslab: I decided to do this because `IOT_SSL_READ_TIMEOUT` was also originally defined in the .c file. I think the reasoning here was to provide reasonable defaults to the application. That being said, I've wrapped these in `#ifndef` so that they can be configured in `aws_iot_config.h`.

platform/linux/mbedtls/network_mbedtls_wrapper.c
yourslab: None of the files in V3 are uncrustified and are mixed up with spaces and tabs... I think we should do another release after to uncrustify all files and update copyright headers. If we do it now though, it will be harder to track changes in this release.

platform/linux/mbedtls/network_mbedtls_wrapper.c
yourslab: Yes, it is similar to `MQTTNoDataAvailable` and not treated as an error. The keep-alive packet is sent after this status is returned provided that the yield timer has also expired.

platform/linux/mbedtls/network_mbedtls_wrapper.c
leegeth: ok sounds good.

platform/linux/mbedtls/network_mbedtls_wrapper.c
```diff
@@ -286,73 +288,120 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
...
-		return NETWORK_SSL_WRITE_ERROR;
-	} else if(has_timer_expired(timer) && written_so_far != len) {
-		return NETWORK_SSL_WRITE_TIMEOUT_ERROR;
+			txLen += ret;
+			pMsg += ret;
+			len -= ret;
+		} else if(ret == MBEDTLS_ERR_SSL_WANT_READ ||
+				ret == MBEDTLS_ERR_SSL_WANT_WRITE) {
+			if(has_timer_expired(&writeTimer)) {
+				return NETWORK_SSL_WRITE_TIMEOUT_ERROR;
```
leegeth, line 333: How about passing the length that is read here back to the calling code so the calling code has a flexibility to decide on whether it needs a retry or a failure?

platform/linux/mbedtls/network_mbedtls_wrapper.c
leegeth: Sounds good. Would it be better if we pass the length of data successfully read back to the calling code in case of a timeout?

platform/linux/mbedtls/network_mbedtls_wrapper.c
```diff
@@ -286,73 +288,120 @@ IoT_Error_t iot_tls_connect(Network *pNetwork, TLSConnectParams *params) {
...
 			len -= ret;
-		} else if (ret == 0 || (ret != MBEDTLS_ERR_SSL_WANT_READ && ret != MBEDTLS_ERR_SSL_WANT_WRITE && ret != MBEDTLS_ERR_SSL_TIMEOUT)) {
+		} else if(ret == MBEDTLS_ERR_SSL_WANT_READ || 
+				ret == MBEDTLS_ERR_SSL_WANT_WRITE ||
+				ret == MBEDTLS_ERR_SSL_TIMEOUT) {
+			if(has_timer_expired(&readTimer)) {
+				if(rxLen == 0U) {
+					return NETWORK_SSL_NOTHING_TO_READ;
+				} else {
+					return NETWORK_SSL_READ_TIMEOUT_ERROR;
```
leegeth, line 394: Would it be better if we pass the length of data successfully read back to the calling code in case of a timeout?

samples/linux/jobs_sample/aws_iot_config.h
```diff
@@ -70,4 +70,9 @@
 
 #define DISABLE_METRICS false ///< Disable the collection of metrics by setting this to true
 
+// TLS configs
+#define IOT_SSL_READ_TIMEOUT 10 ///< Timeout associated with underlying socket of TLS connection (set by mbedtls_ssl_conf_read_timeout)
+#define IOT_SSL_READ_RETRY_TIMEOUT_MS 10 ///< Minimum elapsed time before returning from iot_tls_read when pending data has not yet been received
+#define IOT_SSL_WRITE_RETRY_TIMEOUT_MS 10 ///< Minimum elapsed time before returning from iot_tls_write when pending data has not yet been written
```
leegeth, line 76: Does giving the same value to the retry-timeouts as the IOT_SSL_READ_TIMEOUT attempt a retry, incase the ssl_read times out? Doesn't it need to be higher than IOT_SSL_READ_TIMEOUT to retry at least once in case of a ssl read or write timeout?

yourslab: Sure, changing `IOT_SSL_READ_TIMEOUT` to `3`, allowing 4 maximum tries.

platform/linux/mbedtls/network_mbedtls_wrapper.c
yourslab: Partial writes/reads are never handled by the calling code, but I suppose it doesn't hurt to update the input parameter.

samples/linux/jobs_sample/aws_iot_config.h
```diff
@@ -71,7 +71,7 @@
 #define DISABLE_METRICS false ///< Disable the collection of metrics by setting this to true
 
 // TLS configs
-#define IOT_SSL_READ_TIMEOUT 10 ///< Timeout associated with underlying socket of TLS connection (set by mbedtls_ssl_conf_read_timeout)
+#define IOT_SSL_READ_TIMEOUT 3 ///< Timeout associated with underlying socket of TLS connection (set by mbedtls_ssl_conf_read_timeout)
```
leegeth, line 74: Is this in ms as well, If so can you please add that to the name to indicate the unit as well?

leegeth (approved):

dan4thewin (approved):

aggarg (approved):

leegeth (approved):

PR #1487, opened 2020-12-12
Ota/fix sudo issue

pvyawaha:
*Issue #, if available:*

*Description of changes:*

* This changes fixes the way path is built by using cwd instead of pwd without which the application could not be run in sudo.


platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -1,5 +1,5 @@
 /*
- * OTA PAL V2.0.0 (Release Candidate) for POSIX
+ * OTA PAL for POSIX V2.0.0
```
divekarshubham, line 2:
```suggestion
 * OTA PAL V2.0.0 (Release Candidate) for POSIX
```

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -30,6 +30,7 @@
 #include <errno.h>
 #include <assert.h>
 #include <libgen.h>
+#include <unistd.h>
```
divekarshubham, line 33: Is this still needed?

pvyawaha: Yes for getcwd

divekarshubham (approved):

cobusve (approved):

yanjos-dev (approved):

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -367,18 +373,27 @@ OtaPalStatus_t otaPal_CreateFileForRx( OtaFileContext_t * const C )
...
+                pFileName = getcwd( realFilePath, OTA_FILE_PATH_LENGTH_MAX );
+
+                if( pFileName == NULL )
+                {
+                    LogError( ( "Failed to get current working directory: %s", strerror( errno ) ) );
+                }
+                else
+                {
+                    /* Add the filename . */
+                    realFilePath[ strlen( realFilePath ) ] = '/';
```
gkwicker, line 394: This overwrites the `'\0'` at the end of `realFilePath` – it is no longer a valid NULL-terminated string.

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -367,18 +373,27 @@ OtaPalStatus_t otaPal_CreateFileForRx( OtaFileContext_t * const C )
...
+                    strncat( realFilePath, ( char * ) C->pFilePath, strlen( ( const char * ) C->pFilePath ) + 1U );
```
gkwicker, line 395: This call to `strncat()` will produce unpredictable results since `realFilePath` is no longer NULL-terminated after line 394.  In addition, you haven't verified that the storage for `realFilePath` is sufficient to hold the fully-qualified pathname you're building here.

To fix this, I would:

1. Change line 385 to: `pFileName = getcwd( realFilePath, OTA_FILE_PATH_LENGTH_MAX - 1 );`
1. Change lines 394-395 as follows:

```c
strncat( realFilePath, "/", 1 );  /* realFilePath is still NULL-terminated */
if( strlen( realFilePath ) + strlen( C->pFilePath ) < OTA_FILE_PATH_LENGTH_MAX ) )
{
    strncat( realFilePath, C->pFilePath, strlen( C->pFilePath ) );
}
```

Also note that you don't want to add 1 to the size (3rd argument) passed to `strncat()` here.  `man strncat` to see why.

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -541,22 +556,29 @@ OtaPalStatus_t otaPal_SetPlatformImageState( OtaFileContext_t * const C,
...
-        /* coverity[misra_c_2012_rule_21_6_violation] */
-        pPlatformImageState = fopen( imageStateFile, "w+b" );
+        if( pFileName == NULL )
+        {
+            LogError( ( "Failed to get current working directory: %s", strerror( errno ) ) );
+        }
+        else
+        {
+            /* Add the filename . */
+            strncat( imageStateFile, OTA_PLATFORM_IMAGE_STATE_FILE, sizeof( OTA_PLATFORM_IMAGE_STATE_FILE ) );
```
gkwicker, line 576: See my earlier comment – you're not verifying that `imageStateFile` is of sufficient size to hold both the current working directory and `OTA_PLATFORM_IMAGE_STATE_FILE` before creating the fully-qualified pathname here.  See my earlier suggestion as to how this might be done.

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -626,23 +648,30 @@ OtaPalStatus_t otaPal_ResetDevice( OtaFileContext_t * const C )
...
 
-    ( void ) C;
+    if( pFileName == NULL )
+    {
+        LogError( ( "Failed to get current working directory: %s", strerror( errno ) ) );
+    }
+    else
+    {
+        /* Add the filename . */
+        strncat( imageStateFile, OTA_PLATFORM_IMAGE_STATE_FILE, sizeof( OTA_PLATFORM_IMAGE_STATE_FILE ) );
```
gkwicker, line 669: See my earlier comment – you're not verifying that `imageStateFile` is of sufficient size to hold both the current working directory and `OTA_PLATFORM_IMAGE_STATE_FILE` before creating the fully-qualified pathname here.  See my earlier suggestion as to how this might be done.

gkwicker (changes requested):
See comments – also this new code contains 3 copies of essentially identical code to construct a fully-qualified pathname.  That should be factored into a `static` utility method and reused in this module.  This will make the code less complex, more readable, and easier to test.

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: Thanks, extracted file path generation into `getFilePathFromCWD()` and checked if the buffer is large enough to hold filepath + filename + `\` before performing `strncat()`

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: Thanks, extracted file path generation into `getFilePathFromCWD()` and checked if the buffer is large enough to hold filepath + filename + `\` before performing `strncat()`

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: Thanks, extracted file path generation into `getFilePathFromCWD()` and checked if the buffer is large enough to hold filepath + filename + `\` before performing `strncat()`

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -323,6 +332,33 @@ static OtaPalStatus_t otaPal_CheckFileSignature( OtaFileContext_t * const C )
...
+    else
+    {
+        /* Add the filename . */
+        if( strlen( pCompleteFilePath ) + strlen( pFileName ) + 2U > OTA_FILE_PATH_LENGTH_MAX )
+        {
+            LogError( ( "Insufficient space to generate file path" ) );
+        }
+        else
+        {
+            pCompleteFilePath[ strlen( pCompleteFilePath ) ] = '/';
```
gkwicker, line 356: `pCompleteFilePath` is no longer null-terminated after this line, so the `strncat()` on the next line may produce indeterminate results.  Also, you should __not__ add 1 to the length in the `strncat()` on line 357.  The length is the maximum number of characters you want copied, __not including__ the null terminator which is added after the last character copied.  [`man strncat` for more details](https://en.cppreference.com/w/c/string/byte/strncat).  Please change lines 356 & 357 as follows:

```
strncat( pCompleteFilePath, "/" );
strncat( pCompleteFilePath, pFileName, strlen( pFileName ) );
```

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -323,6 +332,33 @@ static OtaPalStatus_t otaPal_CheckFileSignature( OtaFileContext_t * const C )
...
+    if( pCurrentDir == NULL )
+    {
+        LogError( ( "Failed to get current working directory: %s", strerror( errno ) ) );
+    }
+    else
+    {
+        /* Add the filename . */
+        if( strlen( pCompleteFilePath ) + strlen( pFileName ) + 2U > OTA_FILE_PATH_LENGTH_MAX )
+        {
+            LogError( ( "Insufficient space to generate file path" ) );
```
gkwicker, line 352: You're not returning any status value from `getFilePathFromCWD()`, so calling code will have no idea that this error condition has occurred and will continue as normal (see line 420 for an example).  Please make `getFilePathFromCWD()` return some type of status and have all callers validate it before continuing.

gkwicker (changes requested):
Pointer handling prior to the call to `strncat()` is still wrong - see comments and suggested alternative.  Also, `getFilePathFromCWD()` needs to return some type of status so that callers know when an error has occurred.

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: created `OtaPalPathGenStatus_t` to track path generation status.

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/4504c031ecd34a8fcbc2cbf40250b1b7be519792/platform/posix/ota_pal/source/ota_pal_posix.c#L359

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -409,21 +414,26 @@ OtaPalStatus_t otaPal_CreateFileForRx( OtaFileContext_t * const C )
...
             }
 
+            if( status != OtaPalFileGenSuccess )
+            {
+                LogError( ( "Could not generate the absolute path for the file" ) );
+            }
+
             /* POSIX port using standard library */
             /* coverity[misra_c_2012_rule_21_6_violation] */
             C->pFile = fopen( ( const char * ) realFilePath, "w+b" );
```
gkwicker, line 439: If `status != OtaPalFileGenSuccess` on line 432, you're still opening an invalid filename here.  You should not be continuing here if the filename isn't valid.

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -587,7 +598,12 @@ OtaPalStatus_t otaPal_SetPlatformImageState( OtaFileContext_t * const C,
...
-        getFilePathFromCWD( imageStateFile, OTA_PLATFORM_IMAGE_STATE_FILE );
+        status = getFilePathFromCWD( imageStateFile, OTA_PLATFORM_IMAGE_STATE_FILE );
+
+        if( status != OtaPalFileGenSuccess )
+        {
+            LogError( ( "Could not generate the absolute path for the file" ) );
+        }
 
         /* POSIX port using standard library */
         /* coverity[misra_c_2012_rule_21_6_violation] */
```
gkwicker, line 609: If `status != OtaPalFileGenSuccess` on line 603, you're opening an invalid filename here.  You should not continue processing without a valid filename.

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -664,12 +680,18 @@ OtaPalImageState_t otaPal_GetPlatformImageState( OtaFileContext_t * const C )
...
-    getFilePathFromCWD( imageStateFile, OTA_PLATFORM_IMAGE_STATE_FILE );
+    status = getFilePathFromCWD( imageStateFile, OTA_PLATFORM_IMAGE_STATE_FILE );
+
+    if( status != OtaPalFileGenSuccess )
+    {
+        LogError( ( "Could not generate the absolute path for the file" ) );
+    }
 
     /* POSIX port using standard library */
     /* coverity[misra_c_2012_rule_21_6_violation] */
```
gkwicker, line 697: If `status != OtaPalFileGenSuccess` on line 691, you're opening an invalid filename here.  You should not continue processing in this case.

gkwicker (dismissed):
You're still not doing anything in these error cases.  The code as written will attempt to open invalid filenames.  See comments.

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: Refactored in https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1487/commits/a3d852db7cfbb536eddf4de696d8ec0f070a7702

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: Refactored in https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1487/commits/a3d852db7cfbb536eddf4de696d8ec0f070a7702

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: Refactored in https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1487/commits/a3d852db7cfbb536eddf4de696d8ec0f070a7702

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -317,6 +332,38 @@ static OtaPalStatus_t otaPal_CheckFileSignature( OtaFileContext_t * const C )
...
 }
 
+static OtaPalPathGenStatus_t getFilePathFromCWD( char * pCompleteFilePath,
+                                                 const char * pFileName )
+{
+    char * pCurrentDir = NULL;
+    OtaPalPathGenStatus_t status = OtaPalFileGenSuccess;
+
+    /* Get current directory. */
+    pCurrentDir = getcwd( pCompleteFilePath, OTA_FILE_PATH_LENGTH_MAX - 1 );
```
cobusve, line 342: What is the purpose of getting pCurrentDir here if this is never used in this function?

It seems to me like the intent is to append the pFileName to the current path, and write the result into pCompleteFilePath, but instead this function is appending the destinatioon string to the current File name?

platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -317,6 +332,38 @@ static OtaPalStatus_t otaPal_CheckFileSignature( OtaFileContext_t * const C )
...
+        /* Add the filename . */
+        if( strlen( pCompleteFilePath ) + strlen( pFileName ) + 2U > OTA_FILE_PATH_LENGTH_MAX )
+        {
+            LogError( ( "Insufficient space to generate file path" ) );
+            status = OtaPalBufferInsufficient;
+        }
+        else
+        {
+            strcat( pCompleteFilePath, "/" );
+            strncat( pCompleteFilePath, pFileName, strlen( pFileName ) );
```
cobusve, line 360: The lengths have already been checked above so I think it is more efficient to just do a simple strcat here. Strlen counts the entire length one char at a time a second time here for no additional benefit, we already know the string will fit be cause we checked above did we not?

platform/posix/ota_pal/source/ota_pal_posix.c
pvyawaha: pCurrentDir is used as a status to identify failure cases as getcwd returns NULL when it fails in which case the content of the buffer passed are undefined. The return pointer can optionally used in case of success but it is not required. 

@yanjos-dev  to comment on the test cases.

yanjos-dev: I agree with Prasad, the purpose is just to check the success of the getcwd command. Other than this, it's not being used in this function. The getcwd command is still impacting the code though because it's modifying the "pCompleteFilePath" input. 


As far as the testing, there are the E2E tests and the unit tests. The E2E tests are where we are checking the actual functionality is correct (it's doing the thing it's intended to do. e.g. creating a file). In the unit tests, they are more checking that the code is correctly handling failure and success cases, not that it's actually doing the thing it's supposed to do. So in the case of the unit tests, the getcwd command is being mocked and tested for failure/success values and modifying/not modifying the pCompleteFilePath input. The unit tests are like this due to the PAL having so much to do with file I/O calls and system calls, which are hard to directly force failures in a controlled/safe/reliable way.

platform/posix/ota_pal/source/ota_pal_posix.c
divekarshubham: Thanks, addressed in https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1487/commits/8e4a008c671b3a39f450ac191c011630c4c01f51

cobusve (approved):

yanjos-dev (approved):

Issue #1424, opened 2020-12-03
 ninja: error: unknown target 'install'


vijayrakesh13:
Hi,

I'm trying into integrate aws-iot-device-sdk-embedded-C into Yocto. do_install failed with an error "ninja: error: unknown target 'install'". Kindly please help me to fix the above issue.

`-- Configuring done
 -- Generating done
-- Build files have been written to: /home/poky/bsp/tmp/work/aarch64-poky-linux/aws/1.0-r0/build
[0/2] /home/poky/bsp/tmp/work/aarch64-poky-linux/aws/1.0-r0/recipe-sysroot-native/usr/bin/cmake -P /home/poky/bsp/tmp/work/aarch64-poky-linux/aws/1.0-r0/build/CMakeFiles/VerifyGlobs.cmake
ninja: error: unknown target 'install'
WARNING: /home/poky/bsp/tmp/work/aarch64-poky-linux/aws/1.0-r0/temp/run.do_install.15415:1 exit 1 from 'eval ${DESTDIR:+DESTDIR=${DESTDIR} }VERBOSE=1 cmake --build '/home/poky/bsp/tmp/work/aarch64-poky-linux/aws/1.0-r0/build' "$@" -- ${EXTRA_OECMAKE_BUILD}'
ERROR: Task (/home/poky/src/meta-application/recipes-bleapp/aws/aws_1.0.bb:do_install) failed with exit code '1'`

Thanks & Regards,
Vijay Rakesh.

yourslab:
Hi Vijay,

The `install` target is not yet readily available, but we are working on it:
https://github.com/aws/aws-iot-device-sdk-embedded-C/issues/309

Best regards,
Oscar

vijayrakesh13:
Hi Oscar,

Is there any other way to add AWS IOT SDK libraries to Yocto Build?

Thanks & Regards,
Vijay

yourslab:
You will need to include the sources and includes from the `source` directory of each library. From there, you can make an AWS IoT Recipe to integrate with your Yocto Build as [shown here](https://variwiki.com/index.php?title=Yocto_Hello_World#Sample_C_.22Hello.2C_world.21.22_program):

rpcme:
this is related to this issue (the last couple responses anyway)

https://github.com/aws/meta-aws/issues/89

There will be a recipe created that targets the eSDK.  Otherwise, customers can create a recipe that fulfills the requirement in their custom app layer.

[...]

Issue #1344, opened 2020-10-29
How does the install for the recent release 202009.00 work? 


Kristallfee:
Hello!

I try to include the aws iot emb c sdk into my project, and I saw a cmake-install.cmake file. Installing would make it easier to include. What do I need to configure and to call to install it? 

Thank you 

Simone

aggarw13:
Hi @Kristallfee, we do not support a **cmake-install.cmake** file or an installation target in the most recent release version [202009.00](https://github.com/aws/aws-iot-device-sdk-embedded-C/tree/202009.00). 

To understand your use-case, can you explain what functionality you are looking from an installation target? 
Are you looking for the header files to be installed in a system path (like `/usr/local/include` on Linux platforms) so that you can include them in your project? 
Also, are you looking for the libraries to be installed (as **.a**/static archives or **.so**/dynamic libraries) in the system (at `/usr/local/lib` on Linux platform)?

The CMake build configuration of the 202009.00 release does provide a way to build the **coreMQTT**, **Device Shadow** and **coreJSON** submodules as CMake libraries. Instead, we build the library files directly along with the demo source files (present under the `demos/` folder) in the demo project build configuration. 
For example, the [Device Shadow demo project](https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/202009.00/demos/shadow/shadow_demo_main/CMakeLists.txt#L23-L28) CMake configuration builds all the source files of the Device Shadow, coreMQTT and coreJSON libraries along with the demo source file. 

If you are using the 202009.00 release, you would not be able to build the submodules as libraries and install them from the present build configuration. If you are looking to build and install the libraries as static archives or dynamic libraries in your project, you can refer to `test/CMakeLists.txt` file of the submodule repositories for reference on CMake library build configuration. (For example, coreMQTT repository builds the MQTT library as a CMake library target [here](https://github.com/FreeRTOS/coreMQTT/blob/a867f4cf4819f3bc9a950e0f611e1db5f8ec7c42/test/CMakeLists.txt#L34-L49))

Kristallfee:
Hello @aggarw13, 

thank you for your answer! Yes, I was searching for a way to install the header files in a system path or a own defined path and for the libs to be installed. 

Thank you for your links, I will try to build them with the demos on my own.

Thank you, 
Simone

yourslab:
Hi @Kristallfee

The C SDK now supports installation.

[...]

Issue #309, opened 2019-03-13
cmake install command


shelterz:
CMake install command not found , would it be added in future?
Beacause i want use the sdk as a third party library in my own iot cmake projects.

gordonwang0:
Hi @shelterz,

Thanks for trying out the new SDK!

Yes, we will add an install command in the near future.

shelterz:
That sounds great!

rafajpet:
Hi,
How It looks like with cmake install functionality ?

abhidixi11:
Hello @rafajpet, 
It is not yet implemented, we do have a plan to implement it, but I don't have any official date for you at this time.
Thanks.

yourslab:
Hi @shelterz @rafajpet

The C SDK now supports installation.

Documentation can be found in the [README](https://github.com/aws/aws-iot-device-sdk-embedded-C#installation).

PR #1519, opened 2020-12-23
Create target for installation of libraries and headers into file system

yourslab:
This adds the feature requested in issue #309, #1344, and #1424. It works by creating targets for each individual library and platform abstractions that can then be installed into any directory of a file system.

For instance, the following commands will install only the coreMQTT and coreHTTP libraries in the default system header and library path:
```sh
cmake .. -DINSTALL_LIBS="MQTT;HTTP" -DINSTALL_PLATFORM_ABSTRACTIONS=0 -DINSTALL_TO_SYSTEM=1
sudo make install
```
```
[ 26%] Built target aws_iot_mqtt
[ 46%] Built target aws_iot_http
Install the project...
-- Install configuration: ""
-- Installing: /usr/local/lib/libaws_iot_http.so
-- Installing: /usr/local/lib/libaws_iot_mqtt.so
-- Installing: /usr/local/include/aws/core_http_client.h
-- Installing: /usr/local/include/aws/core_http_config_defaults.h
-- Installing: /usr/local/include/aws/transport_interface.h
-- Installing: /usr/local/include/aws/http_parser.h
-- Installing: /usr/local/include/aws/core_mqtt.h
-- Installing: /usr/local/include/aws/core_mqtt_config_defaults.h
-- Installing: /usr/local/include/aws/core_mqtt_serializer.h
-- Installing: /usr/local/include/aws/core_mqtt_state.h
-- Installing: /usr/local/include/aws/transport_interface.h
```

Once installed, the files can be easily linked and included as a system header:
```c
#include <aws/core_mqtt.h>
#include <aws/core_http_client.h>
```


README.md
```diff
@@ -297,6 +298,66 @@ When building with CMake, submodules are also recursively cloned automatically.
...
+cmake .. -DBUILD_DEMOS=0 -DBUILD_TESTS=0
+make install
+```
+Note that because `make install` will automatically build the `all` target, it may
+be useful to disable building demos and tests with `-DBUILD_DEMOS=0 -DBUILD_TESTS=0`
+unless they have already been configured.
+
+By default, the location of headers will be in the default system include path
+of the OS such as `/usr/local/include`. For libraries, this will likely
+be `/usr/local/lib`.
```
aggarw13, line 315: This should mention about these locations being specific to Linux environments.

platform/posix/CMakeLists.txt
```diff
@@ -9,6 +9,13 @@ target_include_directories( clock_posix
                               PRIVATE
                                 ${PLATFORM_DIR}/include )
 
+# Install clock abstraction as library.
+if(INSTALL_PLATFORM)
+    install(TARGETS
```
aggarw13, line 14: Shouldn't the include path destination be specified as well?

CMakeLists.txt
```diff
@@ -29,14 +29,20 @@ include( "demos/logging-stack/logging.cmake" )
...
+        "Set this to ON to build demo executables."
+        ON )
 option( BUILD_CLONE_SUBMODULES
         "Set this to ON to automatically clone any required Git submodules. When OFF, submodules must be manually cloned."
         ON )
 option( DOWNLOAD_CERTS
         "Set this to ON to automatically download certificates needed to run the demo. When OFF, certificates must be manually downloaded."
         ON )
+option( INSTALL_PLATFORM
+        "Set this to ON to install any platform-related dependencies."
```
aggarw13, line 44:
```suggestion
        "Set this to ON to install to the system the POSIX platform-specific libraries that are used by demos in C-SDK."
```

README.md
```diff
@@ -297,6 +298,66 @@ When building with CMake, submodules are also recursively cloned automatically.
 can be passed as a CMake flag to disable this functionality. This is useful when you'd like to build CMake while using a
 different commit from a submodule.
 
+### Installation
```
aggarw13, line 301: This section should also mention about the `INSTALL_PLATFORM_LIBS` flag

CMakeLists.txt
```diff
@@ -29,14 +29,20 @@ include( "demos/logging-stack/logging.cmake" )
...
+option( BUILD_DEMOS
+        "Set this to ON to build demo executables."
+        ON )
 option( BUILD_CLONE_SUBMODULES
         "Set this to ON to automatically clone any required Git submodules. When OFF, submodules must be manually cloned."
         ON )
 option( DOWNLOAD_CERTS
         "Set this to ON to automatically download certificates needed to run the demo. When OFF, certificates must be manually downloaded."
         ON )
+option( INSTALL_PLATFORM
```
aggarw13, line 43: Pedantic suggestion for better readability
```suggestion
option( INSTALL_PLATFORM_LIBS
```

README.md
```diff
@@ -297,6 +298,66 @@ When building with CMake, submodules are also recursively cloned automatically.
...
+-DDEFENDER_CUSTOM_CONFIG_PATH="defender-config-path"
+-DSHADOW_CUSTOM_CONFIG_PATH="shadow-config-path"
+-DJOBS_CUSTOM_CONFIG_PATH="jobs-config-path"
+-DOTA_CUSTOM_CONFIG_PATH="ota-config-path"
+-DHTTP_CUSTOM_CONFIG_PATH="http-config-path"
+-DJSON_CUSTOM_CONFIG_PATH="json-config-path"
+-DMQTT_CUSTOM_CONFIG_PATH="mqtt-config-path"
+-DPKCS_CUSTOM_CONFIG_PATH="pkcs-config-path"
+```
+Note that the file name of the header should not be included in the path.
```
aggarw13, line 359: Should the name of the configs be changed to `...CONFIG_FILE_FOLDER_PATH` to make it clear? The `CONFIG_PATH` suffix suggests providing the entire path to the config file..

README.md
```diff
@@ -297,6 +298,66 @@ When building with CMake, submodules are also recursively cloned automatically.
 can be passed as a CMake flag to disable this functionality. This is useful when you'd like to build CMake while using a
 different commit from a submodule.
 
+### Installation
+
+The C SDK libraries and platform abstractions can be installed to a file system
```
aggarw13, line 303: How about providing the ability to only install the user provided list of libraries instead of all libraries? Some users may not necessarily want their file system to contain HTTP library if their application only uses MQTT connections

platform/posix/transport/CMakeLists.txt
```diff
@@ -40,6 +40,15 @@ target_link_libraries( openssl_posix
...
                           ${CMAKE_DL_LIBS} )
 
+# Install transport implementations as libraries.
+if(INSTALL_PLATFORM)
+    install(TARGETS
+      openssl_posix
+      plaintext_posix
+      sockets_posix
+      LIBRARY DESTINATION "${CSDK_LIB_INSTALL_PATH}")
+endif()
```
aggarw13, line 50:
```suggestion
if(INSTALL_PLATFORM)
    install(TARGETS
      openssl_posix
      plaintext_posix
      sockets_posix
      LIBRARY DESTINATION "${CSDK_LIB_INSTALL_PATH}"
      INCLUDES DESTINATION )
endif()
```

tools/install.cmake
```diff
@@ -0,0 +1,157 @@
...
+                                    PRIVATE ${${config_prefix}_CUSTOM_CONFIG_PATH})
+    else()
+        target_compile_definitions("${library_name}" PRIVATE -D${config_prefix}_DO_NOT_USE_CUSTOM_CONFIG)
+        # PKCS11 requires a config so include the one from the demos by default.
+        if(${config_prefix} STREQUAL "PKCS")
+            target_include_directories("${library_name}" PRIVATE
+                                        ${DEMOS_DIR}/pkcs11/common/include
+                                        ${LOGGING_INCLUDE_DIRS})
+            target_compile_definitions("${library_name}" PUBLIC
+                                       -DMBEDTLS_CONFIG_FILE="mbedtls_config.h")
```
aggarw13, line 103: If PKCS11 requires a config file, then the user should be required to pass it instead of the one from our demos being used (imo).

I think it would be important to support selective library installation so that the users that do not want to install PKCS11 and also don't want to provide custom configs for any of the other libraries they want, they don't have to be forced to provide custom config for PKCS11

README.md
```diff
@@ -297,6 +298,66 @@ When building with CMake, submodules are also recursively cloned automatically.
...
+unless they have already been configured.
+
+By default, the location of headers will be in the default system include path
+of the OS such as `/usr/local/include`. For libraries, this will likely
+be `/usr/local/lib`.
+
+Upon entering `make install`, the location of each library will be specified first
+followed by the location of all installed headers:
+```
+-- Installing: /usr/local/lib/libaws_iot_defender.so
```
aggarw13, line 320: Should we support building and installing libraries as static archives if user wants? If that increases scope of this PR, it can be done in a follow-up PR.

platform/posix/CMakeLists.txt
yourslab: Yes, it is possible to also install the headers in this file. However, what ends up happening is that the `make install` output will no longer group libraries and headers together. To fix that, the headers are installed here instead:
https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1519/files/96807066b1b35fd543d45f6a57063325cc28dd69#diff-7021d478f66d585e3f9025b7212b1ca40b1db63995592041e28b5dd819c06e53R154-R157
Actually, I did want to also install all libraries in this file but it is not backwards compatible with older versions of CMake.

README.md
yourslab: That is already a feature of CMake and can be specified with `-DBUILD_SHARED_LIBS=0`

platform/posix/CMakeLists.txt
aggarw13: >  install all libraries in this file but it is not backwards compatible with older versions of CMake.
What aspect about installing all libraries in the file is not backwards compatible with older version of CMake?

tools/install.cmake
yourslab: I think it's fine if it is using the configs from the demo as those configs are specific to POSIX. I am supporting selective library installation though.

tools/install.cmake
```diff
@@ -0,0 +1,157 @@
+# Creates an install target to allow users to include CSDK as a set of shared libraries
+
+set(FILEPATH_LOCATIONS
```
aggarw13, line 3: A CI job to validate the `make install` target will be valuable

README.md
yourslab: Actually, I would prefer `CUSTOM_CONFIG_DIRECTORY`. I don't know why it didn't come to mind earlier.

yourslab (commented):
test

platform/posix/CMakeLists.txt
yourslab: The cmake file that calls `add_library()` for a target is also where `install()` needs to be called for that target in older versions of cmake.

.github/workflows/ci.yml
```diff
@@ -66,6 +66,23 @@ jobs:
...
+      - name: Configure CMake build
+        run: |
+          curl https://cmake.org/files/v3.2/cmake-3.2.0-Linux-x86_64.tar.gz -o cmake.tar.gz
+          tar -xf cmake.tar.gz
+          mkdir build && cd build
+          ../cmake-3.2.0-Linux-x86_64/bin/cmake .. \
+          -DBUILD_DEMOS=0 -DBUILD_TESTS=0
+      - name: Install libraries
+        run: |
+          make install
```
aggarw13, line 85: There should be a check for verifying that the libraries and headers have been installed in the target directories. Instead of checking that each of the libraries has been installed, maybe only a handful of libraries can be installed and verified in a separate run of `make install`.
Also, it would be good to test the `CSDK_HEADER_INSTALL_PATH` and `DCSDK_LIB_INSTALL_PATH` variables.

README.md
```diff
@@ -297,6 +298,77 @@ When building with CMake, submodules are also recursively cloned automatically.
...
+following flags through CMake:
+```sh
+cmake .. -DBUILD_DEMOS=0 -DBUILD_TESTS=0 \
+-DCSDK_HEADER_INSTALL_PATH="header-path" -DCSDK_LIB_INSTALL_PATH="lib-path"
+make install
+```
+
+By default, `-DINSTALL_LIBS` will be set so that every C-SDK library is installed,
+but you may pass this flag yourself and exclude any library that you don't need
+installed from this list:
```
aggarw13, line 349:
```suggestion
but you may define this flag to install only a subset of the C-SDK libraries from the following list:
```

.github/workflows/ci.yml
```diff
@@ -66,6 +66,32 @@ jobs:
...
+          mkdir installed-headers-dir installed-libs-dir
+          curl https://cmake.org/files/v3.2/cmake-3.2.0-Linux-x86_64.tar.gz -o cmake.tar.gz
+          tar -xf cmake.tar.gz
+          mkdir build && cd build
+          ../cmake-3.2.0-Linux-x86_64/bin/cmake .. \
+          -DBUILD_DEMOS=0 -DBUILD_TESTS=0 \
+          -DCSDK_LIB_INSTALL_PATH="../installed-headers-dir" -DCSDK_LIB_INSTALL_PATH="../installed-headers-dir"
+      - name: Install libraries
+        run: |
+          sudo make -C build/ install
```
aggarw13, line 87: (Nit) Is `sudo` required for installing in the non-system location?

yourslab: Not necessary but I did this because `sudo` is also used in the README

.github/workflows/ci.yml
```diff
@@ -66,6 +66,31 @@ jobs:
...
+          -DCSDK_HEADER_INSTALL_PATH="../installed-headers-dir" -DCSDK_LIB_INSTALL_PATH="../installed-headers-dir"
+      - name: Install libraries and verify install location
+        run: |
+          sudo make -C build/ install
+          while IFS="" read -r p || [ -n "$p" ]
+          do
+            if [[ $p != "$PWD/installed-headers-dir"* ]] && [[ $p != "$PWD/installed-libs-dir"* ]]; then
+              exit 1
+            fi
+          done <build/install_manifest.txt
```
aggarw13, line 93: It would be good to test the `INSTALL_LIBS` variable as well to check that it only installs the specified libraries

yourslab: Ideally but that is a trickier task... It leads to an SSoT problem where we need to list each header and source of a library then cross-check with what was outputted by the manifest. Anyway, it is unlikely this functionality will change by code changes that are made later on unless `tools/install.cmake` is edited.

yourslab: I have added a comment to `tools/install.cmake` though:
```
When making changes to this file, please verify that -DINSTALL_LIBS="prefix-name" still works.
```

CMakeLists.txt
```diff
@@ -29,14 +29,20 @@ include( "demos/logging-stack/logging.cmake" )
 
 # Configure options to always show in CMake GUI.
 option( BUILD_TESTS
-        "Set this to ON to build both demo and test executables. When OFF, only demo executables are built."
+        "Set this to ON to build test executables."
         OFF )
+option( BUILD_DEMOS
```
aggarw13, line 34: The `build-check-demos` job will have to be updated to pass this flag so that the demo targets are available in the job run

.github/workflows/ci.yml
```diff
@@ -66,6 +66,31 @@ jobs:
...
+          curl https://cmake.org/files/v3.2/cmake-3.2.0-Linux-x86_64.tar.gz -o cmake.tar.gz
+          tar -xf cmake.tar.gz
+          mkdir build && cd build
+          ../cmake-3.2.0-Linux-x86_64/bin/cmake .. \
+          -DBUILD_DEMOS=0 -DBUILD_TESTS=0 \
+          -DCSDK_HEADER_INSTALL_PATH="$PWD/../installed-headers-dir" -DCSDK_LIB_INSTALL_PATH="$PWD/../installed-headers-dir"
+      - name: Install libraries and verify install location
+        run: |
+          cd build && sudo make install
+          while IFS="" read -r p || [ -n "$p" ]
```
aggarw13, line 88: A comment explaining about the of `install_manifest.txt` and the meaning of this loop will be helpful for future reference.

README.md
```diff
@@ -297,6 +298,78 @@ When building with CMake, submodules are also recursively cloned automatically.
...
+following flags through CMake:
+```sh
+cmake .. -DBUILD_DEMOS=0 -DBUILD_TESTS=0 \
+-DCSDK_HEADER_INSTALL_PATH="header-path" -DCSDK_LIB_INSTALL_PATH="lib-path"
+sudo make install
+```
+
+POSIX platform abstractions are used together with the C-SDK libraries in the demos.
+By default, these abstractions are also installed but can be excluded by passing
+the flag: `-DINSTALL_PLATFORM_ABSTRACTIONS=0`.
```
aggarw13, line 357: Shouldn't the installation of the platform libraries be disabled by default to prevent libraries not specified in the `INSTALL_LIBS` variable from being installed to the system?

tools/install.cmake
```diff
@@ -0,0 +1,170 @@
...
+# Install platform abstractions as shared libraries if enabled.
+if(INSTALL_PLATFORM)
+    set(PLATFORM_DIRECTORIES
+            ${COMMON_TRANSPORT_INCLUDE_PUBLIC_DIRS}
+            ${PLATFORM_DIR}/posix/ota_pal/source/include)
+    # Create target for POSIX port of OTA if LIB_RT is installed.
+    if(NOT(${LIB_RT} STREQUAL "LIB_RT-NOTFOUND"))
+        add_library(ota_posix
+                     ${OTA_OS_POSIX_SOURCES})
+        target_link_libraries(ota_posix PUBLIC ${LIB_RT} ota_pal)
```
aggarw13, line 143: Where is the `ota_pal` library built?

yourslab: In `platform/posix/ota_pal`, and it is an implementation specific to POSIX.

README.md
yourslab: Because everything is installed by default unless otherwise specified, I decided to also enable this option by default. Also, I feel like `-DINSTALL_LIBS` should not affect `-DINSTALL_PLATFORM_ABSTRACTIONS` the same way `-DBUILD_DEMOS` does not affect `-DBUILD_TESTS`.

.github/workflows/ci.yml
aggarw13: >  It leads to an SSoT problem where we need to list each header and source of a library then cross-check with what was outputted by the manifest.

That's a valid point. Instead of checking for each header and library file installation in the target directories, a simpler check of just the presence of the library (passed through the `INSTALL_LIBS` flag) in the file names can be added, along with just a sanity check that no file carries the name of a library that is not expected to be installed.

aggarw13 (approved):

.github/workflows/ci.yml
yourslab: Sure, I've added this check for MQTT;DEFENDER

.github/workflows/ci.yml
```diff
@@ -93,6 +93,37 @@ jobs:
             fi
           # Each line of install_manifest.txt contains the location of an installed library or header
           done <install_manifest.txt
+          # Uninstall and reset build.
+          sudo xargs rm < install_manifest.txt || echo "transport_interface.h is installed twice because of duplicate copies in spoke repos."
+          cd .. && rm -rf build
+      - name: Configure CMake build for default install location and selective installation
+        run: |
+          curl https://cmake.org/files/v3.2/cmake-3.2.0-Linux-x86_64.tar.gz -o cmake.tar.gz
+          tar -xf cmake.tar.gz
```
aggarw13, line 102: Cmake is already installed on line 79
```suggestion
          curl https://cmake.org/files/v3.2/cmake-3.2.0-Linux-x86_64.tar.gz -o cmake.tar.gz
          tar -xf cmake.tar.gz
```

.github/workflows/ci.yml
```diff
@@ -93,6 +93,37 @@ jobs:
...
+          mkdir build && cd build
+          ../cmake-3.2.0-Linux-x86_64/bin/cmake .. \
+          -DBUILD_DEMOS=0 -DBUILD_TESTS=0 \
+          -DINSTALL_LIBS="http;defender" -DINSTALL_PLATFORM_ABSTRACTIONS=0
+      - name: Install libraries and verify default install location
```
aggarw13, line 107:
```suggestion
      - name: Install selected libraries and verify that only they are installed in default system locations
```

.github/workflows/ci.yml
```diff
@@ -93,6 +93,37 @@ jobs:
...
+              exit 1
+            fi
+          # Each line of install_manifest.txt contains the location of an installed library or header
+          done <install_manifest.txt
+      - name: Verify filenames from selective build
+        run: |
+          cd build && sudo make install
+          while IFS="" read -r p || [ -n "$p" ]
+          do
+            if [[ $p != *"mqtt"* ]] && [[ $p != *"defender"* ]]; then
```
aggarw13, line 122: Should be checked for HTTP which is passed in `INSTALL_LIBS`
```suggestion
            if [[ $p != *"http"* ]] && [[ $p != *"defender"* ]]; then
```

.github/workflows/ci.yml
yourslab: This step doesn't actually verify that only they are installed in default system locations. That is done after this step.

aggarw13 (approved):

aggarw13 (dismissed):

n9wxu (approved):

sukhmanm (approved):

PR #1529, opened 2021-01-22
Make periodic jobs service requests disabled by default

dan4thewin:
To avoid possible throttling when scaling up code based on this demo, jobs polling and repeated job updates are tied to command line arguments and disabled by default.  This PR also drops redundant subscriptions.

### DIFF
--- a/demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c
+++ b/demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c
@@ -125,7 +125,7 @@ static void usage( const char * programName )
              );
     fprintf( stderr,
              "\nusage: %s "
-             "[-o] -n name -h host [-p port] {--cafile file | --capath dir} --certfile file --keyfile file\n"
+             "[-o] -n name -h host [-p port] {--cafile file | --capath dir} --certfile file --keyfile file [--pollinv seconds] [--updateinv seconds]\n"
              "\n"
              "-o : run once, exit after the first job is finished.\n"
              "-n : thing name\n"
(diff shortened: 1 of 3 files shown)


demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c
```diff
@@ -1200,12 +1214,9 @@ int main( int argc,
...
+    if( subscribe( h, JobsNextJobChanged ) == false )
     {
-        if( subscribe( h, topics[ i ] ) == false )
-        {
-            errx( 1, "fatal error" );
-        }
+        errx( 1, "fatal error" );
     }
 
     if( sendStartNext( h ) == false )
```
mtalreja-a, line 1222: sendStartNext is a POST api which scales separate from Describe Job with $next as the ID which is a GET api. sendStartNext transactions will compete with the Update api which might result in throttling at scale.

dan4thewin: The jobs library only uses MQTT.

demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c
```diff
@@ -125,21 +125,26 @@ static void usage( const char * programName )
...
+             "              certificate based communication.\n"
+             "--capath    : path to a directory containing trusted CA certificates to enable encrypted\n"
+             "              communication.  Defaults to %s.\n"
+             "--certfile  : client certificate for authentication in PEM format.\n"
+             "--keyfile   : client private key for authentication in PEM format.\n",
              DEFAULT_CA_DIRECTORY );
+    fprintf( stderr,
+             "--pollinv   : after this many idle seconds, request a job.  Default is not to poll.\n"
+             "--updateinv : after this many seconds running a job, resend the current status to the jobs service.\n"
+             "              Default is not to resend.\n\n"
```
yourslab, line 146: Without reading the code, it was not immediately obvious that 0 was mapped to this default.
```suggestion
             "              Default value is 0, which disables periodic resends.\n\n"
```
Feel free to change the wording or document the struct instead. Same also applies to `pollinv`.

dan4thewin: changed

yourslab (approved):

n9wxu (approved):

demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c
mtalreja-a: Apologies - I meant to say that the category of the API is for modification instead of retrieval. SendStartNext and DescribeJob are handled and scaled differently.

dan4thewin: I'll try to understand this better.  Any such changes will merit a new PR.

PR #1538, opened 2021-02-10
Update OTA_Shutdown API calls

pvyawaha:
*Issue #, if available:*

*Description of changes:*
This change updates the OTA_Shutdown calls by adding a new parameter for unsubscribe flag.
The change is in both OTA demos - MQTT and HTTP

Dependency for this change is https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/154

OTA Submodule is updated in this PR which has above dependency.


yanjos-dev (commented):
After the change in the OTA repo that this depends on is merged, can you move the submodule pointer to point to it in this PR? That way we don't break the demo in between PRs. I'll approve this after that.

yanjos-dev (approved):

dan4thewin (approved):

PR #1540, opened 2021-02-12
Update subscription handling in OTA demos

pvyawaha:
*Issue #, if available:*

*Description of changes:*

* Uses MQTT match topic utility function for OTA topic type 
* Adds logging for other topics which are not handled 
* Refactors the OTA job callback function

### DIFF
--- a/demos/ota/ota_demo_core_http/ota_demo_core_http.c
+++ b/demos/ota/ota_demo_core_http/ota_demo_core_http.c
@@ -247,7 +247,7 @@
 /**
  * @brief The common prefix for all OTA topics.
  */
-#define OTA_TOPIC_PREFIX                 "$aws/things/"
+#define OTA_TOPIC_PREFIX                 "$aws/things/" CLIENT_IDENTIFIER "/"
 
 /**
  * @brief The string used for jobs topics.
@@ -258,11 +258,6 @@
  * @brief The string used for streaming service topics.
  */
 #define OTA_TOPIC_STREAM                 "streams"
-
-/**
- * @brief The length of #OTA_TOPIC_PREFIX
- */
-#define OTA_TOPIC_PREFIX_LENGTH          ( ( uint16_t ) ( sizeof( OTA_TOPIC_PREFIX ) - 1U ) )
 
 /**
  * @brief HTTP response codes used in this demo.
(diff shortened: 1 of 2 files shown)


demos/ota/ota_demo_core_http/ota_demo_core_http.c
```diff
@@ -247,7 +247,7 @@
 /**
  * @brief The common prefix for all OTA topics.
  */
-#define OTA_TOPIC_PREFIX                 "$aws/things/"
+#define OTA_TOPIC_PREFIX                 "$aws/things/" CLIENT_IDENTIFIER "/"
```
ravibhagavandas, line 250: Do we need to put thing name as part of the prefix here. Some devices could be reading thing name from flash then the preprocessing fails. Could this be replaced with a wildcard (as thing name is always constant for this device). Ex: `$aws/things/+/jobs/$next/get/accepted. 
This way the topic filters can be defined as string constants and demo functions needs to call MQTT_MatchTopic() to get the right filter.

demos/ota/ota_demo_core_http/ota_demo_core_http.c
```diff
@@ -808,35 +800,87 @@ static void otaAppCallback( OtaJobEvent_t event,
             LogDebug( ( "Received invalid callback event from OTA Agent." ) );
     }
 }
+
+jobMessageType_t getJobMessageType( const char * pTopicName,
```
ravibhagavandas, line 804: These functions can be preferably moved to helper functions or within mqtt implementation of lib directory.

demos/ota/ota_demo_core_http/ota_demo_core_http.c
pvyawaha: Update for both OTA MQTT and OTA HTTP demos

demos/ota/ota_demo_core_http/ota_demo_core_http.c
pvyawaha: This will be covered in another PR.

demos/ota/ota_demo_core_mqtt/ota_demo_core_mqtt.c
```diff
@@ -677,36 +673,86 @@ static void otaAppCallback( OtaJobEvent_t event,
...
 
     assert( pPublishInfo != NULL );
     assert( pContext != NULL );
 
     ( void ) pContext;
 
-    LogInfo( ( "Received job message callback, size %ld.\n\n", pPublishInfo->payloadLength ) );
-
-    pData = otaEventBufferGet();
+    jobMessageType = getJobMessageType( pPublishInfo->pTopicName, pPublishInfo->topicNameLength );
```
ravibhagavandas, line 726: seems the handler code `mqttJobCallback` is same for both job message types. could we as well remove this ?

pvyawaha: This is basically to identify the two job topics and other topics , and extending it in future.

ravibhagavandas: Agree, we could then make this a generic helper function in future.

ravibhagavandas (approved):

n9wxu (approved):

PR #1554, opened 2021-02-24
Suppress stringop-truncation warning for HTTP demos

yourslab:
The coreHTTP library uses `strncpy`, causing a warning in the HTTP demos that utilize NULL-terminated strings. This change suppresses that warning for those demos, enabling us to use `ubuntu-latest` for Github Actions.


yourslab:
This can only be fixed by adding it to coreHTTP cmake

PR #1560, opened 2021-02-25
Validate Github tags and releases of spoke repos in release verification script

yourslab:
Since all spoke repos are being tagged and released before execution of the MCM, this updates the script to verify that each of the individual spoke repos has a release and tag named after the version specified in the config.

### DIFF
--- a/tools/release/release-verify.py
+++ b/tools/release/release-verify.py
@@ -135,7 +135,6 @@ def validate_checks(configs) -> list:
     repo_paths.append(f"{CSDK_ORG}/{CSDK_REPO}")
     return repo_paths
 
-
 def validate_ci():
     """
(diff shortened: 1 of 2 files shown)


tools/release/release-verify.py
```diff
@@ -325,6 +354,9 @@ def main():
     # Check that only qualified branches exist in each library repo.
     validate_branches(repo_paths)
 
+    # Check that library repos have a tag and release for each specified version.
+    validate_tags_and_releases(repo_paths, configs)
```
leegeth, line 358: I think it will be better to add cmdl option to turn this off if required. It will be helpful if we ever go back to releasing libraries with CSDK.

leegeth (approved):

aggarg (approved):

PR #1582, opened 2021-03-09
Add example policies for mqtt and shadow

muneebahmed10:
*Description of changes:*
Add minimal policies for MQTT Mutual Auth and Shadow demos for AWS IoT. The `[AWS_REGION]` and `[AWS_ACCOUNT_ID]` placeholders need to be replaced by the user.


rpcme:
For shadow: why are we not using ${iot:Connection.Thing.ThingName} for client id?
For mqtt: is there a reason why we are promoting not using Thing attached to certificate in IoT Core?  I think we should be using ${iot:Connection.Thing.ThingName} in the mqtt policy as well

leegeth:
> For shadow: why are we not using ${iot:Connection.Thing.ThingName} for client id?
> For mqtt: is there a reason why we are promoting not using Thing attached to certificate in IoT Core? I think we should be using ${iot:Connection.Thing.ThingName} in the mqtt policy as well

@rpcme
Any unique client ID can be used for the connection to the AWS IoT MQTT broker. It is not tightly coupled to the ThingName. If we change our demos to use ThingName as the client ID, it may give an impression of only ThingName can be used as client ID. However, we may be able to update our demo documentation to handle that. This [AWS documentation](https://docs.aws.amazon.com/iot/latest/developerguide/audit-chk-conflicting-client-ids.html#audit-chk-conflicting-client-ids-how-to-fix) and it has 2 recommendations for a client ID, ThingName or UUID. 

What are the advantages you see when using ThingName as client ID in our demos? Associating a certificate and policy to a Thing easily?

muneebahmed10:
These policies are for the demos as they are written now, and currently the demos don't use the thing name as the client identifier. We can update the policies accordingly if we want to make that change in the future.

As for the MQTT mutual auth demo, it wasn't designed for AWS IoT specifically, but any mutually authenticated broker, so it doesn't use thing names. That doesn't stop the user from setting the client identifier as the thing name, but it isn't required by either the demo or AWS IoT.

leegeth (dismissed):

leegeth (dismissed):

leegeth (dismissed):

leegeth (approved):

n9wxu (approved):

Issue #1584, opened 2021-03-09
Emit configurable options to demo output


rpcme:
It is useful to emit configurable options to standard output so the individual running the demo knows which options have been consumed by the build process.  For example, if the individual running the demo requires a specific client id, the individual cannot verify that the configuration has been made at runtime.

Specifically, I am running the mqtt_demo_mutual_auth demonstration and have changed the client id in the header file, but the expected behavior did not change. I have no knowledge if the configuration has been included in the build or if I made the configuration in the wrong place and I'm experiencing the same behavior.

yourslab:
The pull request to fix this issue has been merged.
New output is as follows:
```
Using values in test_config.h to define the following macros for http_system_test:
ROOT_CA_CERT_PATH
SERVER_HOST
HTTPS_PORT
Found the following definitions for http_system_test: ROOT_CA_CERT_PATH, SERVER_HOST, HTTPS_PORT
All required definitions for http_system_test were found - Adding to default target.

Using values in test_config.h to define the following macros for mqtt_system_test:
BROKER_PORT
Using the passed CMake arguments to define the following macros for mqtt_system_test:
CLIENT_IDENTIFIER = three
BROKER_ENDPOINT = foo
Found the following definitions for mqtt_system_test: CLIENT_IDENTIFIER, BROKER_ENDPOINT, BROKER_PORT
WARNING: mqtt_system_test missing definitions for macros: ROOT_CA_CERT_PATH, CLIENT_CERT_PATH, CLIENT_PRIVATE_KEY_PATH
To run mqtt_system_test, define required macros in mqtt_system_test/demo_config.h.
```

Issue #1583, opened 2021-03-09
Update MQTT client ID for all MQTT demos via Cmake


rpcme:
Instead of having -DTHING_NAME cmake property to update the thing name for a subset of MQTT related demos, use -DCLIENT_ID property to change this value for all demos.

There are three reasons, and possibly more.
1. The change aligns the demos with best practice for the AWS IoT thing name to match the MQTT Client ID. 
2. The change makes life easier for developers needing to modify the client id for all MQTT demos.
3. The change applies consistent build and runtime behavior across all demos.

yourslab:
The pull request to fix this issue has been merged

PR #1596, opened 2021-03-16
Create a CMake macro for defining aliases for configuration parameters

yourslab:
This change allows aliases to be defined for macros by defining a macro called `set_alias`.

This is done to fix issue #1583, allowing `THING_NAME` to be an alias for `CLIENT_IDENTIFIER`.

It also fixes issue #1584 so that the values of any parameters or the file from which they are retrieved are logged. Here's an example of the log output:
```
Using values in test_config.h to define the following macros for http_system_test:
ROOT_CA_CERT_PATH
SERVER_HOST
HTTPS_PORT
Found the following definitions for http_system_test: ROOT_CA_CERT_PATH, SERVER_HOST, HTTPS_PORT
All required definitions for http_system_test were found - Adding to default target.

Using values in test_config.h to define the following macros for mqtt_system_test:
BROKER_PORT
Using the passed CMake arguments to define the following macros for mqtt_system_test:
CLIENT_IDENTIFIER = three
BROKER_ENDPOINT = foo
Found the following definitions for mqtt_system_test: CLIENT_IDENTIFIER, BROKER_ENDPOINT, BROKER_PORT
WARNING: mqtt_system_test missing definitions for macros: ROOT_CA_CERT_PATH, CLIENT_CERT_PATH, CLIENT_PRIVATE_KEY_PATH
To run mqtt_system_test, define required macros in mqtt_system_test/demo_config.h.
```


tools/cmake/utility.cmake
```diff
@@ -68,3 +68,20 @@ function(set_macro_definitions)
         endif()
     endforeach()
 endfunction()
+
+macro(set_alias config_name)
```
aggarw13, line 72: Can you provide documentation for this function on the effects this function has on definition of configs

demos/defender/defender_demo_json/CMakeLists.txt
```diff
@@ -42,9 +42,9 @@ set_macro_definitions(TARGETS ${DEMO_NAME}
                         "ROOT_CA_CERT_PATH"
                         "CLIENT_CERT_PATH"
                         "CLIENT_PRIVATE_KEY_PATH"
-                        "CLIENT_IDENTIFIER"
                         "THING_NAME"
```
aggarw13, line 45: As part of best practices for connection with IoT Core, the Client ID should be the same as the Thing Name. 
So, should we make the CLIENT_IDENTIFIER as the required config but THING_NAME as the optional config?

yourslab: Required in the context of this function means that a definition is required for it to compile without error. In that case, I think both of these values should be required. In fact, nothing in the defender demo should be optional.

demos/jobs/jobs_demo_mosquitto/CMakeLists.txt
```diff
@@ -53,32 +53,8 @@ target_include_directories(
...
-            CLIENT_CERT_PATH="${CLIENT_CERT_PATH}"
-    )
-endif()
-if(CLIENT_PRIVATE_KEY_PATH)
-    target_compile_definitions(
-        ${DEMO_NAME} PRIVATE
-            CLIENT_PRIVATE_KEY_PATH="${CLIENT_PRIVATE_KEY_PATH}"
-    )
-endif()
 set_macro_definitions(TARGETS ${DEMO_NAME}
```
aggarw13, line 56: The jobs demo does not expose connection setting macros, so the `set_macro_definitions` call can be removed

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog for AWS IoT Device SDK for Embedded C
 
+## main
```
aggarw13, line 3: Thanks for updating the CHANGELOG 🙂

tools/cmake/utility.cmake
```diff
@@ -1,5 +1,11 @@
 include(CheckSymbolExists)
 
+# This function can be called to define a set of required and optional macros for a build target.
+#
+# If a CMake variable is defined with the same name as a required macro, it will be
```
aggarw13, line 5: (Unrelated to the changes in this PR) The `CMAKE_REQUIRED_DEFINITIONS` list in this utility function seems unused. Can that be removed?

demos/jobs/jobs_demo_mosquitto/CMakeLists.txt
yourslab: I believe it can be passed through here
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/main/demos/jobs/jobs_demo_mosquitto/jobs_demo_mosquitto.c#L430-L446

However, I set these parameters as optional as they can be passed at runtime.

tools/cmake/utility.cmake
yourslab: It is actually used by the `check_symbol_exists` function. Although this is not immediately obviously and I must admit that I forgot about it too, so I've left a comment about that.

demos/jobs/jobs_demo_mosquitto/CMakeLists.txt
aggarw13: Interesting

aggarw13 (dismissed):

muneebahmed10:
The `demo_config.h` for Shadow defines both a client identifier: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/31d6147c2a05c2b4dca48295d718f0b48a0c1e53/demos/shadow/shadow_demo_main/demo_config.h#L127 and a thing name: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/31d6147c2a05c2b4dca48295d718f0b48a0c1e53/demos/shadow/shadow_demo_main/demo_config.h#L169 Do these need to be changed?

yourslab:
Perhaps, it would be best practice to set `THING_NAME` to `CLIENT_IDENTIFIER`

aggarw13 (approved):

tools/cmake/utility.cmake
```diff
@@ -43,26 +43,42 @@ function(set_macro_definitions)
...
+        if(DEFINED_MACROS_FROM_FILE_LENGTH GREATER 0)
+            message("Using values in ${MACRO_DEFINITIONS_FILES_TO_CHECK} to define the following macros for ${application_target}:")
+            foreach(defined_macro_from_file ${DEFINED_MACROS_FROM_FILE})
+                message("${defined_macro_from_file}")
+            endforeach()
+        endif()
+        # Log the value of the macros passed through CMake
+        list(LENGTH DEFINED_MACROS_FROM_CMAKE DEFINED_MACROS_FROM_CMAKE_LENGTH)
+        if(DEFINED_MACROS_FROM_CMAKE_LENGTH GREATER 0)
+            message("Passing the following CMake arguments as macros for ${application_target}:")
```
aggarw13, line 76: (Minor) Suggestion to keep the sentence similar to the previous one 🙂
```suggestion
            message("Using the passed CMake arguments to define the following macros for ${application_target}:")
```

tools/cmake/utility.cmake
```diff
@@ -61,10 +87,28 @@ function(set_macro_definitions)
         if(MISSING_REQUIRED_MACROS)
             string(REPLACE ";" ", " MISSING_REQUIRED_MACROS "${MISSING_REQUIRED_MACROS}")
             message("Cannot build ${application_target} because the following required definitions are missing: ${MISSING_REQUIRED_MACROS}")
```
aggarw13, line 89: Suggest making the sentence more concise
```suggestion
            message("ERROR: Macros missing definitions for ${application_target}: ${MISSING_REQUIRED_MACROS}")
```

aggarw13 (approved):

aggarg (approved):

PR #1621, opened 2021-04-20
[MQTT demos C-SDK]MQTT Mutual auth demo tcp disconnection changes when MQTT Disconnects

gshvang:
Description:

This Pr includes the changes For MQTT Mutual Auth demo changes so that now both TCP and MQTT be behind the backoff algorithm, so that if the MQTT connection fails, it will start the TLS connection again with backoff.


demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -1561,7 +1565,8 @@ int main( int argc,
             else
             {
                 /* If TLS session is established, execute Subscribe/Publish loop. */
-                returnStatus = subscribePublishLoop( &mqttContext, &clientSessionPresent );
+                mqttSessionPresent = true;
```
muneebahmed10, line 1568: If this is always true why have it?

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -375,7 +375,10 @@ static uint32_t generateRandomNumber();
  *
  * @return EXIT_FAILURE on failure; EXIT_SUCCESS on successful connection.
  */
-static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext );
+static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext,
+                                              MQTTContext_t * pMqttContext,
+                                              bool * pClientSessionPresent,
+                                              bool * pBrokerSessionPresent );
```
muneebahmed10, line 381: Please update the doxygen comments for these new parameters

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -1305,47 +1334,22 @@ static int initializeMqtt( MQTTContext_t * pMqttContext,
...
-        /* Update the flag to indicate that an MQTT client session is saved.
-         * Once this flag is set, MQTT connect in the following iterations of
-         * this demo will be attempted without requesting for a clean session. */
-        *pClientSessionPresent = true;
-
         /* Check if session is present and if there are any outgoing publishes
          * that need to resend. This is only valid if the broker is
          * re-establishing a session which was already present. */
-        if( brokerSessionPresent == true )
+        if( *pBrokerSessionPresent == true )
```
muneebahmed10, line 1352: I bet you wouldn't need to add all these boolean pointers if resending publishes was part of or just after `establishMqttSession` instead of in `subscribePublishLoop`. This block isn't really part of the subpub loop if you think about it.

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
gshvang: Initially it is false, I have initialed it to false and once The MQTT connection is made it will be set to true.

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -632,6 +643,27 @@ static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext
...
+                /* End TLS session, then close TCP connection. */
+                ( void ) Openssl_Disconnect( pNetworkContext );
+            }
+            else
+            {
+                *pClientSessionPresent = true;
+            }
+        }
+
         if( opensslStatus != OPENSSL_SUCCESS )
```
muneebahmed10, line 667: The MQTT connection changes `returnStatus` but this only tests `opensslStatus`
```suggestion
        if( ( opensslStatus != OPENSSL_SUCCESS ) || ( returnStatus != EXIT_SUCCESS ) )
```

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -371,11 +371,19 @@ static uint32_t generateRandomNumber();
  * Timeout value will exponentially increase until maximum
  * timeout value is reached or the number of attempts are exhausted.
  *
+ * @param[in,out] pClientSessionPresent Pointer to flag indicating if an
+ * MQTT session is present in the client.
+ * @param[in,out] pBrokerSessionPresent Pointer to flag indicating if a
+ * broker session is present to handle outgoing publishes.
+ * @param[out] pMqttContext The output parameter to return the created mqtt context.
  * @param[out] pNetworkContext The output parameter to return the created network context.
```
muneebahmed10, line 379: Parameters should be in order, and some of these descriptions are wrong
```suggestion
 * @param[out] pNetworkContext The output parameter to return the created network context.
 * @param[in] pMqttContext MQTT context pointer.
 * @param[in,out] pClientSessionPresent Pointer to flag indicating if an
 * MQTT session is present in the client.
 * @param[out] pBrokerSessionPresent Session was already present in the broker or not.
 * Session present response is obtained from the CONNACK from broker.
```

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -389,8 +397,7 @@ static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext
  *
  * @return EXIT_FAILURE on failure; EXIT_SUCCESS on success.
  */
-static int subscribePublishLoop( MQTTContext_t * pMqttContext,
-                                 bool * pClientSessionPresent );
+static int subscribePublishLoop( MQTTContext_t * pMqttContext );
```
muneebahmed10, line 400: The parameter documentation in line 395 also needs to be removed

muneebahmed10 (approved):

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -632,7 +641,28 @@ static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext
...
+                /* End TLS session, then close TCP connection. */
+                ( void ) Openssl_Disconnect( pNetworkContext );
+            }
+            else
+            {
+                *pClientSessionPresent = true;
+            }
+        }
+
+        if( ( opensslStatus != OPENSSL_SUCCESS ) || ( returnStatus == EXIT_FAILURE ) )
```
muneebahmed10, line 665: Actually, the `while` loop on 683 needs to be updated as well, and/or maybe this should just be changed to only check `returnStatus`, and we initialize that one to `EXIT_FAILURE`. That way both the TLS and MQTT connection would have to succeed before `returnStatus` is `EXIT_SUCCESS`

muneebahmed10 (changes requested):

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
gshvang: I have update both the while and if condition to ( returnStatus == EXIT_FAILURE )

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -632,7 +641,28 @@ static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext
...
+                /* End TLS session, then close TCP connection. */
+                ( void ) Openssl_Disconnect( pNetworkContext );
+            }
+            else
+            {
+                *pClientSessionPresent = true;
+            }
+        }
+
+        if( ( returnStatus == EXIT_FAILURE ) )
```
muneebahmed10, line 665: Double parenthesis are unnecessary
```suggestion
        if( returnStatus == EXIT_FAILURE )
```

demos/mqtt/mqtt_demo_mutual_auth/mqtt_demo_mutual_auth.c
```diff
@@ -553,7 +558,10 @@ static uint32_t generateRandomNumber()
 }
 
 /*-----------------------------------------------------------*/
-static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext )
+static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext,
+                                              MQTTContext_t * pMqttContext,
+                                              bool * pClientSessionPresent,
+                                              bool * pBrokerSessionPresent )
 {
     int returnStatus = EXIT_SUCCESS;
```
muneebahmed10, line 566: The loop will only work if this is initialized to failure
```suggestion
    int returnStatus = EXIT_FAILURE;
```

muneebahmed10 (changes requested):

muneebahmed10 (approved):

dan4thewin (approved):

PR #1624, opened 2021-04-27
[MQTT Demos Changes ]MQTT demos updated with tcp connection changes

gshvang:
This PR updates MQTT demos with TCP connection changes so that now TCP and MQTT both connect follows backoff algorithm.

### DIFF
--- a/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
+++ b/demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
@@ -270,7 +270,7 @@ static uint32_t generateRandomNumber();
  * timeout value is reached or the number of attempts are exhausted.
  *
  * @param[out] pNetworkContext The output parameter to return the created network context.
- * @param[in] pMqttContext MQTT context pointer.
+ * @param[out] pMqttContext The output to return the created MQTT context.
  * @param[in,out] pClientSessionPresent Pointer to flag indicating if an
  * MQTT session is present in the client.
  * @param[out] pBrokerSessionPresent Session was already present in the broker or not.
@@ -285,9 +285,9 @@ static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext
 
 
 /**
- * @brief A function that connects to MQTT broker,
- * subscribes to a topic, publishes to the same topic
- * MQTT_PUBLISH_COUNT_PER_LOOP number of times, and verifies if it
+ * @brief A function that uses the passed MQTT connection to
+ * subscribe to a topic, publishe to the same topic
+ * MQTT_PUBLISH_COUNT_PER_LOOP number of times, and verify if it
  * receives the Publish message back.
  *
  * @param[in] pMqttContext MQTT context pointer.
@@ -520,10 +520,6 @@ static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext
             {
                 /* End TLS session, then close TCP connection. */
                 ( void ) Openssl_Disconnect( pNetworkContext );
-            }
-            else
-            {
-                *pClientSessionPresent = true;
             }
         }
 
@@ -1397,6 +1393,8 @@ int main( int argc,
             }
             else
             {
+                clientSessionPresent = true;
+
                 /* Check if session is present and if there are any outgoing publishes
                  * that need to resend. This is only valid if the broker is
                  * re-establishing a session which was already present. */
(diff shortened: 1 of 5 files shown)


yourslab (commented):
Ideally, these functions should just be refactored into a common file so that all the demos can reference the same set of functions. As a follow up, it would be a lot easier to make bug fixes in the future if functions shared by MQTT demos are commonized into a single file. This is how it is currently done for HTTP demos.

demos/mqtt/mqtt_demo_subscription_manager/mqtt_demo_subscription_manager.c
```diff
@@ -322,22 +322,24 @@ static uint32_t generateRandomNumber();
...
  * @return EXIT_FAILURE on failure; EXIT_SUCCESS on successful connection.
  */
-static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext );
+static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext,
+                                              MQTTContext_t * pMqttContext );
 
 /**
  * @brief A function that connects to the MQTT broker,
  * subscribes to a topic, publishes to the same topic
  * MQTT_PUBLISH_COUNT_PER_LOOP number of times, and verifies if it
```
aggarw13, line 335: (Minor) The function does not make an MQTT connection anymore
```suggestion
 * @brief A function that uses the passed MQTT connection to
 * subscribe to a topic, publish to the same topic
 * MQTT_PUBLISH_COUNT_PER_LOOP number of times, and verify if it
```

demos/mqtt/mqtt_demo_subscription_manager/mqtt_demo_subscription_manager.c
```diff
@@ -322,22 +322,24 @@ static uint32_t generateRandomNumber();
  * back-off period is reached or the number of attempts are exhausted.
  *
  * @param[out] pNetworkContext The output parameter to return the created network context.
+ * @param[in] pMqttContext MQTT context pointer.
```
aggarw13, line 325:
```suggestion
 * @param[out] pMqttContext The output to return the created MQTT context.
```

demos/mqtt/mqtt_demo_basic_tls/mqtt_demo_basic_tls.c
```diff
@@ -1408,7 +1385,7 @@ int main( int argc,
              * attempts are reached or maximum timeout value is reached. The function
              * returns EXIT_FAILURE if the TCP connection cannot be established to
              * broker after configured number of attempts. */
-            returnStatus = connectToServerWithBackoffRetries( &networkContext );
+            returnStatus = connectToServerWithBackoffRetries( &networkContext, &mqttContext, &clientSessionPresent, &brokerSessionPresent );
```
aggarw13, line 1388: Why is the `clientSessionPresent` passed as an output parameter? Shouldn't it just be an input parameter?

aggarw13, line 1388: Why is the `clientSessionPresent` passed as an output parameter? Shouldn't it just be an input parameter?
It would be more readable if the value of `clientSessionPresent` variable is set after the call to `connectToServerWithBackOffRetries`
```
            returnStatus = connectToServerWithBackoffRetries( &networkContext, &mqttContext, &clientSessionPresent, &brokerSessionPresent );
       
            if( returnStatus == EXIT_FAILURE )
            {
                /* Log error to indicate connection failure after all
                 * reconnect attempts are over. */
                LogError( ( "Failed to connect to MQTT broker %.*s.",
                            BROKER_ENDPOINT_LENGTH,
                            BROKER_ENDPOINT ) );
            }
            else
            { 
       
              /* Update the flag to indicate that an MQTT client session is saved.
               * Once this flag is set, MQTT connect in the following iterations of
                * this demo will be attempted without requesting for a clean session. */ 
               clientSessionPresent = true;     
```

aggarw13:
I also think there is an opportunity to commonize the `connectToServerWithBackoffRetries` into a separate file for all demos to use. 
Will you be able to make that change of creating `mqtt_demo_common.c/h` files containing this function that all the demos can use?

The function can be of the following form, that takes parameters for the caller to mention whether a TLS session should be established.
```
static int connectToServerWithBackoffRetries( NetworkContext_t * pNetworkContext,
                                              bool setupTls,
                                              MQTTContext_t * pMqttContext,
                                              bool clientSessionPresent,
                                              bool * pBrokerSessionPresent )
```

aggarw13 (approved):

dan4thewin (approved):

Issue #1634, opened 2021-05-12
Look for a solution to help us improving MQTT_ProcessLoop 's behaviour in user worker thread.  Thanks!


zlhk100:
Dear AWS C SDK expert,

In our MQTT connecting and incoming message polling sequence,  we are doing something like following straightforward from one worker thread A.
      1>.  MQTT_init 
      2>  Connect to MQTT broker via mutual authN with this API.
             Openssl_Connect(mNetworkContext, &serverInfo, &opensslCredentials,            
                                        SSL_read_write_timeout_ms, SSL_read_write_timeout_ms); 
      3> Establish a clean MQTT session over authN'ed TLS with this API
             MQTT_Connect(&mMqttSessions[handle.idx]->mContext, &connectInfo, NULL, CONNACK_RECV_TIMEOUT_MS,... );
      4>. Then subscribing to some incoming topics over previous established MQTT channel persistent over entire process lifecycle.

During runtime,   from same worker thread A,   other than some other user logics (eg. user queue processing etc),   MQTT_ProcessLoop is invoked every time when worker thread A is scheduled for checking incoming MQTT messages

By our experimenting, We realized that MQTT_ProcessLoop<...> demonstrates below timing behaviour.
     /*  Calling this function blocks the calling context for a time period that
     * depends on the passed @mTimeoutMs(eg. 100ms), and the underlying transport interface SSL_read(...) implementation
     * timeouts @SSL_read_write_timeout_ms(eg. 2000ms), unless an error occurs. 
     *  **_Please blocking time will be up to MAX(mTimeoutMs, SSL_read_write_timeout_ms), which is 2000ms. _**
    status = MQTT_ProcessLoop(&mMqttSessions[handle.idx]->mContext,  mTimeoutMs); 

Here is our problem which need you guys' support to get us un-blocked for our project.   In our case, since we constantly poll the MQTT for pending messages by MQTT_ProcessLoop(...),  we expect it should return right away if nothing is pending.  Otherwise it will block our worker thread A from doing other routine operation with un-accepted latency.

Current hack we put in to fix our issue is like following:
       1> Since step 2> Openssl_Connect(...) will require sufficient timeout to get TLS hand-shaking completed, we leave @SSL_read_write_timeout_ms as is eg. (2000ms)
      2>. After TLS successfully is established,  we dynamically override underneath transport interface SSL_read<>'s timeout value to something new eg. 1ms by using raw posix socket API  like following
[...]

PR #1665, opened 2021-07-13
Poll socket before attempting a speculative read or write

yourslab:
This addresses the problem from issue #1634. Specifically, the POSIX transport implementation is updated to match what is specified in the [transport_interface.h documentation for handling speculative reads](https://github.com/FreeRTOS/coreMQTT/blob/main/source/interface/transport_interface.h#L94-L172).

### DIFF
--- a/platform/posix/transport/src/openssl_posix.c
+++ b/platform/posix/transport/src/openssl_posix.c
@@ -721,7 +721,6 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
 {
     OpensslParams_t * pOpensslParams = NULL;
     int32_t bytesReceived = 0;
-    int32_t sslError = 0;
 
     if( ( pNetworkContext == NULL ) || ( pNetworkContext->pParams == NULL ) )
     {
@@ -734,7 +733,8 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
     }
     else
     {
-        int32_t pollStatus = -1, readStatus = 1;
+        int32_t pollStatus = 1, readStatus = 1, sslError = 0;
+        uint8_t shouldRead = 0U;
         struct pollfd pollFds;
         pOpensslParams = pNetworkContext->pParams;
 
@@ -744,15 +744,37 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
         /* Set the file descriptor for poll. */
         pollFds.fd = pOpensslParams->socketDescriptor;
 
-        /* Check if there is data to read from the socket. */
-        pollStatus = poll( &pollFds, 1, 0 );
+        if( ( bytesToRecv > 1 ) || ( SSL_pending( pOpensslParams->pSsl ) > 0 ) )
+        {
+            shouldRead = 1U;
+        }
+        else
+        {
+            /* Non-speculative read so check if there is data to read from the socket. */
+            pollStatus = poll( &pollFds, 1, 0 );
+        }
 
-        if( ( SSL_pending( pOpensslParams->pSsl ) > 0 ) || ( pollStatus > 0 ) )
+        if( pollStatus < 0 )
+        {
+            bytesReceived = -1;
+        }
+        else if( pollStatus == 0 )
+        {
+            /* No data available to be read from the socket. */
+            bytesReceived = 0;
+        }
+        else
+        {
+            shouldRead = 1U;
+        }
+
+        if( shouldRead == 1U )
         {
             /* SSL read of data. */
             readStatus = ( int32_t ) SSL_read( pOpensslParams->pSsl, pBuffer,
                                                ( int32_t ) bytesToRecv );
 
+            /* Successfully read of application data. */
             if( readStatus > 0 )
             {
                 bytesReceived = readStatus;
@@ -787,12 +809,6 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
                  * retried. */
                 bytesReceived = -1;
             }
-        }
-
-        if( pollStatus < 0 )
-        {
-            LogError( ( "An error occurred while polling." ) );
-            bytesReceived = -1;
         }
     }
 
--- a/platform/posix/transport/src/openssl_posix.c
+++ b/platform/posix/transport/src/openssl_posix.c
@@ -2,22 +2,23 @@
  * AWS IoT Device SDK for Embedded C 202103.00
  * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
  *
- * Permission is hereby granted, free of charge, to any person obtaining a copy of
- * this software and associated documentation files (the "Software"), to deal in
- * the Software without restriction, including without limitation the rights to
- * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
- * the Software, and to permit persons to whom the Software is furnished to do so,
- * subject to the following conditions:
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
  *
- * The above copyright notice and this permission notice shall be included in all
- * copies or substantial portions of the Software.
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
  *
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
- * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
- * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
- * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
  */
 
 /* Standard includes. */
@@ -32,6 +33,7 @@
 
 #include "openssl_posix.h"
 #include <openssl/err.h>
+#include <poll.h>
 
 /*-----------------------------------------------------------*/
 
@@ -74,11 +76,12 @@ struct NetworkContext
 /**
  * @brief Add X509 certificate to the trusted list of root certificates.
  *
- * OpenSSL does not provide a single function for reading and loading certificates
- * from files into stores, so the file API must be called. Start with the
- * root certificate.
+ * OpenSSL does not provide a single function for reading and loading
+ * certificates from files into stores, so the file API must be called. Start
+ * with the root certificate.
  *
- * @param[out] pSslContext SSL context to which the trusted server root CA is to be added.
+ * @param[out] pSslContext SSL context to which the trusted server root CA is to
+ * be added.
  * @param[in] pRootCaPath Filepath string to the trusted server root CA.
  *
  * @return 1 on success; -1, 0 on failure;
@@ -87,9 +90,11 @@ static int32_t setRootCa( const SSL_CTX * pSslContext,
                           const char * pRootCaPath );
 
 /**
- * @brief Set X509 certificate as client certificate for the server to authenticate.
+ * @brief Set X509 certificate as client certificate for the server to
+ * authenticate.
  *
- * @param[out] pSslContext SSL context to which the client certificate is to be set.
+ * @param[out] pSslContext SSL context to which the client certificate is to be
+ * set.
  * @param[in] pClientCertPath Filepath string to the client certificate.
  *
  * @return 1 on success; 0 failure;
@@ -115,7 +120,8 @@ static int32_t setPrivateKey( SSL_CTX * pSslContext,
  * OpenSSL library. If the client certificate or private key is not NULL, mutual
  * authentication is used when performing the TLS handshake.
  *
- * @param[out] pSslContext SSL context to which the credentials are to be imported.
+ * @param[out] pSslContext SSL context to which the credentials are to be
+ * imported.
  * @param[in] pOpensslCredentials TLS credentials to be imported.
  *
  * @return 1 on success; -1, 0 on failure;
@@ -128,7 +134,8 @@ static int32_t setCredentials( SSL_CTX * pSslContext,
  *
  * This function is used to set SNI, MFLN, and ALPN protocols.
  *
- * @param[in] pSsl SSL context to which the optional configurations are to be set.
+ * @param[in] pSsl SSL context to which the optional configurations are to be
+ * set.
  * @param[in] pOpensslCredentials TLS credentials containing configurations.
  */
 static void setOptionalConfigurations( SSL * pSsl,
@@ -174,17 +181,12 @@ static OpensslStatus_t tlsHandshake( const ServerInfo_t * pServerInfo,
         /* Log the absolute directory based on first character of path. */
         if( ( path[ 0 ] == '/' ) || ( path[ 0 ] == '\\' ) )
         {
-            LogDebug( ( "Attempting to open %s: Path=%s.",
-                        fileType,
-                        path ) );
+            LogDebug( ( "Attempting to open %s: Path=%s.", fileType, path ) );
         }
         else
         {
             cwd = getcwd( NULL, 0 );
-            LogDebug( ( "Attempting to open %s: Path=%s/%s.",
-                        fileType,
-                        cwd,
-                        path ) );
+            LogDebug( ( "Attempting to open %s: Path=%s/%s.", fileType, cwd, path ) );
         }
 
         /* Free cwd because getcwd calls malloc. */
@@ -216,8 +218,9 @@ static OpensslStatus_t convertToOpensslStatus( SocketStatus_t socketStatus )
             break;
 
         default:
-            LogError( ( "Unexpected status received from socket wrapper: Socket status = %u",
-                        socketStatus ) );
+            LogError(
+                ( "Unexpected status received from socket wrapper: Socket status = %u",
+                  socketStatus ) );
             break;
     }
 
@@ -233,8 +236,7 @@ static OpensslStatus_t tlsHandshake( const ServerInfo_t * pServerInfo,
     int32_t sslStatus = -1, verifyPeerCertStatus = X509_V_OK;
 
     /* Validate the hostname against the server's certificate. */
-    sslStatus = SSL_set1_host( pOpensslParams->pSsl,
-                               pServerInfo->pHostName );
+    sslStatus = SSL_set1_host( pOpensslParams->pSsl, pServerInfo->pHostName );
 
     if( sslStatus != 1 )
     {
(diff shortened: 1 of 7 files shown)


platform/posix/transport/src/openssl_posix.c
```diff
@@ -729,55 +721,96 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
...
+        struct pollfd pollFds;
         pOpensslParams = pNetworkContext->pParams;
-        /* SSL read of data. */
-        bytesReceived = ( int32_t ) SSL_read( pOpensslParams->pSsl,
-                                              pBuffer,
-                                              ( int32_t ) bytesToRecv );
+
+        /* Initialize the file descriptor. */
+        pollFds.events = POLLIN | POLLPRI | POLLHUP | POLLERR;
+        pollFds.revents = 0;
```
aggarw13, line 743: A comment about the significance of these event flags being set can be helpful

platform/posix/transport/src/openssl_posix.c
```diff
@@ -729,55 +721,96 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
...
+
+        if( ( bytesToRecv > 1 ) || ( SSL_pending( pOpensslParams->pSsl ) > 0 ) )
+        {
+            shouldRead = 1U;
+        }
+        else
+        {
+            /* Non-speculative read so check if there is data to read from the socket. */
+            pollStatus = poll( &pollFds, 1, 0 );
+        }
```
aggarw13, line 755: Comments here to clearly mention that there is a non-blocking read for the case of reading more than one byte whereas  a timed wait when reading a single byte is the behavior represented by this if-else block

platform/posix/transport/src/openssl_posix.c
```diff
@@ -729,55 +721,96 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
...
+        uint8_t shouldRead = 0U;
+        struct pollfd pollFds;
         pOpensslParams = pNetworkContext->pParams;
-        /* SSL read of data. */
-        bytesReceived = ( int32_t ) SSL_read( pOpensslParams->pSsl,
-                                              pBuffer,
-                                              ( int32_t ) bytesToRecv );
+
+        /* Initialize the file descriptor. */
+        pollFds.events = POLLIN | POLLPRI | POLLHUP | POLLERR;
```
aggarw13, line 742: According to the [Linux man page](https://man7.org/linux/man-pages/man2/poll.2.html) for `poll.h`:
"_The bits returned in revents can include any of those specified in events, or one of the values **POLLERR**, **POLLHUP**, or **POLLNVAL**.  (These three bits are meaningless in the events field, and will be set in the revents field whenever the corresponding condition is true.)_"

Thus. explicitly setting `POLLHUP | POLLERR` bits in the `events` member is not required..

platform/posix/transport/src/openssl_posix.c
```diff
@@ -729,55 +721,96 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
...
+        if( pollStatus < 0 )
+        {
+            bytesReceived = -1;
+        }
+        else if( pollStatus == 0 )
+        {
+            /* No data available to be read from the socket. */
+            bytesReceived = 0;
+        }
+        else
```
aggarw13, line 766: There should probably a check that the `pollFds.revent` output parameter has been set to `POLLIN` based on the following documentation for positive return value from `poll` in the [Linux man page](https://man7.org/linux/man-pages/man2/poll.2.html):
"_On success, poll() returns a nonnegative value which is the number of elements in the pollfds whose revents fields have been set to a nonzero value (indicating an event or an error).  A return value of zero indicates that the system call timed out before any file descriptors became read._"

aggarw13: Also, an appropriate log message can be added if the `POLLERR` flag is set in the output parameter of `revents`

platform/posix/transport/src/openssl_posix.c
```diff
@@ -792,47 +825,70 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
...
+
         pOpensslParams = pNetworkContext->pParams;
-        /* SSL write of data. */
-        bytesSent = ( int32_t ) SSL_write( pOpensslParams->pSsl,
-                                           pBuffer,
-                                           ( int32_t ) bytesToSend );
 
-        if( bytesSent <= 0 )
+        /* Initialize the file descriptor. */
+        pollFds.events = POLLOUT | POLLPRI | POLLHUP | POLLERR;
```
aggarw13, line 841: Same comments here as [mentioned earlier](https://github.com/aws/aws-iot-device-sdk-embedded-C/pull/1665/files?file-filters%5B%5D=.c#r669855972) about not needing `POLLHUP`, `POLLERR` flags to be explicitly set here as these are always enabled

platform/posix/transport/src/openssl_posix.c
yourslab: `SSL_pending` will only return a non-zero value after a call to `SSL_read`. It just refers to data that was not yet processed on the last read. It is still necessary to have a mechanism to check if the socket has data to be read.

platform/posix/transport/src/openssl_posix.c
yourslab: We had a chat that the error in such cases would already be handled by `recv` for POSIX and `SSL_read` for OpenSSL, so it is not necessary to check these flags.

platform/posix/transport/src/openssl_posix.c
```diff
@@ -792,47 +827,70 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
...
         }
+        else if( pollStatus < 0 )
+        {
+            /* An error occurred while polling. */
+            LogError( ( "An error occurred while polling." ) );
+            bytesSent = -1;
+        }
+        else
+        {
+            /* Timed out waiting for data to be sent. */
```
aggarw13, line 891: There was no timed wait for the socket to be available as the `timeout` parameter used is `0`. So, this comment needs to be updated
```suggestion
            /* Socket is not available for sending data. Set return code for retrying send. */
```

platform/posix/transport/src/openssl_posix.c
aggarw13: Doesn't `SSL_pendign` data represent all the data buffered _after_ processing input records?

aggarw13: I see a problem if a partial SSL record is received from the network, but `poll` returns a positive return code to represent availability of data on the TCP socket, but `SSL_read` returns zero bytes because the entire SSL record data has not been received. In that case, this function shouldn't return negative return code back to the calling MQTT/HTTP library

platform/posix/transport/src/openssl_posix.c
```diff
@@ -792,47 +827,70 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
...
-                                           ( int32_t ) bytesToSend );
 
-        if( bytesSent <= 0 )
+        /* Initialize the file descriptor. */
+        pollFds.events = POLLOUT;
+        pollFds.revents = 0;
+        /* Set the file descriptor for poll. */
+        pollFds.fd = pOpensslParams->socketDescriptor;
+
+        /* Check if socket descriptor is ready to be written to. */
```
aggarw13, line 848: A comment here about the non-blocking call to `poll` will be valuable

platform/posix/transport/src/openssl_posix.c
```diff
@@ -792,47 +827,70 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
...
+                if( bytesSent == 0 )
+                {
+                    bytesSent = -1;
+                }
             }
         }
+        else if( pollStatus < 0 )
+        {
+            /* An error occurred while polling. */
+            LogError( ( "An error occurred while polling." ) );
```
aggarw13, line 886:
```suggestion
            LogError( ( "Unable to send TLS data on network: An error occurred while checking availability of TCP socket, %d, to send data.",  pOpensslParams->socketDescriptor) );
```

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -115,79 +115,40 @@ int32_t Plaintext_Recv( NetworkContext_t * pNetworkContext,
...
-    /* Check if there is data to read from the socket. */
-    selectStatus = select( pPlaintextParams->socketDescriptor + 1,
-                           &readfds,
-                           NULL,
-                           NULL,
-                           &recvTimeout );
+    /* Speculative read for the start of a payload. */
+    if( bytesToRecv == 1U )
+    {
+        /* Check if there is data to read from the socket. */
```
aggarw13, line 139:
```suggestion
        /* Check if there is data to read (without blocking) from the socket. */
```

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -115,79 +115,40 @@ int32_t Plaintext_Recv( NetworkContext_t * pNetworkContext,
...
+    /* Set the file descriptor for poll. */
+    pollFds.fd = pPlaintextParams->socketDescriptor;
 
-    /* Check if there is data to read from the socket. */
-    selectStatus = select( pPlaintextParams->socketDescriptor + 1,
-                           &readfds,
-                           NULL,
-                           NULL,
-                           &recvTimeout );
+    /* Speculative read for the start of a payload. */
```
aggarw13, line 136:
```suggestion
    /* Speculative read for the start of a payload. */
    /* Note: This is done to avoid blocking behavior of recv() call when no data is available from the network to read. */
```

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -115,79 +115,40 @@ int32_t Plaintext_Recv( NetworkContext_t * pNetworkContext,
...
         bytesReceived = ( int32_t ) recv( pPlaintextParams->socketDescriptor,
                                           pBuffer,
                                           bytesToRecv,
                                           0 );
     }
-    else if( selectStatus < 0 )
+    else if( pollStatus < 0 )
     {
         /* An error occurred while polling. */
         bytesReceived = -1;
```
aggarw13, line 154: The comment on line 158 is invalid about timeout from the `poll` call.

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -198,7 +159,7 @@ int32_t Plaintext_Recv( NetworkContext_t * pNetworkContext,
         bytesReceived = 0;
     }
 
-    if( ( selectStatus > 0 ) && ( bytesReceived == 0 ) )
+    if( ( pollStatus > 0 ) && ( bytesReceived == 0 ) )
```
aggarw13, line 162: (Nit) A comment can be added to mention that a zero value return from `recv` represents closure of TCP connection by the peer. I know that this comment exists within this branch, but it doesn't explicitly mention about it being related to `recv` call

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -224,77 +185,34 @@ int32_t Plaintext_Send( NetworkContext_t * pNetworkContext,
...
-                           &sendTimeout );
-
-    if( selectStatus > 0 )
+    /* Initialize the file descriptor. */
+    pollFds.events = POLLOUT;
+    pollFds.revents = 0;
+    /* Set the file descriptor for poll. */
+    pollFds.fd = pPlaintextParams->socketDescriptor;
+
+    /* Check if there is data to read from the socket. */
```
aggarw13, line 204:
```suggestion
    /* Check if socket is available to send data on the network. This is done to avoid blocking on send() call when the socket is ready to accept more data to send on the network (possibly due to a full TX buffer) . */
```

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -224,77 +185,34 @@ int32_t Plaintext_Send( NetworkContext_t * pNetworkContext,
...
         /* The socket is available for sending data. */
         bytesSent = ( int32_t ) send( pPlaintextParams->socketDescriptor,
                                       pBuffer,
                                       bytesToSend,
                                       0 );
     }
-    else if( selectStatus < 0 )
+    else if( pollStatus < 0 )
     {
         /* An error occurred while polling. */
```
aggarw13, line 217: "Timed out" comment on line 222 is not valid anymore

platform/posix/transport/src/openssl_posix.c
yourslab: We discussed that `poll` is necessary in this case because `SSL_pending` does not actually tell you if data is ready to be read from the socket, it will only return a value >0 if data remains to be read from the previous call to `SSL_read`.

platform/posix/transport/utest/openssl_utest.c
```diff
@@ -848,10 +875,30 @@ void test_Openssl_Recv_Network_Error( void )
     TEST_ASSERT_TRUE( bytesReceived < 0 );
 
     /* Test that a non-retryable zero error code is converted to -1 by the API. */
+    SSL_pending_ExpectAnyArgsAndReturn( 1 );
     SSL_read_ExpectAnyArgsAndReturn( 0 );
     SSL_get_error_ExpectAnyArgsAndReturn( SSL_ERROR_ZERO_RETURN );
-    bytesReceived = Openssl_Recv( &networkContext, opensslBuffer, BYTES_TO_RECV );
+    bytesReceived = Openssl_Recv( &networkContext, opensslBuffer, 1U );
```
aggarw13, line 881: Why was the number of bytes parameter changed?

platform/posix/transport/utest/openssl_utest.c
```diff
@@ -821,10 +842,16 @@ void test_Openssl_Recv_All_Bytes_Received_Successfully( void )
...
     int32_t bytesReceived;
 
+    /* No pending data from last read. */
     opensslParams.pSsl = &ssl;
     SSL_read_ExpectAnyArgsAndReturn( BYTES_TO_RECV );
     bytesReceived = Openssl_Recv( &networkContext, opensslBuffer, BYTES_TO_RECV );
     TEST_ASSERT_EQUAL( BYTES_TO_RECV, bytesReceived );
+
+    /* Pending data from last read. */
+    SSL_read_ExpectAnyArgsAndReturn( BYTES_TO_RECV );
```
aggarw13, line 852: This case seems to be an exact repetition of the case that already exists in this function..?

platform/posix/transport/utest/openssl_utest.c
```diff
@@ -848,10 +875,30 @@ void test_Openssl_Recv_Network_Error( void )
...
+/**
+ * @brief Test that #Openssl_Recv returns 0 when `poll` returns 0.
+ */
+void test_Openssl_Recv_Poll_No_Events( void )
+{
+    int32_t bytesReceived;
+
+    SSL_pending_ExpectAnyArgsAndReturn( 0 );
+    poll_ExpectAnyArgsAndReturn( 0 );
+    bytesReceived = Openssl_Recv( &networkContext, opensslBuffer, 1U );
```
aggarw13, line 900: Should this also test the case when more than one byte is requested but both `SSL_pending` and `poll` return zero?

yourslab: Sure, doesn't hurt to have it

platform/posix/transport/utest/openssl_utest.c
yourslab: I had meant to change this but got coverage in another function. Changed that here.

platform/posix/transport/utest/openssl_utest.c
yourslab: I had forgotten to update the other test that would've attained the same coverage. I reverted this.

platform/posix/transport/utest/openssl_utest.c
yourslab: If more than one byte is requested, neither of those functions are called

platform/posix/transport/src/openssl_posix.c
```diff
@@ -746,13 +746,17 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
...
         if( ( bytesToRecv > 1 ) || ( SSL_pending( pOpensslParams->pSsl ) > 0 ) )
         {
             shouldRead = 1U;
         }
         else
         {
-            /* Non-speculative read so check if there is data to read from the socket. */
+            /* Speculative read for the start of a payload.
+             * Note: This is done to avoid blocking when no
+             * data is available to be read from the socket. */
```
aggarw13, line 759: Formatting of comment is not aligned to the code
```suggestion
           /* Speculative read for the start of a payload.
            * Note: This is done to avoid blocking when no
            * data is available to be read from the socket. */
```

platform/posix/transport/src/openssl_posix.c
```diff
@@ -746,13 +746,17 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
         /* Set the file descriptor for poll. */
         pollFds.fd = pOpensslParams->socketDescriptor;
 
+        /* #SSL_pending returns a value > 0 if application data
+         * from the last processed TLS record remains to be read. */
         if( ( bytesToRecv > 1 ) || ( SSL_pending( pOpensslParams->pSsl ) > 0 ) )
```
aggarw13, line 751: It could be fine to call `SSL_pending` for every call to this function, irrespective of the number of bytes requested. This can act as the first level of filter where the OpenSSL library is queried about pending data, followed by checking the `poll` function for data from the TCP socket

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -201,7 +205,10 @@ int32_t Plaintext_Send( NetworkContext_t * pNetworkContext,
     /* Set the file descriptor for poll. */
     pollFds.fd = pPlaintextParams->socketDescriptor;
 
-    /* Check if there is data to read from the socket. */
+    /* Check if data can be written to the socket.
+     * Note: This is done to avoid blocking on send() call when
+     * the socket is ready to accept more data for transmitting
+     * (possibly due to a full TX buffer). */
```
aggarw13, line 211:
```suggestion
     * Note: This is done to avoid blocking on send() call when
     * the socket is not ready to accept more data for networl
     * transmission (possibly due to a full TX buffer). */
```

platform/posix/transport/utest/openssl_utest.c
```diff
@@ -849,8 +849,9 @@ void test_Openssl_Recv_All_Bytes_Received_Successfully( void )
     TEST_ASSERT_EQUAL( BYTES_TO_RECV, bytesReceived );
 
     /* Pending data from last read. */
+    SSL_pending_ExpectAnyArgsAndReturn( BYTES_TO_RECV );
     SSL_read_ExpectAnyArgsAndReturn( BYTES_TO_RECV );
```
aggarw13, line 853: Is there a reason for returning `BYTES_TO_RECV` from `SSL_read` when only a single byte is requested in the test?

yourslab: I had forgotten to update it

platform/posix/transport/src/openssl_posix.c
yourslab: That is already aligned. What you are suggesting won't pass formatting checks..

platform/posix/transport/src/plaintext_posix.c
```diff
@@ -224,88 +189,48 @@ int32_t Plaintext_Send( NetworkContext_t * pNetworkContext,
...
+    /* Initialize the file descriptor. */
+    pollFds.events = POLLOUT;
+    pollFds.revents = 0;
+    /* Set the file descriptor for poll. */
+    pollFds.fd = pPlaintextParams->socketDescriptor;
+
+    /* Check if data can be written to the socket.
+     * Note: This is done to avoid blocking on send() when
+     * the socket is ready to accept more data for network transmission
+     * (possibly due to a full TX buffer). */
```
aggarw13, line 211:
```suggestion
     * the socket is not ready to accept more data for network 
     * transmission (possibly due to a full TX buffer). */
```

platform/posix/transport/src/openssl_posix.c
```diff
@@ -792,47 +838,72 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
...
-                                           ( int32_t ) bytesToSend );
 
-        if( bytesSent <= 0 )
+        /* Initialize the file descriptor. */
+        pollFds.events = POLLOUT;
+        pollFds.revents = 0;
+        /* Set the file descriptor for poll. */
+        pollFds.fd = pOpensslParams->socketDescriptor;
+
+        /* `poll` checks if the socket is ready to send data. */
```
aggarw13, line 859:
```suggestion
        /* `poll` checks if the socket is ready to send data. 
         * Note: This is done to avoid blocking on SSL_write()
         * when TCP socket is not ready to accept more data for
         * network transmission (possibly due to a full TX buffer). */
```

aggarw13 (dismissed):

aggarw13 (approved):

platform/posix/transport/utest/openssl_utest.c
```diff
@@ -785,10 +788,28 @@ void test_Openssl_Send_Network_Error( void )
     TEST_ASSERT_TRUE( bytesSent < 0 );
 
     /* Test that a non-retryable zero error code is converted to -1 by the API. */
+    poll_ExpectAnyArgsAndReturn( 1 );
     SSL_write_ExpectAnyArgsAndReturn( 0 );
     SSL_get_error_ExpectAnyArgsAndReturn( SSL_ERROR_ZERO_RETURN );
     bytesSent = Openssl_Send( &networkContext, opensslBuffer, BYTES_TO_SEND );
     TEST_ASSERT_EQUAL( -1, bytesSent );
+
+    /* Test that a poll error results in the function under test returning -1. */
```
paulbartell, line 797: A test double would make this test less brittle.

paulbartell (approved):

PR #1666, opened 2021-07-14
Update broken MISRA link

yourslab:
MISRA website was updated so old link gives 404. This only updates it for the hub repo.


muneebahmed10 (approved):

muneebahmed10:
Note that since the link verifier also checks submodules, the check will continue to fail until the submodules are updated

n9wxu (approved):

PR #1670, opened 2021-07-16
Update corePKCS11 demo to read the public key

yourslab:
Update the demo so that it doesn't treat the public key as a private key

### DIFF
--- a/demos/pkcs11/pkcs11_demo_sign_and_verify/pkcs11_demo_sign_and_verify.c
+++ b/demos/pkcs11/pkcs11_demo_sign_and_verify/pkcs11_demo_sign_and_verify.c
@@ -155,7 +155,7 @@ CK_RV PKCS11SignVerifyDemo( void )
     {
         result = xFindObjectWithLabelAndClass( session,
                                                pkcs11demoPRIVATE_KEY_LABEL,
-                                               sizeof( pkcs11demoPRIVATE_KEY_LABEL ),
+                                               sizeof( pkcs11demoPRIVATE_KEY_LABEL ) - 1,
                                                CKO_PRIVATE_KEY,
                                                &privateKeyHandle );
     }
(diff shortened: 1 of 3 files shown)


paulbartell (dismissed):

demos/pkcs11/pkcs11_demo_sign_and_verify/pkcs11_demo_sign_and_verify.c
```diff
@@ -155,7 +155,7 @@ CK_RV PKCS11SignVerifyDemo( void )
     {
         result = xFindObjectWithLabelAndClass( session,
                                                pkcs11demoPRIVATE_KEY_LABEL,
-                                               sizeof( pkcs11demoPRIVATE_KEY_LABEL ),
+                                               sizeof( pkcs11demoPRIVATE_KEY_LABEL ) - 1,
```
aggarw13, line 158: Is this change to discount the `NUL` character? 
The comment in the `pkcs11_demo_objects.c` file mentions that the labels should not be NUL terminated...
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/a6ae98535d1f304830bae1a1cd8029887792868e/demos/pkcs11/pkcs11_demo_objects/pkcs11_demo_objects.c#L337-L340

aggarw13:
A note about this fix in the CHANGELOG would be valuable

demos/pkcs11/pkcs11_demo_sign_and_verify/pkcs11_demo_sign_and_verify.c
yourslab: Yes, it is not NUL terminated in the objects demo as well because when passed to `publicKeyTemplate`/`privateKeyTemplate`, the NUL character is removed through the length.
https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/a6ae98535d1f304830bae1a1cd8029887792868e/demos/pkcs11/pkcs11_demo_objects/pkcs11_demo_objects.c#L358
Hence, NUL character must also be subtracted from the length in the sign and verify demo.

aggarw13: Okay, should the `It should not be NULL terminated.` comment be removed then as it is not fitting in declaration of the byte arrays..

aggarw13 (approved):

xortothemoon (approved):

paulbartell (approved):

PR #1672, opened 2021-07-24
Add MbedTLS submodule and update corePKCS11

archigup:
Adds MbedTLS to third party libraries.
Updates the corePKCS11 submodule.
Fixes corePKCS11 demos and switches them to CSDK MbedTLS submodule.


libraries/3rdparty/CMakeLists.txt
```diff
@@ -46,3 +46,33 @@ set_target_properties(tinycbor PROPERTIES
...
+
+add_library(mbedtls
+    ${MBEDTLS_FILES}
+)
+
+set_target_properties(mbedtls PROPERTIES
+    ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+    POSITION_INDEPENDENT_CODE ON
+    C_STANDARD 99
+)
```
muneebahmed10, line 60: Might I suggest
```suggestion
set_target_properties(mbedtls PROPERTIES
    ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
    POSITION_INDEPENDENT_CODE ON
)

# Use C99 for mbedtls as v2.26.0 is incompatible with C90
if(CMAKE_C_STANDARD LESS 99)
    set_target_properties(mbedtls PROPERTIES C_STANDARD 99)
endif()
```
to allow this to be compiled with a later standard if manually specified?

libraries/3rdparty/CMakeLists.txt
```diff
@@ -46,3 +46,37 @@ set_target_properties(tinycbor PROPERTIES
 target_include_directories(tinycbor PUBLIC
     "${ROOT_DIR}/libraries/3rdparty/tinycbor/src"
 )
+
+file(GLOB MBEDTLS_FILES CONFIGURE_DEPENDS "${ROOT_DIR}/libraries/3rdparty/mbedtls/library/*.c")
+
+add_library(mbedtls
```
muneebahmed10, line 52: You could probably optimize this further and do something similar to the jobs demo by calling find_library: https://github.com/aws/aws-iot-device-sdk-embedded-C/blob/5015cdf04561ce6e9d4e25a3726fe71dc17cdd0a/demos/jobs/jobs_demo_mosquitto/CMakeLists.txt#L18 so that mbedtls is only compiled when it doesn't already exist on the system. I'm unsure if all our code is compatible with the latest mbedtls 3.0.0, though.

muneebahmed10 (approved):

libraries/3rdparty/CMakeLists.txt
archigup: Yeah, newer version didnt compile for me. Also we use some non-default features, so system lib would not likely have them.

aggarg (approved):

PR #1676, opened 2021-07-27
Add build checks for C++ linkage

muneebahmed10:
*Issue #, if available:*

*Description of changes:*
Makes a few changes to the MQTT, HTTP, Shadow, and Defender demos to be able to compile as C++, and then build these demos as a CI check to ensure that the C library functions can be linked into a C++ project. The `-Werror` flag is not passed, as this code was not written with C++ in mind and we don't need to fix warnings that arise from using C++.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
           make -C build/ help | grep demo | tr -d '. ' | xargs make -C build/
           make -C demos/jobs/jobs_demo_mosquitto
   build-check-demos-cpp:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
       - name: Clone This Repo
         uses: actions/checkout@v2
@@ -90,7 +90,6 @@ jobs:
       - name: Build Demos
         run: |
           make -C build/ help | grep demo | tr -d '. ' | xargs make -C build/
-          make -C demos/jobs/jobs_demo_mosquitto
   build-check-system-tests:
     runs-on: ubuntu-18.04
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,46 @@ jobs:
         run: |
           make -C build/ help | grep demo | tr -d '. ' | xargs make -C build/
           make -C demos/jobs/jobs_demo_mosquitto
+  build-check-demos-cpp-mqtt:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Clone This Repo
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Rename source files
+        run: |
+          for source in demos/mqtt/mqtt*/*.c; do mv -- "$source" "${source%.c}.cpp"; done
+      - name: Configure CMake build
+        run: |
+          sudo apt-get install -y libmosquitto-dev
+          curl https://cmake.org/files/v3.2/cmake-3.2.0-Linux-x86_64.tar.gz -o cmake.tar.gz
+          tar -xf cmake.tar.gz
+          mkdir build && cd build
+          ../cmake-3.2.0-Linux-x86_64/bin/cmake .. \
+          -G "Unix Makefiles" \
+          -DBUILD_DEMOS=1 \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DCMAKE_C_FLAGS='-Wall -Wextra' \
+          -DAWS_IOT_ENDPOINT="aws-iot-endpoint" \
+          -DBROKER_ENDPOINT="broker-endpoint" \
+          -DCLIENT_CERT_PATH="cert/path" \
+          -DROOT_CA_CERT_PATH="cert/path" \
+          -DCLIENT_PRIVATE_KEY_PATH="key/path" \
+          -DCLIENT_IDENTIFIER="ci-identifier" \
+          -DTHING_NAME="thing-name" \
+          -DS3_PRESIGNED_GET_URL="get-url" \
+          -DS3_PRESIGNED_PUT_URL="put-url" \
+          -DCLAIM_CERT_PATH="cert/path" \
+          -DCLAIM_PRIVATE_KEY_PATH="key/path" \
+          -DPROVISIONING_TEMPLATE_NAME="template-name" \
+          -DPROVISIONING_CSR_PATH="csr/path" \
+          -DPROVISIONING_PRIVATE_KEY_PATH="key/path" \
+          -DPROVISIONING_CERT_PATH="cert/path"
+      - name: Build Demos
+        run: |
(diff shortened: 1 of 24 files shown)


.github/workflows/ci.yml
```diff
@@ -45,6 +45,52 @@ jobs:
         run: |
           make -C build/ help | grep demo | tr -d '. ' | xargs make -C build/
           make -C demos/jobs/jobs_demo_mosquitto
+  build-check-demos-cpp:
+    runs-on: ubuntu-18.04
```
aggarw13, line 49: We probably should use `ubuntu-latest` wherever we can unless there is a reason of using an older version

.github/workflows/ci.yml
```diff
@@ -45,6 +45,52 @@ jobs:
...
+    steps:
+      - name: Clone This Repo
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Rename source files
+        run: |
+          for demo in mqtt http shadow defender
```
aggarw13, line 57: Instead of using include filter, an exclude filter approach can be used to reduce the nested for loop into a single command as
`find ./demos -name '*.c' -not -path "*ota*" -not -path "*pkcs11*" -type f -exec sh -c 'for f; do mv -- $f ${f%.*}.cpp; done' _ {} +`

.github/workflows/ci.yml
```diff
@@ -45,6 +45,52 @@ jobs:
...
+          -DCLAIM_CERT_PATH="cert/path" \
+          -DCLAIM_PRIVATE_KEY_PATH="key/path" \
+          -DPROVISIONING_TEMPLATE_NAME="template-name" \
+          -DPROVISIONING_CSR_PATH="csr/path" \
+          -DPROVISIONING_PRIVATE_KEY_PATH="key/path" \
+          -DPROVISIONING_CERT_PATH="cert/path"
+      - name: Build Demos
+        run: |
+          make -C build/ help | grep demo | tr -d '. ' | xargs make -C build/
+          make -C demos/jobs/jobs_demo_mosquitto
```
aggarw13, line 93: Seems like this is not required as Jobs is not tested for C++ compliance?

demos/defender/defender_demo_json/CMakeLists.txt
```diff
@@ -12,12 +12,11 @@ include( ${CMAKE_SOURCE_DIR}/libraries/standard/coreJSON/jsonFilePaths.cmake )
 # Include Defender library's source and header path variables.
 include( ${CMAKE_SOURCE_DIR}/libraries/aws/device-defender-for-aws-iot-embedded-sdk/defenderFilePaths.cmake )
 
+file( GLOB DEMO_SRCS "*.c*" )
```
aggarw13, line 15:
```suggestion
# Find all C and CPP source files (
# CPP files are searched for supporting CI build checks that verify C++ build compliance of librariesin C++ builds)
file( GLOB DEMO_SRCS "*.c*" )
```

aggarw13: Similar comment in all other CMake files would be helpful

.github/workflows/ci.yml
muneebahmed10: I copied this from the other checks which use 18.04 due to compiler warnings in the coreHTTP library from gcc versions >= 10. Since this test doesn't use the -Werror flag, it's probably fine to use the latest here though

.github/workflows/ci.yml
muneebahmed10: Thanks, copypasta from the other job

demos/defender/defender_demo_json/CMakeLists.txt
muneebahmed10: Added

.github/workflows/ci.yml
```diff
@@ -45,6 +45,51 @@ jobs:
...
+          submodules: recursive
+      - name: Rename source files
+        run: |
+          for demo in mqtt http shadow defender
+          do
+              for source in demos/$demo/*/*.c
+              do
+                  mv -- "$source" "${source%.c}.cpp"
+              done
+          done
```
muneebahmed10, line 63: Possible change
```suggestion
          find ./demos -name '*.c' -not -path "*ota*" -not -path "*pkcs11*" -not -path "*jobs*" -not -path "*fleet_provisioning*" -type f -exec sh -c 'for f; do mv -- $f ${f%.c}.cpp; done' _ {} +
```

.github/workflows/ci.yml
muneebahmed10: I didn't think a nested loop was any worse than one super long command, and probably easier to read as a loop. One difference is that the command above changes every .c file, including ones in helper directories on a different directory level, while the loop only changes the files that are in the same directory as each demo. Regardless, I made the proper CMake changes to allow the above command to work, and left it as a suggestion in another comment if you decide to commit it

yourslab:
How will this handle the OTA demo considering this new change?
https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/234/files

I suppose we could wrap `#ifdef __cplusplus` in the demo, but it will be a little strange to have that line in a demo meant for C. We can think about those edge cases later on though as this is only being enabled for a subset of the demos.

muneebahmed10:
> How will this handle the OTA demo considering this new change?
> https://github.com/aws/ota-for-aws-iot-embedded-sdk/pull/234/files
> 
> I suppose we could wrap `#ifdef __cplusplus` in the demo, but it will be a little strange to have that line in a demo meant for C. We can think about those edge cases later on though as this is only being enabled for a subset of the demos.

It won't as this PR only enables it for MQTT, HTTP, Shadow, and Device Defender. OTA doesn't even include the `extern "C" {` in its headers, so it can't be linked with C++ as is anyway right now. Also, I don't think the above change would be sufficient for linking C and C++ OTA code together in a project, so it looks like only either C or C++ can be used with OTA, not both

yourslab:
So the strategy works quite well when the demo can be compiled as both C and C++. There's still somewhat of a gap when it comes to testing libraries like OTA that would require a separate application to test its interoperability with C++. However, this is definitely a step above from not having any form of testing at all.

aggarw13 (approved):

.github/workflows/ci.yml
aggarw13: To mimic the same behavior as the nested for loop, the command could use an include filter as
`find ./demos -name '*.c'  -path "demos/mqtt*" -o -path "demos/http*" -o -path "demos/shadow*" -o -path "demos/defender*" -type f -exec sh -c 'for f; do mv -- $f ${f%.*}.cpp; done' _ {} +`

It's a minor style preference I have to limit the number of lines for bash logic in the workflow file. Either way, I'm fine with whichever logic is chosen. It may be beneficial to rename the job's step name to "Rename demo source files to .cpp extension" to make it more understandable.

aggarg (approved):

PR #1680, opened 2021-07-29
Add transport interface using MbedTLS and corePKCS11

archigup:
Adds a new transport interface that will be used by the fleet provisioning demo.


platform/posix/transport/CMakeLists.txt
```diff
@@ -40,12 +40,47 @@ target_link_libraries( openssl_posix
...
+# Include PKCS #11 library's source and header path variables.
+include( ${COREPKCS11_LOCATION}/pkcsFilePaths.cmake )
+
+list(APPEND PKCS_SOURCES
+    "${COREPKCS11_LOCATION}/source/portable/posix/core_pkcs11_pal.c"
+    "${CORE_PKCS11_3RDPARTY_LOCATION}/mbedtls_utils/mbedtls_utils.c"
+)
+
+# Create target for POSIX implementation of MbedTLS.
+add_library( mbedtls_pkcs11_posix
```
aggarw13, line 56: I think it is valuable to mention the `transport_` prefix across all the transport implementation target libraries
```suggestion
add_library( transport_mbedtls_pkcs11_posix
```

platform/posix/transport/CMakeLists.txt
```diff
@@ -40,12 +40,47 @@ target_link_libraries( openssl_posix
...
 
+# Set path to corePKCS11 and it's third party libraries.
+set(COREPKCS11_LOCATION "${CMAKE_SOURCE_DIR}/libraries/standard/corePKCS11")
+set(CORE_PKCS11_3RDPARTY_LOCATION "${COREPKCS11_LOCATION}/source/dependency/3rdparty")
+
+# Include PKCS #11 library's source and header path variables.
+include( ${COREPKCS11_LOCATION}/pkcsFilePaths.cmake )
+
+list(APPEND PKCS_SOURCES
+    "${COREPKCS11_LOCATION}/source/portable/posix/core_pkcs11_pal.c"
```
aggarw13, line 51: These files be can be added in the `pkcsFilePaths.cmake` file under `CORE_PKCS11_LINUX_PAL` sources

platform/posix/transport/CMakeLists.txt
```diff
@@ -40,12 +40,47 @@ target_link_libraries( openssl_posix
...
+
+target_include_directories(
+    mbedtls_pkcs11_posix
+    PUBLIC
+        ${COMMON_TRANSPORT_INCLUDE_PUBLIC_DIRS}
+        ${LOGGING_INCLUDE_DIRS}
+        ${TRANSPORT_INTERFACE_INCLUDE_DIR}
+        ${PKCS_INCLUDE_PUBLIC_DIRS}
+        ${CORE_PKCS11_3RDPARTY_LOCATION}/pkcs11
+        ${DEMOS_DIR}/pkcs11/common/include
```
aggarw13, line 72: Why is inclusion of path from a demo folder required?

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
+/*
+ * FreeRTOS V202107.00
```
aggarw13, line 2: Please update other files as well
```suggestion
 * AWS IoT Device SDK for Embedded C 202103.00
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
aggarw13, line 3:
```suggestion
 * Copyright (C) 2021 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```

platform/posix/transport/CMakeLists.txt
archigup: That has the corePKCS11 config

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+ * 3. Include the header file "logging_stack.h".
+ */
+
+/* Include header that defines log levels. */
+#include "logging_levels.h"
+
+/* Logging configuration for the transport interface implementation which uses
+ * MbedTLS and Sockets. */
+#ifndef LIBRARY_LOG_NAME
+    #define LIBRARY_LOG_NAME     "Transport_MbedTLS_PKCS11_Sockets"
```
aggarw13, line 42: `Sockets` prefix does not seem necessary in the logs. (Possibly that can be removed from logging configuration of other files as well)
```suggestion
    #define LIBRARY_LOG_NAME     "Transport_MbedTLS_PKCS11"
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+#ifdef __cplusplus
+    extern "C" {
+#endif
+/* *INDENT-ON* */
+
+/* Standard includes. */
+#include <stdbool.h>
+
+/* MbedTLS includes. */
+#include "mbedtls/net_sockets.h"
```
aggarw13, line 62: Is inclusion of this header required?

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+ * @param[in] pMbedtlsPkcs11Credentials Credentials for the TLS connection.
+ *
+ * @note A timeout of 0 means infinite timeout.
+ *
+ * @return #MBEDTLS_PKCS11_SUCCESS on success;
+ * #MBEDTLS_PKCS11_INSUFFICIENT_MEMORY, #MBEDTLS_PKCS11_INVALID_CREDENTIALS,
+ * #MBEDTLS_PKCS11_HANDSHAKE_FAILED, #MBEDTLS_PKCS11_INTERNAL_ERROR,
+ * or #MBEDTLS_PKCS11_CONNECT_FAILURE on failure.
+ */
+MbedtlsPkcs11Status_t Mbedtls_Pkcs11_Connect( NetworkContext_t * pNetworkContext,
```
aggarw13, line 150: We have discussed in the past that the Connect/Disconnect calls do not need to use `NetworkContext_t` input type as they are not part of the transport interface, 
```suggestion
MbedtlsPkcs11Status_t Mbedtls_Pkcs11_Connect( MbedtlsPkcs11Context_t * pSslContext,
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * @brief Disable server name indication (SNI) for a TLS session.
+     */
+    bool disableSni;
+
+    const char * pRootCaPath;     /**< @brief String representing a trusted server root certificate. */
+    char * pClientCertLabel;      /**< @brief String representing the PKCS #11 label for the client certificate. */
+    char * pPrivateKeyLabel;      /**< @brief String representing the PKCS #11 label for the private key. */
+    CK_SESSION_HANDLE p11Session; /**< @brief PKCS #11 session handle. */
```
aggarw13, line 132: Why does the PKCS11 session need to be passed as a "credential"?

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+
+/**
+ * @brief Sets up a TLS session on top of a TCP connection using the MbedTLS API.
+ *
+ * @param[out] pNetworkContext The output parameter to return the created network context.
+ * @param[in] pHostName The hostname of the remote endpoint.
+ * @param[in] port The destination port.
+ * @param[in] pMbedtlsPkcs11Credentials Credentials for the TLS connection.
+ *
+ * @note A timeout of 0 means infinite timeout.
```
aggarw13, line 143:
```suggestion
 * @note A timeout of 0 means that the function will block indefinitely.
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+    bool disableSni;
+
+    const char * pRootCaPath;     /**< @brief String representing a trusted server root certificate. */
+    char * pClientCertLabel;      /**< @brief String representing the PKCS #11 label for the client certificate. */
+    char * pPrivateKeyLabel;      /**< @brief String representing the PKCS #11 label for the private key. */
+    CK_SESSION_HANDLE p11Session; /**< @brief PKCS #11 session handle. */
+} MbedtlsPkcs11Credentials_t;
+
+/**
+ * @brief Sets up a TLS session on top of a TCP connection using the MbedTLS API.
```
aggarw13, line 136:
```suggestion
 * @brief Sets up a TLS session on top of a TCP connection using MbedTLS library for TLS establishment and corePKCS11 library for obtaining credentials for the mutually authenticated TLS session.
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+
+/* Transport interface include. */
+#include "transport_interface.h"
+
+/* PKCS #11 includes. */
+#include "core_pkcs11.h"
+
+/**
+ * @brief Definition of the network context for the transport interface
+ * implementation that uses mbedTLS and FreeRTOS+TLS sockets.
```
aggarw13, line 75:
```suggestion
 * @brief Context representing state for the MbedTLS and PKCS11 based
 * transport interface implementation. 
 * @note Applications using this transport interface implementation should 
 * should this structure as the #NetworkContext_t for the  transport interface 
 * configuration passed to coreMQTT or coreHTTP libraries.
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+    mbedtls_ssl_config config;            /**< @brief SSL connection configuration. */
+    mbedtls_ssl_context context;          /**< @brief SSL connection context */
+    mbedtls_x509_crt_profile certProfile; /**< @brief Certificate security profile for this connection. */
+    mbedtls_x509_crt rootCa;              /**< @brief Root CA certificate context. */
+    mbedtls_x509_crt clientCert;          /**< @brief Client certificate context. */
+    mbedtls_pk_context privKey;           /**< @brief Client private key context. */
+    mbedtls_pk_info_t privKeyInfo;        /**< @brief Client private key info. */
+
+    /* PKCS #11. */
+    CK_FUNCTION_LIST_PTR pP11FunctionList;
```
aggarw13, line 89: To maintain symmetry in documentation of members with the mbedTLS members, can you add documentation for the PKCS11 members?

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+    CK_SESSION_HANDLE p11Session;
+    CK_OBJECT_HANDLE p11PrivateKey;
+    CK_KEY_TYPE keyType;
```
aggarw13, line 92: Is this member required in the context? Is it possible to have different values for this member?

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * @param[in] pNetworkContext Network context.
+ */
+void Mbedtls_Pkcs11_Disconnect( NetworkContext_t * pNetworkContext );
+
+/**
+ * @brief Receives data over an established TLS session using the MbedTLS API.
+ *
+ * This can be used as #TransportInterface.recv function for receiving data
+ * from the network.
```
aggarw13, line 167:
```suggestion
 * This represents definition for the the #TransportInterface.recv function of the transport interface
 * to receive data from the network.
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+ */
+int32_t Mbedtls_Pkcs11_Recv( NetworkContext_t * pNetworkContext,
+                             void * pBuffer,
+                             size_t bytesToRecv );
+
+/**
+ * @brief Sends data over an established TLS session using the MbedTLS API.
+ *
+ * This can be used as the #TransportInterface.send function to send data
+ * over the network.
```
aggarw13, line 184:
```suggestion
 * This represents definition for the the #TransportInterface.send function of the transport interface
 * to send data over the network.
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+        }
+
+        /* Free contexts. */
+        contextFree( pMbedtlsPkcs11Context );
+    }
+}
+
+/*-----------------------------------------------------------*/
+
+int32_t Mbedtls_Pkcs11_Recv( NetworkContext_t * pNetworkContext,
```
aggarw13, line 875: We recently made changes to the Plaintext and OpenSSL transport implementations to add non-blocking behavior by using `poll` because we suggest the behavior in our documentation to customers. Please update these recv/send functions to have similar behavior

platform/posix/transport/CMakeLists.txt
aggarw13: Can you add a comment about it?

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -0,0 +1,205 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+#ifndef MBEDTLS_PKCS11_POSIX_H_
+#define MBEDTLS_PKCS11_POSIX_H_
+
```
aggarw13, line 25: Please add a `@brief` comment to help customers know the purpose of this library

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+ * @return #MBEDTLS_PKCS11_SUCCESS, #MBEDTLS_PKCS11_INSUFFICIENT_MEMORY, #MBEDTLS_PKCS11_INVALID_CREDENTIALS,
+ * #MBEDTLS_PKCS11_HANDSHAKE_FAILED, or #MBEDTLS_PKCS11_INTERNAL_ERROR.
+ */
+static MbedtlsPkcs11Status_t configureMbedtls( MbedtlsPkcs11Context_t * pMbedtlsPkcs11Context,
+                                               const char * pHostName,
+                                               const MbedtlsPkcs11Credentials_t * pMbedtlsPkcs11Credentials,
+                                               uint32_t recvTimeoutMs );
+
+/**
+ * @brief Configure the credentials in the MbedTLS SSL context.
```
aggarw13, line 124: Please mention which types certificates (Client, Root CA) are configured by the function

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+ *
+ * @param[in] pMbedtlsPkcs11Context Network context.
+ * @param[in] pHostName Remote host name, used for server name indication.
+ * @param[in] pMbedtlsPkcs11Credentials TLS setup parameters.
+ * @param[in] recvTimeoutMs Receive timeout for network socket.
+ *
+ * @return #MBEDTLS_PKCS11_SUCCESS, #MBEDTLS_PKCS11_INSUFFICIENT_MEMORY, #MBEDTLS_PKCS11_INVALID_CREDENTIALS,
+ * #MBEDTLS_PKCS11_HANDSHAKE_FAILED, or #MBEDTLS_PKCS11_INTERNAL_ERROR.
+ */
+static MbedtlsPkcs11Status_t configureMbedtls( MbedtlsPkcs11Context_t * pMbedtlsPkcs11Context,
```
aggarw13, line 118: If this function actually creates the TLS session, should it be named appropriately as `establishTlsSession`?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+ *
+ * @return #MBEDTLS_PKCS11_SUCCESS on success,
+ * #MBEDTLS_PKCS11_INVALID_CREDENTIALS on error.
+ */
+static MbedtlsPkcs11Status_t configureMbedtlsSniAlpn( MbedtlsPkcs11Context_t * pMbedtlsPkcs11Context,
+                                                      const MbedtlsPkcs11Credentials_t * pMbedtlsPkcs11Credentials,
+                                                      const char * pHostName );
+
+/**
+ * @brief Configure the SNI and Maximum Fragment Length in the MbedTLS SSL context.
```
aggarw13, line 150:
```suggestion
 * @brief Configure the Maximum Fragment Length in the MbedTLS SSL context.
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+    return returnStatus;
+}
+
+/*-----------------------------------------------------------*/
+
+static MbedtlsPkcs11Status_t configureMbedtlsCertificates( MbedtlsPkcs11Context_t * pMbedtlsPkcs11Context,
+                                                           const MbedtlsPkcs11Credentials_t * pMbedtlsPkcs11Credentials )
+
+{
+    MbedtlsPkcs11Status_t returnStatus = MBEDTLS_PKCS11_SUCCESS;
```
aggarw13, line 367: Please add missing asserts for validating input parameters across all functions

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+ * if the code-contains a level-level code; otherwise, using a default string.
+ */
+#define mbedtlsLowLevelCodeOrDefault( mbedTlsCode )       \
+    ( mbedtls_low_level_strerr( mbedTlsCode ) != NULL ) ? \
+    mbedtls_low_level_strerr( mbedTlsCode ) : pNoLowLevelMbedTlsCodeStr
+
+/**
+ * @brief Debug logging level to use for MbedTLS.
+ */
+#define MBEDTLS_DEBUG_LEVEL    0
```
aggarw13, line 89: 4 is the level that for debug logging as can be seen in the mbedTLS documentation
https://github.com/ARMmbed/mbedtls/blob/v2.26.0/include/mbedtls/debug.h#L140-L144
```suggestion
#define MBEDTLS_DEBUG_LEVEL    4
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+ * @brief Helper for reading the specified certificate object, if present,
+ * out of storage, into RAM, and then into an mbedTLS certificate context
+ * object.
+ *
+ * @param[in] pContext Caller TLS context.
+ * @param[in] pLabelName PKCS #11 certificate object label.
+ * @param[in] class PKCS #11 certificate object class.
+ * @param[out] pCertificateContext Certificate context.
+ *
+ * @return Zero on success.
```
aggarw13, line 182: Should this function return a `bool` type if only returns binary information of success or failure?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+                                           CK_OBJECT_CLASS class,
+                                           mbedtls_x509_crt * pCertificateContext );
+
+/**
+ * @brief Helper for setting up potentially hardware-based cryptographic context.
+ *
+ * @param pCtx Caller context.
+ * @param pPrivateKeyLabel PKCS #11 label for the private key.
+ *
+ * @return Zero on success.
```
aggarw13, line 195: Should the documentation mention `CKR_OK` instead? Also, what is the return code for failure?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+ */
+static int32_t readCertificateIntoContext( MbedtlsPkcs11Context_t * pContext,
+                                           char * pLabelName,
+                                           CK_OBJECT_CLASS class,
+                                           mbedtls_x509_crt * pCertificateContext );
+
+/**
+ * @brief Helper for setting up potentially hardware-based cryptographic context.
```
aggarw13, line 190: "_Potentially_ hardware-based cryptographic context" is unclear information. Can you update the comment to be more descriptive for customers?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
aggarw13: Same comment for the `privateKeySigningCallback` function documentation on return information being incomplete

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+        /* Set up the certificate security profile, starting from the default value. */
+        pMbedtlsPkcs11Context->certProfile = mbedtls_x509_crt_profile_default;
+
+        /* Set SSL authmode and the RNG context. */
+        mbedtls_ssl_conf_authmode( &( pMbedtlsPkcs11Context->config ), MBEDTLS_SSL_VERIFY_REQUIRED );
+        mbedtls_ssl_conf_rng( &( pMbedtlsPkcs11Context->config ), generateRandomBytes, pMbedtlsPkcs11Context );
+        mbedtls_ssl_conf_cert_profile( &( pMbedtlsPkcs11Context->config ), &( pMbedtlsPkcs11Context->certProfile ) );
+        mbedtls_ssl_conf_read_timeout( &( pMbedtlsPkcs11Context->config ), recvTimeoutMs );
+        mbedtls_ssl_conf_dbg( &pMbedtlsPkcs11Context->config, mbedtlsDebugPrint, NULL );
+        mbedtls_debug_set_threshold( MBEDTLS_DEBUG_LEVEL );
```
aggarw13, line 312: We don't want to enable logging from mbedTLS by default, instead a new configuration can be exposed as `MBEDTLS_ENABLE_LOGGING` in the header file that is disabled by default. If the user wants to enable logging, they can update the configuration. 
It will be important to add a documentation note in README about the way to enabled mbedTLS logging for debugging TLS connection issues.

archigup: MBEDTLS_DEBUG_LEVEL is defined in that file, and set to 0, so it is not enabled by default. I'll add docs about it there

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+    else
+    {
+        /* Set up the certificate security profile, starting from the default value. */
+        pMbedtlsPkcs11Context->certProfile = mbedtls_x509_crt_profile_default;
+
+        /* Set SSL authmode and the RNG context. */
+        mbedtls_ssl_conf_authmode( &( pMbedtlsPkcs11Context->config ), MBEDTLS_SSL_VERIFY_REQUIRED );
+        mbedtls_ssl_conf_rng( &( pMbedtlsPkcs11Context->config ), generateRandomBytes, pMbedtlsPkcs11Context );
+        mbedtls_ssl_conf_cert_profile( &( pMbedtlsPkcs11Context->config ), &( pMbedtlsPkcs11Context->certProfile ) );
+        mbedtls_ssl_conf_read_timeout( &( pMbedtlsPkcs11Context->config ), recvTimeoutMs );
```
aggarw13, line 310: Does this configure the socket timeout for all network recv operations? In that case, it is important to mention in the API documentation of the `Connect` function that the `recvTimeoutMs` input parameter affects the maximum blocking time of the Recv function implementation of transport interface in this library

platform/posix/transport/src/mbedtls_pkcs11_posix.c
archigup: Yes, that is to have it off by default

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+        returnStatus = configureMbedtlsFragmentLength( pMbedtlsPkcs11Context );
+    }
+
+    if( returnStatus != MBEDTLS_PKCS11_SUCCESS )
+    {
+        contextFree( pMbedtlsPkcs11Context );
+    }
+    else
+    {
+        LogInfo( ( "(Network connection %p) TLS handshake successful.", pNetworkContext ) );
```
aggarw13, line 355: There isn't value to the user in seeing the pointer value, so its better to log the endpoint instead
```suggestion
        LogInfo( ( "TLS handshake successful with host %s", pHostName) );
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+            returnStatus = MBEDTLS_PKCS11_INVALID_CREDENTIALS;
+        }
+    }
+
+    if( returnStatus == MBEDTLS_PKCS11_SUCCESS )
+    {
+        /* Setup the client certificate. */
+        mbedtlsError = readCertificateIntoContext( pMbedtlsPkcs11Context,
+                                                   pMbedtlsPkcs11Credentials->pClientCertLabel,
+                                                   CKO_CERTIFICATE,
```
aggarw13, line 403: Why does this parameter need to be sent when the function mentions that it is meant for reading only certificate types from PKCS11?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+static MbedtlsPkcs11Status_t configureMbedtlsFragmentLength( MbedtlsPkcs11Context_t * pMbedtlsPkcs11Context )
+{
+    MbedtlsPkcs11Status_t returnStatus = MBEDTLS_PKCS11_SUCCESS;
+
+    /* Set Maximum Fragment Length if enabled. */
+    #ifdef MBEDTLS_SSL_MAX_FRAGMENT_LENGTH
+        int32_t mbedtlsError = 0;
+
+        /* Enable the max fragment extension. 4096 bytes is currently the largest fragment size permitted.
+         * See RFC 8449 https://tools.ietf.org/html/rfc8449 for more information.
```
aggarw13, line 477: RFC 8449 represents a different extension, the "record size limit" which seems to be the new extension to replace "maximum fragment length". However, this mbed TLS function seems to be for the "maximum fragment length" extension that is defined in [RFC6066](https://datatracker.ietf.org/doc/html/rfc6066#page-8).
```suggestion
        /* Enable the max fragment extension. 4096 bytes is currently the largest fragment size permitted.
         * See RFC 6066 https://tools.ietf.org/html/rfc6066 for more information.
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+/*-----------------------------------------------------------*/
+
+static int32_t generateRandomBytes( void * pCtx,
+                                    unsigned char * pRandom,
+                                    size_t randomLength )
+{
+    /* Must cast from void pointer to conform to MbedTLS API. */
+    MbedtlsPkcs11Context_t * pContext = ( MbedtlsPkcs11Context_t * ) pCtx;
+    CK_RV xResult;
+
```
aggarw13, line 503: Extra new line can be removed

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+    CK_RV pkcs11Ret = CKR_OK;
+    CK_ATTRIBUTE template = { 0 };
+    CK_OBJECT_HANDLE certificateHandle = 0;
+    int32_t mbedtlsRet = 0;
+
+    /* Get the handle of the certificate. */
+    pkcs11Ret = xFindObjectWithLabelAndClass( pContext->p11Session,
+                                              pLabelName,
+                                              strlen( pLabelName ),
+                                              class,
```
aggarw13, line 531: Seems like this can just be `CKO_CERTIFICATE` instead of being passed as an input parameter?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+    }
+
+    /* Create a buffer for the certificate. */
+    if( CKR_OK == pkcs11Ret )
+    {
+        template.pValue = malloc( template.ulValueLen );
+
+        if( NULL == template.pValue )
+        {
+            pkcs11Ret = CKR_HOST_MEMORY;
```
aggarw13, line 558: Error can be logged about failure in allocating heap memory for certificate

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+    {
+        mbedtlsRet = mbedtls_x509_crt_parse( pCertificateContext,
+                                             ( const unsigned char * ) template.pValue,
+                                             template.ulValueLen );
+    }
+
+    /* Free memory. */
+    free( template.pValue );
+
+    return mbedtlsRet;
```
aggarw13, line 582: If there is an error in the PKCS11 logic, this function will still return `0` because of the default value of `mbedtlsRet`. Instead, why not fix this function by returning a `bool` type?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+}
+
+/*-----------------------------------------------------------*/
+
+static CK_RV initializeClientKeys( MbedtlsPkcs11Context_t * pCtx,
+                                   char * pPrivateKeyLabel )
+{
+    CK_RV ret = CKR_OK;
+    CK_ATTRIBUTE template[ 2 ] = { 0 };
+    mbedtls_pk_type_t keyAlgo = ( mbedtls_pk_type_t ) ~0;
```
aggarw13, line 592: What is the purpose of the `~0` logic? Can you provide a comment for it?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+
+    /* Free memory. */
+    free( template.pValue );
+
+    return mbedtlsRet;
+}
+
+/*-----------------------------------------------------------*/
+
+static CK_RV initializeClientKeys( MbedtlsPkcs11Context_t * pCtx,
```
aggarw13, line 587:
```suggestion
static CK_RV initializeClientKeys( MbedtlsPkcs11Context_t * pContext,
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+        ret = CK_INVALID_HANDLE;
+        LogError( ( "Could not find private key." ) );
+    }
+
+    /* Query the device private key type. */
+    if( ret == CKR_OK )
+    {
+        template[ 0 ].type = CKA_KEY_TYPE;
+        template[ 0 ].pValue = &pCtx->keyType;
+        template[ 0 ].ulValueLen = sizeof( CK_KEY_TYPE );
```
aggarw13, line 612: Shouldn't this be `sizeof( &pCtx->keyType )`?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
aggarw13: Should it be `CKR_ATTRIBUTE_VALUE_INVALID` instead?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+                                          int32_t ( *pRng )( void *,
+                                                             unsigned char *,
+                                                             size_t ),
+                                          void * pRngContext )
+{
+    CK_RV ret = CKR_OK;
+    int32_t result = 0;
+    MbedtlsPkcs11Context_t * pMbedtlsPkcs11Context = ( MbedtlsPkcs11Context_t * ) pContext;
+    CK_MECHANISM mech = { 0 };
+    CK_BYTE toBeSigned[ 256 ];
```
aggarw13, line 668: Why is 256 bytes chosen as the hash len? Is this the standard hash size used across all TLS handshakes for verifying certificate?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+    {
+        /* Use the PKCS#11 module to sign. */
+        ret = pMbedtlsPkcs11Context->pP11FunctionList->C_SignInit( pMbedtlsPkcs11Context->p11Session,
+                                                                   &mech,
+                                                                   pMbedtlsPkcs11Context->p11PrivateKey );
+    }
+
+    if( CKR_OK == ret )
+    {
+        *pSigLen = sizeof( toBeSigned );
```
aggarw13, line 715: This assignment seems unnecessary as `C_Sign` will update the value with the signature length?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+        LogError( ( "Invalid input parameter(s): Arguments cannot be NULL. pNetworkContext=%p, "
+                    "pHostName=%p, pMbedtlsPkcs11Credentials=%p.",
+                    ( void * ) pNetworkContext,
+                    ( const void * ) pHostName,
+                    ( const void * ) pMbedtlsPkcs11Credentials ) );
+        returnStatus = MBEDTLS_PKCS11_INVALID_PARAMETER;
+    }
+    else
+    {
+        snprintf( portStr, 6, "%u", port );
```
aggarw13, line 776: What is the purpose of this `snprintf` call?

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+    }
+
+    /* Clean up on failure. */
+    if( returnStatus != MBEDTLS_PKCS11_SUCCESS )
+    {
+        contextFree( pMbedtlsPkcs11Context );
+    }
+    else
+    {
+        LogInfo( ( "(Network connection %p) Connection to %s established.", pNetworkContext, pHostName ) );
```
aggarw13, line 823:
```suggestion
        LogInfo( ( "Connection to %s established.", pHostName ) );
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+        /* Attempting to terminate TLS connection. */
+        tlsStatus = mbedtls_ssl_close_notify( &( pMbedtlsPkcs11Context->context ) );
+
+        /* Ignore the WANT_READ and WANT_WRITE return values. */
+        if( ( tlsStatus != MBEDTLS_ERR_SSL_WANT_READ ) &&
+            ( tlsStatus != MBEDTLS_ERR_SSL_WANT_WRITE ) )
+        {
+            if( tlsStatus == 0 )
+            {
+                LogInfo( ( "(Network connection %p) TLS close-notify sent.",
```
aggarw13, line 848: The logging of the network context pointer in all the logs can be avoided as it does not provide any debug value

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+            }
+        }
+        else
+        {
+            /* WANT_READ and WANT_WRITE can be ignored. Logging for debugging purposes. */
+            LogInfo( ( "(Network connection %p) TLS close-notify sent; ",
+                       "received %s as the TLS status can be ignored for close-notify."
+                       ( tlsStatus == MBEDTLS_ERR_SSL_WANT_READ ) ? "WANT_READ" : "WANT_WRITE",
+                       pNetworkContext ) );
+        }
```
aggarw13, line 866:
```suggestion

        if( tlsStatus == 0 )
        {
            LogInfo( ( "Closing TLS connection: TLS close-notify sent." ) );
        }
        else if( ( tlsStatus == MBEDTLS_ERR_SSL_WANT_READ ) ||
                    ( tlsStatus == MBEDTLS_ERR_SSL_WANT_WRITE ) )
        {
            /* WANT_READ and WANT_WRITE can be ignored. Logging for debugging purposes. */
            LogInfo( ( "TLS close-notify sent; ",
                        "received %s as the TLS status can be ignored for close-notify."
                        ( tlsStatus == MBEDTLS_ERR_SSL_WANT_READ ) ? "WANT_READ" : "WANT_WRITE" ) );
        }
        else
        {
            LogError( ( " Failed to send TLS close-notify: mbedTLSError= %s : %s.",
                        mbedtlsHighLevelCodeOrDefault( tlsStatus ),
                        mbedtlsLowLevelCodeOrDefault( tlsStatus ) ) );
        }
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -0,0 +1,958 @@
...
+                    mbedtlsLowLevelCodeOrDefault( mbedtlsError ) ) );
+        returnStatus = MBEDTLS_PKCS11_INVALID_CREDENTIALS;
+    }
+    else
+    {
+        mbedtls_ssl_conf_ca_chain( &( pMbedtlsPkcs11Context->config ),
+                                   &( pMbedtlsPkcs11Context->rootCa ),
+                                   NULL );
+        /* Setup the client private key. */
+        result = initializeClientKeys( pMbedtlsPkcs11Context,
```
aggarw13, line 388: Use of client private key can be made optional to support use cases of basic TLS as well as HTTP requests to AWS services that use SigV4

archigup: In those cases you wouldn't be using the PKCS11 transport interface as a key in the HSM is not used.

platform/posix/transport/include/mbedtls_pkcs11_posix.h
archigup: These are required for struct types

platform/posix/transport/include/mbedtls_pkcs11_posix.h
archigup: Yes. Both RSA and EC keys are used

platform/posix/transport/src/mbedtls_pkcs11_posix.c
archigup: It does not; the log was incorrect.

platform/posix/transport/src/mbedtls_pkcs11_posix.c
archigup: It is needed.

platform/posix/transport/src/mbedtls_pkcs11_posix.c
archigup: privateKeySigningCallback needs to return int to match MbedTLS API.

platform/posix/transport/src/mbedtls_pkcs11_posix.c
archigup: That is just a buffer size big enough to hold the types of hashes.

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -86,10 +98,10 @@ typedef struct MbedtlsPkcs11Context
     mbedtls_pk_info_t privKeyInfo;        /**< @brief Client private key info. */
 
     /* PKCS #11. */
-    CK_FUNCTION_LIST_PTR pP11FunctionList;
-    CK_SESSION_HANDLE p11Session;
-    CK_OBJECT_HANDLE p11PrivateKey;
-    CK_KEY_TYPE keyType;
+    CK_FUNCTION_LIST_PTR pP11FunctionList; /**< @brief PKCS #11 function list. */
+    CK_SESSION_HANDLE p11Session;          /**< @brief PKCS #11 session. */
+    CK_OBJECT_HANDLE p11PrivateKey;        /**< @brief PKCS #11 handle for the private key to use for client auth. */
```
aggarw13, line 103:
```suggestion
    CK_OBJECT_HANDLE p11PrivateKey;        /**< @brief PKCS #11 handle for the private key to use for client authentication. */
```

platform/posix/transport/include/mbedtls_pkcs11_posix.h
```diff
@@ -180,8 +195,8 @@ int32_t Mbedtls_Pkcs11_Recv( NetworkContext_t * pNetworkContext,
 /**
  * @brief Sends data over an established TLS session using the MbedTLS API.
  *
- * This can be used as the #TransportInterface.send function to send data
- * over the network.
+ * This function can be used as the #TransportInterface.send implementation of
+ * the transport interface to receive data from the network.
```
aggarw13, line 199:
```suggestion
 * the transport interface to send data over the network.
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -85,8 +81,12 @@ static const char * pNoLowLevelMbedTlsCodeStr = "<No-Low-Level-Code>";
 
 /**
  * @brief Debug logging level to use for MbedTLS.
+ *
+ * @note The default value of 0 disables MbedTLS logging.
+ * See <https://tls.mbed.org/api/debug_8h.html#a6629362e96b43725ace95c8ff01d9985>
```
aggarw13, line 86: Why wrap the URL in angle brackets?
```suggestion
 * See https://tls.mbed.org/api/debug_8h.html#a6629362e96b43725ace95c8ff01d9985
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -176,29 +178,28 @@ static int32_t generateRandomBytes( void * pCtx,
...
-                                           char * pLabelName,
-                                           CK_OBJECT_CLASS class,
-                                           mbedtls_x509_crt * pCertificateContext );
+static bool readCertificateIntoContext( MbedtlsPkcs11Context_t * pContext,
+                                        char * pLabelName,
+                                        mbedtls_x509_crt * pCertificateContext );
 
 /**
- * @brief Helper for setting up potentially hardware-based cryptographic context.
+ * @brief Helper for configuring MbedTLS to use client key from PKCS #11.
```
aggarw13, line 190:
```suggestion
 * @brief Helper for configuring MbedTLS to use client private key from PKCS #11.
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -176,29 +178,28 @@ static int32_t generateRandomBytes( void * pCtx,
...
+ * @param pContext Caller context.
  * @param pPrivateKeyLabel PKCS #11 label for the private key.
  *
- * @return Zero on success.
+ * @return True on success.
  */
-static CK_RV initializeClientKeys( MbedtlsPkcs11Context_t * pCtx,
-                                   char * pPrivateKeyLabel );
+static bool initializeClientKeys( MbedtlsPkcs11Context_t * pContext,
+                                  char * pPrivateKeyLabel );
```
aggarw13, line 198: Shouldn't constness be added to the label parameter?
```suggestion
                                 const char * pPrivateKeyLabel );
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -85,8 +81,12 @@ static const char * pNoLowLevelMbedTlsCodeStr = "<No-Low-Level-Code>";
 
 /**
  * @brief Debug logging level to use for MbedTLS.
+ *
+ * @note The default value of 0 disables MbedTLS logging.
+ * See <https://tls.mbed.org/api/debug_8h.html#a6629362e96b43725ace95c8ff01d9985>
+ * for valid values.
  */
-#define MBEDTLS_DEBUG_LEVEL    0
+#define MBEDTLS_DEBUG_LOG_LEVEL    0
```
aggarw13, line 89: It would be better to expose this at least in the header file of the transport library, if not `demo_config.h` file, so that customers can find a way to enable mbedTLS logging more easily. However, it would still be valuable to mention this in the README under a Troubleshooting section.

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -549,18 +565,19 @@ static int32_t readCertificateIntoContext( MbedtlsPkcs11Context_t * pContext,
...
 
     /* Create a buffer for the certificate. */
-    if( CKR_OK == pkcs11Ret )
+    if( pkcs11Ret == CKR_OK )
     {
         template.pValue = malloc( template.ulValueLen );
 
         if( NULL == template.pValue )
         {
+            LogError( ( "Falied to allocate memory for certificate buffer." ) );
```
aggarw13, line 574:
```suggestion
            LogError( ( "Failed to allocate %u bytes of memory for certificate buffer.", template.ulValueLen ) );
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
```diff
@@ -820,7 +844,7 @@ MbedtlsPkcs11Status_t Mbedtls_Pkcs11_Connect( NetworkContext_t * pNetworkContext
     }
     else
     {
-        LogInfo( ( "(Network connection %p) Connection to %s established.", pNetworkContext, pHostName ) );
+        LogInfo( ( "Connection to %s established.", pHostName ) );
```
aggarw13, line 847:
```suggestion
        LogInfo( ( "TLS connection to %s established.", pHostName ) );
```

platform/posix/transport/src/mbedtls_pkcs11_posix.c
archigup: Not sure about the README, as this is only for this transport interface, but that is a different concern. I'll move it to the header file.

platform/posix/transport/src/mbedtls_pkcs11_posix.c
aggarw13: But what if a customer wants to use the same TLS stack for HTTP operations with S3? They wouldn't want to have 2 different TLS stacks for MQTT and HTTP operations separately

archigup: They would need a separate transport interface anyways. For that we could also have a mbedtls without pkcs11 transport interface. I think that is unrelated to this PR though.

aggarw13 (dismissed):

aggarw13 (dismissed):

cobusve (dismissed):

aggarw13 (approved):

cobusve (approved):

PR #1714, opened 2021-08-31
correct cmake -S command

mcerdeiro:
correct "cmake -S." to "cmake -S ."

*Issue #, if available:*

In my system using cmake 3.10.2 and also from what I can read from the cmake documentation https://cmake.org/cmake/help/latest/manual/cmake.1.html the correct way to call cmake is with space between -S and the path.

*Description of changes:*

Correct documentation according to CMake.


paulbartell:
@mcerdeiro Thanks very much for the contribution. I'll ask the team to keep an eye out for other instances where cmake -S <src_dir> is used without a space.

paulbartell (approved):

cobusve (approved):

PR #1774, opened 2022-01-04
return negative value when byteToSend/Recv is 0

ActoryOu:
*Issue #, if available:*

*Description of changes:*
refer to [API reference](https://aws.github.io/aws-iot-device-sdk-embedded-C/202108.00/libraries/standard/coreMQTT/docs/doxygen/output/html/group__mqtt__callback__types.html)

Openssl_Recv should return negative value when byteToRecv is 0.
Openssl_Send should return negative value when byteToSend is 0.

TransportRecv_t:

If no data is available on the network to read and no error has occurred, zero MUST be the return value. A zero return value SHOULD represent that the read operation can be retried by calling the API function. Zero MUST NOT be returned if a network disconnection has occurred.

TransportSend_t:

If no data is transmitted over the network due to a full TX buffer and no network error has occurred, this MUST return zero as the return value. A zero return value SHOULD represent that the send operation can be retried by calling the API function. Zero MUST NOT be returned if a network disconnection has occurred.

### DIFF
--- a/platform/posix/transport/src/openssl_posix.c
+++ b/platform/posix/transport/src/openssl_posix.c
@@ -725,11 +725,23 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
     if( ( pNetworkContext == NULL ) || ( pNetworkContext->pParams == NULL ) )
     {
         LogError( ( "Parameter check failed: pNetworkContext is NULL." ) );
+        bytesReceived = -1;
     }
     else if( pNetworkContext->pParams->pSsl == NULL )
     {
         LogError( ( "Failed to receive data over network: "
                     "SSL object in network context is NULL." ) );
+        bytesReceived = -1;
+    }
+    else if( pBuffer == NULL )
+    {
+        LogError( ( "Failed to receive data over network: "
+                    "pBuffer is NULL." ) );
+        bytesReceived = -1;
+    }
+    else if( bytesToRecv == 0 )
+    {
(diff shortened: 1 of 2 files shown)


platform/posix/transport/src/openssl_posix.c
```diff
@@ -752,13 +752,11 @@ int32_t Openssl_Recv( NetworkContext_t * pNetworkContext,
     OpensslParams_t * pOpensslParams = NULL;
     int32_t bytesReceived = 0;
 
-    if( bytesToRecv == 0 )
+    if( !isValidNetworkContext( pNetworkContext ) ||
```
chinglee-iot, line 755: Suggest to separate the parameter check condition to clearly indicate error.

platform/posix/transport/src/openssl_posix.c
```diff
@@ -869,13 +867,11 @@ int32_t Openssl_Send( NetworkContext_t * pNetworkContext,
     OpensslParams_t * pOpensslParams = NULL;
     int32_t bytesSent = 0;
 
-    if( bytesToSend == 0 )
-    {
-        LogDebug( ( "bytesToSend is 0." ) );
-    }
-    else if( !isValidNetworkContext( pNetworkContext ) || ( pBuffer == NULL ) )
+    if( !isValidNetworkContext( pNetworkContext ) ||
```
chinglee-iot, line 870: Suggest to separate the parameter check condition to clearly indicate error.

chinglee-iot (approved):

n9wxu (approved):

PR #1819, opened 2022-10-06
Use additional OpensslCredentials_t parameter for hostname checking.

archigup:
This will allow demos to disable hostname checking, allowing them to be used with Greengrass or other endpoints that do not have the hostname in their certificate subject.

### DIFF
--- a/platform/posix/transport/include/openssl_posix.h
+++ b/platform/posix/transport/include/openssl_posix.h
@@ -119,6 +119,11 @@ typedef struct OpensslCredentials
     const char * sniHostName;
 
     /**
+     * @brief If non-zero, don't compare hostname to server certificate subject.
+     */
+    uint8_t disableHostnameCheck;
+
+    /**
      * @brief Set the value for the TLS max fragment length (TLS MFLN)
      *
      * OpenSSL allows this value to be in the range of:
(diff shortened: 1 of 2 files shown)


platform/posix/transport/src/openssl_posix.c
```diff
@@ -244,12 +244,15 @@ static OpensslStatus_t tlsHandshake( const ServerInfo_t * pServerInfo,
     int32_t sslStatus = -1, verifyPeerCertStatus = X509_V_OK;
 
     /* Validate the hostname against the server's certificate. */
-    sslStatus = SSL_set1_host( pOpensslParams->pSsl, pServerInfo->pHostName );
-
-    if( sslStatus != 1 )
+    if( pOpensslCredentials->disableHostnameCheck == 0U )
```
paulbartell, line 247: I would prefer that we add a separate connect() function which takes a sockaddr_in rather than a hostname.

This would encourage the use of SNI whenever possible.

platform/posix/transport/include/openssl_posix.h
```diff
@@ -118,6 +118,11 @@ typedef struct OpensslCredentials
      */
     const char * sniHostName;
 
+    /**
```
paulbartell, line 121: sniHostName == NULL should result in SNI being disabled rather than an explicit disableHostnameCheck.

platform/posix/transport/src/openssl_posix.c
paulbartell: But I think just checking for pOpensslCredentials->sniHostName != NULL should be sufficient here.

PR #1850, opened 2023-02-06
Remove install instructions from README

archigup:
Installation isn't the recommended usage, and causes customer confusion.


johnrhen (approved):

AniruddhaKanhere (approved):

AniruddhaKanhere (commented):
Approved with the below comment/question.
If this is not the recommended method but is still a valid method, then why are we completely removing it? Maybe we should move it to 'Other instructions' section?

archigup:
We do not recommend customers use CSDK in this way. CSDK is designed around compile time configuration and building it into your application. Having these instructions misleads customers; we often get customers confused trying to use it when it doesn't fit their use case. We also plan to remove this target.

PR #1870, opened 2023-06-05
Add file path length check

gmtt:
*Description of changes:*
Add file path length check before read file in ota_pal_posix.c

### DIFF
--- a/platform/posix/ota_pal/source/ota_pal_posix.c
+++ b/platform/posix/ota_pal/source/ota_pal_posix.c
@@ -418,7 +418,7 @@ OtaPalStatus_t otaPal_CreateFileForRx( OtaFileContext_t * const C )
 
     if( C != NULL )
     {
-        if( ( C->pFilePath != NULL ) && ( strlen( ( const char * ) C->pFilePath ) + 1U > OTA_FILE_PATH_LENGTH_MAX ) )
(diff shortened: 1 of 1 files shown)


platform/posix/ota_pal/source/ota_pal_posix.c
```diff
@@ -418,7 +418,12 @@ OtaPalStatus_t otaPal_CreateFileForRx( OtaFileContext_t * const C )
 
     if( C != NULL )
     {
-        if( C->pFilePath != NULL )
+        if( ( C->pFilePath != NULL ) && ( strlen( ( const char * ) C->pFilePath ) + 1U > OTA_FILE_PATH_LENGTH_MAX ) )
```
jasonpcarroll, line 421: hmm the +1 to account for null terminator may not be needed - at least in my mind OTA_FILE_PATH_LENGTH_MAX should be able to store a path that is that long (not including the null terminator), but if this macro defines a specific buffer size and that buffer is expected to be null-terminated this is fine.

jasonpcarroll (dismissed):

AniruddhaKanhere (approved):

AniruddhaKanhere (dismissed):

aggarg (approved):

ActoryOu (approved):

sirnish (approved):
Looks Good

PR #1877, opened 2023-09-03
fleet provisioning with csr demo support for using existing private key and to write certificate to disk

giuspen:
*Issue #, if available:*

*Description of changes:*

I added support in the fleet provisioning with csr demo for using an already existing private key instead of generating a new one (if defining EXISTING_PRIVATE_KEY_PATH) and to write the downloaded certificate to disk (if defining DOWNLOADED_CERT_WRITE_PATH)


Skptak:
Hey @giuspen, thanks for submitting this update to the demo! These seem like useful changes for this demo! I'm reaching out to a team-member who's worked with this to see if they can provide a review

giuspen:
Thanks @Skptak note for completeness that I would also like in future to optionally allow to save the private key to disk when generated in this demo, but that is blocked from the issue https://github.com/aws/aws-iot-device-sdk-embedded-C/issues/1878

paulbartell (changes requested):
@giuspen : The fleet_provisioning_with_csr demo was designed to use PKCS#11 as a secure key storage mechanism. CorePKCS11 is a library which simulates the PKCS#11 api commonly provided by Secure Elements and other Hardware Security Modules.

We would welcome a contribution of a separate fleet provisioning demo which handles cryptographic operations by calling mbedtls or openssl APIs directly and accomplishes key storage by writing to the filesystem. That would be a very different demo and quite a bit different from what you have presented in this PR.

giuspen:
I see @paulbartell thanks for your review

PR #1896, opened 2023-11-07
Update readme to configure TCP keep alive when updating MQTT keep alive time

tony-josi-aws:
While updating the MQTT keep alive time beyond certain value which is above the TCP keep alive time the underlying TCP connection is assumed dead depending on the cofiguration and this causes the device disconnection to go undetected.

This PR updates the readme with instructions to configure TCP keep alive when updating MQTT keep alive time.

### DIFF
--- a/demos/readme.md
+++ b/demos/readme.md
@@ -1,18 +1,54 @@
 
 ## Updating MQTT keep alive time
 
-When you increase MQTT keep alive time and observe that the
-device is getting disconnected before the MQTT keep alive kicks in,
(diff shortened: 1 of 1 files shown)


aggarg (dismissed):

ActoryOu (dismissed):
Overall, it's good enough. Approve in advance. Some indents are not aligned with code. It'd be great to align that.

moninom1 (approved):

ActoryOu (approved):

aggarg (approved):

PR #1922, opened 2024-10-30
Fix frequent MQTT_Ping calls in handleKeepAlive

vishwamartur:
Related to #1915

Update `handleKeepAlive` in `source/core_mqtt.c` to correctly update `pContext->lastPacketTime` when data is received.

* Ensure `MQTT_Ping` is not called frequently due to `pContext->lastPacketTime` not updating.
* Retain the logic for sending ping packets based on both received and sent message intervals.
* Address the issue related to the underlying transport failing to send the data, ensuring reliable data transmission.

### DIFF
--- a/source/core_mqtt.c
+++ b/source/core_mqtt.c
@@ -1,0 +1,1430 @@
+/*
+ * coreMQTT v1.1.0
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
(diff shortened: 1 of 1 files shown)


ActoryOu:
Hi @vishwamartur,
It looks like you're updating [CoreMQTT](https://github.com/FreeRTOS/coreMQTT). Could you help raise a PR there? That's the place to maintain core_mqtt.c.

Thank you.

